PR #25, opened 2020-06-05
Restructured tests

ljaljushkin:
- extracted common test helpers
- separated tests related to range-init


tests/test_helpers.py
```diff
@@ -10,25 +10,23 @@
  See the License for the specific language governing permissions and
```
vshampor, line 10: I think that some of the logic of Pytest's test case collection depends on the files containing test cases having names starting with `test_`. If you intend this file not to contain any actual test cases to be run, perhaps we should rename it to simply `tests\helpers.py`

ljaljushkin: there're still self-check tests for helper: e.g. `test_basic_model_is_valid`, `test_basic_model_has_expected_params`

ljaljushkin: eventually, decided to separate helpers for tests from actual test functions

ljaljushkin:
Jenkins please retry a build

PR #27, opened 2020-06-06
Update documentation

vshampor:


vshampor:
@AlexKoff88 note that this goes directly to master since this is only a documentation update. Afterwards will figure out how to synchronize the develop's documentation state with these changes, if necessary.

PR #32, opened 2020-06-15
Generalize initializing data loader

vshampor:


nncf/initialization.py
```diff
@@ -105,14 +89,25 @@ def write(cls, msg):
...
+                                                 desc=bar_desc, bar_format=bar_format):
                 if num_init_steps is not None and i >= num_init_steps:
                     break
-                self.model(input_, **dataloader_kwargs)
+
+                def is_tensor(obj):
+                    return isinstance(obj, torch.Tensor)
+
+                to_device_fn = partial(torch.Tensor.to, device=device)
+                args, kwargs = objwalk(args_and_kwargs_tuple, is_tensor, to_device_fn)
```
ljaljushkin, line 108: the same changes are needed in hessian_estimator.

nncf/structures.py
```diff
@@ -30,11 +30,13 @@ class QuantizationPrecisionInitArgs(NNCFExtraConfigStruct):
     :param data_loader: 'data_loader' - provides an iterable over the given dataset, instance of descendant
                 of 'torch.utils.data.DataLoader' class. Must return both inputs and targets to calculate loss
                 and gradients.
+    :param device: Device to perform initialization at. Either 'cpu' or 'cuda' (default).
     """
 
-    def __init__(self, criterion: _Loss, data_loader: DataLoader):
+    def __init__(self, criterion: _Loss, data_loader: DataLoader, device: str = 'cuda'):
```
ljaljushkin, line 36: I think, better to use torch.device type

nncf/initialization.py
vshampor: Done

nncf/structures.py
vshampor: Probably `str` is what the users are more used to. Will also help generalize for TF.

vshampor:
@ljaljushkin note that I removed random transforms from the precision init test loaders. Somehow even this refactoring has upset the RNG call sequence and the previous reference traces no longer matched to the test traces. I updated the reference traces by first removing the random transforms on the current `develop` state, then updating this commit's reference traces with the test traces I got in the previous step.

vshampor:
Precommit failed due to CUDA OOM, because another test run was being executed concurrently on the same machine. Anyway, the latest update only affected the transformers integration patch and the associated readme, so I'll be merging this.

PR #39, opened 2020-06-25
Support FP16 in quantization kernels

vshampor:


vshampor:
NB: Ran the benchmark as well, the kernel runtime doesn't seem to be affected in any consistent way (i.e. no global speedup or slowdown).

nncf/binarization/cuda/functions_cuda_kernel.cu
```diff
@@ -29,7 +29,7 @@ __device__ bool last_block(int* counter) {
 
 // support only warp size = 32
 template <typename scalar_t>
-__device__ void sum_warp(volatile scalar_t* sharr) {
```
ljaljushkin, line 32: I was against FP16 kernels only because of this change. Are you sure that CUDA wouldn't mess up sum due to optimization? Probably this is what leads to excessive relative error in tests.

vshampor: The `sum_warp` function is only used in backward's gradient calculation, but the errors occur even in the forward test cases. Anyway, I made a specialization of the function so that FP32 still uses the volatile-warp-synchronous programming, and FP16 uses explicit warp synchronization (no way around doing this if we want to be sure about sums being correct for FP16).

vshampor:
FYI, judging by the runtime of individual tests from the `test_functions.py` suite the FP16 quantization runs 3x slower than the FP32 quantization. This trend remains the same even if I remove all warp thread syncs and use the "unsafe" warp-synchronous code. It is possible that FP16 computations are either not optimized enough for my HW (RTX 2080 Ti), or there is some special way in which FP16 kernels should be written in order to leverage the speedups.

vshampor:
Jenkins please retry a build

PR #42, opened 2020-06-26
Latest configs for mixed precision by staged quantization

ljaljushkin:


AlexKoff88:
@ljaljushkin, do we have these manual configs just to showcase that we have this capability?

ljaljushkin:
> have these manual configs just to showcase that we have
right. it might be useful for users to fine-tune mixed precision faster.

PR #50, opened 2020-07-08
Fix gradient calculation for quantization lower bound

vshampor:
An error was made during simultaneous work on the CUDA kernels way
back in time, and the per activation channel quantization kernel
ended up calculating gradients for lower bound of quantization
incorrectly when the currently quantized value was lower than this
lower bound. This slipped through the tests because they themselves
were too lax and had epsilon values set incommensurably high compared
to the quantized values. Fixed this - also removed the negative range
tests because they will actually always fail with the fixed test setup
due to TuneRange function that is built into AsymmetricQuantize automatically
resetting the negative input_range to positive values.


vshampor:
Jenkins please retry a build

PR #53, opened 2020-07-17
Update result tables

vshampor:


README.md
```diff
@@ -104,50 +104,101 @@ Refer to the [CONTRIBUTING.md](./CONTRIBUTING.md) file for guidelines on contrib
...
+
+|Model|Compression algorithm|Dataset|PyTorch FP32 baseline|PyTorch compressed accuracy|
+| :---: | :---: | :---: | :---: | :---: |
+|RetinaNet-ResNet50-FPN|None|COCO2017|35.6 (avg box mAP)|-|
+|RetinaNet-ResNet50-FPN|INT8|COCO2017|35.6 (avg box mAP)|35.3 (avg box mAP)|
+|RetinaNet-ResNet50-FPN|INT8 + Sparsity 50%|COCO2017|35.6 (avg box mAP)|34.7 (avg box mAP)|
+|RetinaNet-ResNeXt101-64x4d-FPN|None|COCO2017|39.6 (avg box mAP)|-|
+|RetinaNet-ResNeXt101-64x4d-FPN|INT8|COCO2017|39.6 (avg box mAP)|39.1 (avg box mAP)|
+
+
```
vanyalzr, line 201: You could also add Instance Segmentation results:

|Model|Compression algorithm|PyTorch FP32 baseline|PyTorch compressed accuracy|
| ------------- |:-------------:|:-------------:|:-------------:|
|MaskRCNN-ResNet50-FPN|INT8|0.379 (AP@IoU=0.50:0.95)|0.372 (AP@IoU=0.50:0.95) |

vshampor: We'd better provide a configuration file to reproduce it then; in the near future, I think, we will be validating third-party integration checkpoints along with the regular ones.

vanyalzr: These results were obtained by integrating NNCF with torchvision. Hence we'll need to provide a whole separate torchvision patch to validate them. Given that torchvision is not a widely used package for OD models, it might be not reasonable. Alternatively, I can add an mmdetction config for the same Mask-RCNN model and update the results. @AlexKoff88 what do you think?

AlexKoff88: We need to provide a config to claim these numbers. As far as I remember in order to reproduce the results patching of torchvision code is required which means that we need to provide and support this patch. Personally, I don't like this idea.

AlexKoff88: I have a different comment. Why do we need rows with "None" compression type while we provide baseline metrics at the same time?

vanyalzr: I initially reproduced Mask-RCNN quantization within 1% mAP drop in mmdetection. Can provide the corresponding config in the future, if needed.

vshampor: @AlexKoff88 a good point there. Will remove in a follow-up PR.

PR #55, opened 2020-07-17
Extend install testing to building and installing from sdist/bdist_wheel

vshampor:


vshampor:
@AlexKoff88 once this is merged, we should technically be ready for releases to PyPI.

AlexKoff88:
> @AlexKoff88 once this is merged, we should technically be ready for releases to PyPI.
Great!

PR #58, opened 2020-07-20
Fix compressed checkpoint saving and improve log clarity

vshampor:


vshampor:
@ljaljushkin , the previous version of checkpoint saving caused failures in nightly:
http://134.191.240.124/job/nncf/job/nightly/job/nncf_cuda/96/#showFailuresLink

Will run nightly on this PR as well and post the results.

ljaljushkin:
> @ljaljushkin , the previous version of checkpoint saving caused failures in nightly:
> http://134.191.240.124/job/nncf/job/nightly/job/nncf_cuda/96/#showFailuresLink
> 
> Will run nightly on this PR as well and post the results.

Is it ok to bypass unexpected warnings in nightly?

vshampor:
That line only affects the "--weights" loading scenario. We don't assert that the unexpected layer set for nightly checkpoints is empty, anyway.

ljaljushkin:
> That line only affects the "--weights" loading scenario. We don't assert that the unexpected layer set for nightly checkpoints is empty, anyway.

How does `--weights` option cause failures? The checkpoint should be loaded with `force=False` without warnings

vshampor:
In truth, I only introduced this change after I attempted to do a `--weights` run when I should have done a `--resume` run. It looked pretty strange that I had "unexpected layers", but "Loaded 314/314 layers" in console output at the same time. But this kind of strangeness might just be a signal that something went wrong. Will remove the suppression code.

nncf/checkpoint_loading.py
```diff
@@ -56,14 +56,17 @@ def key_normalizer(key):
...
-def process_problematic_keys(is_resume, issues, is_all_saved_loaded):
+def process_problematic_keys(is_resume, issues, is_all_saved_loaded, ignore_unexpected_only=False):
     error_msgs = []
 
     def add_error_msg(name, keys):
         error_msgs.insert(
             0, '{} key(s):\n{}. '.format(name,
                                          ',\n'.join('\t\t"{}"'.format(k) for k in keys)))
 
+    if len(issues["Missing"]) == 0 and ignore_unexpected_only:
```
ljaljushkin, line 67: is it needed?

vshampor: Thanks for spotting this, removed.

nncf/sparsity/magnitude/algo.py
```diff
@@ -41,15 +41,13 @@ def __init__(self, target_model: NNCFNetwork,
...
         super().__init__(target_model, sparsified_module_info)
-        self.sparsity_level = self.threshold = 0
         self.weight_importance = WEIGHT_IMPORTANCE_FUNCTIONS.get(weight_importance)
         scheduler_cls = SPARSITY_SCHEDULERS.get(params.get("schedule", "polynomial"))
         self._scheduler = scheduler_cls(self, params)
 
     def statistics(self):
         stats = super().statistics()
-        stats['sparsity_threshold'] = self.threshold
-        stats['sparsity_level'] = self.sparsity_level
```
ljaljushkin, line 52: why did you remove sparsity_level?

vshampor: This `sparsity_level` report only served to increase confusion. This is not the actual sparsity level, but rather the level that the sparsity scheduler will set at the next call of "set_sparsity_level" (don't remember the exact name). The actual sparsity rate statistic is already reported as "sparsity_rate_for_sparsified_modules". As a result, during evaluation of a completely sparsified checkpoint and calling statistics the user would get seemingly contradictive values of "sparsity_level" (which would be set to some initial scheduled value as specified in the config file) and "sparsity_rate_for_sparsified_modules" (which would be close to the final sparsity value of the config file).

ljaljushkin: yep, there's also `sparsity_rate_for_model` to represent sparsity level for the whole model. then it should be ok

nncf/sparsity/magnitude/algo.py
```diff
@@ -59,17 +57,14 @@ def set_sparsity_level(self, sparsity_level):
...
-        self._set_masks_for_threshold(self.threshold)
-
-    def _select_threshold(self):
+    def _select_threshold(self, sparsity_level):
         all_weights = self._collect_all_weights()
         if not all_weights:
             return 0.0
         all_weights_tensor, _ = torch.cat(all_weights).sort()
-        threshold = all_weights_tensor[int(all_weights_tensor.size(0) * self.sparsity_level)].item()
+        threshold = all_weights_tensor[int((all_weights_tensor.size(0) - 1) * sparsity_level)].item()
```
ljaljushkin, line 67: doesn't it contradict to the previous logic with sparsity rate?

vshampor: Wherein lies the contradiction? This function does more or less the same as before, with added benefit that now it won't fail with an out-of-bounds error for sparsity rate = 1.0

vshampor:
Jenkins please retry a build

nncf/sparsity/magnitude/algo.py
ljaljushkin: technically, it won't sparsify all values, but will keep the maximum values. but sparsity rate=1 is a corner case which shouldn't happen in a real life, so maybe it's ok

PR #60, opened 2020-07-21
Remove redundant result entries in the top-level README

vshampor:


vshampor:
@AlexKoff88

AlexKoff88:
Just a comment that is not related to this MR. Why do we have the master branch by default, not the default branch?

PR #66, opened 2020-07-24
Adjust compression training thresholds for sparsity

vshampor:
Previous thresholds incorrectly tracked global best accuracy
instead of the best accuracy with full compression (i.e. sparsity)
applied.


ljaljushkin:
is it degradation?

vshampor:
No, it is a correction in the testing premises. The overall best accuracy in these test runs is not achieved at the same time as the complete sparsity level, but it is this overall best accuracy that was saved inside the checkpoints as the "best", and the "best" checkpoint that was loaded in the following case did not actually correspond to the final sparsity level specified in configs. Because the best checkpoint export in samples was wrong, this test passing before means that the test itself was wrong.

PR #68, opened 2020-07-27
Create python-publish.yml

vshampor:
Will publish a package to PyPI upon a Github release.


.github/workflows/python-publish.yml
```diff
@@ -0,0 +1,31 @@
+# This workflows will upload a Python Package using Twine when a release is created
```
AlexKoff88, line 1: What about the licensing of such files?

vshampor: No idea. I just used a template file as-is, available from https://github.com/openvinotoolkit/nncf_pytorch/actions. Should we just slap an Apache header onto this file?

AlexKoff88: Probably yes, I actually meant this.

vshampor: Done

Issue #83, opened 2020-08-03
Percentile-based initialization fails in per-channel quantization case


vanyalzr:
`  File "/nncf/initialization.py", line 170, in _apply_initializers
    initializer.apply_init()
  File "/nncf/quantization/init_range.py", line 223, in apply_init
    self.quantize_module.apply_minmax_init(mins_tensor, maxs_tensor, self.log_module_name)
  File "/nncf/quantization/layers.py", line 293, in apply_minmax_init
    self.scale.masked_scatter_(torch.gt(abs_max, SCALE_LOWER_THRESHOLD), abs_max)
RuntimeError: invalid argument 2: source nElements must be == mask `1` elements at /pytorch/aten/src/THC/generic/THCTensorMasked.cu:134`

Should cover this case in pre-commit tests.

PR #86, opened 2020-08-03
Fix percentile per-channel init

vshampor:
Fixes: #83

### DIFF
--- a/tests/quantization/test_range_init.py
+++ b/tests/quantization/test_range_init.py
@@ -10,6 +10,8 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+import itertools
+
 import pytest
 import re
 import torch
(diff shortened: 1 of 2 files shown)


tests/quantization/test_range_init.py
```diff
@@ -319,22 +331,32 @@ def __len__(self):
...
+                assert quantizer.scale.numel() == 1
         else:
-            assert quantizer.input_low.item() == approx(3210, abs=1.0)
-            assert quantizer.input_range.item() == approx(3578, abs=1.0)
-
-    assert_range(act_quantizer_info.quantizer_module_ref)
+            assert torch.allclose(quantizer.input_low, torch.ones_like(quantizer.input_low) * 3210, atol=1.0)
+            assert torch.allclose(quantizer.input_range, torch.ones_like(quantizer.input_low) * 3578, atol=1.0)
+            if per_channel:
+                assert quantizer.input_low.numel() == 3
```
ljaljushkin, line 346: you can avoid condition `if per_channel` by adding numel to params

vshampor: I think this way is more obvious. Would also complicate the parametrization iterable - can't get rid of *all* complexity here

ljaljushkin: why not just use tuples ('symmetric', 1) and ('asymmetric, 3')?

PR #89, opened 2020-08-05
Enable custom range initializers for overriden scopes in schema

vanyalzr:


nncf/config_schema.py
```diff
@@ -231,6 +232,40 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
     "additionalProperties": False,
 }
 
+RANGE_INIT_CONFIG = {
```
vshampor, line 235: Looks like this is a duplicate. Should remove the original, if you plan on extracting these piece to reuse.

nncf/config_schema.py
```diff
@@ -306,7 +341,7 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
             "patternProperties": {
                 ".*": {
                     "type": "object",
-                    "properties": QUANTIZER_CONFIG_PROPERTIES,
+                    "properties": QUANTIZER_CONFIG_INIT_PROPERTIES,
```
vshampor, line 344: Suggest to do the following:
```
"properties": {
**QUANTIZER_CONFIG_PROPERTIES,
**RANGE_INIT_CONFIG_PROPERTIES,
}
```

Might also want to do the same in `QUANTIZER_GROUP_SCHEMA` - so that it is possible to set different inits for entire activation or weight quantizer sets.

nncf/config_schema.py
vshampor: `RANGE_INIT_CONFIG_PROPERTIES`

vanyalzr: Agreed, done

nncf/config_schema.py
vanyalzr: Done. Might do separate initialization configs for quantizer groups in a separate PR.

vanyalzr:
> It doesn't seem that the new config layout is being parsed during quantization building. Some files not added to the commit?

Actually this feature [has been already implemented](https://github.com/openvinotoolkit/nncf_pytorch/blob/404ec51cdc89d739e19e591f4c1513f9b10151dd/nncf/quantization/algo.py#L966) in the quantization algo. It just wasn't enabled in the config schema.

PR #94, opened 2020-08-06
Fix HW config scenario with ops missing in HW config definition

vshampor:


nncf/quantization/quantizer_propagation.py
```diff
@@ -398,7 +398,10 @@ def get_visualized_graph(self):
                     prop_quantizer = node[
                         QuantizerPropagationStateGraph.PROPAGATING_QUANTIZER_NODE_ATTR]  # type: PropagatingQuantizer
                     quant_node_key = "Quantizer #{}".format(prop_quantizer.id)
-                    quant_configs_str_list = [str(conf) for conf in prop_quantizer.potential_quant_configs]
+                    if prop_quantizer.potential_quant_configs:
+                        quant_configs_str_list = [str(conf) for conf in prop_quantizer.potential_quant_configs]
+                    else:
+                        quant_configs_str_list = ["!!! NONE !!!]"]
```
AlexKoff88, line 404: What does it mean ["!!! NONE !!!]"]?

vshampor: This is just a debug visualization code for the quantizer propagation graph. If a propagating quantizer is placed, it currently has to have at least one valid configuration (parsed from config, or defaulted) associated with it. If it has no configurations at all, then currently this is a mistake, and the corresponding spot in the visualization should be drawing attention.

PR #96, opened 2020-08-10
Assign latest tensor shape during ForwardTraceOnly()

vshampor:


nncf/dynamic_graph/patch_pytorch.py
```diff
@@ -51,25 +52,30 @@ def __call__(self, operator: Callable, *args, **kwargs):
             if len(input_traced_tensor_indices) == 1:
                 # Broadcast one and the same creator ID of input to all outputs
                 for out_idx in output_tensors_to_be_traced_indices:
+                    target_meta = deepcopy(fargs[input_traced_tensor_indices[0]].tensor_meta)
```
AlexKoff88, line 55: It is hard to understand what is `target_meta` in this context. 
BTW, what problem does this fix solve?

vshampor: In our "compressed model.dot" graph visualization we draw activation tensor shapes right next to the corresponding edges between the operation nodes. Some ops are purposefully not included in the NNCF graph, such as `split`, `transpose` etc., but they might affect the shape.

GPT2, for instance, has a 3-way `split` operation right after a "Conv1D" operation and a `view` operation to separate the attention heads into a new tensor dimension. We hide the "split" and "view" in the graph, since these ops are not really interesting compression-wise (here). This is how the "compressed_model.dot" looks after the fix:
Screenshot from 2020-08-10 20-15-26

Before the fix, the outgoing red edges would all read `(1, 768, 3072)` - the shape of the hidden operations' *input* tensors rather than output, which looks quite confusing in terms of matrix multiplication dimension requirements when looking at inputs of the node 85 (`matmul`), for example.

AlexKoff88: Now it is clear, thanks!

PR #108, opened 2020-08-18
Add unified scales in HW config operation (via quantizer linking)

vshampor:


AlexKoff88:
It would be also useful to conduct an experiment for QAT w and w/o scales unification.

vshampor:
@ljaljushkin I tried to find the impact points of this new quantizer sharing mechanism on the HAWQ precision init, but could only find one in the `get_bitwidth_graph`, which is essentially a test function. If something comes to your mind that you believe should be paid attention to in HAWQ with respect to the usage with this new mechanism (currently only VPU is impacted), please comment.

vshampor:
For MobileNetV2, a "vpu" INT8 training produces 71.35 top1 w/o unified scales , and 71.3 w/ unified scales.

vshampor:
For ResNet50, the corresponding numbers are 75.67 top1 w/o and 75.56 w/. It's probably safe to assume that unified scale enabling for VPU is a non-breaking change.

PR #111, opened 2020-08-24
Make HW config parsing conform to the implicit rules

vshampor:
(except for the "any supported quantization for the ops in config
without specified quantizations", because they need config wildcarding,
to be implemented as a follow-up)

### DIFF
--- a/nncf/quantization/algo.py
+++ b/nncf/quantization/algo.py
@@ -14,6 +14,7 @@
 # pylint:disable=too-many-lines
 import functools
 from collections import OrderedDict, namedtuple
+from copy import deepcopy
 from enum import Enum
 from pathlib import Path
 from typing import List, Dict, Tuple, Optional, Callable
@@ -24,7 +25,6 @@ import operator
 import shutil
 import torch
 from nncf.layers import NNCFEmbedding
-from texttable import Texttable
 from torch import nn
 
 from nncf.algo_selector import COMPRESSION_ALGORITHMS
@@ -117,6 +117,11 @@ class NonWeightQuantizerInfo:
 
 @COMPRESSION_ALGORITHMS.register('quantization')
 class QuantizationBuilder(CompressionAlgorithmBuilder):
+    DEFAULT_QUANTIZER_CONFIG = QuantizerConfig(bits=8,
+                                               mode=QuantizationMode.SYMMETRIC,
+                                               signedness_to_force=None,
+                                               per_channel=False)
+
     def __init__(self, config, should_init: bool = True):
         super().__init__(config, should_init)
 
@@ -195,10 +200,7 @@ class QuantizationBuilder(CompressionAlgorithmBuilder):
                                       self._hw_precision_constraints)
 
     def __get_default_qconfig(self, constraints: QuantizationConstraints = None):
-        qconfig = QuantizerConfig(bits=8,
-                                  mode=QuantizationMode.SYMMETRIC,
-                                  signedness_to_force=None,
-                                  per_channel=False)
+        qconfig = deepcopy(self.DEFAULT_QUANTIZER_CONFIG)
         if constraints is not None:
             qconfig = constraints.apply_constraints_to(qconfig)
         return qconfig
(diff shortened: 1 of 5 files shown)


nncf/quantization/algo.py
```diff
@@ -117,6 +117,11 @@ def __init__(self, quantizer_module_ref: BaseQuantizer,
 
 @COMPRESSION_ALGORITHMS.register('quantization')
 class QuantizationBuilder(CompressionAlgorithmBuilder):
+    DEFAULT_QUANTIZER_CONFIG = QuantizerConfig(bits=8,
+                                               mode=QuantizationMode.SYMMETRIC,
+                                               signedness_to_force=None,
```
AlexKoff88, line 122: Didn't get what  `signedness_to_force` means? Not a user-friendly name, IMHO.

nncf/quantization/algo.py
```diff
@@ -275,10 +277,12 @@ def get_potential_quantized_modules(self, target_model: NNCFNetwork) -> List[Pot
...
                 metatype = graph_operation[InsertionPointGraph.OPERATOR_METATYPE_NODE_ATTR]
                 qconfig_list = meta_vs_qconfig_map[metatype]
-                if qconfig_list is None:
+                if qconfig_list is not None and len(qconfig_list) == 0:  # Empty list = wildcard quantization
                     qconfig_list = default_qconfig_list
-            quantized_modules_with_potential_qconfig.append(PotentialQuantizedModule(module, module_scope,
-                                                                                     qconfig_list))
+
+            if qconfig_list is not None:
+                quantized_modules_with_potential_qconfig.append(PotentialQuantizedModule(module, module_scope,
```
AlexKoff88, line 284: `PotentialQuantizedModule` does not look self-explainable. Looking at the code I would call it `QuantizableModules`

nncf/quantization/algo.py
```diff
@@ -821,17 +825,20 @@ def __init__(self, target_model: NNCFNetwork,
             self.metric_store = {}
             quantizer_setup_type = self.quantization_config.get('quantizer_setup_type')
             # These metrics are collected here and are updated when the method .statistics() is called
-            self.non_stable_metric_collectors = [NetworkQuantizationShareMetric(target_model, self.weight_quantizers,\
-                 self.non_weight_quantizers, quantizer_setup_type), MemoryСostMetric(target_model, self.weight_quantizers, self.non_weight_quantizers)]
+            self.non_stable_metric_collectors = [NetworkQuantizationShareMetric(target_model, self.weight_quantizers, \
```
AlexKoff88, line 828: Why the `non_stable_metric_collectors ` variable is called so? Are there any problems with these collectors?

nncf/quantization/algo.py
```diff
@@ -275,10 +277,12 @@ def get_potential_quantized_modules(self, target_model: NNCFNetwork) -> List[Pot
                 graph_operation = associated_ops[0]
                 metatype = graph_operation[InsertionPointGraph.OPERATOR_METATYPE_NODE_ATTR]
                 qconfig_list = meta_vs_qconfig_map[metatype]
-                if qconfig_list is None:
+                if qconfig_list is not None and len(qconfig_list) == 0:  # Empty list = wildcard quantization
```
AlexKoff88, line 280: You should explain what is `wildcard quantization` somewhere and how it can be properly specified. Examples are desirable.

nncf/quantization/algo.py
vshampor: This is an optional boolean value. Signedness to force = True meance force signed, False means force unsigned, None means do not force anything. Cannot make the name any more user-friendly than that while keeping its meaning.

nncf/quantization/algo.py
vshampor: @asenina

nncf/quantization/algo.py
AlexKoff88: How about `force_signed=True`?

vshampor: Then `force_signed=False` becomes ambiguous - does it mean "do not force signed" or does it mean "force unsigned"?

nncf/quantization/algo.py
vshampor: Already explained [here](https://github.com/openvinotoolkit/nncf_pytorch/blob/300ccd0998631df8b9be2dafa14ef0973c7518f2/nncf/hw_config.py#L185-L189). All these TODOs are for me, anyway, and I will be resolving these in the next PR.

nncf/quantization/algo.py
vshampor: @asenina

nncf/quantization/algo.py
vshampor: I introduced this `signedness_to_force` name a long time ago, and we have already discussed it then.

AlexKoff88: Not to me at least, it looks like a command-line option I faced many times `--force-smthg`.

PR #119, opened 2020-08-28
Properly handle ops in HW config without quantization configs associated

vshampor:
These get associated with a "wildcard" propagating quantizer, which
will either get merged with any other quantizer during propagation,
or get assigned a default quantization config.


RikAllen (dismissed):
That seems to correspond to what I was thinking - going beyond the parts of the code I understand well enough to comment on

PR #127, opened 2020-09-08
Add per-layer sparsity.

asenina:


asenina:
Jenkins please retry a build

asenina:
@vshampor  @AlexKoff88  please take a look

AlexKoff88:
@asenina, @vshampor, I don't understand the essence of this PR. AFAIK, we have a summary table with a per-layer sparsity ratio. So, what is proposed here? Can we have an example of output, like "Was" | "Now"?

cc'ed @alexsu52

asenina:
@AlexKoff88 Here it is proposed to add the ability to manually (per layer, local) set the sparsity level for each layer separately during training. In the future, the per layer scheduler will used by some specific algo.

nncf/sparsity/rb/loss.py
```diff
@@ -67,3 +67,38 @@ def target_sparsity_rate(self):
 
     def statistics(self):
         return {'mean_sparse_prob': 1 - self.mean_sparse_prob}
+
+    def set_target_sparsity_loss(self, sparsity_level):
+        self.target = 1 - sparsity_level
+
+class PerLayerSparseLoss(SparseLoss):
```
vshampor, line 74: `SparseLossForPerLayerSparsity`

nncf/sparsity/base_algo.py
```diff
@@ -161,3 +161,6 @@ def add_algo_specific_stats(self, stats):
 
     def compression_level(self) -> CompressionLevel:
         return CompressionLevel.FULL
+
+    def do_specific_setting_after_each_epoch(self):
+        pass
```
vshampor, line 166: Perhaps I misunderstand the purpose of this function, 'm not sure that we need it in general. Even now, this function is named too broadly, and the way it is used does not immediately suggest that it is indeed called "after each epoch" - it is also called in init, and in `_set_sparsity_level`. Maybe it should be more fittingly called `_run_post_sparsity_level_update_actions`? Does this reflect the way you wanted this function to be used? Could it be possibly merged with an algo-specific `_set_sparsity_level` then?

vshampor (changes requested):

tests/sparsity/test_common.py
```diff
@@ -225,3 +228,30 @@ def test_polynomial_schedule_per_optimizer_step(self, magnitude_algo_mock, speci
...
+    config = get_empty_config()
+    config['compression'] = {'algorithm': 'rb_sparsity', "params": {"schedule": 'per_layer'}}
+    _, compression_ctrl = create_compressed_model_and_algo_for_test(MockModel(), config)
+
+    # pylint: disable=protected-access
+    assert list(compression_ctrl._loss.per_layer_target.values())[0] == 1
+
+    compression_ctrl._scheduler._set_sparsity_level_for_module('MockModel/NNCFLinear[field]', 0.7)
+
+    assert list(compression_ctrl._loss.per_layer_target.values())[0] == pytest.approx(0.3)
```
vshampor, line 250: Also, in these kinds of tests you should mock the target `F.linear` function call, check its arguments, count the number of zeros in the sparsified tensor and then compare it against the reference. Should also check that setting a sparsity level for a layer does not impact the other layers. Only after this we will be sure that the feature is functional.

AlexKoff88:
@vanyalzr, @ljaljushkin, @alexsu52, need your review as well.

nncf/sparsity/rb/algo.py
```diff
@@ -47,15 +47,19 @@ def __init__(self, target_model: NNCFNetwork,
         super().__init__(target_model, sparsified_module_info)
 
         self._distributed = False
-        self._loss = SparseLoss()  # type: SparseLoss
-        scheduler_cls = SPARSITY_SCHEDULERS.get(params.get("schedule", "exponential"))
-        self._scheduler = scheduler_cls(self, params)
+        schedule_type = params.get("schedule", "exponential")
         sparsify_operations = [m.operand for m in self.sparsified_module_info]
-        self._loss.set_layers(sparsify_operations)
+        if schedule_type == 'per_layer':
```
alexsu52, line 52: What do you think about doing this via the registry?

nncf/sparsity/rb/loss.py
```diff
@@ -67,3 +67,38 @@ def target_sparsity_rate(self):
...
+        sparse_layers_loss = 0
+        for sparse_layer in self._sparse_layers:
+            if not self.disabled and not sparse_layer.sparsify:
+                raise AssertionError(
+                    "Invalid state of SparseLoss and SparsifiedWeight: mask is frozen for enabled loss")
+            if sparse_layer.sparsify:
+                sw_loss = sparse_layer.loss()
+                params_layer = sw_loss.view(-1).size(0)
+                params += params_layer
+                sparse_layers_loss += sw_loss.sum() / params_layer - self.per_layer_target[sparse_layer]
```
alexsu52, line 97: It looks like that the loss does not apply a penalty for the following case `sw_loss.sum() / params_layer` > `self.per_layer_target[sparse_layer]`

nncf/sparsity/schedulers.py
```diff
@@ -236,3 +239,28 @@ def current_sparsity_level(self):
...
+    def initialize(self):
+        for minfo in self.algo.sparsified_module_info:
+            self._set_sparsity_level_for_module(minfo.module_name, self.initial_sparsity)
+
+    def _set_sparsity_level_for_module(self, module_name: str, sparsity_level: float):
+        minfo = [self.sparsified_modules_info_by_modules_name[module_name]]
+        self.algo.set_sparsity_level_for_module(sparsity_level, minfo)
+
+    def epoch_step(self, epoch=None):
+        self.algo.do_specific_setting_after_each_epoch()
```
alexsu52, line 263: Why do you call the batchnorn adaptation algorithm after each epoch?

nncf/sparsity/schedulers.py
```diff
@@ -236,3 +239,28 @@ def current_sparsity_level(self):
...
+        for minfo in self.algo.sparsified_module_info:
+            self.sparsified_modules_info_by_modules_name[minfo.module_name] = minfo
+        self.layered_sparsity_control_algo = None
+        self.initialize()
+
+    def initialize(self):
+        for minfo in self.algo.sparsified_module_info:
+            self._set_sparsity_level_for_module(minfo.module_name, self.initial_sparsity)
+
+    def _set_sparsity_level_for_module(self, module_name: str, sparsity_level: float):
```
alexsu52, line 258: Why is this method private?

nncf/sparsity/schedulers.py
```diff
@@ -236,3 +239,28 @@ def current_sparsity_level(self):
     @property
     def target_sparsity_level(self):
         return self.sparsity_levels[-1]
+
+
+@SPARSITY_SCHEDULERS.register("per_layer")
+class PerLayerSparsityScheduler(SparsityScheduler):
```
alexsu52, line 245: Is there a function in the `SparsityScheduler` to set the sparsity level for the all model as for each layer?

alexsu52 (commented):
> 
> 
> @AlexKoff88 Here it is proposed to add the ability to manually (per layer, local) set the sparsity level for each layer separately during training. In the future, the per layer scheduler will used by some specific algo.

My general comments that this feature as per-layer target sparsity level definition can not be used via a configuration file. Also, at this moment, we have not any implemented algorithm for automatic selection sparsity level for each layer, except the rb-sparsity algorithm as well that doesn't need such functionality.

AlexKoff88:
> > @AlexKoff88 Here it is proposed to add the ability to manually (per layer, local) set the sparsity level for each layer separately during training. In the future, the per layer scheduler will used by some specific algo.
> 
> My general comments that this feature as per-layer target sparsity level definition can not be used via a configuration file. Also, at this moment, we have not any implemented algorithm for automatic selection sparsity level for each layer, except the rb-sparsity algorithm as well that doesn't need such functionality.

I agree with this statement. I am also curious about why we started this task which affects the whole sparsity feature without reviewing any proposal. I would like to avoid this in the future.

vshampor:
Per-layer, local sparsity setting is an absolutely natural extension of our current global sparsity manipulation feature. The regular, global sparsity algo functionality is unaffected. No proposal should be required here.

Manipulating sparsity is procedure that is necessarily done during training; per-layer sparsity setting is something that should be decided by an RL algorithm, for instance, not by a human person crafting a config file. A config file is something short and human-readable - an interface with a human. What is being implemented here is an interface for an algorithm.

AlexKoff88:
Understood. This can be useful for sure. But I am not sure how it aligns with the feature roadmap.

alexsu52:
> 
> 
> Per-layer, local sparsity setting is an absolutely natural extension of our current global sparsity manipulation feature. The regular, global sparsity algo functionality is unaffected. No proposal should be required here.
> 
> Manipulating sparsity is a procedure that is necessarily done during training; per-layer sparsity setting is something that should be decided by an RL algorithm, for instance, not by a human person crafting a config file. A config file is something short and human-readable - an interface with a human. What is being implemented here is an interface for an algorithm.

I would like to see a sample how will be used. BTW, If the user writes 'per_layer' in the config, the `PerLayerSparsityScheduler` instance will be created and used by the user.

vshampor:
@asenina can probably provide a sample usage code in the form of a test.

vshampor:
I think that the scheduler in this case should be a dummy - essentially scheduling a sparsity level for each layer is delegated to the user. The scheduler object may still receive `.step()` and `.epoch_step()` calls, for compatibility with existing pipelines, but should do nothing during these steps. It should still contain a valid sparsity loss - in the RB case, for example.

nncf/sparsity/rb/loss.py
asenina: Done.

nncf/sparsity/base_algo.py
asenina: Done.

nncf/sparsity/rb/loss.py
asenina: Done.

nncf/sparsity/schedulers.py
asenina: This was incorrect. Fixed.

nncf/sparsity/schedulers.py
asenina: This was incorrect. Fixed.

nncf/sparsity/schedulers.py
asenina: No, there isn`t.

nncf/sparsity/magnitude/algo.py
```diff
@@ -49,36 +49,41 @@ def __init__(self, target_model: NNCFNetwork,
...
-        stats['sparsity_threshold'] = self._select_threshold(self.sparsity_rate_for_sparsified_modules)
+        stats['sparsity_threshold'] =\
+             self._select_threshold(self.sparsity_rate_for_sparsified_modules, self.sparsified_module_info)
         return stats
 
     def freeze(self):
         pass
 
     def set_sparsity_level(self, sparsity_level):
+        self.set_sparsity_level_for_module(sparsity_level, self.sparsified_module_info)
```
vshampor, line 60: Now this isn't right. Before this change, our sparsity algorithms set a *global* sparsity level; 60% would mean that 60% of total parameters in sparsified modules would be set to zero, but each module may have a local, per-module sparsity rate other than 60%. Now you've changed it so that setting a 60% sparsity level will always set 60% sparsity locally for each module, both in the global sparsity scenario and in the per-layer sparsity scenario, which is wrong. Fix this so that the current configs still preserve original behaviour - apply *global* sparsity only, not necessarily uniform across the modules (layers).

vshampor (changes requested):

nncf/sparsity/magnitude/algo.py
vshampor: Also, once you are done, provide a statistics table from the console log after running one of NNCF's magnitude sparsity training before the change and after the change, so that we can be sure that the configs we distribute (which don't use per-layer sparsity) still produce the same sparsity distribution as before. It is strange that our existing unit tests did not catch this.

vshampor: Sorry, I guess I misread the code here. Remove the `self.sparsified_module_info` from the signature - you probably don't need to pass an attribute to the method of the same class, just access it in the method. Everywhere where you pass a sparsified modules info list, rename the arguments appropriately: `self.sparsified_module_info` -> `self.sparsified_modules_info_list`. Same with the function - `set_sparsity_level_for_module` seems to be accepting a list of module**s**, and not setting a sparsity level for a single module - either refactor the function, or rename it, or do both.

tests/sparsity/test_common.py
```diff
@@ -225,3 +229,66 @@ def test_polynomial_schedule_per_optimizer_step(self, magnitude_algo_mock, speci
...
+    #  as an interface to the automatic local selection level sparsity algorithm.
+
+    class PerLayerAutomaticControlSparsityLevelForExample():
+        def __init__(self, scheduler: PerLayerSparsityScheduler):
+            self.scheduler = scheduler
+            self.num_update = 0
+
+        def do_update_sparsity_levels(self):
+            for minfo in self.scheduler.algo.sparsified_module_info:
+                self.scheduler.set_sparsity_level_for_module(minfo.module_name, 0.1 + self.num_update * 0.05)
```
vshampor, line 263: I was thinking of a more explicit example, where `set_sparsity_level_for_module` would be called with an explicit, hard-coded model-specific string as an argument, such as `"TwoConvTestModel/NNCFConv2d[conv1]"` or something along these lines.

nncf/sparsity/schedulers.py
```diff
@@ -236,3 +236,52 @@ def current_sparsity_level(self):
...
+    def __init__(self, sparsity_algo, params, sparsity_level_control_algo=None):
+        super().__init__(sparsity_algo, params)
+        self.sparsified_modules_info_by_modules_name = {}
+        self.per_layer_sparsity_levels = {}
+        for minfo in self.algo.sparsified_module_info:
+            self.sparsified_modules_info_by_modules_name[minfo.module_name] = minfo
+        self.layered_sparsity_level_control_algo = sparsity_level_control_algo
```
vshampor, line 249: You don't need to register the control algorithm in the scheduler at all. It is the control algorithm that owns and uses the scheduler according to the scheduler's API, not vice versa.

nncf/sparsity/magnitude/algo.py
asenina: I can't access `self.sparsified_module_info` in the method because the method is used in local sparsity (where we pass a list of one module).  The second argument to  `self.set_sparsity_level_for_module`  is not always `self.sparsified_module_info`. 
In order not to pass its attribute to a class method, I can make  the method as `static`.

vshampor: @asenina this looks wrong and it probably *is* wrong, and can be done much better. Rename the functions properly so that the names reflect what the functions do, rethink the function signatures and level of access (public/protected) and I'm sure that the correct way will present itself on its own.

nncf/config_schema.py
```diff
@@ -483,7 +483,11 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
                                                              "sparsity level will be applied "
                                                              "immediately, so the length of this list "
                                                              "should be larger than the length of the "
-                                                             "'steps' by one.")
+                                                             "'steps' by one."),
+    "mode_sparsity_level":with_attributes(_STRING,
```
vshampor, line 487: `sparsity_level_setting_mode`

vshampor (changes requested):

nncf/sparsity/magnitude/algo.py
asenina: Done.

nncf/sparsity/schedulers.py
asenina: Deleted.

nncf/config_schema.py
asenina: Done.

vshampor (approved):

PR #132, opened 2020-09-09
Fix bug with CompositeCompressionAlgorithmController export_model()

aleksei-kashapov:


nncf/composite_compression.py
```diff
@@ -108,8 +108,9 @@ def statistics(self):
             stats.update(ctrl.statistics())
         return stats
 
-    def export_model(self, filename):
-        self.child_ctrls[-1].export_model(filename)
+    def export_model(self, filename, *args, **kwargs):
+        # The order of algorithms are always such as that the last algorithm is Pruning
+        self.child_ctrls[-1].export_model(filename, *args, **kwargs)
```
AlexKoff88, line 113: @aleksei-kashapov, what caused these changes? Is it a need to pass parameters for 3rd party pipelines that require custom export?

@alexsu52, for your information.

aleksei-kashapov: Exactly. Some 3rd party pipelines such as mmdetection utilize models taking not only a single image for a forward pass but also some args and kwargs.

In export_model() implementation for all other classes, we take this into account except export_model() for CompositeCompressionAlgorithmController.

vshampor: Should be `model_args, model_kwargs` instead of `*args, **kwargs`. Will be more consistent with `torch.onnx.export`. Also please add type hints.

Also I would like you to consider this: with this change, we will go from 2 different ways of handling custom-forward runs to 3:
1) by using `dummy_forward` for graph building, by default parsed from the `"input_info"` section of the config file
2) by using `InitializingDataLoader` for initialization
and now 
3) by passing explicit `model_args` and `model_kwargs` for export now.

Are options 1) and 2) strictly non-applicable here, or, for instance, we can use the same default-generated `dummy_forward` or the same `InitializingDataLoader` for the export? If we can, I suggest that this export function should optionally use ways 1) and 2) (by using Python optional arguments mechanism), in a use case when user's forward arguments in 3) can basically be the same as in 1) and 2).

vshampor: Oh, ok, after reviewing the existing code I recalled that some of the points that I described above are already implemented in the regular `CompressionAlgorithm` objects, and this is really more of a bug fix than a feature improvement. Still, @aleksei-kashapov you might want to think about opening a PR with improvements as described above if you find that they would make NNCF integration easier or with less code.

vshampor (approved):

PR #133, opened 2020-09-10
Introduced compression ratio parameter for Mixed Precision init

ljaljushkin:
It's used for choosing optimal mixed precision configuration for a given ratio.

Compression ratio of mixed precision quantization is calculated by relation to fully INT8 one.
Total compression for the model is sum of compression for each quantized layer, which is multiplication the layer's (Conv, Deconv, Linear) FLOPS and number of bits for its quantization. The ratio is used for estimation of performance boost for quantized model It's a better proxy for amount of calculation then number of parameters multiplied by bitwidth

myplot


nncf/config_schema.py
```diff
@@ -247,6 +247,12 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
                                                              "algorithm. It's calculated  between mean average trace "
                                                              "from previous iteration and current one. 1e-5 by default"
                                                              "bitwidth_per_scope"),
+                    "compression_ratio": with_attributes(_NUMBER,
```
AlexKoff88, line 250: I am ok with the name the description is not clear to me. What does `Minimum compression ratio`? I would rephrase that this is a ratio between bits complexity of fully INT8 model and mixed-precision lower-bit model. Minimum requires additional explanation.

AlexKoff88: I would use the term "Bit complexity" everywhere below as well.

nncf/nncf_network.py
```diff
@@ -666,9 +666,9 @@ def get_parameters_count_in_model(self):
             count = count + param.numel()
         return count
 
-    def get_MACs_in_model(self):
+    def get_ops_count_in_model(self):
```
AlexKoff88, line 669: Please adjust functions' names as well.

AlexKoff88:
As far as I remember we also discussed adding a template for INT4-INT8 quantization with a minimum set of parameters.

AlexKoff88 (approved):

RikAllen:
Is it possible to access the graph you show above from NNCF? Is that generated during an initialisation run? Would be very useful for debug. Thanks

ljaljushkin:
@RikAllen, yes. This graph is accessible in debug mode at <LOG_DIR>/hawq_dumps/Pareto_Frontier_compress_ratio.png
```
from nncf import set_log_level
set_log_level(logging.DEBUG)
```

RikAllen:
Thanks - yes I found it after more code reading. That branch supports --debug to enable which makes life a bit easier for me.

PR #146, opened 2020-09-22
Qconfig subclass 2

RikAllen:
Developed on top of block_floating_point branch. Subclass QuantizerConfig to differentiate between int-n and blockfp configurations. Add factory for construction of them.
Reuse "bits" field of the base class to represent number of bits in either case (was mantissa-bits in block_floating_point).  
Default QuantizerConfigs moved out from a few classes into the QuantizerConfig class itself with a getter based on mode to select int-n or blockfp configs as needed.


RikAllen:
This could be rebased into the original block_floating_point pull request to close the issue created there if preferred. Only last 3 commits differ.

PR #158, opened 2020-09-29
Fix bug with different torch devices

aleksei-kashapov:


ljaljushkin (commented):
it would be great to have a test case that captures this bug

aleksei-kashapov:
I have no idea why it happens. But in my scenario it was like that: I compressed the model without flag `export_to_onnx_standard_ops` then I load the model and tried to export it to ONNX format with `export_to_onnx_standard_ops: True`. In such case, this error occurs.

vshampor:
Jenkins please retry a build

PR #162, opened 2020-10-01
Set data loading workers to 0 across tests to force single process

vshampor:
Could fix the consequences of https://github.com/pytorch/pytorch/issues/39570


vshampor:
Jenkins please retry a build

PR #168, opened 2020-10-06
Add pre-hook functionality to quantization

vshampor:


vshampor:
Cannot be properly integrated without config merging logic change - will move to #302 which has the corresponding change in place.

PR #178, opened 2020-10-09
Renamings

vshampor:


kchechil (commented):
Let's also replace colors to 'blue' in tests/test_sota_checkpoints.py and nncf/quantization/metrics.py

vshampor:
> Let's also replace colors to 'blue' in tests/test_sota_checkpoints.py and nncf/quantization/metrics.py. Or 'brown' if 'blue' already used.

Removed these altogether since black seems to be default.

kchechil (approved):

Issue #164, opened 2020-10-02
Check consistency of all affected quantizers on activation quantizers merge in propagation mode.


ljaljushkin:
By default, the merge of activation quantizers should not happen if they are connected with weight quantizers that have different supported bit-width.
![image](https://user-images.githubusercontent.com/4014476/94909135-6710b900-04ab-11eb-9378-a06076927bad.png)

The merge can happen if the corresponding flag (e.g. allow_different_bitwidth_for_weight_and_activation or something shorter) is specified in the HW Config.
![image](https://user-images.githubusercontent.com/4014476/94909148-6d9f3080-04ab-11eb-8a7d-7c6ad4d07846.png)

@vshampor @asenina @RikAllen

RikAllen:
That sounds quite useful - we would be quantizing activations for storage only once - so in practical terms for storage we would be applying the second case. However - even is stored at e.g. 8 bit activation, we can run them through the convolution as 4 bit anyway if that is optimal for a particular convolution by masking bits off. Merge of quant nodes I think is part of what we need - but I need to figure out what my full ask is.
![image](https://user-images.githubusercontent.com/62760324/94912449-93bed380-049f-11eb-9dc7-4223a3645a85.png)

RikAllen:
(The odd bit depths are because we consider sign:magnitude. Int4 is one bit sign, 3 bits magnitude. Use that twice in long multiplication and we get 1 sign and 2*3 magnitude = 7 bits.

ljaljushkin:
Resolved by introducing liberal and strict modes for assigning bitwidth for activation quantizaer #222 and #191

PR #191, opened 2020-10-13
Consistent bit-width for activations and weight in propagation mode

### LINKED ISSUE #144: Mixed Precision missquantization
While training object detection model through hawq config, I realized there are much more int8 activation quantizations then int8 weight quantizations. According to [netron](https://lutzroeder.github.io/netron/) some convolutions take int8 activations and int4 weights. I suppose that's not how things should be. What do you think?
image

ljaljushkin:
Fixes #144 

Example of a consistent grouping of weight and activation quantizers for Inception-v3 (quantizers circled yellow are from the same group and should have the same bit-width): 

inception

@SKholkin, feel free to review bit-width graphs for other models from `tests\data\reference_graphs\quantized\hawq` folder
@RikAllen, next step is to introduce `allow_different_bitwidth_for_weight_and_activation` (#164) option to HW Config for independent handling bit-width inside group in HAWQ algorithm. 
@vuiseng9, this grouping is implemented for propagation-based mode only:
```json
'quantizer_setup_type': 'propagation_based'
```


ljaljushkin:
Jenkins please retry a build

asenina (approved):

nncf/nncf_network.py
```diff
@@ -74,12 +74,14 @@ class InsertionInfo:
     def __init__(self, op_exec_context: OperationExecutionContext,
                  is_input=False,
                  is_output=False,
-                 shape_to_operate_on=None):
+                 shape_to_operate_on=None,
+                 adjacent_quantizers_locations: 'AdjacentQuantizersLocations' = None):
```
vshampor, line 78: I am against this kind of extension. The new field is quantizer algo-specific, while the rest of the insertion info is more or less compression algorithm-agnostic. Either subclass the `InsertionInfo` struct, or use another means of passing the info between the propagation stage and the quantizer setup stage.

The naming should also be improved. It is not immediately obvious what kind of "adjacency" is meant in the name.

vshampor (changes requested):

nncf/nncf_network.py
ljaljushkin: Subclassed and renamed

nncf/quantization/quantizer_propagation.py
```diff
@@ -214,8 +232,10 @@ def __init__(self, ip_graph: InsertionPointGraph, ignored_scopes=None):
                     self.QUANTIZATION_TRAIT_NODE_ATTR] = QuantizationTrait.NON_QUANTIZABLE
                 qpg_node[self.AFFECTING_PROPAGATING_QUANTIZERS_ATTR] = []
                 qpg_node[self.OPERATOR_METATYPE_NODE_ATTR] = node[InsertionPointGraph.OPERATOR_METATYPE_NODE_ATTR]
-                scope_node = str(node[InsertionPointGraph.REGULAR_NODE_REF_NODE_ATTR][
-                    NNCFGraph.OP_EXEC_CONTEXT_NODE_ATTR].input_agnostic)
+                iap_context = node[InsertionPointGraph.REGULAR_NODE_REF_NODE_ATTR][
+                    NNCFGraph.OP_EXEC_CONTEXT_NODE_ATTR].input_agnostic
+                qpg_node[self.OPERATOR_SCOPE] = iap_context.scope_in_model
+                scope_node = str(iap_context)
```
vshampor, line 238: `node_scope`

vshampor (changes requested):

nncf/quantization/quantizer_propagation.py
ljaljushkin: any other comments to make it worth re-running pre-commit validation?
I can fix it in my next PR's, it looks like a very minor change

vshampor: Not yet, but there will be once I compare this PR and #168; let's wait until then

ljaljushkin: @vshampor any update since then?

PR #194, opened 2020-10-14
Introduced `enabled` parameter to Quantizers

ljaljushkin:
This is required to correctly export to ONNX a checkpoint from the staged quantization regardless of the specified config and scheduler.

### DIFF
--- a/nncf/quantization/algo.py
+++ b/nncf/quantization/algo.py
@@ -860,6 +860,11 @@ class QuantizationController(QuantizationControllerBase):
             self.stable_metric_collectors = [ShareEdgesQuantizedDataPath(target_model)]
             self.update_metric_store(True)
 
+    def prepare_for_export(self):
+        for quantizer_id, quantizer in self.all_quantizations.items():
+            if not quantizer.is_enabled_quantization():
+                nncf_logger.warning('Disabled quantization on export to ONNX: {}'.format(quantizer_id))
+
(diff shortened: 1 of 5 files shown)


vshampor:
Could we perhaps extend our model loading functions that we use inside `create_compressed_model` to notify the user of how many quantizers (or disable-able NNCF modules in general) are actually enabled or disabled as a result of loading the checkpoint? Want to avoid a situation when a user loads a checkpoint with all quantizers disabled, but thinks that they are enabled.

Should also make sure that the disabled quantizers don't end up in the exported ONNX model.

ljaljushkin:
> Could we perhaps extend our model loading functions that we use inside `create_compressed_model` to notify the user of how many quantizers (or disable-able NNCF modules in general) are actually enabled or disabled as a result of loading the checkpoint? Want to avoid a situation when a user loads a checkpoint with all quantizers disabled, but thinks that they are enabled.
> 
> Should also make sure that the disabled quantizers don't end up in the exported ONNX model.

Added warning on exporting with disabled quantizations.
The number of disabled quantizers is printed from statistics: `ratio_of_enabled_quantizations`. The user should know what he loads. IMHO, no need to over-complicate loading functionality, statistics should be enough for reporting.

ljaljushkin:
started 285th build of nightly eval

ljaljushkin:
nightly eval tests are OK
build 289

vshampor (approved):

PR #218, opened 2020-10-25
Add support sparsity freeze epoch for magnitude sparsity.

asenina:


nncf/sparsity/layers.py
```diff
@@ -24,6 +24,7 @@ class BinaryMask(nn.Module):
     def __init__(self, size):
         super().__init__()
         self.register_buffer("_binary_mask", torch.ones(size))
+        self.sparsify = True
```
vshampor, line 27: Rename to `frozen` and update the logic accordingly

vshampor (changes requested):

nncf/sparsity/layers.py
asenina: @vshampor  Ok, but then class `RBSparsifyingWeight` will have two attributes `sparsify` and `frozen`. It looks redundant.

vshampor: Guess you'll have to rename the `RBSparsifytinWeight`'s `sparsify` attribute to `frozen` as well.

asenina: Done.

vshampor (approved):

PR #225, opened 2020-10-28
Change quantization levels for SymmetricQuantizer from 255 to 256

aleksei-kashapov:
Quantization levels are changed for weights only as activations are already use full range of 8 bits

### DIFF
--- a/nncf/quantization/layers.py
+++ b/nncf/quantization/layers.py
@@ -257,8 +257,6 @@ class SymmetricQuantizer(BaseQuantizer):
             level_high = 2 ** num_bits - 1
             level_low = 0
         levels = 2 ** num_bits
-        if is_weights:
-            levels -= 1
         return level_high, level_low, levels
 
     @property
--- a/nncf/quantization/layers.py
+++ b/nncf/quantization/layers.py
@@ -242,7 +242,7 @@ class SymmetricQuantizer(BaseQuantizer):
         self.scale.requires_grad = False
 
     def set_level_ranges(self):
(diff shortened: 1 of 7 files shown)


nncf/quantization/layers.py
```diff
@@ -257,8 +257,6 @@ def calculate_level_ranges(num_bits, signed, is_weights):
             level_high = 2 ** num_bits - 1
             level_low = 0
         levels = 2 ** num_bits
-        if is_weights:
```
ljaljushkin, line 260: It's strange that pre-commit is OK, while reference implementation in tests still has 255 levels. 

At least it should be also corrected in `test_functions.py`

aleksei-kashapov: I suppose the tests were passed as in **test_functions.py** we utilize the same levels-255 for reference and symmetric_quantize.

AlexKoff88:
@aleksei-kashapov, please report the status of E2E validation here as well.

tests/quantization/test_functions.py
```diff
@@ -182,19 +182,15 @@ def calc_scale(input_):
                 return scales
 
         @staticmethod
-        def get_range_level(is_signed, is_weights, bits):
```
AlexKoff88, line 185: We will need this parameter when we introduce changes for the second part of the task that assumes having a workaround for the saturation issue.

vshampor (changes requested):

vshampor (approved):

PR #222, opened 2020-10-28
Liberal bitwidth assignment mode by default on precision initialization

ljaljushkin:
should be merged after #191 and #232

PR #230, opened 2020-10-29
Criterion generalization for HAWQ algorithm

ljaljushkin:
should be merged after #222, #191 and #232


vshampor (changes requested):
More in-code documentation is required to explain why we now need `criterion_fn` at the same time with the `criterion` and what value should the user set for it if they want to achieve their goals.

ljaljushkin:
> ould the user set for it if they want to achieve their go

done

vshampor (approved):

PR #241, opened 2020-11-02
fix quantization range initialization in case of 1 scale channel to a…

krodyush:
…void initialization only by single slice of data (data[0]) and ignoring the other (data[1], data[2],.....)

fix quantization range initialization in case of 1 scale channel to avoid initialization only by single slice of data (data[0]) and ignoring the other (data[1], data[2],.....)


vshampor:
Jenkins please retry a build

vshampor (approved):

PR #242, opened 2020-11-02
Extend input handling flexibility

vshampor:


vshampor:
Jenkins please retry a build

vshampor:
@alexsu52 @AlexKoff88

nncf/dynamic_graph/input_wrapping.py
```diff
@@ -0,0 +1,121 @@
...
+
+        model_args = tuple(args_list)
+    else:
+        # Fallback - ideally the input_infos must be supplied
+        model_args = objwalk(model_args, is_tensor, nncf_model_input)
+        model_kwargs = objwalk(model_kwargs, is_tensor, nncf_model_input)
+    return model_args, model_kwargs
+
+
+def wrap_nncf_model_inputs_with_objwalk(model_args, model_kwargs):
```
AlexKoff88, line 46: Is this used anywhere? Or expected to be used?

nncf/model_creation.py
```diff
@@ -86,6 +87,16 @@ def create_compressed_model(model: Module, config: NNCFConfig,
     has special formats of data loader output or has additional *forward* arguments other than input tensors.
     Otherwise, the *forward* call of the model during graph tracing will be made with mock tensors according
     to the shape specified in the config object.
+    :param wrap_inputs_fn: will be used on the module's input arguments during a regular, non-dummy forward call before
```
AlexKoff88, line 90: What's the point to use the future tense in the description of the function? New trend?

AlexKoff88 (commented):
I'm mostly ok with the changes because they only extend the existing API. The only concern I have is that we don't demonstrate how to use this wrapper (only in tests).

nncf/model_creation.py
vshampor: Edited wording to justify the future tense usage

nncf/dynamic_graph/input_wrapping.py
vshampor: May be useful in the future, IMO.

ljaljushkin: BTW, it can be called in the `else` branch of `wrap_nncf_model_inputs_helper`

tests/data/reference_graphs/const_sparsity/sr_small_model.dot
```diff
@@ -50,41 +50,6 @@ strict digraph  {
 "48 SmallModel/NNCFConv2d[conv_output]/conv2d" [id=48, scope="SmallModel/NNCFConv2d[conv_output]", type=conv2d];
 "49 SmallModel/__mul__" [id=49, scope=SmallModel, type=__mul__];
 "50 SmallModel/add" [id=50, scope=SmallModel, type=add];
-"51 SmallModel/NNCFConv2d[conv_cubic1]/conv2d" [id=51, scope="SmallModel/NNCFConv2d[conv_cubic1]", type=conv2d];
```
ljaljushkin, line 53: strange, that there's the deletion of nodes only

nncf/dynamic_graph/input_wrapping.py
```diff
@@ -0,0 +1,121 @@
...
+@register_operator()
+def nncf_model_input(tensor: 'torch.Tensor'):
+    return tensor
+
+
+def wrap_nncf_model_inputs_helper(model_args, model_kwargs, input_infos: List['ModelInputInfo'] = None,
+                                  force_objwalk=False):
+    """Will wrap the input tensors among the model_args and model_kwargs with a no-op required for
+    correct input tracing by NNCF. If input_infos is supplied and force_objwalk is False, will only
+    affect the tensor args/kwargs as specified in the input_infos (usually populated via a corresponding
```
ljaljushkin, line 22: would be great to test this scenario. As far as I can see, there're tests with direct wrapping all inputs inside `wrap_inputs_fn` (GNMT, SR Small model), but I can't find checking the logic with input_infos.

ljaljushkin (changes requested):

tests/data/reference_graphs/const_sparsity/sr_small_model.dot
vshampor: That's because the reference graph prior to this change actually contained duplicate nodes of the majority of the model - which is fixed by this PR

nncf/dynamic_graph/input_wrapping.py
vshampor: This scenario is actually not required any more - replaced this call with an objwalk-based wrapping in the default dummy forward creation. The branch without `wrap_inputs_fn` in `create_dummy_forward` won't be executed normally when used in `NNCFNetwork` - `NNCFNetwork` always sets the `wrap_inputs_fn` to what it uses in its own forward to ensure consistency.

nncf/dynamic_graph/input_wrapping.py
vshampor: Refactored to use this in `default_dummy_forward` if no `wrap_inputs_fn` is supplied.

nncf/dynamic_graph/input_wrapping.py
ljaljushkin: but there's still a branch with info for missing inputs
https://github.com/openvinotoolkit/nncf/pull/242/files#diff-b2eabe2503ddc6713cecbeb5ff0740babbf6ad0bf3aac57e80f3f36d98a79a14R76

nncf/model_creation.py
```diff
@@ -81,11 +82,21 @@ def create_compressed_model(model: Module, config: NNCFConfig,
...
     :param resuming_state_dict: A PyTorch state dict object to load (strictly) into the compressed model after
     building.
-    :param dummy_forward_fn: will be used instead of a *forward* function call to build
+    :param dummy_forward_fn: if supplied, will be used instead of a *forward* function call to build
     the internal graph representation via tracing. Specifying this is useful when the original training pipeline
     has special formats of data loader output or has additional *forward* arguments other than input tensors.
     Otherwise, the *forward* call of the model during graph tracing will be made with mock tensors according
     to the shape specified in the config object.
+    :param wrap_inputs_fn: if supplied, will be used on the module's input arguments during a regular, non-dummy
+    forward call before passing the inputs to the  underlying compressed model. This is required if the model's input
```
alexsu52, line 91:
```suggestion
    forward call before passing the inputs to the underlying compressed model. This is required if the model's input
```

alexsu52 (approved):
it looks good to me

nncf/dynamic_graph/input_wrapping.py
vshampor: Added tests for wrapping sequence for a given forward function signature, actual call and input infos.

ljaljushkin (approved):

PR #248, opened 2020-11-05
Fix bug with export to Q/DQ

aleksei-kashapov:
Add hack of export processing for our old checkpoints
Add Exception raising for exporting per-channel Q/DQ layers, as PyTorch ONNX exporting supports only per-tensor.


FyodorK:
Jenkins please retry a build

vshampor (approved):

PR #172, opened 2020-10-07
Compress model in eval model

ljaljushkin:
* It helps to avoid issues with BN statistics. No need to care about the corruption of the tracking statistics on model creation and initialization. The model can be inferred with any dummy inputs without accuracy degradation.
* Training time can be reduced by compressing the inference graph without train-specific modules (e.g. AuxOutputs in Inception-v3)

PR #257, opened 2020-11-06
Compress model in eval mode

### LINKED ISSUE #177: NMS CUDA kernel fails when it's running on multiple processes and on different GPUs.
NMS CUDA kernel fails when it's running on multiple processes and different GPU (even without wrapping by DistributedDataParallel and dist.init_process_group)
```
RuntimeError: cuda runtime error (700) : an illegal memory access was encountered at line:
```
```
THCudaCheck(cudaMemcpy(&mask_host[0],
                    mask_dev,
                    sizeof(unsigned long long) * boxes_num * col_blocks,
                    cudaMemcpyDeviceToHost));
```
The same error occurs in multi process DistributedDataParallel mode on multiple GPU when the kernel is running after dist.init_process_group and before wrapping by DistributedDataParallel.

It's OK in a single GPU mode and when it's running from a single process on multiple GPU in DataParallel mode.

The workaround is to call the kernel after wrapping by DistributedDataParallel. But this kernel can be called on the creation of the compressed model which can happen before the wrapping by DistributedDataParallel only. This is where this issue comes from. I wanted to run create_compressed_model for SSD_VGG model in evaluation mode. This mode calls NMS and fails with the mentioned error. 

Compression models in the evaluation may reduce training time by not quantizing auxiliary training branches and prevent errors with corrupting BatchNorm statistics on calling dummy_forward with random inputs for the model in training mode.

@alexsu52 @vshampor @vanyalzr @AlexKoff88

ljaljushkin:
- It helps to avoid issues with BN statistics. No need to care about the corruption of the tracking statistics on model creation and initialization.
The model can be inferred with any dummy inputs without accuracy degradation.
- Training time can be reduced by compressing the inference graph without train-specific modules (e.g. AuxOutputs in Inception-v3)

Also this PR

- fixes #177 
- removes the unnecessary test dot-files

P.S. CI still doesn't work as in the original #172


FyodorK:
Jenkins please retry a build

FyodorK:
Jenkins please retry a build

nncf/nncf_network.py
```diff
@@ -371,6 +371,18 @@ def __init__(self, module, input_infos: List[ModelInputInfo] = None,
...
+        self._graph_builder = GraphBuilder(_orig_graph_build_forward_fn)
+
+        nncf_wrapped_model = self.get_nncf_wrapped_model()
+        train_only_module_scopes = self.collect_train_only_module_scopes(nncf_wrapped_model, self._graph_builder)
+        if not self.ignored_scopes:
+            self.ignored_scopes = []
+        # No need to compress modules that are executed in train mode only, thus adding them to ignored_scopes to avoid
+        # replacing by nncf modules. As a consequence, model can be compressed in eval mode. No need to care about
+        # spoiling BN statistics, as there're disabled in eval mode.
+        self.ignored_scopes.extend(train_only_module_scopes)
```
vshampor, line 384: I think that the `ignored_scopes` attribute, especially if it is public, should be always consistent with what is specified in the config. Perhaps a separate roster of train-only modules should be kept and only concatenated with user's `ignored_scopes` where it is actually necessary to do so, or a separate `_effective_ignored_scopes` attribute should be set up (populated with entries based on `ignored_scopes` and the train module marking) and used internally.

nncf/nncf_network.py
```diff
@@ -747,3 +759,26 @@ def get_MACs_in_model(self):
 
     def get_input_infos(self) -> List[ModelInputInfo]:
         return deepcopy(self.input_infos)
+
+    @staticmethod
+    def collect_train_only_module_scopes(model, graph_builder) -> List[str]:
```
vshampor, line 764: You are iterating over nodes - these do not necessarily represent entire module calls. Consider the situation:
```
*some module's forward*
x = self.op_1(x)
if self.training:
    y = self.train_only_op(x)
x = self.op_2(x)
```
Your code, it seems, will effectively remove `op_1` and `op_2` calls from the compression consideration as well, which may be incorrect.

vshampor (changes requested):

nncf/nncf_network.py
ljaljushkin: Processed separately from `ignored_scope` by `eval_ops_exec_ctx_str`

nncf/nncf_network.py
ljaljushkin: Corrected logic by processing operation execution context.  A module is wrapped, if they're at least a single operation or a module called in the eval mode inside it.

nncf/nncf_network.py
```diff
@@ -747,3 +749,18 @@ def get_MACs_in_model(self):
 
     def get_input_infos(self) -> List[ModelInputInfo]:
         return deepcopy(self.input_infos)
+
+    @staticmethod
+    def collect_eval_only_ops_exec_context(model: nn.Module, graph_builder) -> List[str]:
+        """
+        Returns scopes of the modules which are executed in evaluation mode only.
+        Model is set to eval mode.
```
vshampor, line 757: Should avoid such side effects in the future and keep the model in the state it arrived in.

vshampor (approved):

Issue #131, opened 2020-09-09
Variability in SSD Mixed-Precision Performance


vuiseng9:
We observed large variability in performance with SSD300(VGG) when we tried different combination of precision for weight and activation in test mode. From the collected number below, the best and worst are about 15X gap, inference per batch (size: 128) is 30secs for Int2 weights and Int8 activation for the worst case. The performance should impact fine-tuning mode as well.

NNCF Version: Develop branch with commit 2a681b897f992a898dca463b541ee12d9c9899fe
Similar observations with v1.4
Baseline config: https://github.com/openvinotoolkit/nncf_pytorch/blob/develop/examples/object_detection/configs/ssd300_vgg_voc_int8.json
Platform: V100 GPU

|     weights    |     activations    |     detection elapse                                                                                  |
|----------------|--------------------|-------------------------------------------------------------------------------------------------------|
|     8          |     8              |     Detect for batch: 8/39 **1.847s** Detect for batch: 9/39 **1.844s** Detect for batch: 10/39 **1.876s**        |
|     8          |     4              |     Detect for batch: 8/39 **8.792s** Detect for batch: 9/39 **9.021s** Detect for batch: 10/39 **8.928s**        |
|     8          |     2              |     Detect for batch: 8/39 **17.09s** Detect for batch: 9/39 **17.80s** Detect for batch: 10/39 **17.87s**     |
|     4          |     8              |     Detect for batch: 8/39 **2.283s** Detect for batch: 9/39 **2.105s** Detect for batch: 10/39 **2.296s**        |
|     4          |     4              |     Detect for batch: 8/39 **8.285s** Detect for batch: 9/39 **9.583s** Detect for batch: 10/39 **7.425s**        |
|     4          |     2              |     Detect for batch: 8/39 **11.31s** Detect for batch: 9/39 **11.85s** Detect for batch: 10/39 **12.61s**    |
|     2          |     8              |     Detect for batch: 8/39 **29.40s** Detect for batch: 9/39 **30.75s** Detect for batch: 10/39 **29.30s**     |
|     2          |     4              |     Detect for batch: 8/39 **5.684s** Detect for batch: 9/39 **5.703s** Detect for batch: 10/39 **5.539s**        |
|     2          |     2              |     Detect for batch: 8/39 **5.954s** Detect for batch: 9/39 **6.040s** Detect for batch: 10/39 **6.159s**        |

vshampor:
[...]

PR #270, opened 2020-11-12
Introduce evaluation option of loss function to SSD application

vuiseng9:
Backward compatible changes. Introduce a config key "loss_inference" to specify test with loss evaluation or the default mAP evaluation. It is disabled by default. This patch support issue #131

### DIFF
--- a/examples/object_detection/configs/ssd300_vgg_voc_int8.json
+++ b/examples/object_detection/configs/ssd300_vgg_voc_int8.json
@@ -34,7 +34,8 @@
     "steps": [8, 16, 32, 64, 100, 300],
     "aspect_ratios": [[2], [2, 3], [2, 3], [2, 3], [2], [2]],
     "flip": true,
-    "top_k": 200
+    "top_k": 200,
+    "loss_inference": false
   },
   "compression": {
       "algorithm": "quantization",
--- a/examples/object_detection/main.py
+++ b/examples/object_detection/main.py
@@ -183,7 +183,7 @@ def main_worker(current_gpu, config):
             net.eval()
             if config['ssd_params'].get('loss_inference', False):
                 model_loss = test_net(net, config.device, test_data_loader, distributed=config.distributed,
-                                        loss_inference=True, criterion=criterion)
+                                      loss_inference=True, criterion=criterion)
                 logger.info("Final model loss: {:.3f}".format(model_loss))
             else:
                 mAp = test_net(net, config.device, test_data_loader, distributed=config.distributed)
--- a/examples/object_detection/main.py
+++ b/examples/object_detection/main.py
@@ -181,9 +181,14 @@ def main_worker(current_gpu, config):
         with torch.no_grad():
             print_statistics(compression_ctrl.statistics())
             net.eval()
-            mAp = test_net(net, config.device, test_data_loader, distributed=config.distributed)
-            if config.metrics_dump is not None:
-                write_metrics(mAp, config.metrics_dump)
+            if config['ssd_params'].get('loss_inference', False):
+                model_loss = test_net(net, config.device, test_data_loader, distributed=config.distributed,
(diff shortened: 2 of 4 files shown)


vshampor:
Jenkins please retry a build

vshampor:
@vuiseng9 please fix pylint and then we'll merge this.

vuiseng9:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

examples/object_detection/configs/ssd300_vgg_voc_int8.json
```diff
@@ -34,7 +34,8 @@
     "steps": [8, 16, 32, 64, 100, 300],
     "aspect_ratios": [[2], [2, 3], [2, 3], [2, 3], [2], [2]],
     "flip": true,
-    "top_k": 200
+    "top_k": 200,
+    "loss_inference": false
```
vshampor, line 38: We actually use this config file for checkpoint validation, so we require the proper metric calculation.

vshampor (changes requested):

examples/object_detection/configs/ssd300_vgg_voc_int8.json
vuiseng9: @vshampor When the key is enabled, the training loss is evaluated instead of the mean AP. Disabling the key has the exact behavior as previous implementation. Can you elaborate the metric calculation you refer to?

ljaljushkin: I guess @vshampor meant [this](https://github.com/openvinotoolkit/nncf/blob/develop/examples/object_detection/main.py#L189) metric calculation.
It's used in the nightly checkpoint validation defined in test_sota_checkpoints.py. 

But it shouldn't be affected, as you say: the introduced option keeps the previous behavior with calculation actual mAP metric.

examples/object_detection/main.py
```diff
@@ -181,9 +181,14 @@ def criterion_fn(model_outputs, target, criterion):
...
-            if config.metrics_dump is not None:
-                write_metrics(mAp, config.metrics_dump)
+            if config['ssd_params'].get('loss_inference', False):
+                model_loss = test_net(net, config.device, test_data_loader, distributed=config.distributed,
+                                      loss_inference=True, criterion=criterion)
+                logger.info("Final model loss: {:.3f}".format(model_loss))
+            else:
+                mAp = test_net(net, config.device, test_data_loader, distributed=config.distributed)
+                if config.metrics_dump is not None:
+                    write_metrics(mAp, config.metrics_dump)
```
vshampor, line 191: I think that we won't fall into this branch anymore with the config you've changed. We need that config to execute this code branch so that the example script, as run from within the CI validation run, produces an output file with metrics which is then parsed for a test report.

If you modified the config file to give an example of the option's usage, I suggest that you just create another config file in addition to the original one.

vuiseng9 (commented):
```config['ssd_params'].get('loss_inference', False)```. When the key is not provided, the app would go as default to mean AP evaluation block. so, the original config and CI validation would still works. does this make sense? or am I disconnected at some points?

examples/object_detection/main.py
vuiseng9: ```config['ssd_params'].get('loss_inference', False)```. When the key is not provided, the app would go as default to mean AP evaluation block. so, the original config and CI validation would still works. does this make sense? or am I disconnected at some points?

vuiseng9: The config modification is not meant to be as sample, just laying out the key for reference, ie setting it to false so it is backward compatible. Skipping the key is also handled in the main app.

vshampor (approved):

PR #274, opened 2020-11-13
Pruning of ConvTranspose

AlexanderDokuchaev:


nncf/layer_utils.py
```diff
@@ -14,6 +14,8 @@ def __getattr__(self, name):
 
 
 class _NNCFModuleMixin:
+    target_compression_weight_dim = 0
```
vshampor, line 17: `target_weight_dim_for_compression`

vshampor: Also, I expect you to utilize the same mechanism for correct `ConvTranspose` per-channel quantization, perhaps in a separate PR.

vshampor (changes requested):

vshampor:
@mkaglins

nncf/layer_utils.py
AlexanderDokuchaev: Renamed.
Yes, i will do it in another PR.

nncf/pruning/base_algo.py
```diff
@@ -233,7 +233,7 @@ def pruning_rate_for_filters(minfo: PrunedModuleInfo):
         Calculates sparsity rate for weight filter-wise.
         """
         weight = minfo.module.weight
-        filters_sum = weight.view(weight.size(0), -1).sum(axis=1)
+        filters_sum = weight.view(weight.size(minfo.module.target_weight_dim_for_compression), -1).sum(axis=1)
```
mkaglins, line 236: Here you should transpose weight (in case of target_weight_dim_for_compression != 0) first.
Instead, you will mix filters by reshaping tensor and calculate incorrect filter-wise pruning rate.

nncf/pruning/export_helpers.py
```diff
@@ -169,6 +170,68 @@ def output_prune(self, model: NNCFNetwork, nx_node, graph: NNCFGraph, nx_graph:
...
+        output_mask = nx_node['output_mask']
+        if output_mask is None:
+            return
+
+        bool_mask = torch.tensor(output_mask, dtype=torch.bool)
+        new_num_channels = int(torch.sum(bool_mask))
+
+        nncf_node = graph._nx_node_to_nncf_node(nx_node)
+        node_module = model.get_module_by_scope(nncf_node.op_exec_context.scope_in_model)
+        old_num_clannels = int(node_module.weight.size(0))
```
mkaglins, line 218: Old num channels are the number of output channels, for ConvTrabspose it should be .size(1)

nncf/pruning/filter_pruning/algo.py
```diff
@@ -128,8 +131,9 @@ def _apply_masks(self):
 
         def _apply_binary_mask_to_module_weight_and_bias(module, mask, module_name=""):
             with torch.no_grad():
+                dim = module.target_weight_dim_for_compression if isinstance(module, _NNCFModuleMixin) else 0
```
mkaglins, line 134: Why do you need here this check?
Not all pruned ops are subclasses of `_NNCFModuleMixin`?

AlexanderDokuchaev: Yes, it's need for BatchNorm.

AlexanderDokuchaev:
Jenkins please retry a build

nncf/pruning/base_algo.py
AlexanderDokuchaev: fixed

nncf/layers.py
```diff
@@ -95,6 +104,9 @@ def from_module(module):
 
 
 class NNCFConvTranspose3d(_NNCFModuleMixin, nn.ConvTranspose3d):
+    op_func_name = "conv_transpose3d"
```
vshampor, line 107: If op_func_name is a required attribute for all NNCF-wrapped modules now, it should be located in a base class. It should also be documented in code there.

vshampor (changes requested):

mkaglins (approved):

vshampor (approved):

PR #277, opened 2020-11-17
Add pass_meta_from_tensor to trace torch non in-place operations

aleksei-kashapov:
Currently there is no option to trace tensors if the operations on them return a new tensor, non in-place.
Examples of such operations:` torch.tensor.exp(), torch.tensor.view()`

I add a new method and give an example of how to overload a `torch.tensor.view()` method to be able to still trace such operations

### DIFF
--- a/nncf/dynamic_graph/trace_tensor.py
+++ b/nncf/dynamic_graph/trace_tensor.py
@@ -51,6 +51,9 @@ class TracedTensor(torch.Tensor):
         tensor.__class__ = TracedTensor
         return tensor
 
+    def view(self, size):
+        tensor = super().view(size)
+        return pass_meta_from_tensor(tensor, self)
 
 def is_iterable(item):
     non_iterable_types = (str, bytes, bytearray, torch.Tensor, np.ndarray)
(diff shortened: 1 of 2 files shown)


nncf/dynamic_graph/trace_tensor.py
```diff
@@ -95,3 +98,9 @@ def make_input_infos(inputs):
         else:
             input_infos.append(None)
     return input_infos
+
+
+def pass_meta_from_tensor(to_tensor, from_tensor):
+    if not isinstance(from_tensor, TracedTensor):
+        return to_tensor
+    return TracedTensor.from_torch_tensor(to_tensor, from_tensor.tensor_meta)
```
ljaljushkin, line 106: This should happen outside of TracedTensor [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/dynamic_graph/wrappers.py#L64), no need to add it to TracedTensor explicitly.

ljaljushkin (changes requested):
`view` should be wrapped by `ForwardTraceOnly`. I'd suggest debugging this class to root-cause the issue.

vshampor:
@aleksei-kashapov I'm pretty sure that this is handled in NNCF using these kinds of mechanisms:
https://github.com/openvinotoolkit/nncf/blob/3e81fe39bed412211d11f9dfca8b5f961ba315eb/nncf/dynamic_graph/operator_metatypes.py#L482

PR #288, opened 2020-11-21
Proper BN momentum parameter and train mode setting in BN adaptation

vanyalzr:
Fix BN adaptation due to #257

### DIFF
--- a/nncf/initialization.py
+++ b/nncf/initialization.py
@@ -11,7 +11,7 @@ from nncf.nncf_logger import logger as nncf_logger
 from nncf.quantization.init_range import MinMaxInitializer, ThreeSigmaInitializer, MeanMinMaxInitializer
 from nncf.quantization.init_range import PercentileInitializer
 from nncf.structures import QuantizationPrecisionInitArgs, QuantizationRangeInitArgs, BNAdaptationInitArgs
-from nncf.utils import objwalk, is_tensor
+from nncf.utils import objwalk, is_tensor, training_mode_switcher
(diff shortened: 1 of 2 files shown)


nncf/initialization.py
```diff
@@ -210,6 +224,8 @@ def set_bn_momentum(module, momentum_value):
             args_kwargs_tuple = data_loader.get_inputs(loaded_item)
             self._infer_batch(args_kwargs_tuple, device)
 
+        self.model.eval()
```
ljaljushkin, line 227: Not sure we should always set the model to eval mode. It can be not obvious for the users of this init functionality.

I'd add a context manager or decorator for secure safe switching modes.

The idea is always to keep the original state of the model with activating the chosen mode inside a function's body. 
```
def _run_model_inference
    # doesn't matter what is a current training mode of the model, it would be returned to the original state outside of the context manager
    with training_mode_switcher(model, is_training=True) as model:
         # work with the model in a training mode
         model.infer()
    # outside of context manager model is in the original sate
    model.infer()
```

That would be helpful in multiple places in NNCF.

vshampor:
Jenkins please retry a build

nncf/initialization.py
vshampor: @vanyalzr are you going to address this?

vanyalzr: Sure, I agree with the comment, will address

ljaljushkin (approved):

PR #291, opened 2020-11-23
Fix H/W on an image in the mock coco dataset

vshampor:


vshampor:
Changes are in non-precommit-executable files, precommit failures have already been fixed in #292

PR #294, opened 2020-11-24
Set proper workdir path for Mask-RCNN

vshampor:


vshampor:
Changes are in non-precommit-affected code only, aborted the corresponding runs.

PR #296, opened 2020-11-24
Add input/output-names parameters to export_model function.

asenina:


nncf/compression_method_api.py
```diff
@@ -231,7 +234,9 @@ def export_model(self, filename, *args, **kwargs):
         # pylint:disable=unexpected-keyword-arg
         with torch.no_grad():
             torch.onnx.export(model, tuple(input_tensor_list),
-                              filename, verbose=True, enable_onnx_checker=False, opset_version=10)
+                              filename, input_names=input_names,
```
AlexKoff88, line 237: Can we also strip the debug information from the operators' description? I noticed that ONNX files contain stack trace in that description.

We also have an agreement to introduce a domain for custom operators: "org.openvinotoolkit". It can be easily done by introducing a `custom_opsets={"org.openvinotoolki": 1}`. But I don't mind doing that in the separate PR because we need to validate it as well, e.g. inspecting exported ONNX file.

asenina: Removed `verbose=True`.

Ok, we will add `custom_opsets` option in another PR.

vshampor (approved):

PR #299, opened 2020-11-25
Add quantizer merging logic to the propagation mode

vshampor:


vshampor:
Moved to a combined PR #302

PR #302, opened 2020-11-26
Merging and pre hooks

vshampor:


vshampor:
This PR does two things:
1) Allows input tensors for NNCF-wrapped PyTorch functions to be pre-processed separately, e.g. introduces the "pre-hook" functionality; previously only the function output tensors could be post-processed.
2) Introduces a new algorithm for the merging of the propagating quantizers that have reached one and the same downward-branching node.

Will add more commentaries in-place.

nncf/dynamic_graph/context.py
```diff
@@ -13,24 +13,130 @@
...
+
+    def __call__(self, value):
+        tmp_list = list(self._current_tuple)
+        tmp_list[self._idx_to_set] = value
+        new_tuple = tuple(tmp_list)
+        self._current_tuple = new_tuple
+        self._previous_level_setter(new_tuple)
+
+
 class OperatorInput:
```
vshampor, line 57: A PyTorch function input is not necessarily a single raw `torch.Tensor` for each input argument of the function. Consider `torch.cat([x1, x2])`, which has the input tensors `x1` and `x2` located inside a container, which itself is being passed as an argument. This function call, however, conceptually takes two tensors as its input, and if this is not traced properly, the NNCF graph will become disjoint. In a previous implementation this was circumvented by flattening the node inputs - a node's input signature, which identifies it among the other NNCF nodes, is a flat list of all tensors in the input arguments nesting hierarchy. 

The same approach is taken here - a flat view of the function's input tensors is being constructed, and each tensor is conceptually assigned a separate "input port ID" ordinal. However, this flat view must be a *view*, i.e. it should allow the tensors to be replaced with the results of pre-processing while preserving the original nested structure. Sadly, Python has no real notion of writeable iterators; therefore, it is necessary for each tensor in a flat view to store and use associated getter and setter functions, which are built using the closest owning container's `__getitem__` and `__setitem__` and the associated index of the tensor in the said container.

nncf/dynamic_graph/graph.py
```diff
@@ -234,7 +234,7 @@ def add_node(self, op_exec_context: OperationExecutionContext, inputs) -> NNCFNo
             self._nx_graph.add_edge(parent, node_key)
             has_traced_inputs = True
             self._nx_graph.edges[parent, node_key][NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR] = info.shape
-            self._nx_graph.edges[parent, node_key][NNCFGraph.IN_PORT_NAME] = i
+            self._nx_graph.edges[parent, node_key][NNCFGraph.IN_PORT_NAME_EDGE_ATTR] = i
```
vshampor, line 237: Since PyTorch/Python combo is essentially dynamic, there is no clear definition of how many inputs (in terms of NNCF graph nodes) does a given function function take (see `torch.cat` example above). The function's "input ports" therefore are not something static; rather, these are enumerated according to the number of traced tensors arriving as the function (node) input.

nncf/nncf_network.py
```diff
@@ -74,34 +74,49 @@ def __eq__(self, other):
...
-        return self.op_exec_context.__hash__()
+        return hash(str(self))
 
 
 class InsertionPoint:
     def __init__(self, ia_op_exec_context: InputAgnosticOperationExecutionContext,
-                 insertion_type: InsertionType):
+                 insertion_type: InsertionType,
+                 input_port_id: int = None):
         self.ia_op_exec_context = ia_op_exec_context
```
vshampor, line 107: An insertion point now stores an input port to affect, in order to represent the pre-hook semantics as well. `None` specifies a post-hook insertion.

nncf/quantization/algo.py
```diff
@@ -112,10 +112,10 @@ class QuantizerGroup(Enum):
 
 class NonWeightQuantizerInfo:
     def __init__(self, quantizer_module_ref: BaseQuantizer,
-                 affected_ia_op_exec_contexts: List[InputAgnosticOperationExecutionContext],
+                 affected_insertions: List[QuantizerInsertionInfo],
                  quantizers_between_quantizable_layers: QuantizersBetweenQuantizableLayers = None):
         self.quantizer_module_ref = quantizer_module_ref
-        self.affected_ia_op_exec_contexts = affected_ia_op_exec_contexts
+        self.affected_insertions = affected_insertions
```
vshampor, line 118: The `InputAgnosticOperationExecutionContext` is an address of a *node*, but the insertion of a hook for a node can now proceed in two ways - either post- or pre-. Therefore a position of a quantizer is no longer defined by a node, but by a node and the insertion type.

nncf/dynamic_graph/context.py
```diff
@@ -235,19 +357,26 @@ def pop_scope(self):
...
-    def register_pre_hooks(self, fn_list: List[Callable], ia_op_exec_context: InputAgnosticOperationExecutionContext):
-        if ia_op_exec_context in self._pre_hooks:
-            raise KeyError("Pre hook for context {} is already registered".format(str(ia_op_exec_context)))
-        self._pre_hooks[ia_op_exec_context] = fn_list
+    def register_pre_hooks(self, fn_list: List[Callable], ia_op_exec_context: InputAgnosticOperationExecutionContext,
+                           input_port_id: int):
+        pre_hook_id = PreHookId(ia_op_exec_context, input_port_id)
+        if pre_hook_id in self._pre_hooks:
+            raise KeyError("Pre hook for context {} is already registered".format(str(pre_hook_id)))
+        self._pre_hooks[pre_hook_id] = fn_list
```
vshampor, line 365: 1) Pre-hooks is what will allow to quantize branches in the network separately.
Below, on the left, is the result of the MobileNetV2 quantization prior to this PR, and on the right - with this PR.
Screenshot from 2020-12-01 19-33-00

Notice that the `__add__` node used to have a single, nominal pre-hook insertion point, while in fact it takes two tensor activations. The pre-hook insertion points were also non-functional, i.e. the actual insertion could only occur at a post-hook point, which led to a situation where either both of the branches would be quantized with a single quantizer (if a FQ is inserted in the top-most post-hook insertion point), or none of them would. 

With the PR, the node has two pre-hook insertion points, corresponding to the number of inputs (i.e. "ports"). Furthermore, it is possible to separately quantize the `conv2d` input, using the `conv2d`'s pre-hook 0, and the residual `__add__` input, using the `__add__`'s pre-hook 0.

nncf/quantization/algo.py
```diff
@@ -416,9 +411,35 @@ def _quantize_activations(self, target_model: NNCFNetwork) -> List[InsertionComm
                     err_msg += str(ia_op_exec_context)
                     raise RuntimeError(err_msg)
 
+                final_quantizer_config_setup[insertion_info] = quantizer_config
+
+            finalized_proposal = quantization_proposal.finalize(final_quantizer_config_setup)
+            final_quantizer_setup = prop_graph_solver.get_final_quantizer_setup(finalized_proposal)
```
vshampor, line 417: 2) The quantizer propagation algorithm has been modified to work in two stages. This is related to the new quantizer configuration merging algorithm. Basically the first step of the propagation results in a proposal - a set of the points at which a quantizer should be inserted in order to quantize all operations specified in HW config with a supported configuration, and, for each point, a list of possible quantizer configurations out of which one and only one should be chosen. The choice could be either affected by utilizing the NNCF config options/presets, or it could be delegated to a specialized algorithm, such as HAWQ in case multiple bitwidths are present in the quantizer configuration list, or an RL algorithm. 

Once the final quantizer configurations are chosen, an extra step is necessary to merge the quantizers that are redundant - e.g. if one and the same final quantizer configuration has been chosen both for the global quantizer of all downstream branches and for the local quantizer on a branch.

nncf/quantization/quantizer_propagation.py
```diff
@@ -120,22 +124,21 @@ def __eq__(self, other):
...
 
 class PropagationStrategy(Enum):
     CONSERVATIVE = 0  # While propagating up through a downward-branching node,
-                      # do not propagate if the propagation results in narrowing the list of
-                      # quantization variants available to quantizers on neighbouring branches
-    AGGRESSIVE = 1
+                      # do not merge at all, or ...
+    MODERATE = 1      # ... only merge for exact matches
+    AGGRESSIVE = 2    # ... merge common parts, and if a branch quantizer has options for scope narrowing in addition to
+                      # the common part, keep the quantizer on branch
```
vshampor, line 142: The `AGGRESSIVE` strategy will be used by default, and it is the one most likely to garner performance benefits.

nncf/quantization/quantizer_propagation.py
```diff
@@ -590,28 +735,35 @@ def traverse_graph(self, curr_node_key: str,
...
-        insertion_type = node[QuantizerPropagationStateGraph.INSERTION_POINT_DATA_NODE_ATTR].insertion_type
-        if insertion_type == InsertionType.OPERATOR_POST_HOOK:
-            return True
-        return False
+        # node = self.nodes[ip_node_key]
+        # insertion_type = node[QuantizerPropagationStateGraph.INSERTION_POINT_DATA_NODE_ATTR].insertion_type
+        # if insertion_type == InsertionType.OPERATOR_POST_HOOK:
+        #     return True
+        # return False
+        return True
```
vshampor, line 766: Now that all insertion points, both pre- and post-hook, are functional, this function is mostly obsolete.

nncf/quantization/quantizer_propagation.py
```diff
@@ -798,11 +1174,28 @@ def propagation_step(self, curr_prop_quantizer: PropagatingQuantizer,
...
+                filter(lambda x: x in self._active_propagating_quantizers_queue or x is curr_prop_quantizer, dom_pqs))
+            waiting_pqs = self._quantizers_waiting_for_branch_merge.get_waiting_quantizers_for_branching_node_key(
+                branching_node_key)
+            if waiting_pqs == active_dom_pqs:
+                self._active_propagating_quantizers_queue.append(curr_prop_quantizer)
+                self._handle_quantizer_merge(waiting_pqs, quant_prop_graph, branching_node_key)
+            else:
+                # Not all of the dominated quantizers have reached the blocking node yet
+                self._active_propagating_quantizers_queue.appendleft(curr_prop_quantizer)
+            return quant_prop_graph
```
vshampor, line 1197: 2) The new propagation algorithm will make all quantizers that are trying to pass up through a downward-branching node wait until the quantizers on the neighbouring branches reach that node as well. Only then will the merge process occur, which may result in the new merge quantizer being added above the branching node to propagate further, and in the existing branch quantizers being discarded. For the `PropagationStrategy.AGGRESSIVE`, it is also possible that neither of the branch quantizers is discarded, but the merge quantizer is still created - for instance, if one of the branches will potentially require requantization, in case the user/specialized algorithm chooses a corresponding configuration.

aggressive_merge

Here, the merge quantizer will have a single 8-bit configuration associated since this is the only option that is supported by all the downstream branches. If, at the final quantization config selection stage of the setup, some of the branches also end up having 8-bit quantization chosen for them, the corresponding branch quantizers will be discarded since the global quantizer already has 8-bit quantization. The user is still, however, presented with a possibility to choose non-8-bit quantization for all the branches, in which case requantization will occur on the branches.

nncf/quantization/quantizer_id.py
```diff
@@ -57,14 +57,16 @@ class NonWeightQuantizerId(QuantizerId):
...
+                 input_port_id=None):
         self.ia_op_exec_context = ia_op_exec_context
+        self.input_port_id = input_port_id
 
     def get_base(self) -> 'InputAgnosticOperationExecutionContext':
         return self.ia_op_exec_context
 
     def get_suffix(self) -> str:
-        return ''
+        return '|OUTPUT' if self.input_port_id is None else '|INPUT{}'.format(self.input_port_id)
```
vshampor, line 69: The string representation of a quantizer has changed since now it needs to reflect the insertion type - either as an input quantizer to an operation or as an output quantizer. This may affect the code that relies upon it - please comment if you know of such pieces of code.

vshampor:
@alexsu52 @AlexKoff88 @ljaljushkin @asenina, add others as necessary.

asenina:
Jenkins please retry a build

tests/data/reference_graphs/quantized/hawq/inception_v3_device_VPU__mode_liberal__setup_type_propagation_based.dot
```diff
@@ -1,527 +1,532 @@
...
+"61 Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/NNCFConv2d[conv]/ModuleDict[pre_ops]/UpdateWeight[0]/SymmetricQuantizer[op]/symmetric_quantize" [color=green, id=61, label="8_bit__WFQ_#61", scope="Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/NNCFConv2d[conv]/ModuleDict[pre_ops]/UpdateWeight[0]/SymmetricQuantizer[op]", style=filled, type=symmetric_quantize];
+"62 Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/NNCFConv2d[conv]/conv2d" [color=lightblue, id=62, label="conv2d_#62", scope="Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/NNCFConv2d[conv]", style=filled, type=conv2d];
+"63 Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/BatchNorm2d[bn]/batch_norm" [id=63, label="batch_norm_#63", scope="Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/BatchNorm2d[bn]", style=filled, type=batch_norm];
+"64 Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]/RELU" [id=64, label="RELU_#64", scope="Inception3/InceptionA[Mixed_5b]/BasicConv2d[branch_pool]", style=filled, type=RELU];
+"65 Inception3/InceptionA[Mixed_5b]/cat" [id=65, label="cat_#65", scope="Inception3/InceptionA[Mixed_5b]", style=filled, type=cat];
```
ljaljushkin, line 67: This graph looks strange now 
image

1) input for AvgPool can't be INT4 by HW config.
2) previously, we had the same quantization parameters on all inputs and outputs of the concat node, which is HW-friendly. now, we allow uncoupled bitwidth of activations on the branches.

how it was before
image

nncf/quantization/precision_init/hawq_debug.py
```diff
@@ -105,30 +185,9 @@ def get_bitwidth_graph(algo_ctrl, model, all_quantizers_per_full_scope,
...
-                for next_nx_node_key in nncf_graph.get_successors(post_hooked_nx_node_key):
-                    activation_fq_node = nncf_graph.get_nx_node_by_key(next_nx_node_key)
-                    activation_quantizer = non_weight_quantizers[quantizer_id].quantizer_module_ref
-                    bits = activation_quantizer.num_bits
-
-                    activation_fq_node['color'] = bits_color_map[bits]
-                    activation_fq_node['style'] = 'filled'
-                    node_id = activation_fq_node[NNCFGraph.ID_NODE_ATTR]
-                    if grouped_mode:
-                        node_id = groups_of_adjacent_quantizers.get_group_id_for_quantizer(activation_quantizer)
```
ljaljushkin, line 127: please add a group number to the node name. It was very useful for debugging purposes

ljaljushkin (changes requested):

vshampor:
Jenkins please retry a build

nncf/quantization/precision_init/hawq_debug.py
vshampor: But it's still there, new line 155, is it not?

ljaljushkin: yes, this code is ok. For some reason, `groups_of_adjacent_quantizers` is empty now, which is not expected. Probably, this is also the reason, why graphs look incorrect

nncf/pruning/filter_pruning/algo.py
```diff
@@ -111,11 +115,12 @@ def get_in_out_channels(module):
         graph = self._model.get_original_graph()
         for nncf_node in graph.get_all_nodes():
             node_module = self._model.get_module_by_scope(nncf_node.op_exec_context.scope_in_model)
+            scope = nncf_node.op_exec_context.scope_in_model
             in_channels, out_channels = get_in_out_channels(node_module)
             if in_channels:
-                self.modules_in_channels[nncf_node.node_id] = in_channels
+                self.modules_in_channels[scope] = in_channels
```
vshampor, line 121: @mkaglins I had to refactor some of your code to use module scopes instead of NNCF graph node keys when referring to prunable modules. The tests are passing, but still please review.

tests/data/reference_graphs/quantized/hawq/mobilenet_v2_device_VPU__mode_liberal__ratio_2.5__setup_type_propagation_based.dot
```diff
@@ -1,273 +1,273 @@
...
-"46 MobileNetV2/Sequential[features]/InvertedResidual[4]/Sequential[conv]/ConvBNReLU[0]/NNCFConv2d[0]/ModuleDict[pre_ops]/UpdateWeight[0]/SymmetricQuantizer[op]/symmetric_quantize" [color=purple, id=46, label="2_bit__WFQ_#35", scope="MobileNetV2/Sequential[features]/InvertedResidual[4]/Sequential[conv]/ConvBNReLU[0]/NNCFConv2d[0]/ModuleDict[pre_ops]/UpdateWeight[0]/SymmetricQuantizer[op]", style=filled, type=symmetric_quantize];
+"42 MobileNetV2/Sequential[features]/InvertedResidual[3]/Sequential[conv]/BatchNorm2d[3]/batch_norm" [id=42, label="batch_norm_#42", scope="MobileNetV2/Sequential[features]/InvertedResidual[3]/Sequential[conv]/BatchNorm2d[3]", style=filled, type=batch_norm];
+"43 MobileNetV2/Sequential[features]/InvertedResidual[3]/Sequential[conv]/BatchNorm2d[3]/AsymmetricQuantizer/asymmetric_quantize" [color=purple, id=43, label="2_bit__AFQ_#43_G44", scope="MobileNetV2/Sequential[features]/InvertedResidual[3]/Sequential[conv]/BatchNorm2d[3]/AsymmetricQuantizer", style=filled, type=asymmetric_quantize];
```
ljaljushkin, line 45: That's not expected for VPU config. The minimum acceptable bitwidth for the add layer is 4, not 2.

image

ljaljushkin (changes requested):

tests/data/reference_graphs/quantized/hw/VPU/mobilenet_v2.dot
```diff
@@ -152,123 +152,123 @@ strict digraph  {
 "150 MobileNetV2/adaptive_avg_pool2d" [id=150, scope=MobileNetV2, type=adaptive_avg_pool2d];
 "151 MobileNetV2/Sequential[classifier]/Dropout[0]/dropout" [id=151, scope="MobileNetV2/Sequential[classifier]/Dropout[0]", type=dropout];
 "152 MobileNetV2/Sequential[classifier]/NNCFLinear[1]/linear" [id=152, scope="MobileNetV2/Sequential[classifier]/NNCFLinear[1]", type=linear];
-"/nncf_model_input_0" [color=purple, id=0, label="Quantizer: [B:8 M:A SGN:ANY W:N PC:N] [B:4 M:S SGN:ANY W:N PC:N] ", scope="", type=nncf_model_input];
```
ljaljushkin, line 155: Only int8 bitwidth is left after propagation, how does HAWQ choose int4 and int2 now? Is it called in the controller, not after the "experimental" stage of propagation?

vshampor: This is the graph that is built after the model has left `create_compressed_model`, and the information about the possible configurations at algorithm build time is essentially lost, so now this kind of (test) graph only reflects the final, possibly precision-inited quantized graph state.

tests/data/reference_graphs/quantized/hawq/mobilenet_v2_device_VPU__mode_liberal__ratio_2.5__setup_type_propagation_based.dot
vshampor: Fixed a bug in liberal bitwidth assignment and added a runtime check to validate final config against the initial proposed search space set of configs.

nncf/pruning/filter_pruning/algo.py
mkaglins: @vshampor Looks good.

tests/data/reference_graphs/quantized/hawq/inception_v3_device_VPU__mode_liberal__setup_type_propagation_based.dot
vshampor: After enabling heterogenous quantizer configuration handling for HAWQ, I managed to restore most of the reference bitwidth distribution graphs to the same state that is currently present on develop.
Current state of the same piece of the graph that we considered in this discussion:
Screenshot from 2021-01-07 19-53-58
Note that 4-bit AFQ requantizations appeared on the branches. This is due to the fact top-level FQ is 8-bit symmetric per-channel, which is the only configuration listed as supported by avg_pool2d in `vpu.json`, while the conv2d inputs only support 4-bit symmetric per-tensor and 8-bit asymmetric per-tensor quantizations; requantization is unavoidable for current VPU config.

vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

PR #303, opened 2020-11-26
Remove .dot extension in the HW config test case descriptor

vshampor:


ljaljushkin (approved):
forgot to remove, thanks 👍

PR #304, opened 2020-11-26
Don't call MLFlow if it's not enabled

ljaljushkin:
Required to avoid mlflow.exceptions.MlflowException: Could not create run under non-active experiment with ID 0.


SKholkin (approved):
Good to go

PR #308, opened 2020-11-28
Removed custom SqueezeNet model for better user experience

ljaljushkin:
Originally we had a modified copy of the SqueezeNet model to work around a bug in ONNX exporter with converting MaxPool with ceil_mode=True.
This bug isn't actual now for torch 1.5 and there's an almost identical SqueezeNet model in torchvision > 0.6.
That's why custom SqueezeNet was deleted as not needed to remove confusion.

There are no changes in the corresponding NNCF graph.
Previously trained checkpoints for custom SqueezeNet can be loaded and evaluated with SqueezeNet from torchvision. INT8 model has the same accuracy, the mixed model differs only by ~0.01 in maximum.

Should be merged after #307


vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

vshampor (approved):

ljaljushkin:
@vuiseng9

PR #310, opened 2020-11-30
Fix pruning l2norm

AlexanderDokuchaev:
Problem:
- l2norm module has weights that was not prune before
- l2norm associated with two nodes ('div' and 'rmul')

Done:
- handling case when one module associated with several operations.
- added `ignored_algorithms' parameters for modules.
- updated reference files of ssd models.


nncf/pruning/export_helpers.py
```diff
@@ -345,6 +345,23 @@ def mask_propagation(self, model: NNCFNetwork, nx_node, graph: NNCFGraph, nx_gra
             assert all([torch.allclose(input_masks[0], mask) for mask in input_masks])
         nx_node['output_mask'] = input_masks[0]
 
+    def input_prune(self, model: NNCFNetwork, nx_node: dict, graph: NNCFGraph, nx_graph: nx.DiGraph):
```
vshampor, line 348: This is not a proper way to register a user module as having trainable parameters to compress.

Use the `@nncf.register_module` decorator on a target module, such as `L2Norm`.

vshampor (changes requested):

nncf/pruning/export_helpers.py
vshampor: @AlexanderDokuchaev could you successfully use the proposed way? Is the PR unnecessary now, or are investigations still underway?

AlexanderDokuchaev: Updated with using decorator @register_module

vshampor:
Jenkins please retry a build

nncf/nncf_network.py
```diff
@@ -514,6 +514,9 @@ def get_nncf_modules(self) -> Dict['Scope', torch.nn.Module]:
         nncf_module_names_list = NNCF_MODULES + [x.__name__ for x in NNCF_WRAPPED_USER_MODULES_DICT.values()]
         return get_all_modules_by_type(self.get_nncf_wrapped_model(), nncf_module_names_list)
 
+    def get_nncf_modules_by_module_names(self, nncf_module_names_list) -> Dict['Scope', torch.nn.Module]:
```
vshampor, line 517: Type hint for the argument missing

nncf/utils.py
```diff
@@ -146,6 +147,12 @@ def get_state_dict_names_with_modules(model, str_types=None, prefix=''):
             found.update(sub_found)
     return found
 
+def get_nncf_module_names_for_algorithm(algorithm_name=None):
```
vshampor, line 150: Type hints are missing. Should we really use names (strings) instead of class objects themselves?

nncf/layer_utils.py
```diff
@@ -14,8 +14,18 @@ def __getattr__(self, name):
...
+        Attributes:
+            op_func_name    Name of corresponding torch function.
+            target_weight_dim_for_compression   Target dimension of weights that will be compressed in some algorithms.
+            filter_algorithms   List of algorithms that will use the module.
+                                If None - module will use for all algorithms.
+        """
+
     op_func_name = ""
     target_weight_dim_for_compression = 0
+    filter_algorithms = None
```
vshampor, line 28: Rename to "ignored_algorithms" and adjust the logic.

nncf/compression_method_api.py
```diff
@@ -267,6 +267,8 @@ class CompressionAlgorithmBuilder:
     on an NNCFNetwork object wrapping a target PyTorch model (torch.nn.Module).
     """
 
+    _registered_name = None
```
vshampor, line 270: Type hint required

nncf/compression_method_api.py
```diff
@@ -278,6 +280,7 @@ def __init__(self, config: NNCFConfig, should_init: bool = True):
         if not isinstance(self.config, list):
             self.ignored_scopes = self.config.get('ignored_scopes')
             self.target_scopes = self.config.get('target_scopes')
+        self.compressed_nncf_module_names = get_nncf_module_names_for_algorithm(self._registered_name)
```
vshampor, line 283: `_nncf_module_types_to_compress`

examples/object_detection/layers/modules/l2norm.py
```diff
@@ -16,8 +16,14 @@
 import torch.nn.init as init
 
 from nncf.utils import add_domain
+from nncf import register_module
 
+
+@register_module
 class L2Norm(nn.Module):
+
+    filter_algorithms = ['filter_pruning']
```
vshampor, line 25: Should instead be supplied as an argument to the `@register_module` invocation.
`@register_module(ignored_algorithms=['filter_pruning'])`

nncf/layers.py
```diff
@@ -199,7 +199,7 @@ def register_module(cls, *quantizable_field_names: str):
     # customly named attributes if it becomes necessary
     UNWRAPPED_USER_MODULES.registry_dict[cls.__name__] = cls
     nncf_wrapped_module_class_name = 'NNCFUser{}'.format(cls.__name__)
-    NNCF_WRAPPED_USER_MODULES_DICT[cls] = type(nncf_wrapped_module_class_name, (_NNCFModuleMixin, cls), {})
+    NNCF_WRAPPED_USER_MODULES_DICT[cls] = type(nncf_wrapped_module_class_name, (cls, _NNCFModuleMixin), {})
```
vshampor, line 202: I suppose that once you use the `@register_module` extra parameters for describing the user module to NNCF, this change could be reverted.

vshampor (changes requested):

examples/object_detection/layers/modules/l2norm.py
vshampor: This is related to the problem of user modules having weights in attributes that are named other than `weight`. While you're at it, you may also lay down another argument to `@register_module` that will allow the user to specify just that - the attribute name(s) of the module's weight(s).

nncf/utils.py
AlexanderDokuchaev: In my mind, for custom layers using string names is easier to use. Using classes looks weird  because you should import all classes of algorithms to ignore them.

vshampor (approved):

PR #315, opened 2020-12-01
Introduced a separate batch size for initialization

### LINKED ISSUE #170: Configure batch size for HAWQ separately from training batch size
HAWQ analysis may require more GPU memory, hence it would be beneficial to have different batch sizes for training and for precision initialization 
@RikAllen

ljaljushkin:
Separate data_loader is registered for initialization via `register_default_init_args`
resolves #170 
@vuiseng9

### DIFF
--- a/examples/classification/main.py
+++ b/examples/classification/main.py
@@ -137,10 +137,9 @@ def main_worker(current_gpu, config: SampleConfig):
     else:
         # Data loading code
         train_dataset, val_dataset = create_datasets(config)
-        train_loader, train_sampler, val_loader = create_data_loaders(config, train_dataset, val_dataset)
+        train_loader, train_sampler, val_loader, init_loader = create_data_loaders(config, train_dataset, val_dataset)
 
-        nncf_config = register_default_init_args(nncf_config, train_loader, criterion, train_criterion_fn,
-                                                 config.device)
+        nncf_config = register_default_init_args(nncf_config, init_loader, criterion, train_criterion_fn, config.device)
 
     # create model
     model = load_model(model_name,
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ model = resnet50()
 nncf_config = NNCFConfig.from_json("resnet50_int8.json")
 
 # Provide data loaders for compression algorithm initialization, if necessary
-nncf_config = register_default_init_args(nncf_config, train_loader, loss_criterion)
+register_default_init_args(nncf_config, train_loader, loss_criterion)
```
vshampor, line 44: The return value is not redundant. We need to modify one of the arguments of a function, but it is better to do so explicitly by returning the modified value instead of relying upon pass-by-reference mechanism. I'd rather if you modified `register_default_init_args` so that it modifies a copy of the original config instead of the argument directly.

vshampor (changes requested):

README.md
ljaljushkin: agree as it's a library function

vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

README.md
ljaljushkin: But with an internal copy of nncf_config, users would have to change their code as well.
Otherwise, the sample config wouldn't have an updated nncf_config with registered structures.
```
config.nncf_config = register_default_init_args(nncf_config, train_loader, loss_criterion)
```
So kept as is, without copy and with returning the argument.

ljaljushkin:
Jenkins please retry a build

vshampor (approved):

PR #322, opened 2020-12-04
Use mock cifar for sanity

vshampor:


vshampor:
`test_sanity_sample` runtime seems to have impoved from ~30 mins to ~15 mins.

PR #323, opened 2020-12-04
Add domain for custom operators in onnx-model.

asenina:


nncf/compression_method_api.py
```diff
@@ -241,9 +253,16 @@ def export_model(self, filename, input_names: List[str] = None, output_names: Li
                               output_names=output_names,
                               enable_onnx_checker=False,
                               opset_version=10)
+
+        onnx_model = onnx.load(filename)
+        for op_node in onnx_model.graph.node:
+            if op_node.op_type in CUSTOM_OPERATIONS_NODE_TYPE:
+                op_node.domain = DOMAIN_CUSTOM_OPS_NAME
+        onnx.save(onnx_model, filename)
```
vshampor, line 261: I don't think that this is acceptable at all. @AlexKoff88

vshampor (changes requested):

nncf/compression_method_api.py
AlexKoff88: @vshampor, what is wrong with that? Unfortunately, another approach does not work in our case, as far as I understood.

@asenina, can we extract this code into a separate function and call, e.g. `add_custom_operators_domain`. BTW, it is called operator in ONNX notation (not operation).

asenina:
Jenkins please retry a build

asenina:
Jenkins please retry a build

asenina:
Jenkins please retry a build

nncf/compression_method_api.py
AlexKoff88: I guess, it is not relevant anymore. I am ok with the changes.

AlexKoff88 (approved):

vshampor (approved):

PR #278, opened 2020-11-18
add sclae_log parameters for quantization.

krodyush:
add sclae_log parameters for quantization. its allow to increse convergence speed for high scales and increase accuracy for low scales.

PR #329, opened 2020-12-06
Add an option to optimize logarithms of quantizer scales instead of scales directly

vshampor:


krodyush (approved):
looks like it does the same what I requested in  https://github.com/openvinotoolkit/nncf/pull/278

vshampor:
@krodyush could you verify with your use cases that this indeed improves convergence?

krodyush:
Yes I could. But it require time for preparing and run experiments.

vshampor:
> Yes I could. But it require time for preparing and run experiments.

Please start the experiments.

krodyush:
checked. it still works. 1 epoch tune for packed bert-qa model.
exact_match metric for "logarithm_scale": true better by ~1%

vshampor:
@krodyush , thanks.

PR #330, opened 2020-12-07
Add quantization support for nn.EmbeddingBag

vuiseng9:


vuiseng9:
@vshampor @ljaljushkin @AlexKoff88 

This addition is mainly for DLRM embedding implementation. Local quantization exercise works fine. Could you review if anything is missing to add support for a new operator? 
dlrm_s_quantized_graph

vshampor:
@vuiseng9 should probably also add the new metatype to `DEFAULT_QUANT_TRAIT_TO_OP_DICT` next to the regular `Embedding` in `nncf/quantization/quantizer_propagation.py`. Furthermore, please extend the single-layer tests (https://github.com/openvinotoolkit/nncf/blob/516d8c981ba43763ade17ec307bc6543ad6d46e5/tests/test_compressed_graph.py#L683) with a new case that covers the EmbeddingBag scenario - should be pretty straightforward.

vuiseng9:
@vshampor, @ljaljushkin  I am trying to wrap my head around `TorchBinaryMethodDesc`, why embedding is wrapped in this class instead of `SingleLayerModelDesc` ?

https://github.com/openvinotoolkit/nncf/blob/516d8c981ba43763ade17ec307bc6543ad6d46e5/tests/test_compressed_graph.py#L652

Node that update quantized weight is not shown synthetic model dump below.
https://github.com/openvinotoolkit/nncf/blob/516d8c981ba43763ade17ec307bc6543ad6d46e5/tests/data/reference_graphs/quantized/synthetic_model/embedding.dot#L1-L7

Shouldn't embedding graph contain quantize nodes like conv2d?
https://github.com/openvinotoolkit/nncf/blob/516d8c981ba43763ade17ec307bc6543ad6d46e5/tests/data/reference_graphs/quantized/synthetic_model/Conv2d.dot#L1-L9

vshampor:
@vuiseng9 I'm ok if your change will result in both the Embedding and the EmbeddingBag scenarios being tested as modules (layers) instead of free functions - this way there will be quantizer nodes in the reference graphs (correponding to weight quantization). Currently if `F.embedding` is encountered outside a module, its inputs won't be quantized at all since there is currently no way to discern the integer input from the weight input in this case.

ljaljushkin:
`TorchBinaryMethodDesc` is a special case of `SingleLayerModelDesc` for testing "free" functions with 2 inputs,  e.g. from nn.functional (F.embedding) or methods of torch (torch.add), while SingleLayerModelDesc is more general and works with single input modules by default.

I'd recommend adding `nn.Embedding` and `nn.EmbeddingBag` via `TorchBinaryMethodDesc` and keeping the test case with `F.embedding`

vshampor:
Can't we have both the layer case and the free function case?

ljaljushkin:
> Can't we have both the layer case and the free function case?

Tthis is exactly what I meant in a previous message. Let's have both cases covered in tests. @vuiseng9 pay attention to the name of dot-file, it should be different to match correctly.

vuiseng9:
Thanks! @vshampor @ljaljushkin, make sense to me now. Will work on the tests.

vuiseng9:
The default dot file generation has Functional and Module differentiated by first character of the file, e.g. embedding.dot vs Embedding.dot. I suppose this is okay or good enough?

vuiseng9:
@vshampor @ljaljushkin the last open - the synthetic model quantization tests for nn.Embedding and nn.EmbeddingBag do not have any quantize node in the dot files. As we don't quantize the integer lookup input, we should see quantize node for updating embedding weight table? what could be missing here?

vshampor:
@vuiseng9 the lack of the weight quantizers currently is due to the fact that embeddings are not present as an operation type in the `cpu.json` HW config, which is used by default for synthetic tests. Originally we strived for the HW config operations being specified in the IR namespace, but `nn.Embedding` layers seem to be translated to `Gather` operations in the IR and that brings us to the problem that not every `Gather` operation is an embedding. Therefore we seem to be moving away from the aforementioned requirement; I extended the `cpu.json` config file with the custom, NNCF Pytorch-specific `"Embedding"` name, and associated the metatype with that name in #347. The HW config file parsing rules are so that the new change will lead to embedding layers being quantized using the default, 8-bit symmetric per-tensor configuration; should there be any further CPU-specific configurations for this layer, then these may be specified in the HW config file directly.

I suggest you rebase your PR upon #347 once it gets merged. Consider if, fundamentally, the `EmbeddingBagMetatype` as a separate metatype is really necessary (e.g. that there may arise a situation where `nn.EmbeddingBag` may require different treatment w.r.t. compression than `nn.Embedding` in the same model, or different quantizer configuration search space could be required for the two). If it *is* necessary, then add another entry to the HW config using #347 as reference. If not, then I propose that you register the `embedding_bag` function under the same `EmbeddingMetatype` - then both the bag and non-bag functions will share a metatype and, correspondingly, a HW config definition.

vuiseng9:
@vshampor Thanks for the detailed explanation!

Regarding `EmbeddingBag` to roll under `EmbeddingMetatype`, from quantization point of view, both `embedding` and `embedding_bag` should be the same, only weight quantization and no quantization on the input(s). But functionally, `embeddingbag` works as multiple embedding lookup, followed by a reduce operation (mean/sum/max). I wonder the implication of grouping them together from HW mapping perspective, does that mean both map to same HW kernel/operator? if that is the case, they shouldn't be grouped?

vshampor:
Jenkins please retry a build

tests/data/reference_graphs/quantized/synthetic_model/EmbeddingBag.dot
```diff
@@ -0,0 +1,5 @@
+strict digraph  {
+"0 /nncf_model_input" [id=0, scope="", type=nncf_model_input];
+"1 TestModel/NNCFEmbeddingBag[_layer]/embedding_bag" [id=1, scope="TestModel/NNCFEmbeddingBag[_layer]", type=embedding_bag];
```
vshampor, line 3: This still does not contain the weight quantizer node. Have you registered `EmbeddingBag` in the HW config in any way that I proposed previously?

vshampor (changes requested):

vuiseng9:
@vshampor sorry, I was waiting a decision if we need to grouping them. I have added EmbeddingBag to the CPU HW cfg. :)

vshampor:
Jenkins please retry a build

vshampor (approved):

tests/test_compressed_graph.py
```diff
@@ -653,6 +656,12 @@ def forward(self, x):
 
     TorchBinaryMethodDesc(model_name='embedding', torch_method=F.embedding,
                           input_info=[{"sample_size": [1], "type": "long"}, {"sample_size": [2]}]),
+    TorchBinaryMethodDesc(model_name='embedding_bag', torch_method=F.embedding_bag,
+                          wrap_inputs_fn=partial(n_inputs_fn, nargs=3),
```
ljaljushkin, line 660: @vuiseng9, please don't do that. because `TorchBinaryMethodDesc` is supposed to be used for 2 inputs only. But you extend it for an arbitrary number of inputs. There's a `SingleLayerModelDesc` class for it.
Please, just replace it with `SingleLayerModelDesc` here. It should work, as it's a base class and it has all this arguments

ljaljushkin (changes requested):

tests/test_compressed_graph.py
vuiseng9: `SingleLayerModelDesc` for `nn.EmbeddingBag` is added earlier. See line566-567

tests/test_compressed_graph.py
```diff
@@ -499,9 +499,9 @@ def forward(self, *args, **kwargs):
 
 
 class TorchBinaryMethodDesc(SingleLayerModelDesc):
-    def __init__(self, model_name: str, torch_method: Callable, input_info=None):
+    def __init__(self, model_name: str, torch_method: Callable, wrap_inputs_fn=n_inputs_fn, input_info=None):
```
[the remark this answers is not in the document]
ljaljushkin: SingleLayerModelDesc has wrap_inputs_fn and input_info
TorchBinaryMethodDesc just overrides the init method. All functionality is already in a base class.

just to this:
```
SingleLayerModelDesc(model_name='embedding_bag', torch_method=F.embedding_bag,
                          wrap_inputs_fn=partial(n_inputs_fn, nargs=3),
```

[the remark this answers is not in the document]
vuiseng9: @ljaljushkin Thanks for lighting the way, it didn't occur to me :) . Pushed the change and revert the `TorchBinaryMethodDesc` class definition.

ljaljushkin (approved):

vshampor:
Jenkins please retry a build

vuiseng9:
Do we have different set of rules between jenkins linter and nncf/.pylintrc? I do run pylint before pushing the codes. There seems to be discrepancy as I don't encounter those warnings locally.

vshampor:
Jenkins please retry a build

ljaljushkin:
> Do we have different set of rules between jenkins linter and nncf/.pylintrc? I do run pylint before pushing the codes. There seems to be discrepancy as I don't encounter those warnings locally.

CI uses certain version of pylint: 2.4.4. It might be the reason why you encountered these warnings

PR #332, opened 2020-12-07
Change tensorboardX to pytorch.utils.tensorboard

aleksei-kashapov:


setup.py
```diff
@@ -50,7 +50,7 @@ def find_version(*file_paths):
                     "graphviz",
                     "jsonschema",
                     "pydot",
-                    "tensorboardX",
+                    "tensorboard",
```
vshampor, line 53: Now that we have removed the external `tensorboardX` dependency and no repository code file is using it, why do we introduce another unused dependency?

vshampor (changes requested):

setup.py
aleksei-kashapov: pytorch.utils.tensorboard utilizes tensorboard package

vshampor: Ok, please set a specific version via `==`

aleksei-kashapov:
Jenkins please retry a build

vshampor (approved):

PR #336, opened 2020-12-07
Move to Python 3.8

FyodorK:
- Adding definitive version of required packages to setup.py, 
- Moving to python3.8, 
- Updating ReadMe.md file against changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,30 +41,30 @@ def find_version(*file_paths):
     raise RuntimeError("Unable to find version string.")
 
 
-INSTALL_REQUIRES = ["ninja",
-                    "addict",
-                    "pillow",
-                    "texttable",
+INSTALL_REQUIRES = ["ninja==1.10.0.post2",
+                    "addict==2.4.0",
+                    "pillow==8.0.1",
+                    "texttable==1.6.3",
                     "scipy==1.3.2",
-                    "networkx",
-                    "graphviz",
-                    "jsonschema",
-                    "pydot",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -73,8 +73,8 @@ def find_version(*file_paths):
 if python_version[0] < 3:
     print("Only Python > 3.5 is supported")
     sys.exit(0)
-elif  python_version[1] < 5:
-    print("Only Python > 3.5 is supported")
+elif  python_version[1] < 8:
+    print("Only Python > 3.8 is supported")
```
vshampor, line 77: This is probably asking the user too much. Let's not explicitly prohibit 3.5-3.7 here.

setup.py
```diff
@@ -41,30 +41,30 @@ def find_version(*file_paths):
...
+                    "opencv-python==4.4.0.46",
+                    "pytest-mock==3.3.1",
+                    "prettytable==2.0.0",
+                    "mdutils==1.3.0",
+                    "yattag==1.14.0",
+                    "jsonschema==3.2.0",
+                    "wheel==0.36.1",
+                    "defusedxml==0.7.0rc1",
+                    "mlflow==1.12.1",
                     "returns==0.14"]
```
vshampor, line 68: Separate these changes into another PR.

vshampor (changes requested):

setup.py
FyodorK: Should these lines be deleted from script?

PR #338, opened 2020-12-07
Fk/update packages versions

FyodorK:
Add definitive versions of the packages to setup.py only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ See [third_party_integration](./third_party_integration) for examples of code mo
 
 ## System requirements
 - Ubuntu\* 18.04 or later (64-bit)
-- Python\* 3.8 or later
+- Python\* 3.5 or later
 - NVidia CUDA\* Toolkit 10.2
 - PyTorch\* 1.7 or later.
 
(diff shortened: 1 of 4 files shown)


vshampor (approved):

ljaljushkin:
@vshampor, @FyodorK 

With the latest develop branch after installation of required dependencies by `python setup.py install` I have a broken environment.
Any pip command fails with an error:
```
AttributeError: '_NamespacePath' object has no attribute 'sort'
```
my Python version is 3.6.9

Looks like this is a bug with pip. People recommend([1](https://github.com/googleapis/google-cloud-python/issues/2990), [2](https://github.com/pypa/pip/issues/4216)) upgrading to the latest pip with setuptools >= 34.3.0
Probably, we should check this in a setup.py to prevent errors like this.

ljaljushkin:
I've created a new virtual environment, upgraded pip and `python setup.py install` fails with error:
```
  File "/tmp/easy_install-s0gu5pna/pandas-1.2.0rc0/setup.py", line 20, in <module>
    import os
ModuleNotFoundError: No module named 'numpy'
```
Do we support Python3.8 only? What Python version is used in CI?

FyodorK:
According to requirements it is "Ubuntu* 18.04 or later (64-bit)"

vshampor:
@ljaljushkin looks like there are indeed problems with the current setup.py, as indicated by multiple nightly test failures. @FyodorK will investigate and provide a fix.

PR #339, opened 2020-12-08
Corrected grouping of activation quantizers

ljaljushkin:
Not merged FQ's for activations should be at the different groups, if unmerged activation FQ on the branch goes directly after another FQ for activation (common input for different branches).
@vuiseng9


ljaljushkin:
reference graph for SSD looks like this. The number at the end of the node label is a group id.
image

vshampor (approved):

PR #340, opened 2020-12-08
Fix per-layer sparsity. Add simple scheduler.

asenina:


nncf/compression_method_api.py
```diff
@@ -291,3 +291,13 @@ def build_controller(self, target_model: NNCFNetwork) -> CompressionAlgorithmCon
 
     def _should_consider_scope(self, scope_str: str) -> bool:
         return should_consider_scope(scope_str, self.target_scopes, self.ignored_scopes)
+
+
+class SimpleCompressionScheduler(CompressionScheduler):
+    """
+    Should be create in per-layer sparsity mode.
+    This scheduler is just a stub and has no meaning.
```
vshampor, line 299: Should avoid such statements - looks bad. Remove "has no meaning". Also, there is no *requirement* for it to be created only in the per-layer sparsity mode. The scheduler is pretty much independent, so document it as such. Finally, good naming is worth a hundred comments, so if you just call it `StubCompressionScheduler`, there will be no need for the docstring at all.

vshampor (changes requested):

vshampor:
Jenkins please retry a build

nncf/compression_method_api.py
asenina: Done.

vshampor (approved):

PR #347, opened 2020-12-10
Add Embedding to the CPU HW config definition

vshampor:


vshampor:
Jenkins please retry a build

PR #349, opened 2020-12-10
Fixed protobuf error

pfinashx:
Fixed error: ModuleNotFoundError: No module named 'google.protobuf' in parallel cpu jobs

### DIFF
--- a/tests/test_sota_checkpoints.py
+++ b/tests/test_sota_checkpoints.py
@@ -8,6 +8,7 @@ import pytest
 import subprocess
 import re
 import shlex
+import sysconfig
 from prettytable import PrettyTable
 from collections import OrderedDict
 from yattag import Doc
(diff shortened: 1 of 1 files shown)


FyodorK:
Jenkins please retry a build

pfinashx:
@vshampor Please take a look

tests/test_sota_checkpoints.py
```diff
@@ -457,6 +458,10 @@ def openvino_preinstall(openvino):
         subprocess.run("pip install scikit-image==0.17.2", check=True, shell=True)
         subprocess.run("{} setup.py install".format(sys.executable), cwd=ACC_CHECK_DIR, check=True, shell=True)
 
+        # Workaround to fix protobuf error
+        subprocess.run("touch __init__.py", cwd=os.path.join(sysconfig.get_paths()["purelib"], 'google'),
+                       check=True, shell=True)
```
vshampor, line 463: @pfinashx is this problem limited to OpenVINO/AC runs or will it also appear when using our example scripts?

pfinashx: Yes, it's appear when using our scripts, the problem is in installation of google protobuf package - it has missing __init__.py in ../site-packages/google/ folder. It does not depend on the number of running jobs. And it's reproduced only for CPU-only install. So it's depends from torch-cpu (see PR#332, where tensorboard was changed)

vshampor: @pfinashx I expect you will follow up with the fixes for our scripts as well. I will merge this PR in its current state only because it unblocks CI runs.

PR #350, opened 2020-12-10
Update max threshold values for checkpoint test

aleksei-kashapov:


vshampor:
@aleksei-kashapov this should be part of your 7-bit PR since it has related changes in checkpoints.

PR #365, opened 2020-12-23
Small scheduler fixes

mkaglins:
1. Deleted set_pruning_rate step from pruning schedulers initialization.
2. Added `step` to pruning schedulers and algo to allow mask applying on every training step
3. Added message about potential long time for filter norms calculations

@AlexKoff88 @vshampor Please, take a look.

### DIFF
--- a/nncf/pruning/filter_pruning/algo.py
+++ b/nncf/pruning/filter_pruning/algo.py
@@ -243,9 +243,8 @@ class FilterPruningController(BasePruningAlgoController):
 
     def set_pruning_rate(self, pruning_rate):
         # Pruning rate from scheduler can be flops pruning rate or percentage of params that should be pruned
-        same_pruning_rate = np.isclose(pruning_rate, self.pruning_rate)
         self.pruning_rate = pruning_rate
-        if not self.frozen and not same_pruning_rate:
+        if not self.frozen:
             nncf_logger.info("Computing filter importances and masks...")
             if self.all_weights:
                 if self.prune_flops:
--- a/nncf/pruning/filter_pruning/algo.py
+++ b/nncf/pruning/filter_pruning/algo.py
@@ -246,8 +246,7 @@ class FilterPruningController(BasePruningAlgoController):
         same_pruning_rate = np.isclose(pruning_rate, self.pruning_rate)
         self.pruning_rate = pruning_rate
         if not self.frozen and not same_pruning_rate:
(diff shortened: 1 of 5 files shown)


nncf/pruning/filter_pruning/algo.py
```diff
@@ -240,8 +243,11 @@ def _find_layerwise_pruning_rate(self, target_flops_pruning_rate):
 
     def set_pruning_rate(self, pruning_rate):
         # Pruning rate from scheduler can be flops pruning rate or percentage of params that should be pruned
+        same_pruning_rate = np.isclose(pruning_rate, self.pruning_rate)
         self.pruning_rate = pruning_rate
-        if not self.frozen:
+        if not self.frozen and not same_pruning_rate:
+            nncf_logger.info("Computing filter importances and masks,"
```
AlexKoff88, line 249: I would keep only the first part with an ellipsis:
"Computing filter importances and masks..."

nncf/pruning/filter_pruning/algo.py
```diff
@@ -240,8 +243,10 @@ def _find_layerwise_pruning_rate(self, target_flops_pruning_rate):
 
     def set_pruning_rate(self, pruning_rate):
         # Pruning rate from scheduler can be flops pruning rate or percentage of params that should be pruned
+        same_pruning_rate = np.isclose(pruning_rate, self.pruning_rate)
         self.pruning_rate = pruning_rate
-        if not self.frozen:
+        if not self.frozen and not same_pruning_rate:
```
AlexKoff88, line 248: I didn't get why do we need this `same_pruning_rate`. To me, here we should be controlled only the desired pruning level and set masks.

vshampor: It is prudent to skip setting one and the same pruning rate to save computation time. However, @mkaglins, now I'm wondering - if the PR schedule sets up a constant PR over many epochs, but the model's weights evolve drastically over these epochs (e.g. because of a high LR setting), should `set_pruning_rate` be called anyway? The algorithm that calculates *which* filters to prune depends on the values of the weights inside the affected filters, does it not?

mkaglins: @vshampor You are right, `set_pruning_rate` should be called anyway, I will fix it.
Moreover, deleting `set_pruning_rate` from the scheduler is enough to save computation time.

vshampor:
Jenkins please retry a build

PR #377, opened 2020-12-29
Fix eval failures due to BN adaptation enabled by default

vshampor:


ljaljushkin:
@vshampor which exactly failure does it fix? Was there a hang on evaluation?

vshampor:
@ljaljushkin getting errors in log output due to yet unfixed multiprocessing distributed data loader random exceptions on associated object destruction.

PR #378, opened 2020-12-29
Use a reduced number of BN adaptation samples for sanity testing

vshampor:


tests/quantization/test_precision_init.py
```diff
@@ -208,6 +208,10 @@ def create_hawq_test_config(batch_size=10, num_data_points=100, image_size=10):
                 },
                 'range': {
                     'num_init_samples': 1
+                },
+                'batchnorm_adaptation': {
+                    'num_bn_adaptation_samples': 0,
```
ljaljushkin, line 213: Does it mean that BN Adaptation won't be called?
I'd prefer calling it, but for the smaller number of samples, e.g. 1

ljaljushkin (changes requested):

tests/quantization/test_precision_init.py
vshampor: This tests precision init. It should be functional without BN adaptation - my reasoning was that this test should not involve too many other entities.

PR #380, opened 2020-12-29
Reduce BN adaptation samples count in HAWQ sanity configs

vshampor:


ljaljushkin (approved):
looks good, it should accelerate sanity tests for now

But @vuiseng9 need to revert your changes because this part was implemented in another way:

https://github.com/openvinotoolkit/nncf/pull/250/files#diff-4a44cb8f5d057f52c164551eb471a993a5a737f1904d2ea2248690770ac619fdR520-R535

Issue #201, opened 2020-10-16
Handle frozen layers properly in sparsity and quantization


vanyalzr:
Some training pipelines (e.g. Mask-RCNN-R50 in mmdetection) have certain layers frozen by default. Sparsification and low-bitwidth quantization for the frozen layers might lead to accuracy degradation, so these layers have to be skipped. INT8 quantization is expected to be fine (given that FakeQuantized parameters are tuned).

ljaljushkin:
As we discussed internally if frozen layers are not specified in `ignored_scopes`, NNCF should throw an error.

ljaljushkin:
relates to #199

vshampor:
Should be resolved by #455.

PR #382, opened 2020-12-29
Not compressing frozen layers

SKholkin:
Not compressing frozen layers with weights (requires_grad=False)


nncf/utils.py
```diff
@@ -376,3 +377,7 @@ def compute_FLOPs_hook(module, input_, output, dict_to_save, name):
 def add_domain(name_operator: str) -> str:
     from nncf.compression_method_api import DOMAIN_CUSTOM_OPS_NAME
     return DOMAIN_CUSTOM_OPS_NAME + "::" + name_operator
+
+
+def is_layer_frozen(module):
+    return not reduce(lambda x, y: x and y.requires_grad, module.parameters(), True)
```
ljaljushkin, line 383: let's check for `weight` parameter only

nncf/pruning/base_algo.py
```diff
@@ -205,12 +203,11 @@ def _is_module_prunable(self, target_model: NNCFNetwork, module, module_scope: S
         input_non_pruned_modules = get_first_pruned_modules(target_model, pruned_types + ['linear'])
         output_non_pruned_modules = get_last_pruned_modules(target_model, pruned_types + ['linear'])
         module_scope_str = str(module_scope)
-
-        if self.ignore_frozen_layers and not module.weight.requires_grad:
+        if not is_nncf_module(module):
             msg = "Ignored adding Weight Pruner in scope: {} because"\
-                    " the layer appears to be frozen (requires_grad=False)".format(module_scope_str)
+                    " module is not wrapped".format(module_scope_str)
```
ljaljushkin, line 208: 'is not wrapped for compression'

ljaljushkin (changes requested):

nncf/dynamic_graph/transform_graph.py
```diff
@@ -34,25 +34,47 @@ def is_nncf_module(module):
...
         if module.__class__.__name__ == module_type.__name__:
-            nncf_module = module
+            replaced_module = module
             if not module.__class__.__name__ == nncf_module_type.__name__:
-                nncf_module = nncf_module_type.from_module(module)
-            return nncf_module
+                replaced_module = nncf_module_type.from_module(module)
+
     from nncf.layers import UNWRAPPED_USER_MODULES
     for _, user_module_type in UNWRAPPED_USER_MODULES.registry_dict.items():
```
SKholkin, line 46: With `UNWRAPPED_USER_MODULES` logic stays the same as before with ignored scope

ljaljushkin (approved):

nncf/dynamic_graph/transform_graph.py
```diff
@@ -34,24 +34,46 @@ def is_nncf_module(module):
     return False
 
 
-def replace_module_by_nncf_module(module: nn.Module):
+def replace_module_by_nncf_module(module: nn.Module, child_scope, eval_ops_exec_ctx_str=None):
```
vshampor, line 37: Type hints are missing. Also, do you really have to pass the string representation of op contexts here? Can't you pass the objects themselves?

nncf/dynamic_graph/transform_graph.py
```diff
@@ -34,24 +34,46 @@ def is_nncf_module(module):
...
-            nncf_module = deepcopy(module)
-            return add_nncf_functionality_to_user_module(nncf_module)
-    return module
+            replaced_module = deepcopy(module)
+            replaced_module = add_nncf_functionality_to_user_module(replaced_module)
+
+    if module is not replaced_module:
+        if not module.weight.requires_grad:
+            nncf_logger.info("Ignored wrapping modules in scope: {} because "
+                             "the layer appears to be frozen (requires_grad=False)".format(child_scope))
```
vshampor, line 54: As stated in #201:
>That's why all frozen layers must be **explicitly** removed from the compression via ignored_scope attribute.
>If there are frozen layers not specified in ignored_scope, the error should happen with the message suggesting to >add frozen layers to ignored_scope or to make them unfrozen.

The logic I'm seeing, however, implicitly uses `requires_grad=False` to exclude modules from compression instead of checking whether the module is also within `ignored_scopes`. Has the strategy changed?

nncf/dynamic_graph/transform_graph.py
vshampor: `child_scope`, `eval_ops_exec_ctx_str` are all names from which it is not immediately obvious how the corresponding argument is related to what the function does. Rename the arguments appropriately. Alternatively, add a verbose docstring using the `""":param """` style.

vshampor (changes requested):

nncf/quantization/precision_init/manual_init.py
```diff
@@ -50,12 +52,15 @@ def apply_init(self) -> SingleConfigQuantizerSetup:
                 raise ValueError('Invalid format of bitwidth per scope: [int, str] is expected')
             bitwidth = pair[0]
             scope_name = pair[1]
+            print(str(scope_name))
```
ljaljushkin, line 55: please remove debug code

nncf/utils.py
```diff
@@ -354,6 +355,22 @@ def should_consider_scope(scope_str: str, target_scopes: List[str], ignored_scop
...
+def get_frozen_nncf_modules(model, convert_scopes_back_from_nncf_modules=False):
+    frozen_layers = []
+    for scope, module in model.get_nncf_modules().items():
+        if not module.weight.requires_grad:
+            frozen_layers.append((scope, module))
+    return frozen_layers
+    pass
+
+
+def convert_scope_back_from_nncf(scope):
```
ljaljushkin, line 367: is it used? should be removed, if not

ljaljushkin (changes requested):

nncf/dynamic_graph/transform_graph.py
ljaljushkin: not actual, as logic was changed

nncf/dynamic_graph/transform_graph.py
ljaljushkin: not actual, as logic was changed

nncf/model_creation.py
```diff
@@ -141,6 +142,27 @@ def create_compressed_model(model: Module, config: NNCFConfig,
 
     for builder in compression_algo_builder_list:
         compressed_model = builder.apply_to(compressed_model)
+
+    frozen_layers = []
```
ljaljushkin, line 146: probably, it makes sense to separate this logic in a two methods
```python
frozen_layers = get_frozen_layers(compressed_model, ignored_scopes, target_scopes)
handle_frozen_layers(compression_algo_builder_list, frozen_layers)
```

nncf/model_creation.py
```diff
@@ -141,6 +142,27 @@ def create_compressed_model(model: Module, config: NNCFConfig,
...
+            if not builder.are_frozen_layers_allowed():
+                problem_algos.append(builder.algo_name())
+        if len(problem_algos) > 0:
+            raise RuntimeError(
+                f'Frozen layers are not allowed for {problem_algos}. '
+                f'Please unfreeze them or put into the Ignored Scope.\n'
+                f'Frozen Layers: {[str(pair[0]) for pair in frozen_layers]}')
+        else:
+            logger.warn('Frozen layers detected. Using ' +
+                        reduce(lambda x, y: x + " and " + y,
```
ljaljushkin, line 162: not sure it's a good idea to insert code when you form a string in-place. this code may fail, but the error traceback would be confusing. 
I'd recommend doing some code outside of logging or raising an error (as in the code above).

ljaljushkin (changes requested):

tests/test_frozen_layers.py
```diff
@@ -0,0 +1,187 @@
...
+                    "type": "manual"
+                }
+            }
+        },
+        "target_scopes": ['TwoConvTestModel/Sequential[features]/Sequential[0]/Conv2d[0]']
+    })
+    , model_creator=TwoConvTestModel, is_error=True),
+    FrozenLayersTestStruct(config=get_quantization_config_without_range_init(),
+                           model_creator=TwoConvTestModel, is_error=False),
+    FrozenLayersTestStruct(config=update_config(get_quantization_config_without_range_init(), {
```
ljaljushkin, line 79: I'd do it in another way.
```
class FrozenLayersTestStruct
    config_base=get_quantization_config_without_range_init
       
    def create_config():
```
for most of the cases, you just provide a config update, responsibility for creation of config is inside class.
```
FrozenLayersTestStruct(config_update={})
```
if you need another config_base (rare case), just override it.
```
FrozenLayersTestStruct(config_base=get_basic_pruning_config, config_update={})
```

Note, that you don't even need to create an instance of the final config, you just specify instructions on how to create it, final creation happens only once on `create_config` call

`config_base` is just a pointer to function, not a dictionary.

ljaljushkin (changes requested):

tests/test_frozen_layers.py
```diff
@@ -0,0 +1,187 @@
...
+                "prune_last_conv": True
+            }
+        }
+    }), model_creator=TwoConvTestModel, is_error=True),
+    FrozenLayersTestStruct(config=get_basic_sparsity_config(),
+                           model_creator=TwoConvTestModel, is_error=True)
+]
+
+
+@pytest.mark.parametrize('params', TEST_PARAMS, ids=[str(p.config['model']) for p in TEST_PARAMS])
```
ljaljushkin, line 173: forgot to test `binarization` algorithm

ljaljushkin (changes requested):

ljaljushkin:
Jenkins please retry a build

tests/test_frozen_layers.py
ljaljushkin: also please don't forget to give better names for the test cases

ljaljushkin:
Jenkins please retry a build

ljaljushkin:
Jenkins please retry a build

nncf/quantization/algo.py
```diff
@@ -1088,7 +1088,30 @@ def _add_single_activation_quantizer(self, target_model: NNCFNetwork,
                                              input_port_id=curr_insertion_info.in_port_id)
             insertion_commands.append(
                 InsertionCommand(insertion_point, hook, OperationPriority.QUANTIZATION_PRIORITY))
-        return quantizer_id, insertion_commands
+            return quantizer_id, insertion_commands
```
ljaljushkin, line 1091: 👀

ljaljushkin: but the latest pre-commit was OK.  How? 😲

SKholkin: Interesting. Also develop branch uses out-of-loop version

ljaljushkin (approved):

nncf/utils.py
```diff
@@ -354,6 +354,15 @@ def should_consider_scope(scope_str: str, target_scopes: List[str], ignored_scop
                and not in_scope_list(scope_str, ignored_scopes)
 
 
+def get_frozen_nncf_modules(model, ignored_scopes=None, target_scopes=None):
+    frozen_layers = []
+    for scope, module in model.get_nncf_modules().items():
+        if not module.weight.requires_grad:
+            if should_consider_scope(str(scope), target_scopes, ignored_scopes):
+                frozen_layers.append((scope, module))
+    return frozen_layers
```
vshampor, line 363: Move to NNCFNetwork.

nncf/compression_method_api.py
```diff
@@ -329,6 +329,12 @@ def _nncf_module_types_to_compress(self) -> List[str]:
                 filtered_nncf_module_names_list.append(module.__name__)
         return filtered_nncf_module_names_list
 
+    def are_frozen_layers_allowed(self) -> bool:
+        return False
+
+    def algo_name(self) -> str:
```
vshampor, line 335: Shouldn't extend API just for that. Find a way to reuse the string values that are given as `@COMPRESSION_ALGORITHM.register()` parameters.

nncf/model_creation.py
```diff
@@ -158,3 +164,19 @@ def create_compressed_model(model: Module, config: NNCFConfig,
             graph = compressed_graph_builder.build_graph(compressed_model, compressed_model.get_tracing_context())
             graph.visualize_graph(osp.join(config.get("log_dir", "."), "compressed_graph.dot"))
     return compression_ctrl, compressed_model
+
+
+def handle_frozen_layers(frozen_layers, compression_algo_builder_list):
```
vshampor, line 169: Move to the base `CompressionAlgorithmBuilder.apply_to`.

nncf/compression_method_api.py
```diff
@@ -329,6 +329,12 @@ def _nncf_module_types_to_compress(self) -> List[str]:
                 filtered_nncf_module_names_list.append(module.__name__)
         return filtered_nncf_module_names_list
 
+    def are_frozen_layers_allowed(self) -> bool:
```
vshampor, line 332: Will there be a real use case where a user would need to query this in runtime (i.e. during training)? If not, consider making this a protected method.

vshampor (changes requested):

nncf/quantization/algo.py
```diff
@@ -1090,6 +1090,27 @@ def _add_single_activation_quantizer(self, target_model: NNCFNetwork,
                 InsertionCommand(insertion_point, hook, OperationPriority.QUANTIZATION_PRIORITY))
         return quantizer_id, insertion_commands
 
+    def are_frozen_layers_allowed(self) -> bool:
+        for quantizer in self._weight_quantizers.values():
+            if not quantizer.quantizer_module_ref.num_bits == 8:
+                return False
+        if self.config.get('initializer', {}).get('precision', None) is not None:
```
vshampor, line 1097: Use the value of `self._precision_init_args` instead of poking into the config, here and elsewhere.

nncf/compression_method_api.py
ljaljushkin: First, this decorator has nothing to do with the actual Builder class. It just put the name on the registry.
Second, the name can be not obvious, some logic may require -
https://github.com/openvinotoolkit/nncf/pull/382/files#diff-c2d882888dac8c1721e41bf7d650e6e3fc137181f13f835c25c3e0b27809eefeR1101

vshampor: The user won't interact with this function in any feasible way. This is a builder function, after all, and what user finally gets are controllers. Remove this function, and instead extend the `are_frozen_layers_allowed` function return value with a string specifying the reason why these are not allowed. After an invocation of `are_frozen_layers_allowed` (which will be done somewhere inside the builder code anyway), print the reason as a warning message.

ljaljushkin:
all comments are resolved in #455

PR #383, opened 2021-01-11
Fix object detection sample.

asenina:


vshampor (approved):

vanyalzr:
@asenina did you have a chance to check the `ssd_mobilenet_voc_rb_sparsity_int8.json` config? We'd want to update it as well as change the sparsity algo from RB to magnitude in case RB is not working properly there.

asenina:
@vanyalzr No, i didnt. I will check it shortly and report the results.

PR #400, opened 2021-01-17
Simplify paths to configs

andrey-churkin:
It looks better. I think we can do the same for NNCF PyTorch.
@vshampor what do you think?


alexsu52 (approved):
@andrey-churkin It looks better.

vshampor:
I think that paths relative to the repository root are more explicit.

andrey-churkin:
@vshampor My arguments to use links to configs:
- It is more user-friendly. You should not scroll the table because it fits completely.
- You can see the contents of the configuration quickly. You need just click to link.

Also, you can get a path relative to the repository root easily. For example, click on `inception_v3_imagenet_int8.json` and copy this line
screenshot

cc @AlexKoff88

PR #401, opened 2021-01-18
[TF] Add handling of non-distributed strategy

andrey-churkin:


beta/examples/tensorflow/common/object_detection/evaluation/coco_evaluator.py
```diff
@@ -227,10 +227,6 @@ def update(self, predictions, groundtruths=None):
                 groundtruths['height'] = sizes[:, 0]
                 groundtruths['width'] =  sizes[:, 1]
 
-            if 'num_detections' not in groundtruths:
```
alexsu52, line 230: Could you comment on why we need these changes?

beta/examples/tensorflow/segmentation/train.py
```diff
@@ -45,7 +44,8 @@ def get_argument_parser():
                                         save_checkpoint_freq=False,
                                         export_args=False,
                                         print_freq=False,
-                                        dataset_type=False)
+                                        dataset_type=False,
+                                        cpu_only=False)
```
alexsu52, line 48: Do you figure out the root of the training issue on CPU?

beta/examples/tensorflow/common/object_detection/evaluation/coco_evaluator.py
andrey-churkin: This code is not necessary now. In this place, `'num_detection'` always is in `groundtruths` and already has the required format.

beta/examples/tensorflow/segmentation/train.py
andrey-churkin: I tried. The problem sometimes is not reproduced. I could try to fix it again.

alexsu52 (approved):

PR #422, opened 2021-01-20
WIP: [tf]Add sota_checkpoints tf test 

daniil-lyakhov:


daniil-lyakhov:
`--pretrained` parameter works similar with pytorch sample param: nothing happened if it's omitted, but if it's present - sample have to use pretrained model. I updated docstring to make `update_from_args` call clearer

daniil-lyakhov:
Jenkins please retry a build

beta/examples/tensorflow/common/argparser.py
```diff
@@ -28,6 +28,10 @@ def get_common_argument_parser(**flags):
         help='Path to a config file with task/model-specific parameters.',
         required=True)
 
+    parser.add_argument('--metrics-dump',
```
alexsu52, line 31: metric_dump argument has not to be used in the train.py of the segmentation example. Thus, I propose to make it optional by adding it to the argument parser using the add_argument function. You should also modify the train.py of the segmentation example so that metric_dump is not used.

alexsu52 (changes requested):

alexsu52:
@daniil-lyakhov Could you add a test that demonstrates how you use these arguments?

PR #428, opened 2021-01-21
Changed collecting reference fp32 metric

pfinashx:
Now, if reference FP32 model not executed, will be collect "target" metric from JSON file.


vshampor (changes requested):
This situation must be specifically indicated to the reader of the table. If the reference had to be taken from the .JSON, put the value in parentheses when displaying.

pfinashx:
> This situation must be specifically indicated to the reader of the table. If the reference had to be taken from the .JSON, put the value in parentheses when displaying.

Done, please look at ubuntu20 job #12

PR #443, opened 2021-01-25
Updated HW config: added pertensor quantization of activations

AlexKoff88:
Updated HW config: added pertensor quantization of activations for schemes compatibility


ljaljushkin:
@AlexKoff88, shouldn't we target it to the release branch?

AlexKoff88:
> @AlexKoff88, shouldn't we target it to the release branch?

No rush with that.

vshampor (approved):

Issue #427, opened 2021-01-21
Repeated sota checkpoints test_eval launch failure


daniil-lyakhov:
During testing I notice that second launch of TestSotaCheckpoints::test_eval assert keyerror because metric_dump directrory isn't empty. This hasn't any impact on CI work pipeline but appear during hand testing.

Do we need to fix it and clean metrics_dump dir every launch or that's ok?

daniil-lyakhov:
I can patch it if needed

vshampor:
@daniil-lyakhov please provide a PR so that I could review it.

daniil-lyakhov:
I have a suggestion: let's add new parameter --metrics-dump-dir with default value `os.dir.join(PROJECT_ROOT,"test_results","metrics_dump" + str(timestamp))`. If this dir isn't empty - test will be failed

vshampor:
Closed via #445

PR #30, opened 2020-06-10
Generalized binarization pipeline for quantization

ljaljushkin:

PR #445, opened 2021-01-25
Clean previous metrics dump before sota test launch

daniil-lyakhov:
https://github.com/openvinotoolkit/nncf/issues/427

### DIFF
--- a/tests/test_sota_checkpoints.py
+++ b/tests/test_sota_checkpoints.py
@@ -2,6 +2,7 @@ import os
 import json
 import sys
 import csv
+import datetime
 from typing import Tuple, List, Optional
 
 import pytest
@@ -170,7 +171,7 @@ class TestSotaCheckpoints:
             writer.writerow({'model': filename, 'launcher': '-', 'device': '-', 'dataset': '-', 'tags': '-',
                              'metric_name': '-', 'metric_type': '-', 'metric_value': error_message})
 
-    def write_results_table(self, init_table_string):
+    def write_results_table(self, init_table_string, path):
         result_table = PrettyTable()
         result_table.field_names = init_table_string
         for key in self.row_dict:
@@ -205,7 +206,7 @@ class TestSotaCheckpoints:
                             i = '-'
                         with tag('td'):
                             text(i)
-        f = open('results.html', 'w')
+        f = open(path / 'results.html', 'w')
         f.write(doc.getvalue())
         f.close()
 
@@ -328,8 +329,8 @@ class TestSotaCheckpoints:
             pytest.skip('Path to datasets is not set')
         test = "eval"
         metric_file_name = self.get_metric_file_name(model_name=eval_test_struct.model_name_)
-        metrics_dump_file_path = METRICS_DUMP_PATH / metric_file_name
-        log_dir = METRICS_DUMP_PATH / "logs"
+        metrics_dump_file_path = pytest.metrics_dump_path / metric_file_name
+        log_dir = pytest.metrics_dump_path / "logs"
         cmd = self.CMD_FORMAT_STRING.format(sys.executable, 'test', conf=eval_test_struct.config_name_,
                                             dataset=sota_data_dir,
                                             data_name=eval_test_struct.dataset_name_,
@@ -355,7 +356,8 @@ class TestSotaCheckpoints:
         if eval_test_struct.reference_ is not None:
             fp32_metric = self.ref_fp32_dict[str(eval_test_struct.reference_)]
             metric_type_from_json = True
-            reference_metric_file_path = METRICS_DUMP_PATH / self.get_metric_file_name(eval_test_struct.reference_)
+            reference_metric_file_path = pytest.metrics_dump_path / \
+                                         self.get_metric_file_name(eval_test_struct.reference_)
             if os.path.exists(reference_metric_file_path):
                 with open(str(reference_metric_file_path)) as ref_metric:
                     metrics = json.load(ref_metric)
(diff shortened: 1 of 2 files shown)


pfinashx (changes requested):

tests/test_sota_checkpoints.py
```diff
@@ -498,6 +498,13 @@ def openvino_preinstall(openvino):
                        check=True, shell=True)
 
 
+@pytest.fixture(autouse=True, scope="class")
+def clean_previous_metrics_dump_dir():
+    if os.path.isdir(METRICS_DUMP_PATH):
+        subprocess.run("rm -f *.json", cwd=METRICS_DUMP_PATH, check=True, shell=True)
```
alexsu52, line 504: It does not look like a safe method and is unlikely to pass the SDL checks.  The second concern, that it is not a portable solution. What do you think about using the standard module, for example, `shutil`?

vshampor: @pfinashx, `"rm -rf *"` looks scary indeed. I also think that the `os.` approach should be used, just so that the test invocation does not accidentally use '/' as the current working directory. 

@daniil-lyakhov with the `os` approach, should either filter for `.json` files or delete the directory and recreate it instead of removing all files inside it.

daniil-lyakhov: Completely agree with that. I used to remove only `.json` files because test keeps log dir inside `metrics_dump` dir. Shall we remove all previous run logs before new launch?

pfinashx: If subprocess does not safe (although more safe file mask can be used, like "*.metrics.json")
We can use construction with glob.glob to find files and os.remove to delete they - 
`for file in glob.glob(f'{METRICS_DUMP_PATH}/*metrics.json'):
    os.remove(file)`

pfinashx: Please run NNCF eval with your branch and add job number here.

tests/test_sota_checkpoints.py
```diff
@@ -507,20 +509,27 @@ def openvino_preinstall(openvino):
...
-def clean_previous_metrics_dump_dir():
-    if os.path.isdir(METRICS_DUMP_PATH):
-        subprocess.run("rm -f *.json", cwd=METRICS_DUMP_PATH, check=True, shell=True)
+def make_metrics_dump_path(metrics_dump_dir):
+    if pytest.metrics_dump_path is None:
+        data = datetime.datetime.now()
+        pytest.metrics_dump_path = PROJECT_ROOT / "test_results" / \
+            f"metrics_dump_{'_'.join([str(getattr(data, atr)) for atr in ['year', 'month', 'day', 'hour', 'minute', 'second']])}"
+    else:
+        pytest.metrics_dump_path = Path(pytest.metrics_dump_path)
```
pfinashx: I think we need specify basename of metrics path like PROJECT_ROOT - 
`else:  pytest.metrics_dump_path = PROJECT_ROOT / Path(pytest.metrics_dump_path)
`
to put not full path in command line, but only the folder inside .../nncf/

pfinashx (changes requested):

tests/test_sota_checkpoints.py
daniil-lyakhov: But when you specifying --sota-test-path and --sota-checkpoint-path it uses global path. As it does in our samples. It's a bit controversial for me as a test user. I think we need others opinion on this @asuhov @vshampor.

tests/test_sota_checkpoints.py
daniil-lyakhov: Could you please provide job link?

pfinashx: I started job #30 on Ubuntu20

tests/test_sota_checkpoints.py
pfinashx: In general, I don't think "naming output folder" is a good idea, test_sota_crckpoints in the main needed for CI, I think end users will run the models separately.

tests/test_sota_checkpoints.py
```diff
@@ -506,14 +506,28 @@ def openvino_preinstall(openvino):
...
+    if pytest.metrics_dump_path is None:
+        data = datetime.datetime.now()
+        pytest.metrics_dump_path = PROJECT_ROOT / "test_results" / "metrics_dump_" \
+            f"{'_'.join([str(getattr(data, atr)) for atr in ['year', 'month', 'day', 'hour', 'minute', 'second']])}"
+    else:
+        pytest.metrics_dump_path = Path(pytest.metrics_dump_path)
+    assert not os.path.isdir(pytest.metrics_dump_path) or not os.listdir(pytest.metrics_dump_path), \
+                                    f"metrics_dump_path dir should be empty: {pytest.metrics_dump_path}"
+    print(f"metrics_dump_path: {pytest.metrics_dump_path}")
+    yield
```
alexsu52, line 520: Why do you use yield here?

daniil-lyakhov: Yeah, I can omit it

tests/test_sota_checkpoints.py
alexsu52: @pfinashx I agree that test_sota_checkpoints is mainly needed for CI, but we should not forget about developers/contributors who want to validate their changes also on the checkpoints locally and control results of each run. For example, It can be helpful for bumping up the version of any third-party components. If you have not critical concerns I suggest giving developers the ability to specify any path where need to save results.

daniil-lyakhov:
Jenkins please retry a build

vshampor (approved):

vshampor:
@pfinashx can this be merged right now?

pfinashx:
> @pfinashx can this be merged right now?

Waiting test job (# 56), and if it pass, PR can be merged.

pfinashx (approved):

PR #446, opened 2021-01-25
Folding conv2d and BN2d

asenina:


nncf/layers.py
```diff
@@ -61,6 +63,46 @@ def from_module(module):
...
+            else:
+                conv = self._nncf_conv_forward(input, self.weight)
+
+            if self.scale_factor is not None:
+                conv_orig = conv / self.scale_factor.reshape(bias_shape)
+            else:
+                conv_orig = conv
+            if self.bias is not None:
+                conv_orig = conv_orig + self.bias.reshape(bias_shape)
+            #conv = self.bn(conv_orig)
```
Zehaos, line 91: Should bn be involved?

PR #455, opened 2021-01-27
Handling Frozen Layers

ljaljushkin:
resolves comments from #382 and issue #201

### DIFF
--- a/nncf/composite_compression.py
+++ b/nncf/composite_compression.py
@@ -14,9 +14,11 @@
 from typing import List
 
 import torch.nn
+from copy import deepcopy
 
 from nncf.compression_method_api import CompressionLoss, CompressionScheduler, \
-    CompressionAlgorithmController, CompressionLevel
+    CompressionAlgorithmController, CompressionLevel, CompressionAlgorithmBuilder
+from nncf.hw_config import HWConfigType, HW_CONFIG_TYPE_TARGET_DEVICE_MAP
 from nncf.nncf_network import NNCFNetwork
 from nncf.pruning.base_algo import BasePruningAlgoController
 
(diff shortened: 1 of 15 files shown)


nncf/model_creation.py
```diff
@@ -136,17 +103,18 @@ def create_compressed_model(model: Module, config: NNCFConfig,
                                    scopes_without_shape_matching=scopes_without_shape_matching)
 
     should_init = resuming_state_dict is None
-    compression_algo_builder_list = create_compression_algorithm_builders(config, should_init=should_init)
+    composite_builder = CompositeCompressionAlgorithmBuilder(config, should_init=should_init)
+
+    composite_builder.apply_to(compressed_model)
+    composite_builder.handle_frozen_layers()
```
vshampor, line 108: Applying the algorithm without handling frozen layers should not be allowed. The `handle_frozen_layers` call should be moved inside `apply_to`, and the method itself be made protected.

nncf/utils.py
```diff
@@ -354,6 +354,15 @@ def should_consider_scope(scope_str: str, target_scopes: List[str], ignored_scop
                and not in_scope_list(scope_str, ignored_scopes)
 
 
+def get_wrapped_frozen_nncf_modules(model, ignored_scopes=None, target_scopes=None):
+    frozen_layers = []
+    for scope, module in model.get_nncf_modules().items():
+        if not module.weight.requires_grad:
+            if should_consider_scope(str(scope), target_scopes, ignored_scopes):
+                frozen_layers.append((scope, module))
+    return frozen_layers
```
vshampor, line 363: Please stop adding functions to utils when a proper place for them is inside `NNCFNetwork` (or any other specialized class, for that matter).

nncf/composite_compression.py
```diff
@@ -78,6 +82,71 @@ def load_state_dict(self, state_dict):
...
+    @property
+    def child_builders(self):
+        return self._child_builders
+
+    def apply_to(self, target_model: NNCFNetwork) -> NNCFNetwork:
+        for ctrl in self._child_builders:
+            target_model = ctrl.apply_to(target_model)
+        return target_model
+
+    def handle_frozen_layers(self):
```
vshampor, line 133: With this kind of a solution, a single-builder NNCF usage such as:
```
builder = MyBuilder()
builder.apply_to(nncf_network)
ctrl = nncf_network.commit_compression_changes()
```
 won't be handling frozen layers at all. 

I think it would be safe to instead call `_handle_frozen_layers()` for each of the child_builders. When you move the call itself to `apply_to`, implementing a separate `_handle_frozen_layers()` for the composite builder will not be necessary.

vshampor (changes requested):

nncf/utils.py
ljaljushkin: just forgot to remove

nncf/model_creation.py
ljaljushkin: done

nncf/composite_compression.py
ljaljushkin: done

vshampor (approved):

PR #456, opened 2021-01-27
Correctly construct Q-DQ config for E2E tests

vshampor:


vshampor:
Jenkins please retry a build

tests/test_sota_checkpoints.py
```diff
@@ -90,7 +90,21 @@ def get_metric_file_name(model_name: str):
     @staticmethod
     def q_dq_config(config):
         nncf_config = NNCFConfig.from_json(config)
-        nncf_config["export_to_onnx_standard_ops"] = True
+        if "compression" in nncf_config:
```
pfinashx, line 93: So, if there is no compression in config (FP32 model), "export_to_onnx_standard_ops":True will not applied?

vshampor: Correct, and in fact there is no need to do run the corresponding accuracy check since the ONNX would be the same. Same for non-INT8 models

tests/test_sota_checkpoints.py
```diff
@@ -90,7 +90,21 @@ def get_metric_file_name(model_name: str):
...
+            compression_config = nncf_config["compression"]
+            quantization_config = None
```
pfinashx, line 95: I have error TypeError: 'NoneType' with resnet18_imagenet_binarization_dorefa.json: It has "compression" but it not list, and "algorithm" is not "quantization", need more conditions.

pfinashx (changes requested):

tests/test_sota_checkpoints.py
vshampor: Done

PR #461, opened 2021-01-29
Release v160

vshampor:


vshampor:
Will ignore the failure in the single test in the precommit which is known to be sporadic and relates more to the unfixable PyTorch caching memory allocation mechanism.

PR #469, opened 2021-02-01
Update install instruction

aleksei-kashapov:


tests/test_install.py
```diff
@@ -20,22 +20,34 @@
...
     version_string = "{}.{}".format(sys.version_info[0], sys.version_info[1])
-    subprocess.call("virtualenv -ppython{} {}".format(version_string, venv_path), shell=True)
     python_executable_with_venv = ". {0}/bin/activate && {0}/bin/python".format(venv_path)
     pip_with_venv = ". {0}/bin/activate && {0}/bin/pip".format(venv_path)
 
+    if venv_type == 'virtualenv':
+        subprocess.call("virtualenv -ppython{} {}".format(version_string, venv_path), shell=True)
+    elif venv_type == 'venv':
+        subprocess.call("python{} -m venv {}".format(version_string, venv_path), shell=True)
+        subprocess.call("{} install --upgrade pip".format(pip_with_venv), shell=True)
```
vshampor, line 49: You are running this line in a wrong venv.

vshampor (changes requested):

aleksei-kashapov:
@vshampor please have a look

vshampor:
@aleksei-kashapov please restore the additional installation test that used -m venv.

aleksei-kashapov:
> @aleksei-kashapov please restore the additional installation test that used -m venv.

Do we really need them? They take time to pass

vshampor:
@aleksei-kashapov we really need them since `-m venv` is in fact becoming a standard way to create new virtual environments. However, we shouldn't remove `virtualenv` testing either - it is still a viable option for users and, as you've discovered, it may act differently than `-m venv`.

vshampor (approved):

PR #480, opened 2021-02-03
Python Style Guide

alexsu52:


AlexanderDokuchaev:
How about use https://github.com/psf/black?

daniil-lyakhov (approved):

daniil-lyakhov:
I have a question about nested function call. For example:
```python
def k(input):
    result = f(input)
    return g(result)
```
or
```python
def k(input):
    return g(f(input))
```
What is better?

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+  - Closing over a local variables.
+  ```python
+  # Correct:
+  def make_scaling_fn(scale):
+      def mul(x):
+          return scale * x
+      return mul
+  ```
+
+- Inner classes are fine when it creates more readble and simple code.
```
andrey-churkin, line 161: readble --> readable

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+
+- Do not nest a function just to hide it from users of a module. Instead, prefix its name with an \_ at the module
```
andrey-churkin, line 163: Replace \_ to `_` for consistency with line 140:
> If needed, globals should be declared at the module level and made internal to the module by prepending an `_` to the

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+  from foo.bar.yourclass import YourClass
+  ```
+  If this spelling causes local name clashes, then spell them explicitly:
+  ```python
+  import myclass
+  import foo.bar.yourclass
+  ```
+  and use `myclass.MyClass` and `foo.bar.yourclass.YourClass`.
+
+- Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present
```
andrey-churkin, line 121: Maybe (from `<module>` import *) --> (`from module import *`)

and `<module>` is not displayed now

andrey-churkin (changes requested):

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+container while iterating over it.
+
+```python
+# Correct:
+for key in adict: ...
+if key not in adict: ...
+if obj in alist: ...
+for line in afile: ...
+for k, v in adict.items(): ...
+for k, v in six.iteritems(adict): ...
```
andrey-churkin, line 204: I think we can remove it. `six` is a Python 2 and 3 compatibility library. We use only Python 3.

andrey-churkin (changes requested):

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+
+Be sure to use the right style for module, function, method docstrings and
+inline comments.
+
+<a id="s3.5.1-modules"></a>
+<a id="351-modules"></a>
+<a id="modules"></a>
+#### 3.5.1 Modules 
+
+Every file should contain license boilerplate.
```
andrey-churkin, line 467: license boilerplate --> a license boilerplate

andrey-churkin (changes requested):

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+    when loading. If True, the function will raise an exception if it cannot match the state_dict_to_load
+    parameters to the model's parameters (i.e. if some parameters required by model are missing in
+    state_dict_to_load, or if state_dict_to_load has parameters that could not be matched to model parameters,
+    or if the shape of parameters is not matching). If False, the exception won't be raised.
+    Usually is_resume is specified as False when loading uncompressed model's weights into the model with
+    compression algorithms already applied, and as True when loading a compressed model's weights into the model
+    with compression algorithms applied to evaluate the model.
+    :return: The number of state_dict_to_load entries successfully matched and loaded into model.
+    """
+```
```
andrey-churkin, line 517: I think it should be formatted because it is very hard to read

```python
def load_state(model: torch.nn.Module, state_dict_to_load: dict, is_resume: bool = False) -> int:
    """Used to load a checkpoint containing a compressed model into an NNCFNetwork object, but can
    be used for any PyTorch module as well. Will do matching of state_dict_to_load parameters to
    the model's state_dict parameters while discarding irrelevant prefixes added during wrapping
    in NNCFNetwork or DataParallel/DistributedDataParallel objects, and load the matched parameters
    from the state_dict_to_load into the model's state dict.

    :param model: The target module for the state_dict_to_load to be loaded to.
    :param state_dict_to_load: A state dict containing the parameters to be loaded into the model.
    :param is_resume: Determines the behavior when the function cannot do a successful parameter match
        when loading. If True, the function will raise an exception if it cannot match the state_dict_to_load
        parameters to the model's parameters (i.e. if some parameters required by model are missing in
        state_dict_to_load, or if state_dict_to_load has parameters that could not be matched to model parameters,
        or if the shape of parameters is not matching). If False, the exception won't be raised.
        Usually is_resume is specified as False when loading uncompressed model's weights into the model with
        compression algorithms already applied, and as True when loading a compressed model's weights into the 
        model with compression algorithms applied to evaluate the model.
    :return: The number of state_dict_to_load entries successfully matched and loaded into model.
    """
```

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+# BAD COMMENT: Now go through the b array and make sure whenever i occurs
+# the next element is i+1
+```
+
+<a id="s3.6-strings"></a>
+<a id="36-strings"></a>
+<a id="strings"></a>
+### 3.6 Strings 
+
+Use 'something' instead of "something"
```
andrey-churkin, line 584: Maybe
Use 'something' instead of "something" --> Use `'something'` instead of `"something"`

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+nncf_logger.info('Test message: %s', nncf.__version__)
+```
+
+<a id="s3.8-error-messages"></a>
+<a id="38-error-messages"></a>
+<a id="error-messages"></a>
+### 3.8 Error Messages 
+
+Error messages (such as: message strings on exceptions like `ValueError`, or
+messages shown to the user) should follow three guidelines:
```
andrey-churkin, line 630: Three? Only two were provided.

I think it is necessary to add that the message should start with a capital letter.

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+<a id="s3.9-todo-comments"></a>
+<a id="39-todo-comments"></a>
+<a id="todo-comments"></a>
+### 3.9 TODO Comments 
+
+Use `TODO` comments for code that is temporary, a short-term solution, or
+good-enough but not perfect.
+
+A `TODO` comment begins with the string `TODO` in all caps and a parenthesized
+name, e-mail address, or other identifier
```
andrey-churkin, line 643: other identifier --> another identifier

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,820 @@
...
+- `PTClassName` for Torch
+- `TFClassName` for TensorFlow
+
+<a id="s3.10.4-file-naming"></a>
+<a id="3104-file-naming"></a>
+<a id="file-naming"></a>
+#### 3.10.4 File Naming 
+
+Python filenames must have a `.py` extension and must not contain dashes (`-`).
+This allows them to be imported and unittested.
```
andrey-churkin, line 807: Maybe
unittested --> unit tested

alexsu52:
> 
> 
> I have a question about nested function call. For example:
> 
> ```python
> def k(input):
>     result = f(input)
>     return g(result)
> ```
> 
> or
> 
> ```python
> def k(input):
>     return g(f(input))
> ```
> 
> What is better?

docs/styleguide/PyGuide.md
alexsu52: cc' @vshampor

alexsu52: I've updated the formatting.

gadylshintr (approved):

vshampor (approved):

asenina (approved):

vanyalzr (approved):

ljaljushkin (approved):

docs/styleguide/PyGuide.md
```diff
@@ -0,0 +1,821 @@
...
+
+- Wildcard imports (`from module import *`) should be avoided, as they make it unclear which names are present 
+  in the namespace, confusing both readers and many automated tools.
+
+<a id="s2.3-3rd-party-packages"></a>
+<a id="23-3rd-party-packages"></a>
+<a id="3rd-party-packages"></a>
+### 2.3 3rd party packages 
+
+Do not add new third-party dependencies unless absolutely necessary. All things being equal, give preference to built-in packages.
```
AlexKoff88, line 129: I guess we should oblige people to provide a full list of code borrwings/new dependencies in the PR description when they introduce new PRs.

AlexKoff88 (commented):
Good stuff!

docs/styleguide/PyGuide.md
alexsu52: What do you think If I add this to the contribution guide?

andrey-churkin (approved):

PR #488, opened 2021-02-04
Force nncf build

aleksei-kashapov:
WIP: writing a test


nncf/extensions/__init__.py
```diff
@@ -1,5 +1,41 @@
...
+        if class_type.extension_type() != ext_type:
+            continue
+        class_type.load()
+
+
+def force_build_cpu_extensions():
+    _force_build_extensions(ExtensionsType.CPU)
+
+
+def force_build_cuda_extensions(arch_list=None):
```
vshampor, line 27: Should not give an option to the user here to set  - let them set the environment variables on their own.

nncf/extensions/__init__.py
```diff
@@ -1,5 +1,41 @@
...
+class ExtensionsType(enum.Enum):
+    CPU = 0
+    CUDA = 1
+
+
+def _force_build_extensions(ext_type):
+    for class_type in EXTENSIONS.registry_dict.values():
+        if class_type.extension_type() != ext_type:
+            continue
+        class_type.load()
```
vshampor, line 20: How do you even know that `class_type` will have a `.load()` method?

nncf/binarization/extensions.py
```diff
@@ -36,15 +34,34 @@
     os.path.join(BASE_EXT_DIR, "cuda/functions_cuda_impl.cu")
 ]
 
-BinarizedFunctionsCPU = load(
-    'binarized_functions_cpu', CPU_EXT_SRC_LIST, extra_include_paths=EXT_INCLUDE_DIRS,
-    verbose=False
-)
+
+@EXTENSIONS.register()
+class BinarizedFunctionsCPULoader:
```
vshampor, line 39: You should make this and other related classes implement a specific abstract class interface instead of just creating the same number of methods with the same signature.

BTW, if you fancy, you may try using metaclasses or `__init_subclass__` to handle registration automatically, but this is not a strict requirement - `.register()` already does the trick.

tests/test_extensions_build.py
```diff
@@ -0,0 +1,38 @@
...
+
+    subprocess.call("virtualenv -ppython{} {}".format(version_string, venv_path), shell=True)
+
+    subprocess.run(
+        "{python} {nncf_repo_root}/setup.py develop".format(
+            python=python_executable_with_venv,
+            nncf_repo_root=PROJECT_ROOT),
+        check=True,
+        shell=True,
+        cwd=PROJECT_ROOT)
```
vshampor, line 29: This is mostly a copy-and-paste of a code from `test_install.py`. Reuse this setup code using pytest fixtures.

vshampor (changes requested):

nncf/extensions/__init__.py
aleksei-kashapov: Done

tests/test_extensions_build.py
aleksei-kashapov: It has some similar lines, but test_extensions_build.py uses other arguments. So I don't see a reason to try to unite test_install.py and test_extensions_build.py it will save a few lines but will demand to create new functions/file

nncf/binarization/extensions.py
aleksei-kashapov: Implemened; please take a look at the updated commit

nncf/extensions/__init__.py
aleksei-kashapov: Now using an abstract class; Does this resolve an issue?

nncf/extensions/__init__.py
```diff
@@ -1,5 +1,43 @@
...
+    def extension_type():
+        pass
+
+    @staticmethod
+    @abstractmethod
+    def load():
+        pass
+
+
+def _force_build_extensions(ext_type):
```
vshampor, line 27: Type hints

tests/extensions_build_checks.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Set CUDA Architecture
+# See cmake/Modules_CUDA_fix/upstream/FindCUDA/select_compute_arch.cmake
+os.environ['TORCH_CUDA_ARCH_LIST'] = '7.5+PTX'
+force_build_cuda_extensions()
+
+cuda_ext_dir = (torch_ext_dir / 'quantized_functions_cuda')
+assert cuda_ext_dir.exists()
+cuda_ext_so = (cuda_ext_dir / 'quantized_functions_cuda.so')
+assert cuda_ext_so.exists()
```
vshampor, line 32: All assertions should be done in the controlling test code. This should merely run force_build functions.

vshampor (changes requested):

tests/test_extensions_build.py
vshampor: Creating a tmp_venv seems to be a common task in the install tests. Create a corresponding fixture and extract the tmp venv creation and activation code there.

nncf/extensions/__init__.py
aleksei-kashapov: Done

tests/extensions_build_checks.py
aleksei-kashapov: Done; please take a look

aleksei-kashapov:
Jenkins please retry a build

aleksei-kashapov:
This version was tested by me and Mateusz. So if there are no issues, please, merge it

tests/conftest.py
```diff
@@ -140,3 +142,41 @@ def onnx_dir(request):
 @pytest.fixture(scope="module")
 def ov_config_dir(request):
     return request.config.getoption("--ov-config-dir")
+
+
+@pytest.fixture(scope="function")
+def create_venv(install_type, tmp_path, package_type, venv_type):  # pylint:disable=redefined-outer-name
```
vshampor, line 148: Either remove the NNCF installation step from this fixture, or rename the fixture to `tmp_venv_with_nncf`

nncf/dynamic_graph/patch_pytorch.py
```diff
@@ -204,3 +204,30 @@ def unpatch_torch_operators():
...
+
+
+def patch_extension_build_function():
+    """
+    The function patches PyTorch and fix a bug inside CUDA extensions building;
+    The bug must be fixed with a new PyTorch 1.8.0
+    """
+    import torch.utils.cpp_extension
+
+    if torch.__version__ == '1.8.0':
```
vshampor, line 216: Should check for version larger than 1.8.0 instead of checking version equality.

vshampor (changes requested):

nncf/dynamic_graph/patch_pytorch.py
aleksei-kashapov: Torch developers could change function name or somehow change it, so I think that it is better to keep the versions in which we are sure

tests/conftest.py
aleksei-kashapov: Done

nncf/dynamic_graph/patch_pytorch.py
vshampor: [Semantic versioning](https://www.python.org/dev/peps/pep-0440/) in Python packages is a well-defined concept, and once a project starts using it, it rarely deviates from it. Expecting a format change here is unreasonable.

vshampor: It is more likely that during the next update we will transition directly to a torch version 1.8.1 should it be available at that moment, and your code may break.

aleksei-kashapov: Updated

aleksei-kashapov: Reupdated

vshampor (approved):

PR #491, opened 2021-02-08
The "concave" parameter is misleading

andrey-churkin:
Let's consider test cases from NNCF tests

https://github.com/openvinotoolkit/nncf/blob/452ff2d7714a4e3da4b5da6484a8a13e27cbb2cc/tests/sparsity/test_common.py#L150-L156

and draw these dependencies, namely

```python
epochs = [0, 1, 2, 3, 4]
concave = [0.1, 13 / 90, 25 / 90, 0.5, 0.5]  # "concave": true
convex = [0.1, 29 / 90, 41 / 90, 0.5, 0.5]  # "concave": false

# plt.plot(epochs, convex, '-ro', label='Convex')
plt.plot(epochs, concave, '-bo', label='Concave')
plt.legend()
plt.ylabel('Sparsity level')
plt.grid(True)
plt.xlabel('Epochs')
plt.savefig('concave.png')
```

I got the following charts:

**"concave": true**
concave

**"concave": false"**
convex

From these charts, we can see when we set the `"concave"` parameter to `true` we get the [convex function](https://en.wikipedia.org/wiki/Convex_function) and the [concave function](https://en.wikipedia.org/wiki/Concave_function) otherwise.

### DIFF
--- a/nncf/sparsity/schedulers.py
+++ b/nncf/sparsity/schedulers.py
@@ -72,7 +72,7 @@ class PolynomialSparseScheduler(SparsityScheduler):
     def __init__(self, sparsity_algo, params=None):
         super().__init__(sparsity_algo, params)
         self.power = self._params.get('power', 0.9)
-        self.concave = self._params.get('concave', False)
+        self.concave = self._params.get('concave', True)
         self._update_per_optimizer_step = self._params.get('update_per_optimizer_step', False)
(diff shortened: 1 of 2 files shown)


andrey-churkin:
cc' @alexsu52

vshampor (approved):

PR #507, opened 2021-02-16
Use a unique state for each scheduler

andrey-churkin:


tests/composite/test_composite_scheduler.py
```diff
@@ -0,0 +1,42 @@
...
+        for scheduler in schedulers:
+            composite_scheduler.add(scheduler)
+        return composite_scheduler
+
+    expected = _create_composite_scheduler([1, 2])
+    actual = _create_composite_scheduler([0, 0])
+    actual.load_state_dict(expected.state_dict())
+
+    for expected_child, actual_child in zip(expected.child_schedulers, actual.child_schedulers):
+        assert expected_child.delta == actual_child.delta
```
ljaljushkin, line 42: Did this test fail before your fix?

andrey-churkin: Yes. The reason is that we use one common dictionary for parameters of child schedulers. It is okay for `current_step` or `current_epoch` because it should be the same for all schedulers. But there is another possible case. For example, two schedulers have their own state parameters but these parameters have the same name. These parameters can have different values. The current approach does not work properly in this case.

Also, all child schedulers contain all state parameters as attributes (not just the ones required by them) after calling `PTCompositeCompressionScheduler.load_state_dict()` method in the current approach.

ljaljushkin (approved):

andrey-churkin:
Jenkins please retry a build

alexsu52:
@andrey-churkin Could you check that the current checkpoints is still loading?

andrey-churkin:
I moved these changes to https://github.com/openvinotoolkit/nncf/pull/511 because both PRs require checkpoints update. I close this PR.

PR #510, opened 2021-02-16
Common API for transformation commands and transformation layout

alexsu52:
Introduces the common API for transformation commands and transformation layout and implementation in the TF-backend.

### DIFF
--- a/beta/nncf/tensorflow/graph/model_transformer.py
+++ b/beta/nncf/tensorflow/graph/model_transformer.py
@@ -144,7 +144,7 @@ class TFModelTransformer(ModelTransformer):
     def _insert(self, target_point, insertion_objects):
         target_layer_name = self._get_target_layer_name(target_point)
 
-        if target_point.type == TargetType.LAYER_WEIGHT_OPERATION:
+        if target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
             weight_operations = [
                 WeightOperations(target_point.weights_attr_name, insertion_objects)]
             self._insert_weight_operations(target_layer_name, weight_operations)
@@ -165,7 +165,7 @@ class TFModelTransformer(ModelTransformer):
         weight_operations = []
         for cmd in commands:
             if cmd.type != TransformationType.INSERT or \
-                    cmd.target_point.type != TargetType.LAYER_WEIGHT_OPERATION:
+                    cmd.target_point.type != TargetType.OPERATION_WITH_WEIGHTS:
                 raise TypeError('Multiple insertion transform does not support command: '
                                 'command type - {}; target point type - {}'
                                 .format(cmd.type, cmd.target_point.type))
@@ -180,7 +180,7 @@ class TFModelTransformer(ModelTransformer):
     def _remove(self, target_point):
         target_layer_name = self._get_target_layer_name(target_point)
 
-        if target_point.type == TargetType.LAYER_WEIGHT_OPERATION:
+        if target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
             self._remove_weight_operation(
                 target_layer_name,
                 target_point.weights_attr_name,
--- a/beta/nncf/tensorflow/graph/model_transformer.py
+++ b/beta/nncf/tensorflow/graph/model_transformer.py
@@ -144,7 +144,7 @@ class TFModelTransformer(ModelTransformer):
     def _insert(self, target_point, insertion_objects):
         target_layer_name = self._get_target_layer_name(target_point)
 
-        if target_point.type == TargetType.WEIGHT_OPERATION:
+        if target_point.type == TargetType.LAYER_WEIGHT_OPERATION:
             weight_operations = [
                 WeightOperations(target_point.weights_attr_name, insertion_objects)]
             self._insert_weight_operations(target_layer_name, weight_operations)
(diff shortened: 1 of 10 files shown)


nncf/common/graph/transformations/commands.py
```diff
@@ -11,14 +11,17 @@
...
 from nncf.common.utils.ordered_enum import OrderedEnum
 
 
 class TransformationPriority(OrderedEnum):
     DEFAULT_PRIORITY = 0
+    FP32_TENSOR_STATISTICS_OBSERVATION = 1
     SPARSIFICATION_PRIORITY = 2
     QUANTIZATION_PRIORITY = 11
-    PRUNING_PRIORITY = 1
+    PRUNING_PRIORITY = 3
```
ljaljushkin, line 24: How does it work before with the first priority? We didn't check it in combination with other algorithms, did we?

nncf/common/graph/transformations/commands.py
```diff
@@ -31,17 +34,22 @@ class TargetType(OrderedEnum):
     LAYER = 0
     BEFORE_LAYER = 1
     AFTER_LAYER = 2
-    WEIGHT_OPERATION = 3
+    PRE_LAYER_OPERATION = 3
```
ljaljushkin, line 37: Can't see usage of `PRE_LAYER_OPERATION` and `POST_LAYER_OPERATION`. Is it just reserved?
Does it match to PRE_OP and POST_OP in NNCf for Pytorch?

beta/nncf/tensorflow/graph/model_transformer.py
```diff
@@ -144,7 +144,7 @@ def _get_target_layer_name(self, target_point):
     def _insert(self, target_point, insertion_objects):
         target_layer_name = self._get_target_layer_name(target_point)
 
-        if target_point.type == TargetType.WEIGHT_OPERATION:
+        if target_point.type == TargetType.LAYER_WEIGHT_OPERATION:
```
AlexKoff88, line 147: What is the intent of renaming this? To me, OPERATION_WITH_WEIGHTS sound more meaningful.

beta/nncf/tensorflow/graph/transformations/commands.py
```diff
@@ -0,0 +1,235 @@
...
+from typing import Any, Callable, List, Optional
+
+from nncf.common.graph.transformations.commands import TargetPoint
+from nncf.common.graph.transformations.commands import TargetType
+from nncf.common.graph.transformations.commands import TransformationCommand
+from nncf.common.graph.transformations.commands import TransformationPriority
+from nncf.common.graph.transformations.commands import TransformationType
+
+
+class TFLayerPoint(TargetPoint):
```
AlexKoff88, line 23: Can we have short descriptions of these classes that can explain the intent to newcomers? As far as I understand this defines spots for insertion of optimization method-specific operations.

nncf/common/graph/transformations/commands.py
alexsu52: I have two concerns which I address by changing PRUNING_PRIORITY from 1 to 3:
- FP32_TENSOR_STATISTICS_OBSERVATION and PRUNING_PRIORITY have the same priority. This can lead to non-deterministic behavior.
- When Pruning and Sparsity algorithms are stacked. My opinion is that the sparsity mask should be calculated for the pruned weights that have already been pruned by the filter pruning algorithm. Otherwise, the sparsity rate for all models in the sparsity algorithm will be incorrect.

This change influences only the interaction of the sparsity and pruning algorithms. I didn't check it in combination. But If you agree with my statement, I propose to prove this change.

nncf/common/graph/transformations/commands.py
alexsu52: Yes, It is reserved for NNCF with Pytorch backend.
NNCF_MODULE_PRE_OP -> PRE_LAYER_OPERATION 
NNCF_MODULE_POST_OP -> POST_LAYER_OPERATION 

The naming is open. Please, provide your variant. Also, I can exclude these types of target points from this PR. @ljaljushkin What do you think?

ljaljushkin: I am ok with naming. I was a bit confused to see duplication of similar entities in NNCF PyTorch, like InsertionType (for TransformationType) and OperationPriority(for TransformationPriority).

It was made intentionally, the next step is to re-implement PyTorch part with new classes, right?

nncf/common/graph/transformations/commands.py
ljaljushkin: Sounds reasonable. 
It makes sense to fix [test_priority](https://github.com/openvinotoolkit/nncf/blob/develop/tests/test_nncf_network.py#L257) with new behaviour then.

beta/nncf/tensorflow/graph/model_transformer.py
alexsu52: I agree, thanks. I have renamed it.

beta/nncf/tensorflow/graph/transformations/commands.py
alexsu52: Done

nncf/common/graph/transformations/commands.py
alexsu52: Yes, you are right.

nncf/common/graph/transformations/commands.py
alexsu52: I have added tests. Please take a look.

ljaljushkin (approved):

vshampor (approved):

PR #511, opened 2021-02-17
Use unified API for schedulers implementation

andrey-churkin:


andrey-churkin:
Should be merged after https://github.com/openvinotoolkit/nncf/pull/507

nncf/sparsity/schedulers.py
```diff
@@ -88,7 +88,7 @@ def epoch_step(self, next_epoch=None):
                 self._steps_per_epoch = self._steps_in_current_epoch
 
                 # Reset step and epoch step counters
-                next_epoch = -1
+                next_epoch = 0
```
ljaljushkin, line 91: let's create test for this or check whether existing one fails

nncf/sparsity/schedulers.py
```diff
@@ -150,6 +150,10 @@ def epoch_step(self, next_epoch=None):
     def current_sparsity_level(self):
         if self.sparsity_target_epoch == 0:
             return self.sparsity_target
+
+        if self.current_epoch == -1:
```
ljaljushkin, line 154: Should we do the same for other schedulers? Is there a test for this behaviour?

ljaljushkin (changes requested):

beta/examples/tensorflow/classification/main.py
```diff
@@ -115,12 +115,26 @@ def load_checkpoint(model, ckpt_path):
     return None
 
 
-def resume_from_checkpoint(model, compression_ctrl, ckpt_path, steps_per_epoch):
+def get_scheduler_state(num_steps, steps_per_epoch, config):
```
ljaljushkin, line 118: Shouldn't it be part of the scheduler API? As for me it looks complicated for the user to implement this. Moreover, this code looks identical for all samples.

IMO, it's a scheduler responsibility to create a scheduler state from some user parameters (e.g. num_steps, steps_per_epoch)

ljaljushkin (changes requested):

beta/examples/tensorflow/classification/main.py
```diff
@@ -115,12 +115,26 @@ def load_checkpoint(model, ckpt_path):
...
+
+
+def resume_from_checkpoint(model, compression_ctrl, ckpt_path, steps_per_epoch, config):
     if load_checkpoint(model, ckpt_path) == 0:
         return 0
     initial_step = model.optimizer.iterations.numpy()
     initial_epoch = initial_step // steps_per_epoch
-    compression_ctrl.scheduler.load_state(initial_step, steps_per_epoch)
+
+    scheduler_state = get_scheduler_state(initial_step, steps_per_epoch, config)
```
ljaljushkin, line 135: do you really need to calculate parameters for getting scheduler state? why not just load them from the checkpoint in a resume mode?

beta/examples/tensorflow/classification/main.py
andrey-churkin: It is a temporary solution. We are planning to save and load the scheduler's state to the checkpoint in the future. It will be implemented as separate PR.

beta/examples/tensorflow/classification/main.py
andrey-churkin: Thanks, it is a good idea. I create a separate issue (CVS-49551) to implement the saving and loading scheduler's state from the checkpoint. It will be implemented as separate PR.

beta/examples/tensorflow/classification/main.py
alexsu52: @andrey-churkin I support your plan, but anyway, could you create a temporal helper function that creates the scheduler state by input parameters.

nncf/api/compression.py
```diff
@@ -122,16 +122,14 @@ def epoch_step(self, next_epoch: Optional[int] = None) -> None:
             next_epoch = self.current_epoch + 1
         self.current_epoch = next_epoch
 
-    def load_state(self, step: int, epoch: int) -> None:
+    def load_state(self, state: Dict[str, int]) -> None:
```
alexsu52, line 125:
```suggestion
    def load_state(self, state: Dict[str, object]) -> None:
```

nncf/api/compression.py
```diff
@@ -122,16 +122,14 @@ def epoch_step(self, next_epoch: Optional[int] = None) -> None:
...
 
-        :param step: The index of the global training step.
-        :param epoch: The training epoch index.
+        :param state: Output of `get_state()` method.
         """
-        self.current_step = step
-        self.current_epoch = epoch
+        self.__dict__.update(state)
 
     def get_state(self) -> Dict[str, int]:
```
alexsu52, line 134:
```suggestion
    def get_state(self) -> Dict[str, object]:
```

nncf/api/compression.py
```diff
@@ -122,16 +122,14 @@ def epoch_step(self, next_epoch: Optional[int] = None) -> None:
...
         Loads the compression scheduler state, but does not update the state of the
         compression method.
 
-        :param step: The index of the global training step.
-        :param epoch: The training epoch index.
+        :param state: Output of `get_state()` method.
         """
-        self.current_step = step
-        self.current_epoch = epoch
+        self.__dict__.update(state)
```
alexsu52, line 132: It looks like a potential bug.

andrey-churkin: What is the problem?

beta/examples/tensorflow/classification/main.py
andrey-churkin: Done

nncf/sparsity/schedulers.py
andrey-churkin: Yes, we should. I suggest doing it in a separate pull request. For example, when we will move all schedulers implementation to a common part.

nncf/sparsity/schedulers.py
andrey-churkin: The test already exists. The reason for these changes is removing this code
`
return 0 if self._current_epoch == -1 else self._current_epoch
`

nncf/api/compression.py
alexsu52: The explicit assignment is preferable to the implicit assignment. The state value is an object, at least you should make a copy. But I believe the explicit assignment is better.

vshampor (approved):

nncf/api/compression.py
andrey-churkin: Done

nncf/sparsity/schedulers.py
ljaljushkin: I can't find the test and how it was adapted to the new behaviour in your PR. Could you point to the source code of it?

nncf/sparsity/schedulers.py
ljaljushkin: Could you create a follow-up ticket for this then?

alexsu52 (approved):

vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

nncf/sparsity/schedulers.py
andrey-churkin: I created a separate story for this task. (Ref: 49837)

nncf/sparsity/schedulers.py
andrey-churkin: My PR does not change previous behavior. Everything works as before.

https://github.com/openvinotoolkit/nncf/blob/2b4a93a9ff3e3c44e26a9eee8275462204a732f7/tests/sparsity/test_common.py#L200-L202

ljaljushkin (approved):

andrey-churkin:
@vshampor All questions were resolved. Could you merge it, please?

PR #512, opened 2021-02-17
Add AutoQ documentation

vuiseng9:


vuiseng9:
@ljaljushkin Could you review? :)

ljaljushkin:
LGTM
@AlexKoff88 please also review

docs/Algorithms.md
```diff
@@ -11,6 +11,7 @@ Each compression method receives its own hyperparameters that are organized as a
   - Mixed-bitwidth quantization
   - Automatic bitwidth assignment based on HAWQ
   - Automatic quantization parameter selection and activation quantizer setup based on HW config preset
+  - Automatic bitwidth assignment based on AutoQ, a Deep Reinforcement Learning algorithm to select best mixed precision given quality metric and HW type.
```
AlexKoff88, line 14: Suggest changing `...on AutoQ, a Deep...` to `...on AutoQ - a Deep...`

AlexKoff88 (commented):
Requirements look goof to me from the POT standpoint. I guess we need to align and fix versions (`==`) for all the components. With this change, we can guarantee the results on validated platforms.

docs/compression_algorithms/Quantization.md
```diff
@@ -230,6 +231,56 @@ file.
 
 ---
 
+#### AutoQ
+Different model architectures possess varying redundancy and sensitivity towards accuracy. Different device architecture may provide different performance capabilities. A mixed-precision model performs efficiently on a device may behave differently on another. It is a scaling challenge to find respective optimal precision configuration for a range of platforms (CPU, GPU, VPU etc and SKUs). To this end, NNCF provides an alternate mode, namely AutoQ, for mixed-precision automation. Based on HAQ, it employs deep reinforcement learning algorithm to automate the explorations and eventually learn the quantization policy that achieves optimal trade-off between accuracy, performance and user-specified constraints (e.g. model size, latency, compute complexity, power).
```
AlexKoff88, line 235: It looks like good stuff for the blogpost ) But can we make it shorter in the documentation? Like,
"AutoQ provides an alternate mode for mixed-precision quantization that utilizes an actor-critic algorithm, Deep Deterministic Policy Gradient (DDPG) for efficient search over the bitwidth space..."
It is probably worth pointing to the paper where the original implementation is disclosed.

docs/Algorithms.md
vuiseng9: Pushed the following description, imo we should reference HAQ here too.

- Automatic bitwidth assignment mode AutoQ, based on HAQ, a Deep Reinforcement Learning algorithm to select best mixed precision given quality metric and HW type.

docs/compression_algorithms/Quantization.md
vuiseng9: Condensed first two paragraphs into one. ) The remaining paragraphs contain algorithm description, keeping them for the documentation. 

Yes, that was a miss. Link to HAQ is embedded now.

AlexKoff88 (approved):

PR #517, opened 2021-02-19
Added wheel install in venv

pfinashx:
Since in docker was added python3.7-venv, all install tests passed, except of bdist_wheel setup.
While creating venv virtual environment wheel does not install automatically, like in virtualenv virtual environment, so it need install manually.


pfinashx:
Test: http://213.221.44.203/job/nncf/job/precommit/job/manual_precommit/233/

aleksei-kashapov (approved):
Good job! Thank you for your contribution!

PR #518, opened 2021-02-20
Adjust Padding feature to support accurate execution of U4 on VPU

ljaljushkin:
Tickets: 44506

better to merge after https://github.com/openvinotoolkit/nncf/pull/532


ljaljushkin:
Jenkins please retry a build

ljaljushkin:
Jenkins please retry a build

ljaljushkin:
@vshampor,@alexsu52 and @AlexKoff88 please review

nncf/hw_configs/vpu.json
```diff
@@ -35,6 +35,9 @@
     "operations": [
         {
             "type": "Convolution",
+            "attributes": {
+                "adjust_padding": true
```
AlexKoff88, line 39: The problem I see is that there is no connection of this attribute with the precision that it has an impact on.

ljaljushkin: the precision would be not enough, need to connect with a specific quantization config:
```
 "bits": 4,
 "mode": "symmetric",
 "granularity": "pertensor"
 "signed": true
 ```
I supposed we won't need this feature for any other combinations in a future. It's a current workaround for a specific case of U4 activations, which most probably will be deleted. Do you really think we should invest more time to make it flexible for other combinations?

**Comment by ljaljushkin** [comment]
any objections to merge?

File: nncf/quantization/adjust_padding.py [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+
+
+class AdjustPaddingArgs(NamedTuple):
+    weight_bitwidth: int
+    activation_quantizer: BaseQuantizer
+    quantized_module: nn.Module
+    module_scope: Scope
+
+
+class AdjustPadding:
```
vshampor, line 38: Since it does not actually adjust padding values, should probably rename it to `CalculatePaddingAdjustment`.

nncf/quantization/algo.py
```diff
@@ -994,6 +1001,40 @@ def _add_single_weight_quantizer(self, target_model: NNCFNetwork, insertion_poin
         command = InsertionCommand(insertion_point, op, OperationPriority.QUANTIZATION_PRIORITY)
         return quantizer_id, command
 
+    def _get_adjust_padding_args(self, quantization_point: SingleConfigQuantizationPoint,
+                                 quantizer_module_id: NonWeightQuantizerId,
+                                 target_model: NNCFNetwork,
+                                 quantization_points: List[SingleConfigQuantizationPoint]):
+        result = []
+        if quantization_point.is_adjust_padding_applicable:
```
vshampor, line 1009: As discussed, should remove the `is_adjust_padding_applicable` from quantization points and instead determine a matching between a weight-quantized module and its associated activation quantizer from the rest of the quantizer setup.

nncf/quantization/quantizer_setup.py
```diff
@@ -1,20 +1,31 @@
...
+from nncf.tensor_statistics.statistics import TensorStatistic
 from nncf.utils import get_scale_shape
 
 QuantizationPointId = int
 
 
 class QuantizationPointBase:
-    def __init__(self, insertion_point: InsertionPoint):
+    def __init__(self, insertion_point: InsertionPoint,
+                 scopes_of_directly_quantized_operators=None, is_adjust_padding_applicable=True):
```
vshampor, line 23: `scopes_of_directly_quantized_operators` should be a required parameter - it may be set to an empty list to signify the situation when a quantizer really does not affect any node in the NNCFGraph (such as when an output quantizer is being set up), but otherwise this information should be valid at all times.

vshampor (changes requested):

nncf/layers.py
```diff
@@ -60,6 +90,22 @@ def from_module(module):
         dict_update(nncf_conv.__dict__, module.__dict__)
         return nncf_conv
 
+    def custom_forward(self, input_):
```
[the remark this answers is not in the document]
ljaljushkin: `forward` of _NNCFModuleMixin won't be called then

nncf/layers.py
```diff
@@ -47,9 +50,36 @@ def from_module(module):
...
+        padding=0,
+        dilation=1,
+        groups: int = 1,
+        bias: bool = True,
+        padding_mode: str = 'zeros'
+    ):
+        super().__init__(in_channels, out_channels, kernel_size, stride, padding, dilation, groups,
+                         bias, padding_mode)
+        self._set_padding_value(Parameter(torch.zeros([1]), requires_grad=False))
+        self.custom_forward_fn_ = self.custom_forward.__func__
```
vshampor, line 75: You are actually setting an instance attribute here. The rest of the code seems to be written as if the `custom_forward_fn` should be a class attribute. I propose keeping it a class attribute, if this does not break anything, i.e. just do `def custom_forward_fn_` in the NNCF... classes. Perhaps should also rename `custom_forward_fn_` to `_custom_forward_fn`.

vshampor (changes requested):

nncf/quantization/algo.py
ljaljushkin: Done

nncf/layers.py
ljaljushkin: Done

nncf/layers.py
```diff
@@ -12,16 +12,19 @@
...
 import torch
 import torch.nn.functional as F
 import warnings
 from torch import nn
+from torch.nn import Parameter
 from torch.nn import init
 from torch.nn.utils.rnn import PackedSequence
 
 from nncf.common.utils.registry import Registry
+from .checkpoint_loading import OPTIONAL_PARAMETERS_REGISTRY
```
vshampor, line 27: Please use absolute imports

nncf/nncf_network.py
```diff
@@ -672,18 +684,22 @@ def get_insertion_point_graph(self) -> InsertionPointGraph:
...
-                module = self.get_module_by_scope(scope)
-                if module is not None:
-                    subtype = op_arch.determine_subtype(containing_module=module)
-                    if subtype is not None:
-                        op_arch = subtype
+                op_arch = self.get_op_arch_by_graph_node(nncf_graph_node_ref)
                 ip_graph_node[InsertionPointGraph.OPERATOR_METATYPE_NODE_ATTR] = op_arch
         return ip_graph
 
+    def get_op_arch_by_graph_node(self, nncf_graph_node_ref):
```
vshampor, line 691: Add type hints.

nncf/nncf_network.py
```diff
@@ -672,18 +684,22 @@ def get_insertion_point_graph(self) -> InsertionPointGraph:
...
+        op_exec_context = nncf_graph_node_ref[NNCFGraph.OP_EXEC_CONTEXT_NODE_ATTR]
+        op_name = op_exec_context.operator_name
+        scope = op_exec_context.scope_in_model
+        op_arch = OPERATOR_METATYPES.get_operator_metatype_by_op_name(op_name)
```
vshampor, line 695: Will OperatorSubtype resolution work here? For instance, it is impossible to tell DW-conv and conv apart based on the function name only, you need the module for that.

nncf/quantization/algo.py
```diff
@@ -1062,8 +1106,56 @@ def _build_insertion_commands_list_for_quantizer_setup(self,
...
+                                                                qp_id_vs_quant_module_id_dict, quantizer_setup,
+                                                                target_model)
+
+        commands = self._add_adjust_padding_ops(adjust_padding_args, target_model)
+        if commands:
+            insertion_commands += commands
+
         return insertion_commands, qp_id_vs_quant_module_id_dict
 
+    def _collect_adjust_padding_args(self, non_unified_scales_quantization_point_ids, qp_id_vs_quant_module_id_dict,
```
vshampor, line 1120: Type hints

nncf/quantization/algo.py
```diff
@@ -1062,8 +1106,56 @@ def _build_insertion_commands_list_for_quantizer_setup(self,
...
+        for wqp_id in weight_qps:
+            wqp = quantizer_setup.quantization_points[wqp_id]
+            ip = wqp.insertion_point
+            module_scope = ip.module_scope
+            nncf_graph = target_model.get_original_graph()
+            graph_node = nncf_graph.find_node_in_nx_graph_by_scope(module_scope)
+            if not graph_node:
+                raise RuntimeError(f'Internal error: failed to find node for by scope={module_scope}')
+
+            op_arch = target_model.get_op_arch_by_graph_node(graph_node)
```
vshampor, line 1140: Check the `NNCFNetwork.get_insertion_point_graph()` and see if you can either reuse the info from `InsertionPointGraph`, or the code inside it that determines correct metatypes for a node.

vshampor (changes requested):

nncf/nncf_network.py
ljaljushkin: subtype resolution happens after that:
https://github.com/openvinotoolkit/nncf/pull/518/files/8244eb755114545fe58c6bef20cb69f5298ea80d#diff-acec2cd12abd71c59568f29765b23819248f804127cf70a32640c80b9ab733dbR696-R701

vshampor: Talk about selective vision XD

vshampor (approved):

PR #526, opened 2021-02-25
Fix bug with cuda kernel

aleksei-kashapov:
If input tensor has only 1 channel incorrect shape of scale tensor is chosen;

### DIFF
--- a/nncf/extensions/src/quantization/cuda/functions_cuda_impl.cu
+++ b/nncf/extensions/src/quantization/cuda/functions_cuda_impl.cu
@@ -39,6 +39,10 @@ ScaleType get_scale_type(const at::Tensor& input, const at::Tensor& input_low, c
             TORCH_CHECK(input_range.size(1) == scale_count, "Scale shape is not flat");
             return  ScaleType::PER_ACTIVATION_CHANNEL;
         }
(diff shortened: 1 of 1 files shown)


aleksei-kashapov:
@vshampor please take a look

nncf/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -39,6 +39,7 @@ ScaleType get_scale_type(const at::Tensor& input, const at::Tensor& input_low, c
             TORCH_CHECK(input_range.size(1) == scale_count, "Scale shape is not flat");
             return  ScaleType::PER_ACTIVATION_CHANNEL;
         }
+        return ScaleType::PER_WEIGHT_CHANNEL;
```
vshampor, line 42: Shouldn't skip the TORCH_CHECKS like that. Please add ` || scale_count == 1` in the condition on line 30 instead.

vshampor (changes requested):

nncf/extensions/src/quantization/cuda/functions_cuda_impl.cu
aleksei-kashapov: Update to a discussed version

vshampor (approved):

PR #529, opened 2021-02-25
YOLOv4 functional support

negvet:


beta/examples/tensorflow/common/object_detection/architecture/darknet.py
```diff
@@ -0,0 +1,65 @@
...
+        no_bias_kwargs = {'use_bias': False}
+        no_bias_kwargs.update(kwargs)
+        return nn_ops.compose(
+            nn_ops.DarknetConv2D(*args, **no_bias_kwargs),
+            tf.keras.layers.experimental.SyncBatchNormalization(),
+            tf.keras.layers.Activation(self.mish))
+
+
+    def csp_resblock_body(self, x, num_filters, num_blocks, all_narrow=True):
+        '''A series of resblocks starting with a downsampling Convolution2D'''
```
alexsu52, line 37: Could you please use one format for docstrings in the example code. I mean using or `"""` or `'''`.

beta/examples/tensorflow/object_detection/models/yolo_v4_config.py
```diff
@@ -0,0 +1,26 @@
...
+
+from beta.examples.tensorflow.common.sample_config import SampleConfig
+
+YOLOV4_CONFIG = SampleConfig({
+    "num_feature_layers": 3,
+    "num_classes": 80,
+    "model_params": {
+        "architecture": {
+            "backbone": {
+                "name": "darknet"
```
alexsu52, line 22: Does Yolo v4 support another backbone? I mean we have resnet-50 backbone.  What happens if I set "resnet" as the backbone in the config.

negvet: It is possible to enable support for other backbones in the future. For now, we support only the darknet backbone. Fix: the ValueError will raise in the case of the other backbone is set.

alexsu52 (approved):

PR #532, opened 2021-02-26
Optional parameters support on loading PT checkpoint

ljaljushkin:
Also refactored load_data functionality

This feature is needed for #518 to not update all checkpoints when new parameter is introduced.
Instead, if old checkpoint doesn't have this new optional attribute, it can be successfully loaded even in a resume mode. The new attribute will be initialized by default (e.g. padding_value=0)

### DIFF
--- a/nncf/checkpoint_loading.py
+++ b/nncf/checkpoint_loading.py
@@ -1,5 +1,5 @@
 """
- Copyright (c) 2019-2020 Intel Corporation
+ Copyright (c) 2019-2021 Intel Corporation
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
  You may obtain a copy of the License at
@@ -10,12 +10,15 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+from enum import Enum
+from typing import Dict
+from typing import List
+from typing import Set
 
 import re
 import torch
 
 from nncf.common.utils.logger import logger as nncf_logger
-from nncf.nncf_network import MODEL_WRAPPED_BY_NNCF_ATTR_NAME
 
 
 def load_state(model: torch.nn.Module, state_dict_to_load: dict, is_resume: bool = False) -> int:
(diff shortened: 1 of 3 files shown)


nncf/checkpoint_loading.py
```diff
@@ -38,97 +41,154 @@ def load_state(model: torch.nn.Module, state_dict_to_load: dict, is_resume: bool
...
+    key_matcher.handle_problematic_keys()
     nncf_logger.info("Loaded {}/{} layers".format(num_loaded_layers, len(model_state_dict.items())))
 
     model.load_state_dict(new_dict, strict=False)
     return num_loaded_layers
 
 
-def process_problematic_keys(is_resume, issues, is_all_saved_loaded):
-    error_msgs = []
+class OptionalParametersRegistry:
```
vshampor, line 57: Do any of the current parameters in our checkpoints fall into the category of "optional"?

vshampor (changes requested):
Please add class descriptions via docstrings to conform to the code style guidances.

nncf/checkpoint_loading.py
ljaljushkin: not now, but I think it makes sense to add 'enabled' and '_num_bits' of Quantizer. I will add them.

If there're any other candidates of the optional parameters (e.g. in pruning? @mkaglins, @AlexanderDokuchaev), please let me know or feel free to register throw `OptionalParametersRegistry`

nncf/checkpoint_loading.py
```diff
@@ -38,97 +41,154 @@ def load_state(model: torch.nn.Module, state_dict_to_load: dict, is_resume: bool
...
+        all_processed_keys = []
+        params_suffixes_to_skip = tuple(OPTIONAL_PARAMETERS_REGISTRY.get_optional_parameters_suffixes())
+        for keys in self._keys.values():
+            all_processed_keys.extend(keys)
+
+        for key in model_state_dict.keys():
+            if key not in all_processed_keys:
+                if key.endswith(params_suffixes_to_skip):
+                    self.add_key(key, ProcessedKeyStatus.SKIPPED)
+                    nncf_logger.warning("The optional parameter {} is missed in the loaded state".format(key))
```
ljaljushkin, line 101: @vshampor warning 👀

vshampor (changes requested):
@ljaljushkin please add type hints to the functions, as well as the function argument and return value descriptions in the docstrings conforming to the agreed-upon format.

vshampor (approved):

PR #534, opened 2021-02-28
Common transformations

vshampor:


nncf/common/graph/transformations/commands.py
```diff
@@ -31,14 +31,19 @@ class TransformationType(OrderedEnum):
...
+    BEFORE_LAYER = 1  # additional TF-node in TF-graph before TF-layer op
+    AFTER_LAYER = 2  # additional TF-node in TF-graph after TF-layer op
+
+    # These three add a TF-operation in TF, not a TF-node. In PT, these map to module pre- and post-ops.
+    PRE_LAYER_OPERATION = 3  # A pre-operation that can access attrs of the associated PT-module or TF-layer
+    POST_LAYER_OPERATION = 4  # A post-operation that can access attrs of the associated PT-module or TF-layer
+    OPERATION_WITH_WEIGHTS = 5  # Same as PRE_LAYER_OPERATION, but targets weights of the layer/module specifically
+
+    OPERATOR_PRE_HOOK = 6  # A pre-operation in PT that does not in general access module attrs - N/A in TF
+    OPERATOR_POST_HOOK = 7  # A post-operation in PT that does not in general access module attrs - N/A in TF
```
alexsu52, line 46: It looks like these comments are not a docstring and could not be added in the documentation automatically way if we will decide to do it. I mean `TargetType.__doc__` does not have any description. I can suggest adding a description as the docstring  at the beginning of the class:
```
class TargetType(OrderedEnum):
    """
    Enumeration of the target point types.

    Target point types define the meaning and operations that can be performed
    on the target point in the model graph.

    `LAYER` - add, remove or update a TF Keras Layer.  Supports only in the TF-backend.
    ....
    """
```
Of course, you can add a docstring for each type, but, as I know, It requires additional implementation. 

Could you add a description for other enumerations in the commands.py.

beta/tests/tensorflow/test_transformations.py
```diff
@@ -71,7 +71,7 @@ def test_removal_command_union():
 
 def test_add_insertion_command_to_multiple_insertion_commands_same():
     check_fn = lambda src, dst: \
-        dst.type ==  TargetType.OPERATION_WITH_WEIGHTS and \
+        dst.type == TargetType.OPERATION_WITH_WEIGHTS and \
```
AlexKoff88, line 74: I don't like the name `TargetType` because it does not have any sense to me at least. Can we come up with something more self-explainable? The previous name was better, IMHO.
NNCF code is getting more and more complicated. Providing self-explainable names to basic entities and classes is one way to simplify it.

nncf/dynamic_graph/transformations/commands.py
```diff
@@ -0,0 +1,62 @@
...
+
+from nncf.common.graph.transformations.commands import TargetPoint
+from nncf.common.graph.transformations.commands import TargetType
+from nncf.common.graph.transformations.commands import TransformationCommand
+from nncf.common.graph.transformations.commands import TransformationPriority
+from nncf.common.graph.transformations.commands import TransformationType
+from nncf.dynamic_graph.graph import InputAgnosticOperationExecutionContext
+
+
+class PTInsertionPoint(TargetPoint):
```
AlexKoff88, line 11: There is no PyTorch specific call/class in this file so I wonder why everything is called `PT...`?

beta/tests/tensorflow/test_transformations.py
alexsu52: Let's discuss it offline and try to find a more simple and suitable solution.

beta/nncf/tensorflow/api/compression.py
```diff
@@ -89,8 +89,11 @@ def apply_to(self, model: ModelType) -> ModelType:
         :return: The model with additional modifications necessary to enable
             algorithm-specific compression during fine-tuning.
         """
+
```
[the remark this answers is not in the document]
vshampor: Done

nncf/common/graph/transformations/commands.py
vshampor: Done

beta/tests/tensorflow/test_transformations.py
vshampor: The problem here is that NNCF TF operation seems to be planning to use "removal" commands in addition to "insertion" operations, so `InsertionType` could no longer be used as a single general enum applicable both for PT and TF.

nncf/dynamic_graph/transformations/commands.py
vshampor: The InputAgnosticOperationExecutionContext right now is a PT-specific entity, and other entities in this file are highly dependent on this at the moment.

vshampor:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

beta/tests/tensorflow/test_transformations.py
AlexKoff88: Let revise it in the future.

nncf/dynamic_graph/transformations/commands.py
AlexKoff88: @vshampor, do you mean that it will be revised during one of the future steps of integration?

vshampor: Sure, we are moving towards a state where all framework-specific entities are collocated in framework-specific folders, but we are doing so gradually.

nncf/common/graph/transformations/commands.py
```diff
@@ -17,20 +17,47 @@
...
-    MULTI_INSERT = 1
-    REMOVE = 2
-
-
 class TargetType(OrderedEnum):
+    """
+    Describes the types of actions that can be applied to the NNCF internal graph representation
+    of the compressed model.
+
+    Definitions used below: TF-nodes - visible in TF-graph, TF-operations - invisible in TF-graph
```
alexsu52, line 46: The description looks confusing to me. Some points are below: 
- TargetType is not a type of action, but It is an object or location where the action will be applied.
- What is TF-graph? It can be confused for developers because TF has several graphs DefGraph, KerasGraph and etc.
- What are visible and invisible nodes?
- What do you mean by hooking in TF?

nncf/common/graph/transformations/commands.py
```diff
@@ -17,20 +17,47 @@
...
+    'compressed' model, i.e. a model that supports compression-aware training and
+    export. Certain compression algorithms may need to act on one and the same
+    spot in the model's control flow graph, and proper priority-based ordering is
+    required for functionally correct results that are invariant w.r.t. compression
+    algorithm application ordering.
+
+    Rationales:
+    * quantization should occur after sparsification/pruning, otherwise the dependency
+    of sparsity level on the value threshold becomes discontinuous which impacts the
+    corresponding algos.
```
alexsu52, line 32:
```suggestion
    corresponding algorithms.
```

nncf/common/graph/transformations/commands.py
```diff
@@ -41,6 +68,16 @@ class TargetType(OrderedEnum):
     OPERATOR_POST_HOOK = 7
 
 
+class TransformationType(OrderedEnum):
+    """
+    Defines the sub-types of transformations w.r.t. `TargetType`.
```
alexsu52, line 73:
```suggestion
    Defines the types of transformations w.r.t. `TargetType`.
```

nncf/common/graph/transformations/commands.py
vshampor: Done

nncf/common/graph/transformations/commands.py
vshampor: Done

nncf/common/graph/transformations/commands.py
vshampor: I modified the description to resolve some of your bullets here, but my knowledge of TF graph and node specifics is lacking in order to give a correct description of TF entities. Please suggest exact wording of the descriptions, containing exact TF entity identifiers (such as qualified class names), so that there is no ambiguity in the description. I tried to follow the existing code to try and obtain this knowledge on my own, but either the code is too convoluted, or it does not explicitly involve "operations" or "nodes".

vshampor:
Jenkins please retry a build

nncf/common/graph/transformations/commands.py
```diff
@@ -17,20 +17,53 @@
...
+                            for which the local attributes of said PT-module or TF-layer are accessible
+    `OPERATION_WITH_WEIGHTS` - same as PRE_LAYER_OPERATION, but targets weights of the layer/module specifically
+    `OPERATOR_PRE_HOOK` - a location before a function call in PT without access to specific
+                          module attributes - N/A in TF
+    `OPERATOR_POST_HOOK` - a location after a function call in PT without access to specific
+                           module attributes - N/A in TF
+
+    Notes: `PRE_LAYER_OPERATION`, `POST_LAYER_OPERATION` and `OPERATION_WITH_WEIGHTS` add a TF-operation in TF, not a
+            TF-node. In PT, these map to module pre- and post-ops.
+    """
```
alexsu52, line 66:
```suggestion
    """
    Describes the types of locations in the model that can be modified using NNCF
    in order to create a compressed model.

    `LAYER` - a location corresponding directly to an existing layer in the model
    `BEFORE_LAYER` - a location before the associated model layer,
                     implemented by inserting an additional layer in the TF model
    `AFTER_LAYER` - a location after the associated model layer, 
                    implemented by inserting an additional layer in the TF model
    `PRE_LAYER_OPERATION` - a location before the associated PT-module or TF-layer
                            execution, for which the local attributes of said 
                            PT-module or TF-layer are accessible
    `POST_LAYER_OPERATION` - a location before the associated PT-module or TF-layer 
                             execution, for which the local attributes of said 
                             PT-module or TF-layer are accessible
    `OPERATION_WITH_WEIGHTS` - same as PRE_LAYER_OPERATION, but targets weights 
                               of the layer/module specifically
    `OPERATOR_PRE_HOOK` - a location before a function call in PT without access to 
                          specific module attributes - N/A in TF
    `OPERATOR_POST_HOOK` - a location after a function call in PT without access to 
                           specific module attributes - N/A in TF

    Notes: Adding operations to a PT-module or TF-layer implemented by wrapping 
    the original PT-module or TF-layer and registering operations that are executed 
    before/after calling the original PT-module or TF-layer according to the 
    registration location:`PRE_LAYER_OPERATION`, `POST_LAYER_OPERATION` and 
    `OPERATION_WITH_WEIGHTS`.
    """
```

alexsu52 (approved):

PR #535, opened 2021-02-28
Fix using tmp venv with NNCF installed in tests

vshampor:


aleksei-kashapov (commented):
Why don't you like the name?

tests/conftest.py
```diff
@@ -145,7 +145,7 @@ def ov_config_dir(request):
 
 
 @pytest.fixture(scope="function")
-def create_tmp_venv_with_nncf(install_type, tmp_path, package_type, venv_type):  # pylint:disable=redefined-outer-name
+def tmp_venv_with_nncf(install_type, tmp_path, package_type, venv_type):  # pylint:disable=redefined-outer-name
```
aleksei-kashapov, line 148: Why don't you like the name?

tests/test_extensions_build.py
```diff
@@ -10,13 +10,13 @@
...
                          [('venv', 'develop', 'GPU')])
-def test_force_cuda_build(create_tmp_venv_with_nncf, install_type, tmp_path, package_type):
+def test_force_cuda_build(tmp_venv_with_nncf, install_type, tmp_path, package_type):
     '''
     Check that CUDA Extensions weren't initially built and \
     then with TORCH_CUDA_ARCH_LIST were forced to be built
     '''
 
-    venv_path = tmp_path / 'venv'
+    venv_path = tmp_venv_with_nncf
```
aleksei-kashapov, line 19: Anyway we have `tmp_path` as input. It is better to use only one way `tmp_path` or return value of `tmp_venv_with_nncf`

tests/conftest.py
vshampor: The built-in fixtures such as `tmp_path` and `tmp_dir` do not have `create` in them, so I decided to go for consistency here.

tests/test_extensions_build.py
vshampor: The tmp path with venv and the tmp path with the checks scripts should not necessarily be the same. In fact, it is even beneficial this way - less chance of common PYTHONPATH influence.

aleksei-kashapov (approved):

PR #543, opened 2021-03-03
Corrected bitwidth graph dump for HAWQ

ljaljushkin:


nncf/quantization/algo.py
```diff
@@ -1764,6 +1762,12 @@ def get_quantizer_setup_for_current_state(self) -> SingleConfigQuantizerSetup:
             retval.quantization_points[qp_id] = new_qp
         return retval
 
+    def apply_quantizer_setup_to_current_state(self, quantizer_setup: SingleConfigQuantizerSetup):
```
vshampor, line 1765: But we have `apply_new_quantizer_setup` for that.

vshampor (changes requested):

nncf/quantization/algo.py
ljaljushkin: IMHO, `apply_new_quantizer_setup` is not  much for that. It creates builder, controller... why not just map?

vshampor: Your `apply_new_quantizer_setup_to_current_state` does not actually apply the setup, it merely sets up bitwidths while leaving the rest of the quantizer configuration as-is, which leads to a model ending up in an inconsistent state with respect to the quantizer configuration constraints.

ljaljushkin: ok, makes sense. used `apply_new_quantizer_setup`

vshampor (approved):

PR #545, opened 2021-03-03
Support pruning for Instance Normalization

AlexanderDokuchaev:


nncf/dynamic_graph/graph.py
```diff
@@ -172,6 +171,10 @@ class BaseModuleAttributes:
     def __init__(self, weight_requires_grad: bool):
         self.weight_requires_grad = weight_requires_grad
 
+    @staticmethod
+    def from_module(module):
+        return BaseModuleAttributes(module.weight.requires_grad)
```
vshampor, line 176: Type hints will be necessary, ultimately, which will result in the explicit dependency of this class on PyTorch. @gadylshintr , do we need this dependency?

gadylshintr: Currently I use the same `ModuleAttributes` classes for Pytorch and TF, hence I'd prefer to get rid of this dependency if possible.

AlexanderDokuchaev: Added `_get_module_attributes` in nncf_network.py

vshampor (approved):

gadylshintr (approved):

PR #553, opened 2021-03-05
Mark module attributes at wrap_operator stage

vshampor:


nncf/dynamic_graph/graph.py
```diff
@@ -200,9 +200,11 @@ class NNCFNode:
     def __init__(self,
                  node_id: int,
                  op_exec_context: OperationExecutionContext,
-                 module_attributes: BaseModuleAttributes = None):
+                 module_attributes: BaseModuleAttributes = None,
+                 op_type: 'OperatorMetatype' = None):
         self.node_id = node_id
         self.op_exec_context = op_exec_context
+        self.op_type = op_type  # type: OperatorMetatype
```
vshampor, line 207: @gadylshintr your PR would probably need to be adjusted so that your `type` attribute is an OperatorMetatype. What do you think?

nncf/quantization/algo.py
```diff
@@ -1140,8 +1140,8 @@ def weight_qp_filter_fn(qp_id_):
             if not graph_node:
                 raise RuntimeError(f'Internal error: failed to find node for by scope={module_scope}')
 
-            op_arch = target_model.get_op_arch_by_graph_node(graph_node)
-            is_adjust_padding_applicable = op_arch in adjust_padding_operation_set
+            op_type = graph_node[NNCFGraph.OP_TYPE_NODE_ATTR]
```
vshampor, line 1143: Will separately refactor accesses such as this one to work with NNCFNode's instead of the underlying nx dict.

nncf/dynamic_graph/context.py
```diff
@@ -425,9 +429,14 @@ def in_operator(self, val):
...
     @property
-    def scope_modules(self):
+    def module_call_stack(self) -> List[torch.nn.Module]:
         self._init_thread_local()
-        return self._thread_local.scope_modules
+        return self._thread_local.module_call_stack
+
+    def get_current_module(self) -> Optional[torch.nn.Module]:
+        if self.module_call_stack:
+            return self.module_call_stack[-1]
```
ljaljushkin, line 438: any issues with multi-threading? I wonder how reliable this approach to get current module.

nncf/nncf_network.py
```diff
@@ -673,38 +675,8 @@ def get_input_shape_for_insertion_point(self, insertion_point: InsertionPoint) -
...
-        for node_key in ip_graph.nodes:
-            ip_graph_node = ip_graph.nodes[node_key]
-            ip_graph_node_type = ip_graph_node[InsertionPointGraph.NODE_TYPE_NODE_ATTR]
-            if ip_graph_node_type == InsertionPointGraphNodeType.OPERATOR:
-                nncf_graph_node_ref = ip_graph_node[InsertionPointGraph.REGULAR_NODE_REF_NODE_ATTR]
-                op_arch = self.get_op_arch_by_graph_node(nncf_graph_node_ref)
-                ip_graph_node[InsertionPointGraph.OPERATOR_METATYPE_NODE_ATTR] = op_arch
         return ip_graph
 
-    def get_op_arch_by_graph_node(self, nncf_graph_node_ref: dict) -> 'OperatorMetatype':
```
ljaljushkin, line 696: 👍

nncf/dynamic_graph/context.py
vshampor: The access is being done to a thread-local variable, as is evident from the code somewhat below these lines. Also notice that the existing code uses scopes in order to launch operator hooks, and it gets said scopes directly from the module object class names, so basically if it works for module class names, it also works for the module objects themselves.

ljaljushkin (approved):

vshampor:
Jenkins please retry a build

nncf/dynamic_graph/wrappers.py
```diff
@@ -46,13 +53,37 @@ def wrapped(*args, **kwargs):
...
+                if subtype is not None:
+                    op_type = subtype
+
+            module_attrs = None
+            # Collect module attributes, if required
+            if op_name in OP_NAMES_REQUIRING_MODULE_ATTRS:
+                if curr_module is None:
+                    raise RuntimeError("Operation {} requires module attributes, "
+                                       "but it was executed outside any module".format(op_name))
+                module_attrs = _get_module_attributes(curr_module, op_name)
```
vshampor, line 74: @gadylshintr, also FYI

nncf/dynamic_graph/graph.py
vshampor: After discussions we decided to keep `OperatorMetatype` local to quantization algo for the time being. The marking of nodes with `OperatorMetatype` values will be done at the quantization algo level, not at the wrap_operator stage.

vshampor:
Jenkins please retry a build

alexsu52 (approved):

PR #554, opened 2021-03-05
Remove version agnostic operator name mapping

vshampor:
The PyTorch public API is quite stable now, so there is no need to do this bookkeeping. This will, however, impact the majority of the existing .pth checkpoints as they would have to be converted so that the activation quantizer identifiers have the version agnostic `RELU` replaced with either `relu` or `relu_`.


vshampor:
Pytorch eval build 381 OK with existing checkpoints

nncf/torch/checkpoint_loading.py
```diff
@@ -268,8 +268,26 @@ def run(self) -> Dict[str, torch.Tensor]:
...
+        if has_version_agnostic_names:
+            warnings.warn('Legacy NNCF-enabled .pth checkpoint has been loaded! '
+                          'The version-agnostic `RELU` operator name entries in the state dict have been deprecated. '
+                          'The loader will try to match these entries to the correspoindig `relu` and `relu_` op '
+                          'names. The newly exported checkpoints will be adjusted to the new format.',
+                          category=DeprecationWarning)
+
         for normalized_key_to_load in normalized_keys_to_load:
             key_to_load = normalized_keys_to_load.get_orig_key(normalized_key_to_load)
+            normalized_key_to_load = cross_match_key_map.get(normalized_key_to_load, normalized_key_to_load)
```
ljaljushkin, line 289: This function call looks strange. should arguments really be duplicated ?

nncf/torch/checkpoint_loading.py
```diff
@@ -268,8 +268,26 @@ def run(self) -> Dict[str, torch.Tensor]:
         normalized_keys_to_load = NormalizedKeys(list(self.state_dict_to_load.keys()))
 
         has_legacy_storage_keys = False
+
+        has_version_agnostic_names = False
```
ljaljushkin, line 272: please move this functionality to `NormalizedKeys` class as it was done in my PR for legacy_storage_keys: 
https://github.com/openvinotoolkit/nncf/pull/682/files#diff-40c3a1a297ae11ac63166f5c9a2117bb25bd2e0c548e6d36e07edf64bd9848dfR241

ljaljushkin (changes requested):

nncf/torch/checkpoint_loading.py
vshampor: I tried to do the same using an `in` check, and Pylint triggered on it, advising to use `get`. `dict.get` returns the second argument value if the first argument value was not found among the keys of the dict.

nncf/torch/checkpoint_loading.py
vshampor: The matching done in this PR cannot be done in `NormalizedKeys` because it requires both the model and the checkpoint key sets - we don't know whether `RELU` in a given model corresponds to `relu` or `relu_`, so we have to iterate over alternatives.

ljaljushkin (approved):

vshampor:
Jenkins please retry a build

PR #557, opened 2021-03-07
[TF] Custom activation fix

negvet:
The proper way to add a custom activation is to add a custom activation function, rather than a custom activation class.


daniil-lyakhov:
Validated successfully

PR #559, opened 2021-03-07
Add __init__ scripts to scheduler folders

vshampor:


vshampor:
@andrey-churkin FYI since this causes package installation failure.

PR #560, opened 2021-03-07
Polynomial scheduler should not update sparsity in epoch_step() method when update_per_optimizer_step is True

andrey-churkin:


nncf/common/sparsity/schedulers.py
```diff
@@ -92,7 +92,8 @@ def epoch_step(self, next_epoch=None):
 
         self._steps_in_current_epoch = 0
         super().epoch_step(next_epoch)
-        self._set_sparsity_level()
+        if not self._update_per_optimizer_step:
+            self._set_sparsity_level()
```
vshampor, line 96: The motivation is understandable, but couldn't it be solved in another way, so that an `epoch_step` call would just set the internal state of the scheduler to step 0 of the next epoch instead of doing nothing? Should, however, think about whether this behaviour is general enough (i.e. what to do with adaptive schedulers or other schedulers whose schedule is determined in runtime).

vshampor (approved):

PR #562, opened 2021-03-09
Compression loss device bug

SKholkin:
ticket: 49963

### DIFF
--- a/tests/test_algo_common.py
+++ b/tests/test_algo_common.py
@@ -18,8 +18,10 @@ import onnx
 import os
 import pytest
 from functools import partial
+from functools import reduce
 from torch import nn
 from torch.nn import DataParallel
+from torch import cuda
 
 from nncf import NNCFConfig
 from nncf.checkpoint_loading import load_state
@@ -381,3 +383,58 @@ def test_can_export_compressed_model_with_specified_domain_for_custom_ops(tmp_pa
             count_custom_ops += 1
 
     assert count_custom_ops == 4
+
+
+def charnge_compression_algorithms_order(config):
+    # changes order of compression algorithms in config
+    def shift_list(list_for_shift):
+        shifted_list = [list_for_shift.pop()] + list_for_shift
+        return shifted_list
+
+    config_compression = list(config.get('compression', {}))
+    shifted_config_compression = shift_list(config_compression)
+    config.update({'compression': shifted_config_compression})
+    return config
+
+
+def get_basic_rb_sparsity_int8_config():
+    config = get_basic_sparsity_config()
+    config.update({
+        "compression": [
+            {
+                "algorithm": "rb_sparsity",
+                "sparsity_init": 0.02,
+                "params":
+                    {
+                        "schedule": "polynomial",
+                        "sparsity_target": 0.5,
+                        "sparsity_target_epoch": 2,
+                        "sparsity_freeze_epoch": 3
+                    },
+            },
+            {
+                "algorithm": "quantization"
+            }
+        ]
+    }
+    )
+    return config
+
+
+comp_loss_configs = [
+    get_basic_rb_sparsity_int8_config(),
+    charnge_compression_algorithms_order(get_basic_rb_sparsity_int8_config())
+]
+
+
+@pytest.mark.parametrize("device_type", ['cpu', 'cuda'], ids=['CPU', 'CUDA'])
+@pytest.mark.parametrize("config", comp_loss_configs,
+                         ids=[reduce(lambda x, y: x + "_" + y.get("algorithm", ""), config.get('compression', []),
+                                     'compression')
+                              for config in comp_loss_configs])
+def test_compression_loss_device_compatibility(config, device_type):
+    model = BasicConvTestModel()
+    if device_type == 'cuda':
+        model.to(cuda.current_device())
+    _, compression_ctrl = create_compressed_model_and_algo_for_test(model, config)
+    compression_ctrl.loss()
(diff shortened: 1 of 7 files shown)


nncf/compression_method_api.py
```diff
@@ -51,14 +51,20 @@ class PTCompressionLoss(nn.Module, CompressionLoss):
     sparsity algorithm calculates the number of non-zero weights in convolutional
     and fully-connected layers to construct the loss function.
     """
+    def __init__(self, default_device=None):
+        """
+        :param default_device: torch device being used for computing internal tensors
```
ljaljushkin, line 56: it's not actually for computing, it's just for creating a default value of loss on the target device

tests/test_algo_common.py
```diff
@@ -381,3 +382,60 @@ def test_can_export_compressed_model_with_specified_domain_for_custom_ops(tmp_pa
...
+                        "sparsity_target_epoch": 2,
+                        "sparsity_freeze_epoch": 3
+                    },
+            }
+        ]
+    })
+                      ]
+
+
+@pytest.mark.parametrize("is_cpu", [True, False], ids=['is_cpu_' + 'True', 'is_cpu_' + 'False'])
```
ljaljushkin, line 431: cpu and cuda is more meaningful

tests/test_algo_common.py
```diff
@@ -381,3 +382,60 @@ def test_can_export_compressed_model_with_specified_domain_for_custom_ops(tmp_pa
...
+
+def get_updated_config(config, update):
+    config.update(update)
+    return config
+
+
+comp_loss_configs = [get_updated_config(get_basic_sparsity_config(), {
+    "compression": [
+        {
+            "algorithm": "rb_sparsity",
```
ljaljushkin, line 395: please remove duplication of code by defining common parts

examples/classification/configs/quantization/mobilenet_v2_imagenet_int8.json
```diff
@@ -15,12 +15,13 @@
...
+    "compression": [
+        {
         "algorithm": "quantization",
         "initializer": {
             "range": {
                 "num_init_samples": 2560
             }
         }
-    }
+        }]
```
ljaljushkin, line 26: no need to change it, please revert

ljaljushkin (changes requested):

nncf/compression_method_api.py
```diff
@@ -51,14 +51,20 @@ class PTCompressionLoss(nn.Module, CompressionLoss):
     sparsity algorithm calculates the number of non-zero weights in convolutional
     and fully-connected layers to construct the loss function.
     """
+    def __init__(self, default_device=None):
+        """
+        :param default_device: torch device being used for creating default value loss tensor
```
ljaljushkin, line 56: for creating a default value of loss on the target device

tests/quantization/test_scheduler.py
```diff
@@ -33,7 +34,9 @@ def create_staged_scheduler(ctrl_spy, w_start=2, a_start=1):
 class QuantizationCtrlBaseSpy:
     #pylint:disable=no-member
     def __init__(self, mocker):
-        self._mocked_ctrl = QuantizationControllerBase(mocker.stub)
+        mock_model = mocker.MagicMock(spec=NNCFNetwork)
```
ljaljushkin, line 37: 👍

ljaljushkin (changes requested):

tests/test_algo_common.py
```diff
@@ -381,3 +385,58 @@ def test_can_export_compressed_model_with_specified_domain_for_custom_ops(tmp_pa
             count_custom_ops += 1
 
     assert count_custom_ops == 4
+
+
+def charnge_compression_algorithms_order(config):
```
ljaljushkin, line 390: please be more attentive! 
charnge -> change

nncf/compression_method_api.py
```diff
@@ -51,14 +51,20 @@ class PTCompressionLoss(nn.Module, CompressionLoss):
     sparsity algorithm calculates the number of non-zero weights in convolutional
     and fully-connected layers to construct the loss function.
     """
+    def __init__(self, default_device=None):
+        """
+        :param default_device: torch device being used for creating a default value of loss on the target device
```
vshampor, line 56: I am fairly sure that everywhere we have to fall back on the default value of the loss we also know the device of the model to be compressed. I suggest that instead of changing the API for this you keep the API as-is and instead define a new `ZeroCompressionLoss(PTCompressionLoss)` with an init function taking a device. Instantiate it explicitly in all algos that do not define their specific compression loss.

vshampor (changes requested):

nncf/compression_method_api.py
ljaljushkin: With current design I see the problem with implementation of PTCompressionAlgorithmController:
```
def __init__(self, target_model: ModelType):
     self._loss = PTCompressionLoss()
```
Algorithms are not required to define a new ZeroCompressionLoss. That still would lead to errors with device.

vshampor: Even the `PTCompressionAlgorithmController` should be an interface, not a real base class. The real controller class specifying no compression is `NoCompressionAlgorithmController`.

ljaljushkin: it means that we need to change current design first (probably in a separate PR), because this class has a lot of implemented methods: `export_model`, `run_batchnorm_adaptation`, `statistics`

tests/test_algo_common.py
vshampor: I can still see `charnge` here.

vshampor (changes requested):

vshampor:
Jenkins please retry a build

nncf/compression_method_api.py
vshampor: Will be done in a separate PR.

vshampor (approved):

alexsu52 (approved):

PR #574, opened 2021-03-15
Propagate masks before apply masks

AlexanderDokuchaev:
REF: 50604

In some case (like DenseNet) when batch_norm layer is after not convolution layer we should propagate masks to apply actual mask for this batch_norm layer

```
conv1 conv2 
  \    / 
  concat
    | 
batch_norm 
```

### DIFF
--- a/docs/compression_algorithms/Pruning.md
+++ b/docs/compression_algorithms/Pruning.md
@@ -2,7 +2,7 @@
 
 #### Filter pruning    
 
- Filter pruning algorithm zeros output filters in Convolutional layers based on some filter importance criterion  (filters with smaller importance are pruned).     
+Filter pruning algorithm zeros output filters in Convolutional layers based on some filter importance criterion  (filters with smaller importance are pruned).     
 The framework contains three such criteria: `L1`, `L2` norm, and `Geometric Median`. Also, different schemes of pruning application are presented by different schedulers.    
 
 #### Filter importance criterions **L1, L2**
@@ -90,7 +90,7 @@ sparsity and filter pruning algorithms. It can be enabled by setting a non-zero 
         "prune_first_conv": false, // Whether to prune first Convolutional layers or not. First means that it is a convolutional layer such that there is a path from model input to this layer such that there are no other convolution operations on it. `False` by default. 
         "prune_last_conv": false, // Whether to prune last Convolutional layers or not.  Last means that it is a Convolutional layer such that there is a path from this layer to the model output such that there are no other convolution operations on it. `False` by default. 
         "prune_downsample_convs": false, // Whether to prune downsample Convolutional layers (with stride > 1) or not. `False` by default.
-        "prune_batch_norms": false, // Whether to nullifies parameters of Batch Norm layer corresponds to zeroed filters of convolution corresponding to this Batch Norm. `False` by default.
+        "prune_batch_norms": true, // Whether to nullifies parameters of Batch Norm layer corresponds to zeroed filters of convolution corresponding to this Batch Norm. `True` by default.
         "zero_grad": true // Whether to setting gradients corresponding to zeroed filters to zero during training, `True` by default.    
     },
 
@@ -123,4 +123,3 @@ To do this, the special `can_prune` attribute is propagated through the network.
 The result is a valid graph for the pruned and non-pruned parts: that is, there is no operation that does not accept a pruned input.
 
 3. On the final stage, there is a direct pruning of operations according to the obtained `can_prune` attribute values.
-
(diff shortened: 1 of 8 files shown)


nncf/pruning/export_helpers.py
```diff
@@ -460,7 +466,6 @@ def mask_propagation(self):
             node_type = self.graph.node_type_fn(node)
             cls = self.get_class_by_type_name(node_type)
             cls.mask_propagation(self.model, node, self.graph, self.nx_graph)
-        nncf_logger.info('Finished mask propagation in graph')
```
mkaglins, line 463: Maybe logging in `get_class_by_type_name` function should be also disabled?

nncf/pruning/filter_pruning/algo.py
```diff
@@ -449,17 +455,28 @@ def _apply_binary_mask_to_module_weight_and_bias(module, mask, module_scope):
...
+
+        # 1. Propagate masks for all modules
+        graph = self.model.get_original_graph()
+        # pylint: disable=protected-access
+        nx_graph = graph._nx_graph
+        model_pruner = ModelPruner(self.model, graph, nx_graph)
+        model_pruner.mask_propagation()
+
+        # 2. Apply masks
+        node_types = [v.op_func_name for v in NNCF_GENERAL_CONV_MODULES_DICT] + ['group_norm']
```
mkaglins, line 467: Please, rename to something more clear like `masked_node_types` or `types_to_apply_mask`.

nncf/pruning/filter_pruning/algo.py
```diff
@@ -449,17 +455,28 @@ def _apply_binary_mask_to_module_weight_and_bias(module, mask, module_scope):
...
+        if self.prune_batch_norms:
+            node_types.append('batch_norm')
+        nncf_nodes = [nx_graph.nodes[name] for name in nx_graph]
+        pruned_node_modules = list()
```
mkaglins, line 471: Why `pruned_node_modules` is never updated?

nncf/pruning/filter_pruning/algo.py
```diff
@@ -449,17 +455,28 @@ def _apply_binary_mask_to_module_weight_and_bias(module, mask, module_scope):
...
+        # pylint: disable=protected-access
+        nx_graph = graph._nx_graph
+        model_pruner = ModelPruner(self.model, graph, nx_graph)
+        model_pruner.mask_propagation()
+
+        # 2. Apply masks
+        node_types = [v.op_func_name for v in NNCF_GENERAL_CONV_MODULES_DICT] + ['group_norm']
+        if self.prune_batch_norms:
+            node_types.append('batch_norm')
+        nncf_nodes = [nx_graph.nodes[name] for name in nx_graph]
```
mkaglins, line 470: `nncf_nodes = graph.get_all_nodes()` will be better

docs/compression_algorithms/Pruning.md
```diff
@@ -90,7 +90,7 @@ sparsity and filter pruning algorithms. It can be enabled by setting a non-zero
         "prune_first_conv": false, // Whether to prune first Convolutional layers or not. First means that it is a convolutional layer such that there is a path from model input to this layer such that there are no other convolution operations on it. `False` by default. 
         "prune_last_conv": false, // Whether to prune last Convolutional layers or not.  Last means that it is a Convolutional layer such that there is a path from this layer to the model output such that there are no other convolution operations on it. `False` by default. 
         "prune_downsample_convs": false, // Whether to prune downsample Convolutional layers (with stride > 1) or not. `False` by default.
-        "prune_batch_norms": false, // Whether to nullifies parameters of Batch Norm layer corresponds to zeroed filters of convolution corresponding to this Batch Norm. `False` by default.
+        "prune_batch_norms": true, // Whether to nullifies parameters of Batch Norm layer corresponds to zeroed filters of convolution corresponding to this Batch Norm. `True` by default.
```
mkaglins, line 93: @AlexKoff88 Are we still need `prune_batch_norms `as a parameter? It looks like without the `prune_batch_norms=True` model will be pruned incorrectly after training (non-zero BatchNorm params will be deleted).

mkaglins (approved):

daniil-lyakhov:
Hi there! I'm expecting some problems with pruning algo.

There is place In resnetX_v2 topology:

conv1     conv2
   \               /
        add
          |
        batch norm

I have a hypothesis that in such case our current pruning algo prune batch norm twice (add two binary masks on one weight). I reproduced it in TF realization of filter pruning. @AlexanderDokuchaev, could you please test resnet50_v2 topology in PyTorch on twice batchnorm pruning? 

Minimal topology to reproduce such behavior:
image
Highlighted batchnorm pruned twice.

AlexanderDokuchaev:
@daniil-lyakhov https://github.com/openvinotoolkit/nncf/pull/609 fixed this problem.
It's was happened because one bathnorm layer was associated with two convolution as https://github.com/openvinotoolkit/nncf/blob/e05ffa17ee5ff3d24fdbc9b4600e2eafc5eee4b6/beta/nncf/tensorflow/pruning/base_algorithm.py#L142. 
Now this batchnnorm will be prune only ones after mask propagation.

PR #584, opened 2021-03-18
[TF] Use common pruning schedulers in NNCF TF

andrey-churkin:


alexsu52:
@gadylshintr Could you take a look?

beta/nncf/tensorflow/pruning/base_algorithm.py
```diff
@@ -240,6 +244,8 @@ def _check_pruning_rate(self, params):
         pruning_flops_target = params.get('pruning_flops_target', None)
         if pruning_target and pruning_flops_target:
             raise ValueError('Only one parameter from \'pruning_target\' and \'pruning_flops_target\' can be set.')
+        if pruning_flops_target:
+            self.prune_flops = True
```
gadylshintr, line 248: Since pruning by flops is not supported in TF version yet I suppose it would be better to throw an exception here and not to introduce `self.prune_flops` yet.

andrey-churkin: Ok, I will add an exception. But, the `prune_flops` attribute is required for schedulers.
https://github.com/openvinotoolkit/nncf/blob/aaaf45ef21967e26ee2edb8e256d04028b38c83d/nncf/common/pruning/schedulers.py#L38-L42

andrey-churkin: Done

andrey-churkin:
@evgeniya-egupova, FYI

gadylshintr (approved):

alexsu52 (approved):

PR #589, opened 2021-03-21
Fix exponential scheduler with bias

andrey-churkin:
Looks like that current implementation does not work properly.
Consider the following example:
```python
a, b, k = _init_exp(10, 0.1, 0.7)
x = a * np.exp(-k * 10) + b
print(x)
```
I expected that `x` is equal to `0.7` but I got `0.666666666666667`.

### DIFF
--- a/nncf/common/pruning/schedulers.py
+++ b/nncf/common/pruning/schedulers.py
@@ -161,16 +161,24 @@ class ExponentialPruningScheduler(PruningScheduler):
 @PRUNING_SCHEDULERS.register("exponential_with_bias")
 class ExponentialWithBiasPruningScheduler(PruningScheduler):
     """
-    Calculates pruning rate progressively according to the formula
-    P = a * exp(- k * epoch) + b
-    Where:
-    epoch - epoch number
-    P - pruning rate for current epoch
-    a, b, k - params
+    Pruning scheduler which calculates pruning rate for the current epoch
+    according to the formula:
+
+        current_level = a * exp(-k * epoch_idx) + b,
+
+    where a, b, k is a params.
     """
+
     def __init__(self, controller, params: dict):
+        """
+        Initializes a pruning scheduler with an exponential (with bias) decay schedule.
+
+        :param controller: Sparsity algorithm controller.
+        :param params: Parameters of the scheduler.
+        """
         super().__init__(controller, params)
-        self.a, self.b, self.k = self._init_exp(self.num_pruning_epochs, self.initial_level, self.target_level)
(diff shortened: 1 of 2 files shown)


andrey-churkin:
Jenkins please retry a build

nncf/common/pruning/schedulers.py
```diff
@@ -194,11 +202,12 @@ def get_b(a, k):
             return p_min - a
```
alexsu52, line 202: def get_b(a, k): -> def get_b(a):

alexsu52 (changes requested):

nncf/common/pruning/schedulers.py
andrey-churkin: Done.

alexsu52 (approved):

mkaglins (approved):

vanyalzr:
@mkaglins @AlexKoff88 I suppose we could re-run some filter pruning benchmarks to see diff in accuracy for this scheduler once this fix is introduced?

mkaglins:
> @mkaglins @AlexKoff88 I suppose we could re-run some filter pruning benchmarks to see diff in accuracy for this scheduler once this fix is introduced?

This `exponential_with_bias` scheduler is not used in any officially claimed results, but we can run a comparison of `exponential` (that often used in Imagenet experiments) with this fixed version.

nncf/common/pruning/schedulers.py
```diff
@@ -190,17 +198,18 @@ def _init_exp(epoch_idx, p_min, p_max, factor=0.125):
...
         def get_a(k):
-            return (3 / 4 * p_max - p_min) / (np.exp(-k * factor * epoch_idx) - 1)
+            return (p_max - p_min) / (np.exp(-k * epoch_idx) - 1)
 
         def f_to_solve(x):
-            y = np.exp(factor * x * epoch_idx)
-            return 1 / 3 * y + 1 / (y ** 7) - 4 / 3
+            c = (0.75 * p_max - p_min) / (p_max - p_min)
+            y = np.exp(-x * epoch_idx)
+            return y ** factor - c * y + c - 1
```
vanyalzr, line 210: Should we make the `factor` parameter adjustable via config? I suppose the 1/8 value is based on what is set in the paper, not on our empirical results.

mkaglins: Yes, it is a good idea to set better default value based on empirical results on our models (firstly get these results). But I am not sure about the importance of this param in the config for users.

PR #591, opened 2021-03-22
Set abs criterion for magnitude pruning by default

vanyalzr:


alexsu52:
@vanyalzr Will users achieve the claimed results using provided config files after this change?

vanyalzr:
> @vanyalzr Will users achieve the claimed results using provided config files after this change?

Abs importance criterion gives better performing models in most cases in my experiments than normed abs, will update results accordingly.

vshampor:
@vanyalzr is this still relevant?

vanyalzr:
closing for now

PR #592, opened 2021-03-22
Changes to fix batchnorm adaptation algorithm in DDP mode

aleksei-kashapov:
I think that BatchnormAdaptatiomn should run in all processes during distributed regime. Currently without that it freezes a distributed  training in mmdetection for complex 2-stage models


vshampor:
It is strange that this has not been discovered by our regular testing yet, especially seeing that the batch norm adaptation is enabled by default. @aleksei-kashapov could you please
1) find out why there are no hangs in the `test_sanity_sample` test suite, despite the fact that it uses BN adaptation *and* runs in DDP mode
2) if the sanity tests are insufficient to catch this problem, devise a better, additional test that would check for hangs during run_batchnorm_adaptation with DDP mode specifically? The expectation here is that this test will fail without your fix and pass with it.

aleksei-kashapov:
@vshampor I presented a new approach to the problem of synchronization of the processes in multiprocessing mode.
1) All processes synchronize before quitting `create_compressed_model` which resolves a problem in integration with mmdetection.
2) Statistics collection for quantization performs on a single process too. It reduces the amount of computations and also makes logs more user friendly (before every process printed statistics)

aleksei-kashapov:
This PR fixes a problem with the BatchNorm adaptation algorithm. Before the algorithm, incorrectly changed the training states of the model.

Now:
1) Only BatchNorm layers are changed training state during adaptation
2) Fix incorrect behavior of `training_mode_switcher `which didn't return an initial state to the model
3) Add `barrier()` before returning compressed model in `create_compressed_model `to guarantee synchronization of the processes in DDP mode.
4) Make collecting tensor statistics only on the main process in DDP mode

This fully fixes the problem of freezing during DDP launches in integration with mmdetection

nncf/initialization.py
```diff
@@ -164,39 +164,66 @@ def __init__(self, model, init_device: str, num_bn_forget_steps):
...
+    def _bn_training_state_switcher(self):
+        def save_original_bn_training_state(module):
+            self.original_training_state[module] = module.training
+
+        def set_bn_training_state(module, state):
+            module.training = state
+
+        def restore_original_bn_training_state(module):
+            module.training = self.original_training_state[module]
+
```
vshampor, line 189: Why isn't using the modified `training_mode_switcher` enough?

nncf/model_creation.py
```diff
@@ -141,4 +143,17 @@ def create_compressed_model(model: Module, config: NNCFConfig,
             graph = compressed_graph_builder.build_graph(compressed_model, compressed_model.get_tracing_context())
             graph.visualize_graph(osp.join(config.get("log_dir", "."), "compressed_graph.dot"))
 
+    # Synchronize all processes if run in distributed mode
+    if is_dist_avail_and_initialized():
+        try:
+            barrier()
+        except RuntimeError as err:
```
vshampor, line 150: In which cases may a `RuntimeError` occur during a barrier call? Can NNCF really continue working in this scenario?

vshampor (changes requested):

nncf/initialization.py
aleksei-kashapov: The approach using modified `training_mode_switcher` is not fully correct. In BatchNorm adaptation algorithm we only want to update BatchNorm statistics, while switching the whole model in training mode can also affect on other model's parts.

nncf/model_creation.py
aleksei-kashapov: According to a source code- https://github.com/pytorch/pytorch/blob/master/torch/distributed/distributed_c10d.py#L2483
`barrier()` handles `RuntimeError` if a DDP backend doesn't support this function. For example, gloo backend doens't provide a suppor of  `barrier()` for GPU.
If the `RuntimeError` occurs NNCF can not guarantee that the processes are synchronized after `create_compressed_model()`, but in many cases, it should work fine (where working pipeline with a model doesn't demand process synchronization, e.g. examples in NNCF).

nncf/initialization.py
```diff
@@ -164,39 +164,66 @@ def __init__(self, model, init_device: str, num_bn_forget_steps):
...
+    def _bn_momentum_switcher(self):
         def set_bn_momentum(module, momentum_value):
             module.momentum = momentum_value
 
-        def save_original_bn_momenta(module):
+        def save_original_bn_momentum(module):
             self.original_momenta_values[module] = module.momentum
 
-        def restore_original_bn_momenta(module):
+        def restore_original_bn_momentum(module):
```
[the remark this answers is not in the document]
aleksei-kashapov: Unfortunately, I didn't have a chance to learn Latin, so thank you for the enlightenment. 
I was following the original parameter name in PyTorch docs of BatchNorm layer.

nncf/initialization.py
vanyalzr: I think `training_mode_switcher` could be sufficient in practice, but switching BN `training` flags only is a safer approach, since we potentially preserve inference-time behavior of other layers such as DropOut while refreshing BN statistics.

nncf/initialization.py
```diff
@@ -164,39 +164,66 @@ def __init__(self, model, init_device: str, num_bn_forget_steps):
...
+                                   torch.nn.modules.batchnorm.BatchNorm3d)):
                 func(module)
 
         return func_apply_to_bns
 
-    def _run_model_inference(self, data_loader, num_init_steps, device):
-        num_bn_forget_steps = self.num_bn_forget_steps
+    @contextmanager
+    def _bn_training_state_switcher(self):
+        def save_original_bn_training_state(module):
```
vshampor, line 181: Type hints

nncf/initialization.py
```diff
@@ -164,39 +164,66 @@ def __init__(self, model, init_device: str, num_bn_forget_steps):
...
+        self.model.apply(self._apply_to_batchnorms(save_original_bn_training_state))
+        self.model.apply(self._apply_to_batchnorms(partial(set_bn_training_state, state=True)))
+        try:
+            yield
+        finally:
+            self.model.apply(self._apply_to_batchnorms(restore_original_bn_training_state))
 
+    @contextmanager
+    def _bn_momentum_switcher(self):
         def set_bn_momentum(module, momentum_value):
```
vshampor, line 199: Type hints

nncf/utils.py
```diff
@@ -362,12 +362,25 @@ def should_consider_scope(scope_str: str, target_scopes: List[str], ignored_scop
...
 def training_mode_switcher(model: torch.nn.Module, is_training: bool = True):
-    is_original_mode_training = model.training
+    saved_state = {}
+
+    def helper_save_state(module):
+        for ch in module.children():
+            saved_state[ch] = ch.training
+            helper_save_state(ch)
+
+    def helper_load_state(module, state):
```
vshampor, line 372: Type hints

nncf/utils.py
```diff
@@ -362,12 +362,25 @@ def should_consider_scope(scope_str: str, target_scopes: List[str], ignored_scop
...
+            helper_load_state(ch, state)
+
+    helper_save_state(model)
     model.train(is_training)
     try:
         yield
     finally:
-        model.train(is_original_mode_training)
+        helper_load_state(model, saved_state)
+        del saved_state
```
vshampor, line 383: Remove this `del` call

vshampor (changes requested):

nncf/utils.py
aleksei-kashapov: Done

nncf/utils.py
aleksei-kashapov: Done

nncf/initialization.py
aleksei-kashapov: Done

nncf/initialization.py
aleksei-kashapov: Done

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        for ch in module.children():
+            saved_model_state[ch] = ch.training
+            save_model_training_state(ch)
+
+    def compare_saved_model_state_and_current_model_state(module):
+        for ch in module.children():
+            assert saved_model_state[ch] == ch.training
+            compare_saved_model_state_and_current_model_state(ch)
+
+    def random_change_model_training_state(module):
```
vanyalzr, line 24: `randomly_change_model_training_state`

nncf/nncf_network.py
```diff
@@ -504,10 +504,14 @@ def _set_nncf_wrapped_model(self, value):
...
-                           self._user_dummy_forward_fn, self._wrap_inputs_fn,
-                           self.scopes_without_shape_matching, self.ignored_scopes, self.target_scopes,
-                           reset=True)
+        from nncf.utils import save_module_training_state, load_module_training_state
+        saved_state = save_module_training_state(self)
+        model_copy = NNCFNetwork(self.get_nncf_wrapped_model(), self.input_infos,
+                    self._user_dummy_forward_fn, self._wrap_inputs_fn,
+                    self.scopes_without_shape_matching, self.ignored_scopes, self.target_scopes,
+                    reset=True)
+        load_module_training_state(model_copy, saved_state)
```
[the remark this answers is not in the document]
aleksei-kashapov: In our examples we have Inception model, which has different forward method depending on training state

tests/test_utils.py
aleksei-kashapov: Done

nncf/nncf_network.py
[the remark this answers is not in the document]
aleksei-kashapov: I am sorry, I was mistaken. In out example ICNET is a model that has different forward pass in training mode.

https://github.com/openvinotoolkit/nncf/blob/3722466085f7202da34fe6539b20f0aef25bc6da/examples/common/models/segmentation/icnet.py#L211-L213

https://github.com/openvinotoolkit/nncf/blob/3722466085f7202da34fe6539b20f0aef25bc6da/examples/common/models/segmentation/icnet.py#L312-L321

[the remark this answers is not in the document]
aleksei-kashapov: > Also, why not make it a context manager as well?
@vanyalzr using context manager is not convenient here. Because we create a copy of a model while load and save an original model state. Therefore, we have two instances that need to give as arguments to this context manager, this doesn't make this context manager usable in any other part of the code.

nncf/model_creation.py
vshampor: Please add a short comment in code explaining this situation.

nncf/initialization.py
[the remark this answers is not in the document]
vshampor: Ok, but the function really does not only save and restore a single *momentum*, it actually does so for all *momenta*.

vshampor (approved):

PR #598, opened 2021-03-24
Cuda extension fix

aleksei-kashapov:
This will fix skipping this test

### DIFF
--- a/tests/test_extensions_build.py
+++ b/tests/test_extensions_build.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 import pytest
 import pathlib
 import shutil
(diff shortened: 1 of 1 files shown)


vshampor (changes requested):
For starters, the following call seems to not work at all:
```
 pytest tests -k test_force_cuda_build
```

```
/bin/sh: 1: .: Can't open /tmp/pytest-of-vshampor/pytest-4/test_force_cuda_build_venv_dev0/venv/bin/activate
```

Second call with the same command line gives:
```
 RuntimeError: Ninja is required to load C++ extensions
```
and inspecting my virtual environment's bin/ninja file with a text editor shows that the interpreter to the ninja script is being somehow set to the tmp_venv's python, which means that there is an error in your test virtual environment setup code (test runs should not impact the user's main virtual env, that's what we make tmp_venv's for):
```
#!/tmp/pytest-of-vshampor/pytest-3/test_force_cuda_build_venv_dev0/venv/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'ninja','console_scripts','ninja'
__requires__ = 'ninja'
import re
...
```
And third, finally, is that your new code in this PR still does not cover the general situation of a user with a GPU-powered machine and installed pytorch. CUDA, when installed, may not expose `nvcc` via PATH, but torch itself is still able to find it. My machine shows nothing when `which nvcc` is invoked, yet it is able to compile the extensions and run GPU NNCF.

The first two items do not seem to be regressions, but rather the problems in the test itself. I'd like you to fix them in this same PR.

vshampor:
It seems that ninja is being broken because setuptools, even when run inside a temporary venv, will install entrypoints to the packages that use them (such as ninja) into the "base" folder with the python interpreter and site-packages that was used to create the temporary environment. This behaviour is quite stupid and un-fixable. We will have to live with the fact that the locally launched tests that do NNCF installation into a `python -m venv` virtual environment will break the main local development virtual environment (because after each such test the ninja entrypoint will attempt to be run with the temporary interpreter from a tmp venv, as seen above). An alternative would be to use proper docker images for each test_install case, but this seems to be impossible without moving the parameterization of `test_install` outside the pytest framework and into the CI side.

aleksei-kashapov:
@vshampor please take a look at new version. Locally it works on my machine in CPU mode

PR #599, opened 2021-03-25
Add pip support

a-ignatyev:


setup.py
```diff
@@ -57,20 +57,13 @@ def find_version(*file_paths):
...
 
-DEPENDENCY_LINKS = []
-
 python_version = sys.version_info[:2]
-if python_version[0] < 3:
-    print("Only Python > 3.5 is supported")
-    sys.exit(0)
-elif python_version[1] < 6:
+if python_version[0] < 3 or python_version[1] < 6:
     print("Only Python > 3.6 is supported")
```
daniil-lyakhov, line 62: You mean >= 3.6 ?

a-ignatyev: Yes. Fixed.

setup.py
```diff
@@ -57,20 +57,13 @@ def find_version(*file_paths):
...
-DEPENDENCY_LINKS = []
-
 python_version = sys.version_info[:2]
-if python_version[0] < 3:
-    print("Only Python > 3.5 is supported")
-    sys.exit(0)
-elif python_version[1] < 6:
-    print("Only Python > 3.6 is supported")
+if python_version[0] < 3 or python_version[1] < 6:
+    print("Only Python >= 3.6 is supported")
```
vshampor, line 62: Our current README states support for >=3.7 only, should probably update this condition as well

vshampor (changes requested):

setup.py
alexsu52: What reason to bump up the python version? Tensorflow 2.4 requires python>=3.6 and PyTorch 1.7 too.

vshampor: We actually validate on 3.7 only.

vshampor: Ok, since TF requires 3.6, will postpone updating setup.py.

vshampor:
Install build 253 passed.

vshampor (approved):

PR #601, opened 2021-03-25
Add per-tensor configurations for CPU weights

vshampor:


nncf/hw_configs/cpu.json
```diff
@@ -26,12 +26,12 @@
                 "mode": "symmetric",
                 "level_low": -128,
                 "level_high": 127,
-                "granularity": "perchannel"
+                "granularity": ["perchannel", "pertensor"]
```
AlexKoff88, line 29: How about other configs? Since it is obvious that pertensor is automatically supported if perchannel is in the support and we would like just formally identify this in the confugs, it is worth changing all the impacted configs.

vshampor: Done

nncf/hw_configs/vpu.json
```diff
@@ -40,28 +47,28 @@
             },
             "quantization": {
                 "activations": ["q8_tn", "q4_tn"],
-                "weights": ["q8_ch","q4_ch", "q2_ch"]
+                "weights": ["q8_w", "q4_w", "q2_w"]
```
AlexKoff88, line 50: What are `q2_w and q4_w`? they are not defined.

vshampor: Thanks for spotting this, fixed.

AlexKoff88 (approved):

PR #602, opened 2021-03-26
Make training settings sensible for CIFAR10/100 in classification sample

vanyalzr:
@mkaglins


vshampor (approved):

vanyalzr:
Finetuning ResNet50 w/ 40% filters pruned on CIFAR100: orange - before, blue - after this diff (both evaluated w/o resize & center-crop)

PR #542, opened 2021-03-02
[TF] RB sparsity algorithm implementation

daniil-lyakhov:

PR #604, opened 2021-03-26
[TF] Fixed training flag propagation

alexsu52:


daniil-lyakhov:
I find out that `fit` method doesn't set learning phase to 1 except manual setup. It seems like learning_phase as global value is depricated in tf2 (I confirm I got warnings on `set_learning_phase` call). They suggest using only `training` argument, but in case of `NNCFWrapper` it always None( Thus I've add this workaround to train rb_sparse models https://github.com/openvinotoolkit/nncf/pull/542/commits/f99c8ee9e3eb867ddbdd7295d6963a3e31dcc68a).

Please check https://github.com/tensorflow/tensorflow/issues/25754
https://colab.research.google.com/drive/1fCfDu3jHjCrN0CBPz1xJHotzl_2899WW?usp=sharing

daniil-lyakhov (approved):

alexsu52:
> 
> 
> I find out that `fit` method doesn't set learning phase to 1 except manual setup. It seems like learning_phase as global value is depricated in tf2 (I confirm I got warnings on `set_learning_phase` call). They suggest using only `training` argument, but in case of `NNCFWrapper` it always None( Thus I've add this workaround to train rb_sparse models [f99c8ee](https://github.com/openvinotoolkit/nncf/commit/f99c8ee9e3eb867ddbdd7295d6963a3e31dcc68a)).
> 
> Please check [tensorflow/tensorflow#25754](https://github.com/tensorflow/tensorflow/issues/25754)
> https://colab.research.google.com/drive/1fCfDu3jHjCrN0CBPz1xJHotzl_2899WW?usp=sharing

The PR solves the issue that the `training` argument is always None for `NNCFWrapper` call. I know that the learning phase from the keras.backend module is the deprecated method but all Keras layers are using it at least in TF 2.4.0.

PR #607, opened 2021-03-27
Do not quantize embeddings twice; unify embedding weight scales if ne…

vshampor:
…cessary


vshampor:
Jenkins please retry a build

nncf/hw_configs/vpu.json
```diff
@@ -310,6 +310,23 @@
         {"type": "Crop"},
         {"type": "Transpose"},
         {"type": "Tile"},
-        {"type": "StridedSlice"}
+        {"type": "StridedSlice"},
+        // NNCF-specific extensions are below:
+        {
+            "type": "Embedding",
+            "quantization": {
```
ljaljushkin, line 317: Shouldn't we explicitly specify here?: 
```
"scales": "unified"
```

vshampor: Well, that depends on our convention, since we basically own the format of our own HW configs now. Everywhere else "scales": "unified" is applied to multi-input ops and refers to the unification of weights between the inputs specifically. The embeddings, instead, have their weights unified, and only in specific conditions when they get "linked" by propagating activation quantizers of downstream real unified scales operations. Anyway I think that specifying this as an option would be probably useful

nncf/quantization/algo.py
```diff
@@ -929,91 +886,164 @@ def _build_commands_for_single_unified_scale_group(self,
...
+        :param: target_model - the model to be quantized.
+        :param: insertion_points - a list of target points for quantization using one
+        quantizer module
+        :param: qconfig - the QuantizerConfig for the resulting quantizer module
+        :param: range_init_minmax_values - a pair of minimum and maximum values of input statistics
+        for initializing the quantizer's trainable parameters
+        :return: A tuple with the identifier of the new quantizer module and a list of
+        insertion commands registering this module for quantization at spots described by
+        insertion_points."""
+        #pylint:disable=too-many-branches
```
ljaljushkin, line 920: it seems this function can be splitted into two to process activation and weight separately without significant code duplication

nncf/nncf_network.py
```diff
@@ -638,7 +638,7 @@ def get_compression_modules_by_type(self, compression_module_type: ExtraCompress
     def _compression_module_type_to_attr_name(compression_module_type: ExtraCompressionModuleType):
         """Required for backward compatibility with checkpoints that store function and activation
         quantizers directly under corresponding attributes of NNCFNetwork."""
-        if compression_module_type == ExtraCompressionModuleType.ACTIVATION_QUANTIZER:
+        if compression_module_type == ExtraCompressionModuleType.EXTERNAL_QUANTIZER:
             return "activation_quantizers"
```
ljaljushkin, line 642: There's still `activation` and `external` at the same time. 
I'd rename everywhere at once. BTW, why previous name is not suitable now?

ljaljushkin (changes requested):

nncf/quantization/structs.py
```diff
@@ -16,6 +16,8 @@ def __init__(self, quantizer_module_ref: BaseQuantizer,
 class WeightQuantizerInfo:
     def __init__(self,
                  quantizer_module_ref: BaseQuantizer,
-                 quantized_module: torch.nn.Module):
+                 quantized_module: torch.nn.Module,
+                 affected_insertions: List[PTTargetPoint]):
         self.quantizer_module_ref = quantizer_module_ref
         self.quantized_module = quantized_module
+        self.affected_insertions = affected_insertions
```
ljaljushkin, line 23: please remove, if it's not used anywhere

ljaljushkin (changes requested):

nncf/quantization/algo.py
vshampor: If it *could* be split, I would have left the two functions. As it is, it cannot be split because it is now required to share a single quantizer module between both activation and weight quantization points in general. If you have a solution that I am not seeing, please let me know.

nncf/nncf_network.py
vshampor: Cannot rename the string representation that easily since it goes into the state dict and, by corollary into the checkpoints. Renamed the entity name in code because the meaning of the underlying storage has changed - it stores quantizer modules that are potentially used both for activations and weights (so the designation "activation quantizers" is not longer applicable), and it does so externally to the underlying model.

nncf/quantization/structs.py
vshampor: It is necessary to know which points in the graph a given quantizer module quantizes. A single quantizer module may apply its quantization to different points in the model's graph. This is the mechanism behind unified scales.

nncf/quantization/algo.py
ljaljushkin: why not to process weights first, return quantized module and pass it to the next function for processing activations?

nncf/quantization/structs.py
ljaljushkin: in that case it makes sense to define interface or base class (e.g. 'QuantizerInfo') to access common fields

vshampor: Done

nncf/nncf_network.py
vshampor: Adjusted the code to use the new `external_quantizer` naming going forward, and still be able to load the old-style checkpoints (displaying a deprecation warning in this case).

nncf/quantization/algo.py
vshampor: The function looks terrifying, I agree, but I do not currently see a solution that would make it more beautiful code-wise without overdesigning. Perhaps in the future, if a new use case appears, it will be more clear as to how to decompose the function's body into potentially reusable blocks.

ljaljushkin (approved):

vshampor:
Jenkins please retry a build

PR #609, opened 2021-03-28
Add common mask propagation

AlexanderDokuchaev:


nncf/common/pruning/mask_propagator.py
```diff
@@ -0,0 +1,59 @@
...
+ limitations under the License.
+"""
+
+from nncf.common.graph.graph import NNCFGraph
+from nncf.common.graph.graph import NNCFNode
+from nncf.common.pruning.export_helpers import DefaultMetaOp
+from nncf.common.pruning.utils import PruningOperationsMetatypeRegistry
+
+
+class MaskPropagator:
```
alexsu52, line 20: I would like to suggest for discussion the following name of the pruning mask propagation algorithm `MaskPropagationAlgorithm` instead of `MaskPropagator`. 
@AlexanderDokuchaev  @evgeniya-egupova @mkaglins @ljaljushkin @vshampor what do you think?

alexsu52, line 20: Please add a short description of the class and methods. Please, guided by [style guide](https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md) for writing docstrings here and anywhere in this PR.

alexsu52 (changes requested):
I guess, the PR will not be merge without support for the pruning mask propagation algorithm in the TensorFlow backend.

nncf/common/pruning/mask_propagator.py
AlexanderDokuchaev: Yes, MaskPropagationAlgorithm is better 👍

beta/nncf/tensorflow/graph/converter.py
```diff
@@ -234,7 +238,7 @@ def _update_graph_with_raw_nodes(graph: Union[nx.DiGraph, NNCFGraph],
     for original_name, instances in raw_nodes.items():
         for i, attributes in instances.items():
             node_name = get_expanded_node_name(original_name, i, attributes['is_shared'])
-            graph.add_node(node_name, **attributes)
+            graph.add_node(node_name, original_name=original_name, **attributes)
```
andrey-churkin, line 241: Why do you use the keyword argument?

beta/nncf/tensorflow/pruning/base_algorithm.py
```diff
@@ -87,6 +84,7 @@ def __init__(self, config):
                                                           self._prune_downsample_convs)
 
         self._pruned_layer_groups_info = None
+        self._pruned_spec_layers = list()
```
andrey-churkin, line 87:
```suggestion
        self._pruned_spec_layers = []
```

beta/nncf/tensorflow/pruning/export_helpers.py
```diff
@@ -65,6 +79,19 @@ def accept_pruned_input(cls, node: NNCFNode):
...
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
+        input_masks = get_input_masks(node, graph)
+        output_mask = node.data.get('output_mask', None)
+
+        if is_grouped_conv(node):
+            if is_depthwise_conv(node):
+                output_mask = input_masks[0]
+            else:
+                output_mask = None
```
andrey-churkin, line 91:
```suggestion
            output_mask = None
            if is_depthwise_conv(node):
                output_mask = input_masks[0]
```

beta/nncf/tensorflow/pruning/export_helpers.py
```diff
@@ -92,6 +137,59 @@ class TFConcat(DefaultMetaOp):
     def accept_pruned_input(cls, node: NNCFNode):
         return True
 
+    @classmethod
+    def check_concat(cls, node: NNCFNode, graph: NNCFGraph):
+        """
+        Return whether all input sources of node is convolutions or not
+        :param node: node to determine it's sources
+        :param graph: NNCF graph to work with
+        """
```
andrey-churkin, line 146: Can you refactor the docstrings according to the style guide, please?

beta/tests/tensorflow/pruning/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+ Copyright (c) 2020 Intel Corporation
```
andrey-churkin, line 2:
```suggestion
 Copyright (c) 2021 Intel Corporation
```

nncf/common/pruning/export_helpers.py
```diff
@@ -30,6 +31,15 @@ def accept_pruned_input(cls, node: NNCFNode):
         """
         raise NotImplementedError
 
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
+        """
+        Propagates the pruning mask through a node using pruning masks of all inputs and the current node (if any).
+        :param node: The graph node to propagate mask through it
+        :param graph: The model graph to prune
+        """
```
andrey-churkin, line 40: Can you refactor the docstrings according to the style guide, please?

nncf/common/graph/graph.py
```diff
@@ -171,6 +171,19 @@ def get_previous_nodes(self, node: NNCFNode) -> List[NNCFNode]:
         nx_node_keys = self._nx_graph.pred[self._node_id_to_key_dict[node.node_id]]
         return [self._nx_node_to_nncf_node(self._nx_graph.nodes[key]) for key in nx_node_keys]
 
+    def get_input_edges(self, node: NNCFNode) -> List[dict]:
+        """
+        Returns shapes of input tensors
+
+        :param node: Consumer node.
+        :return: List of shapes of input tensors.
+        """
```
andrey-churkin, line 180: Can you refactor the docstrings according to the style guide, please?

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,74 @@
...
+from nncf.common.pruning.export_helpers import DefaultMetaOp
+from nncf.common.pruning.utils import PruningOperationsMetatypeRegistry
+
+
+class MaskPropagationAlgorithm:
+    """
+    Algorithm responsible for propagation masks across all nodes in the graph.
+    Before mask_propagation() you need set node.data['output_masks']
+    for nodes that have masks already defined.
+    """
```
andrey-churkin, line 25: Can you add an empty line after the class docstring?

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        Initializes MaskPropagationAlgorithm.
+
+        :param graph: Graph to work with.
+        :param pruning_operator_metatypes: registry with operation metatypes pruning algorithm is aware of, i.e.
+                metatypes describing operations with common pruning mask application and propagation properties, e.g.
+                IdentityMaskForwardOps unifies operations that propagate pruning masks as is (relu, swish etc.), whereas
+                Convolution unifies different convolution operations (conv1d, conv2d, conv3d) which accepts some input masks
+                and provide some output masks.
+        """
```
andrey-churkin: Can you refactor the docstrings according to the style guide, please?

andrey-churkin (changes requested):

beta/nncf/tensorflow/graph/converter.py
AlexanderDokuchaev: In case of shared layers (when one layer calls several time), node_name is 'nncf_wrapped_conv1^0', 'nncf_wrapped_conv1^1'.
To find corresponding nncf_node we should know original_name,

andrey-churkin: I mean, why we can not write something like this

```suggestion
            graph.add_node(node_name, original_name, **attributes)
```

AlexanderDokuchaev: `graph.add_node(node_name, original_name=original_name, **attributes)` is short analog 
```
attributes['original_name'] = original_name
graph.add_node(node_name, **attributes)
```

And `graph.add_node(node_name, original_name, **attributes)`  looks like TypeError, because 
```
def add_node(self, node_for_adding, **attr):
```

andrey-churkin: Ok, thanks.

alexsu52 (commented):
Some reference graphs were changed in the graph test. Would you short explain a reason?

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,75 @@
...
+        Initializes MaskPropagationAlgorithm.
+
+        :param graph: Graph to work with.
+        :param pruning_operator_metatypes: registry with operation metatypes pruning algorithm is aware of, i.e.
+               metatypes describing operations with common pruning mask application and propagation properties.
+        """
+        self._graph = graph
+        self._pruning_operator_metatypes = pruning_operator_metatypes
+
+    def get_class_by_type_name(self, type_name: str) -> DefaultMetaOp:
```
alexsu52, line 38:
```suggestion
    def get_meta_operation_by_type_name(self, type_name: str) -> DefaultMetaOp:

```

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,75 @@
...
+    Before call mask_propagation() you need set node.data['output_masks']
+    for nodes that have masks already defined.
+    """
+
+    def __init__(self, graph: NNCFGraph, pruning_operator_metatypes: PruningOperationsMetatypeRegistry):
+        """
+        Initializes MaskPropagationAlgorithm.
+
+        :param graph: Graph to work with.
+        :param pruning_operator_metatypes: registry with operation metatypes pruning algorithm is aware of, i.e.
```
alexsu52, line 32: Please write the docstrings here and below in this file following the style guide. The example from the style guide:
```
def load_state(model: torch.nn.Module, state_dict_to_load: dict, is_resume: bool = False) -> int:
    """
    Used to load a checkpoint containing a compressed model into an NNCFNetwork object, but can
    be used for any PyTorch module as well. Will do matching of state_dict_to_load parameters to
    the model's state_dict parameters while discarding irrelevant prefixes added during wrapping
    in NNCFNetwork or DataParallel/DistributedDataParallel objects, and load the matched parameters
    from the state_dict_to_load into the model's state dict.

    :param model: The target module for the state_dict_to_load to be loaded to.
    :param state_dict_to_load: A state dict containing the parameters to be loaded into the model.
    :param is_resume: Determines the behavior when the function cannot do a successful parameter 
        match when loading. If True, the function will raise an exception if it cannot match 
        the state_dict_to_load parameters to the model's parameters (i.e. if some parameters required 
        by model are missing in state_dict_to_load, or if state_dict_to_load has parameters that 
        could not be matched to model parameters, or if the shape of parameters is not matching). 
        If False, the exception won't be raised. Usually is_resume is specified as False when loading 
        uncompressed model's weights into the model with compression algorithms already applied, and 
        as True when loading a compressed model's weights into the model with compression algorithms 
        applied to evaluate the model.
    :return: The number of state_dict_to_load entries successfully matched and loaded into model.
    """
```

AlexanderDokuchaev:
@alexsu52  Reference graphs changes because now TFInsertionCommand added to all batchnorm layer.
It's need because mask for batch norm after not convolution layer exists only after call mask propagation.
```
   conv1 conv2
     \   /
add \ mul \ concat
       |
  batch norm
```
 
And in RetinaNet InsertionsCommand was added twice for this case (https://github.com/openvinotoolkit/nncf/pull/574#issuecomment-811730939).

vshampor:
Jenkins please retry a build

nncf/pruning/export_helpers.py
```diff
@@ -10,58 +10,94 @@
...
         raise NotImplementedError
 
     @classmethod
-    def input_prune(cls, model: NNCFNetwork, nx_node: dict, graph: PTNNCFGraph, nx_graph: nx.DiGraph):
+    def input_prune(cls, model: NNCFNetwork, node: PTNNCFNode, graph: PTNNCFGraph):
         """
-        Prune nx_node by input_masks (if masks is not none and operation support it).
+        Prune node by input_masks (if masks is not none and operation support it).
+
+        :param model: NNCF nectork.
```
mkaglins, line 88: typo:  `NNCF nectork` -> `NNCF network`

mkaglins (approved):

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        for node in self._graph.topological_sort():
+            cls = self.get_meta_operation_by_type_name(node.node_type)
+            cls.mask_propagation(node, self._graph)
+
+
+def get_input_masks(node: NNCFNode, graph: NNCFGraph):
+    """
+    Return input masks for all inputs of nx_node.
+    """
```
alexsu52, line 63:
```suggestion
    """
    Returns input masks for all inputs of nx_node.
    
    :return: Input masks.
    """
```

nncf/common/pruning/mask_propagation.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        Mask propagation in graph:
+        to propagate masks run method mask_propagation (of metaop of current node) on all nodes in topological order.
+        """
+        for node in self._graph.topological_sort():
+            cls = self.get_meta_operation_by_type_name(node.node_type)
+            cls.mask_propagation(node, self._graph)
+
+
+def get_input_masks(node: NNCFNode, graph: NNCFGraph):
```
alexsu52, line 60: please, add a type hint for the return value

AlexanderDokuchaev: @alexsu52  Using `__nccf_backend__`  and condition to define `Tensor` , is it ok? Or can be done better.

```
from nncf.common.utils.backend import __nncf_backend__
if __nncf_backend__ == 'Torch':
    from torch import Tensor
elif __nncf_backend__ == 'TensorFlow':
    from tensorflow import Tensor
```

alexsu52 (approved):

nncf/common/pruning/mask_propagation.py
alexsu52: Thanks for the comment! It looks like a workaround. I would suggest using NumPy representation in the common part or the custom type hints.
```
```

AlexanderDokuchaev: Updated to typevar, i think better use framework representation, in case with nympy need to add logic to convert nympy <-> framework with saving device type. It's overkill to fix "typing" problem.

AlexanderDokuchaev:
@alexsu52 @andrey-churkin @evgeniya-egupova all comments was fixed

alexsu52 (approved):

andrey-churkin (approved):

PR #28, opened 2020-06-09
Align documentation with latest master

vshampor:

PR #613, opened 2021-03-29
Fixed test_hawq_on_single_conv_without_quantizers within CPU-only precommit

ljaljushkin:
Also mocked torch sampling to produce identical tensors on CPU and GPU

REF: 51638

Gradient is calculated slightly differently on CPU and GPU for identical inputs.
image


ljaljushkin:
started nncf_cpu_only_precommit #28

vshampor (approved):

PR #614, opened 2021-03-29
Fix bug with no_nncf_trace context

aleksei-kashapov:


aleksei-kashapov:
@LeonidBeynenson

nncf/dynamic_graph/context.py
```diff
@@ -516,11 +517,16 @@ def reset_graph(self):
...
+    if ctx:
+        if ctx.is_tracing:
+            ctx.disable_tracing()
+            yield
+            if ctx:
+                ctx.enable_tracing()
+        else:
+            yield
+    else:
+        yield
```
vshampor, line 529: This code looks worse that the one above.
1) Avoid nestings if possible, use compound conditions.
2) Do not do `if a:` when you actually try to test if `a` name points to None. Do `if a is not None`.
3) What problem does this solve? Such explanations should be present in the commit message.

vshampor (changes requested):

aleksei-kashapov:
This PR resolves a problem when we have a nested `no_nncf_trace()` context inside another `no_nncf_trace()`.
Without changes it works as follows: when the program exits the nested context we update a global tracing variable to a disabled tracing state. In that case, another context will be influenced and NNCF starts tracing while it is not expected behavior. 

With this PR nested context checks whether we need to update state from tracing to non_tracing if not we just yield nothing and it doesn't influence another no_nncf_trace() context

nncf/dynamic_graph/context.py
aleksei-kashapov: 1. Updated
2. Updated
3. Added a description. I hope, now it is more clear

vshampor: Please also add a simple test for this logic. Sorry I did not mention it in the previous message.

aleksei-kashapov: Add a test to `test_graph_building.py`

vshampor (approved):

PR #619, opened 2021-03-31
Fix input channel axes

evgeniya-egupova:


beta/nncf/tensorflow/layers/data_layout.py
```diff
@@ -39,7 +42,16 @@ def get_data_format(layer):
...
+    original_layer = layer.layer if isinstance(layer, NNCFWrapper) else layer
     data_format = get_data_format(layer)
+    class_name = original_layer.__class__.__name__
+    if class_name in GENERAL_CONV_LAYERS:
+        return -1 if data_format == 'channels_last' else -1 - original_layer.rank
+    if isinstance(original_layer, (tf.keras.layers.BatchNormalization,
+                                   tf.keras.layers.LayerNormalization)):
+        return original_layer.axis
+    # works for Dense, Cropping family, Pooling family, SeparableConv family,
+    # SpatialDropout family, UpSampling family, ZeroPadding family
```
alexsu52, line 54: I recommend don't add such comments, because if you want to check layer type you should do it in the code.

alexsu52 (changes requested):
Could you add a model with inputs for which the number of tensor dimensions > tensor rank to the graph test to check if the quantization algorithm handles this case correctly.

alexsu52 (approved):

PR #622, opened 2021-04-01
Test sota checkpoint, batch size fix

negvet:
What's new:

`batch` key in `beta/tests/sota_checkpoints_eval.json`  renamed to `batch_per_gpu`. 

If `batch_per_gpu` is provided - `global_batch` is getting calculated at `beta/tests/test_sota_checkpoints.py` using the `BATCH_COEFF`.

If `batch_per_gpu` is NOT provided in `beta/tests/sota_checkpoints_eval.json` - the `global_batch` is getting from main model config.

### DIFF
--- a/beta/tests/test_sota_checkpoints.py
+++ b/beta/tests/test_sota_checkpoints.py
@@ -18,6 +18,7 @@ import csv
 import datetime
 from typing import Tuple, List, Optional
 
+import tensorflow as tf
 import pytest
 import subprocess
 import re
@@ -57,6 +58,8 @@ DATASET_TYPE_AVAILABILITY = {
     "segmentation": False,
 }
 
+num_gpus = len(tf.config.list_physical_devices('GPU'))
+BATCH_COEFF = num_gpus if num_gpus else 1
 
 class EvalRunParamsStruct:
     def __init__(self,
@@ -312,10 +315,10 @@ class TestSotaCheckpoints:
                     resume_file = model_dict[model_name].get('resume', {})
                 else:
                     resume_file = None
-                if model_dict[model_name].get('batch', {}):
-                    batch = model_dict[model_name].get('batch', {})
+                if model_dict[model_name].get('batch_per_gpu', {}):
+                    global_batch = model_dict[model_name]['batch_per_gpu'] * BATCH_COEFF
                 else:
-                    batch = None
+                    global_batch = None
                 if model_dict[model_name].get('mean_value', {}):
                     mean_val = model_dict[model_name].get('mean_value', {})
                 else:
@@ -340,7 +343,7 @@ class TestSotaCheckpoints:
                                                           sample_type_=sample_type_,
                                                           resume_file_=resume_file,
                                                           weights_=weights,
-                                                          batch_=batch,
+                                                          batch_=global_batch,
                                                           mean_val_=mean_val,
                                                           scale_val_=scale_val,
                                                           diff_fp32_min_=diff_fp32_min,
--- a/beta/tests/sota_checkpoints_eval.json
+++ b/beta/tests/sota_checkpoints_eval.json
@@ -161,7 +161,7 @@
                     "weights": "retinanet/retinanet.h5",
                     "metric_type": "mAP",
                     "model_description": "retinanet",
-                    "batch": 40
+                    "batch_per_gpu": 1
                 },
                 "retinanet_int8_w_sym_t_a_sym_t": {
                     "config": "examples/tensorflow/object_detection/configs/quantization/retinanet_coco_int8.json",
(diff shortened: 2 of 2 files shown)


alexsu52:
@pfinashx, please, review.

alexsu52 (approved):

daniil-lyakhov (approved):

beta/tests/test_sota_checkpoints.py
```diff
@@ -312,10 +315,10 @@ def read_metric(metric_file_name: str):
                     resume_file = model_dict[model_name].get('resume', {})
                 else:
                     resume_file = None
-                if model_dict[model_name].get('batch', {}):
-                    batch = model_dict[model_name].get('batch', {})
+                if model_dict[model_name].get('batch_per_gpu', {}):
+                    global_batch = model_dict[model_name]['batch_per_gpu'] * BATCH_COEFF
```
pfinashx, line 319: So, if we have 2 GPU in system, global_batch for, example, retinanet will be _**2**_ versus 40 in previous version?

negvet: That is correct. From time to time, retinanet runs out of memory with batch_size 40. That is the reason why we propose to decrease the batch size for some models, including retinanet.

pfinashx: Did you measure time difference between bathes 2 and 40?

pfinashx: Here job with batch 40 for model retinanet_int8_w_sym_t_a_sym_t (in fact model evaluate twice for tfds and tfrecords), it's took 21 min
And here job with your PR - https://adas-ci-icv.inn.intel.com/job/NNCF/job/nightly/job/eval_tf/50/ it's took 2h 11min.
You job with full scope of models took more then 6 hours, whereas usually this job takes 2-3 hours.

evgeniya-egupova: @negvet Batch 40 was meant to be used by 2 GPUs with 12Gb of memory, so ±20 images is a reasonable batch size for 1 GPU.

evgeniya-egupova: However, it looks like the minimum configuration for this test is 2 GPUs of 8 GB each. @pfinashx, please confirm. We probably should determine a new maximum batch size.

pfinashx: @evgeniya-egupova Yes, 2 GPU 8GB each is minimum configuration in our CI.

In my opinion - it will be enough to decrease batch to 32 minimum, to keep balance between execution time and errors.
In common I agree, that good idea to calculate batch size depending on GPU quantity in system.

negvet: In the case of 8 GB limit per GPU, the following works on my local machine:
RetinaNet FP32 - batch_per_gpu=4
YOLOv4 FP32 - batch_per_gpu=2
These are the upper limits.

andrey-churkin (approved):

beta/tests/test_sota_checkpoints.py
negvet: In the case of 8GB per GPU, I tested the following approx upper limits:
- RetinaNet FP32 - `batch_per_gpu=15` (30 `global_batch` for 2 GPUs)
- Mask-RCNN FP32 - `batch_per_gpu=9` (18 `global_batch` for 2 GPUs)

In the case of memory issues, I propose to incrementally decrease `batch_per_gpu`

pfinashx: > RetinaNet FP32 - batch_per_gpu=15 (30 global_batch for 2 GPUs)
> Mask-RCNN FP32 - batch_per_gpu=9 (18 global_batch for 2 GPUs)
Sounds good to me, let test it in job.

pfinashx (approved):

negvet:
One more issue, that seems to be solved by this PR, is that when the global batch size does not get divided by the number of GPUs without remainder - there are errors, such as

```
    ValueError: Dimension 0 in both shapes must be equal, but are 5 and 6. Shapes are [5,1000] and [6,1000].
        From merging shape 0 with other shapes. for '{{node maskrcnn/tf_op_layer_multilevel_propose_rois/top_k_rois/top_k_boxes/stack/multilevel_propose_rois/top_k_rois/top_k_boxes/stack}} = Pack[N=2, T=DT_INT32, _cloned=true, axis=-1](maskrcnn/tf_op_layer_multilevel_propose_rois/top_k_rois/top_k_boxes/stack/multilevel_propose_rois/top_k_rois/top_k_boxes/mul, maskrcnn/tf_op_layer_multilevel_propose_rois/top_k_rois/top_k_boxes/TopKV2/multilevel_propose_rois/top_k_rois/top_k_boxes/TopKV2:1)' with input shapes: [5,1000], [6,1000].
```

I am able to reproduce it with Mask-RCNN when batch=16 and 3 GPUs are available for example.

vshampor:
Jenkins please retry a build

PR #625, opened 2021-04-02
Make NNCF API interfaces abstract

asenina:


nncf/sparsity/rb/algo.py
```diff
@@ -127,3 +127,10 @@ def statistics(self, quickly_collected_only=False):
         if self._distributed and self._check_sparsity_masks:
             stats['masks_consistents'] = self._check_distributed_masks()
         return stats
+
+    def get_sparsity_init(self):
+        return self.sparsity_init
```
andrey-churkin, line 132: Why should it be added?

nncf/sparsity/magnitude/algo.py
```diff
@@ -124,3 +124,14 @@ def compression_level(self) -> CompressionLevel:
         if self.scheduler.current_sparsity_level >= self.scheduler.target_level:
             return CompressionLevel.FULL
         return CompressionLevel.PARTIAL
+
+    def get_sparsity_init(self):
+        return self.sparsity_init
```
andrey-churkin, line 129: Why should it be added?

nncf/quantization/algo.py
```diff
@@ -1276,7 +1278,7 @@ def _get_local_init_range_config(self, scope: Scope, scope_overrides: Dict[str,
         return module_init_range_config
 
     def statistics(self, quickly_collected_only=False):
-        stats = super().statistics(quickly_collected_only)
+        stats = super().statistics()
```
andrey-churkin, line 1281: Why was it changed?

andrey-churkin (changes requested):

nncf/sparsity/rb/algo.py
asenina: Deleted.

nncf/sparsity/magnitude/algo.py
asenina: Deleted.

nncf/quantization/algo.py
asenina: Fixed.

asenina:
@alexsu52 @AlexKoff88  @ljaljushkin please take a look.

beta/nncf/tensorflow/api/compression.py
```diff
@@ -104,4 +97,11 @@ def build_controller(self, model: ModelType) -> TFCompressionAlgorithmController
             algorithm-specific compression during fine-tuning.
         :return: The instance of the `CompressionAlgorithmController`.
         """
-        return TFCompressionAlgorithmController(model)
+
+
+class TFZeroCompressionLoss(CompressionLoss):
```
alexsu52, line 102: It looks like the wrong module for `TFZeroCompressionLoss`.

nncf/api/composite_compression.py
```diff
@@ -258,10 +267,6 @@ def build_controller(self, model: ModelType) -> CompositeCompressionAlgorithmCon
          algorithm-specific compression during fine-tuning.
         :return: The instance of the `CompositeCompressionAlgorithmController`.
         """
-        composite_ctrl = CompositeCompressionAlgorithmController(model)
-        for builder in self.child_builders:
-            composite_ctrl.add(builder.build_controller(model))
-        return composite_ctrl
```
alexsu52, line 264: Could you explain why you deleted the implementation of the `build_controller` method.

alexsu52 (commented):
I would like to clarify the status of the composite compression APIs. Should it be abstract? 

As I see, CompositeCompressionLoss has the same issue with zero.

nncf/api/composite_compression.py
asenina: `CompositeCompressionAlgorithmController`  and `CompositeCompressionAlgorithmBuilder` are interfaces. This classes can't be instantiated because not all abstract methods are implemented. The code that was deletedd was never executed. Method `build_controller` implemented in  `TFCompositeCompressionAlgorithmBuilder` and `PTCompositeCompressionAlgorithmBuilder`.

alexsu52: Thank you for the answer. I understand, Should we remove remove ` def build_controller(self, model: ModelType) -> CompositeCompressionAlgorithmController:` from the CompositeCompressionAlgorithmBuilder abstraction class. It looks like the `get_transformation_layout` method has the same issue.

nncf/api/compression.py
```diff
@@ -58,7 +57,7 @@ def __call__(self, *args, **kwargs) -> Any:
         return self.calculate(*args, **kwargs)
 
 
-class CompressionScheduler:
+class CompressionScheduler(ABC):
```
andrey-churkin, line 60: Do we really want to make this class abstract? Seems like it has the concrete implementation and I can not imagine other implementation.
@alexsu52 @vshampor What do you think?

vshampor: Leaving this as an interface won't hurt.

nncf/api/compression.py
```diff
@@ -224,14 +201,14 @@ def statistics(self, quickly_collected_only: bool = False) -> Dict[str, object]:
             need to keep track of statistics on each training batch/step/iteration.
         :return: A dictionary of printable statistics.
         """
-        return self._loss.statistics(quickly_collected_only)
```
andrey-churkin, line 227: Please confirm that all controllers where `self._loss` is used add `self._loss.statistics()` to their statistics. Seems like it is not true now. (For example, see `RBSparsityController`).

beta/nncf/tensorflow/api/compression.py
asenina: Done.

vshampor (approved):

nncf/api/composite_compression.py
asenina: Fixed the same issue for `get_transformation_layout`.

nncf/api/compression.py
asenina: Done.

asenina:
@alexsu52 @andrey-churkin @AlexKoff88 @ljaljushkin please take another look. Please request changes or approved.

nncf/api/composite_compression.py
alexsu52: Thanks. I just want to understand. Should we override all methods of the abstraction class in the inhered class if the inhered class is abstract too? It looks like you should remove `build_controller` and `transformation_layout` methods or add `apply_to` method to the CompositeCompressionAlgorithmController for consistency.

nncf/api/composite_compression.py
```diff
@@ -224,6 +225,15 @@ def prepare_for_export(self) -> None:
             stripped_model = ctrl.strip_model(stripped_model)
         self._model = stripped_model
 
+    @property
+    def loss(self) -> CompressionLoss:
+        return self._loss
```
alexsu52, line 230: Could you check that using 0 value does not lead to "zero_loss" issue here https://github.com/openvinotoolkit/nncf/blob/29a596c98cab03a08591e5e1006430f671d50f1c/nncf/api/composite_compression.py#L57

andrey-churkin:
Why `loss` and `scheduler` properties should be abstract? What is the value of these changes? @vshampor Could you please provide a justification?

I have the following concern about these changes:
We copy-paste the same implementation of the `loss` and `scheduler` properties ~10 times. I think it isn't good.

I understand we can not create an instance of the `CompressionLoss`/`CompressionScheduler` class in the `__init__()` method of
the `CompressionAlgorithmController` class in the current implementation. Also, it isn't a good idea to return `None` in the properties.

I suggest passing the `loss` and `scheduler` objects as the parameters to the `__init__()` method:

```python
class CompressionAlgorithmController(ABC):
    def __init__(self, target_model: ModelType, scheduler: CompressionScheduler, loss: CompressionLoss):
        ...
```

We have a builder that is responsible for the creation of a controller. Every builder has enough information about how to initialize its controller. So, the builder can initialize its controller correctly.

@alexsu52, @AlexKoff88 Could you please provide your opinion about this?

nncf/api/compression.py
andrey-churkin: Only for my understanding, what is the value of these changes?

vshampor:
@andrey-churkin the abstract classes should not have an `__init__` function at all, and if they do in this PR, then this is bad design which should be rectified in the scope of this PR, or in the future if, due to historical reasons, this cannot be done right now.

With the controllers we seem to have a following obligation to the user - from every controller it is possible to obtain an associated scheduler and a loss. This obligation is a contract, *an interface*. The interface invocations are usually function calls; therefore, the obligation translates to the interface declaring `get_loss()` and `get_scheduler()` functions. In Python, specifying corresponding `loss` and `scheduler` abstract properties is fundamentally equivalent to declaring these functions.

If you find yourself *implementing* the `loss` and `scheduler` in specific classes multiple times in exactly the same manner, then use an NNCF-internal `BaseCompressionAlgorithm(CompressionAlgorithmAlgorithm)` class that defines `loss` and `scheduler` once in that same manner, and inherit exact compression algorithms from this base class.

andrey-churkin:
> the abstract classes should not have an `__init__` function at all

@vshampor I disagree with this statement. I think that in order to accept this limitation, arguments are needed.
In my opinion, it is absolutely normal. For example, see these precedents [numpy](https://github.com/numpy/numpy/blob/main/numpy/polynomial/_polybase.py#L290), [matplotlib](https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/animation.py#L178), [pytorch_1](https://github.com/pytorch/pytorch/blob/master/torch/quantization/observer.py#L56), [pytorch_2](https://github.com/pytorch/pytorch/blob/master/torch/distributed/elastic/multiprocessing/api.py#L140), [pytorch_3](https://github.com/pytorch/pytorch/blob/master/torch/distributed/elastic/timer/api.py#L111).

If I understood correctly, you suggest making classes fully abstract without any implementation. What is the motivation to do these changes?

vshampor:
@andrey-churkin let me correct myself. A class defining an *interface* should not have an `__init__` function. We are trying to turn most of the API into actual interfaces.

beta/nncf/tensorflow/pruning/filter_pruning/algorithm.py
```diff
@@ -231,3 +236,16 @@ def _set_binary_masks_for_all_pruned_layers(self, pruning_rate: float):
             nncf_node = [n for n in nncf_sorted_nodes if wrapped_layer.name == n.node_name][0]
             if nncf_node.data['output_mask'] is not None:
                 self._set_operation_masks([wrapped_layer], nncf_node.data['output_mask'])
+
+    @property
+    def scheduler(self) -> CompressionScheduler:
```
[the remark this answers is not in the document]
asenina: Done.

beta/nncf/tensorflow/pruning/filter_pruning/algorithm.py
```diff
@@ -231,3 +236,16 @@ def _set_binary_masks_for_all_pruned_layers(self, pruning_rate: float):
...
+        return self._scheduler
+
+    @property
+    def loss(self) -> CompressionLoss:
+        return self._loss
+
+    def statistics(self, quickly_collected_only=False):
+        stats = super().statistics(quickly_collected_only)
+        stats.update(self._loss.statistics(quickly_collected_only))
+        return stats
```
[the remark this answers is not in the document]
asenina: Done.

nncf/api/compression.py
andrey-churkin: It was discussed offline. I understand arguments to use this way.

beta/nncf/tensorflow/sparsity/magnitude/algorithm.py
```diff
@@ -218,3 +230,7 @@ def raw_statistics(self):
             })
 
         return raw_sparsity_statistics
+
+    def statistics(self, quickly_collected_only=False):
+        stats = super().statistics(quickly_collected_only)
+        return stats
```
alexsu52, line 236:
```suggestion
```

beta/nncf/tensorflow/sparsity/rb/algorithm.py
```diff
@@ -160,3 +170,7 @@ def raw_statistics(self):
             })
 
         return raw_sparsity_statistics
+
+    def statistics(self, quickly_collected_only=False):
+        stats = super().statistics(quickly_collected_only)
+        return stats
```
alexsu52, line 176:
```suggestion
```

nncf/api/composite_compression.py
```diff
@@ -225,6 +236,7 @@ def prepare_for_export(self) -> None:
         self._model = stripped_model
 
 
+
```
alexsu52, line 239:
```suggestion
```

alexsu52 (approved):

andrey-churkin (approved):

vshampor:
Jenkins please retry a build

asenina:
Jenkins please retry a build

PR #626, opened 2021-04-02
Switch gnc_graph calls in test to using positional args

vshampor:
The random_index mechanism in networkx looks sketchy as it only
inspects args for a seed value (aka "random_state"), and not kwargs.
This is a workaround for "random_state_index is incorrect" errors.


andrey-churkin (approved):

asenina (approved):

vshampor:
Jenkins please retry a build

PR #627, opened 2021-04-03
Separate traced graph

vshampor:


tests/data/reference_graphs/const_sparsity/inception_v3.dot
```diff
@@ -416,16 +416,6 @@ strict digraph  {
...
-"420 Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv0]/RELU" [id=420, scope="Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv0]", type=RELU];
-"421 Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]/Conv2d[conv]/conv2d" [id=421, scope="Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]/Conv2d[conv]", type=conv2d];
-"422 Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]/BatchNorm2d[bn]/batch_norm" [id=422, scope="Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]/BatchNorm2d[bn]", type=batch_norm];
-"423 Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]/RELU" [id=423, scope="Inception3/InceptionAux[AuxLogits]/BasicConv2d[conv1]", type=RELU];
-"424 Inception3/InceptionAux[AuxLogits]/adaptive_avg_pool2d" [id=424, scope="Inception3/InceptionAux[AuxLogits]", type=adaptive_avg_pool2d];
-"425 Inception3/InceptionAux[AuxLogits]/Linear[fc]/linear" [id=425, scope="Inception3/InceptionAux[AuxLogits]/Linear[fc]", type=linear];
-"426 /nncf_model_output" [id=426, scope="", type=nncf_model_output];
```
ljaljushkin, line 428: Probably output of the Aux branch would be useful to catch via `nncf_model_output` node in the future.
It might be needed for Knowledge Distillation, not sure we shouldn't trace auxiliary graphs.

BTW, could you please point to the code which enforces to not add nodes for auxiliary graphs?

vshampor: I didn't debug this exhaustively, but it seems to me that this is a result of the model being switched to eval before compression; the `model.eval()` call is located at the very start of the `create_compressed_model` function.

ljaljushkin: It might be related, but this change was [previously introduced](https://github.com/openvinotoolkit/nncf/pull/257/files#diff-efe67aa3336974c237f5b42e96b8b218087b5df3c9ce97ad473676b95a5c7a3fR109) just to not compress auxiliary subgraphs, but not to remove these nodes from the graph.

vshampor: @ljaljushkin I restored checking reference graphs against the train-state of the model for tests. For the user, however, the compressed_graph.dot built right after create_compressed_model remains as if it were built for the eval state (it has always been this way, in fact). We may change this behavior later, if it is inconvenient.

ljaljushkin (approved):

asenina (approved):

PR #630, opened 2021-04-04
Add env var for regen .dot files. 

asenina:


alexsu52 (changes requested):
RB sparsity algorithm and filter pruning algorithm do not support models with custom layers such as mobilenet_v2_slim. This model should be excluded from the test for these algorithms. Could you file an issue for the finding or fix it directly?

alexsu52 (approved):

PR #636, opened 2021-04-06
Remove optimizer from a checkpoint

negvet:


beta/examples/tensorflow/common/prepare_checkpoint.py
```diff
@@ -0,0 +1,106 @@
...
+from beta.examples.tensorflow.common.utils import configure_paths
+from beta.examples.tensorflow.object_detection.models.model_selector import get_predefined_config
+from beta.examples.tensorflow.object_detection.models.model_selector import get_model_builder
+
+
+def get_argument_parser():
+    parser = get_common_argument_parser(precision=False,
+                                        save_checkpoint_freq=False)
+
+    parser.add_argument(
```
alexsu52, line 31: It looks like, the mode command line attribute is redundant here.

alexsu52 (changes requested):
Does the object detection example resume the model from the checkpoint without optimizer parameters?

beta/examples/tensorflow/common/prepare_checkpoint.py
negvet: That is true, actually, the whole `get_argument_parser` function here is redundant.

negvet:
> Does the object detection example resume the model from the checkpoint without optimizer parameters?

Yes, it does resume

beta/examples/tensorflow/common/prepare_checkpoint.py
```diff
@@ -0,0 +1,85 @@
...
+    checkpoint = tf.train.Checkpoint(model=compress_model)
+    load_checkpoint(checkpoint, config.ckpt_path)
+
+    checkpoint_manager = tf.train.CheckpointManager(checkpoint, config.checkpoint_save_dir, max_to_keep=None)
+    save_path = checkpoint_manager.save()
+    logger.info('Saved checkpoint: {}'.format(save_path))
+
+
+def main(argv):
+    parser = get_common_argument_parser()
```
alexsu52, line 78: Could you check that the help (prepare_checkpoint.py --help) for the prepare_checkpoint.py script is correct?

negvet: I created a separate argument parser to make help correct

alexsu52 (approved):

PR #639, opened 2021-04-08
Added End2End test for TF

pfinashx:
Finished E2E tests for TF part.


beta/requirements.txt
```diff
@@ -7,6 +7,7 @@ tensorflow_addons==0.12.0
 tensorflow_hub
 texttable
 opencv-python
+pyyaml
```
alexsu52, line 10: It looks like a new dependency for the NNCF. @vshampor  Is it an approved dependency?

vshampor: It is not, and in fact it should not be used for security reasons. YAML parsing in general is insecure, so we should avoid having to handle .yml files ourselves.

pfinashx: This package is used to modify AC configs.
Now we have configs in beta/tests/tensorflow/data/ac_configs/ only for FP32 baselines models, and in tests we modify them on-the-fly for compressed models, this modification was discussed with developers by @daniil-lyakhov

vshampor: This requirement is not for the NNCF package then, but for the TF tests. Have the requirements.txt been separated for TF yet?

pfinashx: And in fact it used other OpenVino frameworks and tools, like training_extensions and Accuracy Checker.
I can separate req's and add requirements.txt in tests/ or we need add full configs for all models.

pfinashx: Updated

pfinashx:
Jenkins please retry a build

pfinashx:
Jenkins please retry a build

alexsu52 (approved):

alexsu52:
@vshampor , please, merge if you have no comments.

PR #645, opened 2021-04-08
Fix pruning by flops for models with concat layers

AlexanderDokuchaev:


nncf/pruning/filter_pruning/algo.py
```diff
@@ -207,12 +223,16 @@ def _calculate_flops_pruned_model_by_masks(self) -> float:
...
                         group.nodes)
             new_out_channels_num = int(sum(group.nodes[0].operand.binary_filter_pruning_mask))
+            num_of_sparse_elems = len(group.nodes[0].operand.binary_filter_pruning_mask) - new_out_channels_num
             for node in group.nodes:
                 tmp_out_channels[node.module_scope] = new_out_channels_num
             # Prune in_channels in all next nodes of cluster
             next_nodes = self.next_nodes[group.id]
-            for node_module_scope in next_nodes:
-                tmp_in_channels[node_module_scope] = new_out_channels_num
+            for node_module_scope, is_after_concat in next_nodes:
```
evgeniya-egupova, line 231: Can we instead of assigning the new output shape to the input shape, simply subtract the number of pruned filters from the input shape of next nodes?

AlexanderDokuchaev: No, in case of using operation like `add` (when number of output channels each layer before `add` equal number of input next of layer), number of input channels will be reduced several times, but should only once.
```
conv1 conv2
  \   /
   add
    |
  conv3
```

evgeniya-egupova: Then shouldn't they be in one pruning group with one mask? We will subtracts filters only once.

evgeniya-egupova: In TF elementwise ops like add and multiply require inputs with the same shape, and therefore should be grouped.

AlexanderDokuchaev: Hm... yes, you are right.

mkaglins (approved):

evgeniya-egupova (approved):

PR #646, opened 2021-04-10
Unify concat scales

vshampor:
Fixes: #421


nncf/quantization/structs.py
```diff
@@ -24,3 +25,18 @@ def __init__(self,
...
+    UNIFY_ONLY_PER_TENSOR - only results in scale unification if per-tensor quantization is ultimately applied.
+    This is the target scenario for concat unified scales since the channel count between the concatenated tensors
+    may be mismatching and, more importantly, the concatenation might occur on exactly the channel dimension which
+    means that the concatenated tensor must reuse all quantization scales of the input per-channel
+    quantized tensors.
+    UNIFY_ALWAYS - results in scale unification for both per-channel and per-tensor quantization. This is the
+    target scenario for eltwise unified scales, as it is assumed that the eltwise ops have matching input
+    tensor shapes and therefore the quantization channel count is the same.
+    """
+    UNIFY_ONLY_PER_TENSOR = 0
```
AlexKoff88, line 41: Good point, BTW. There can be a case when we have a DW Conv after Concat op, which means that we can quantize activations per channel and there is no need to align the inputs of Concat.

nncf/quantization/algo.py
```diff
@@ -1600,7 +1602,10 @@ def apply_new_quantizer_setup(self, quantizer_setup: SingleConfigQuantizerSetup)
                 quant_module.num_bits = qp.qconfig.num_bits
             return self, self._target_model_ref
         new_model = self._target_model_ref.get_clean_shallow_copy()
-        new_builder = ExperimentalQuantizationBuilder(quantizer_setup, self._tensor_stats, self.hw_config)
+        new_builder = ExperimentalQuantizationBuilder(self._quantizer_setup,
```
AlexKoff88, line 1605: I wonder why it is called `Experimental...`?

vshampor: This kind of builder produces a special controller/model pair that is able to produce differently quantized models out of a certain predefined set; namely, at given point in model you can choose between one of possible quantizer configs (the set of possible quantizer configs is determined by the propagation algo) and then obtain a model that is quantized in selected manner. 

The concat scale unification impacted the experimental builder because a change in the selection of a quantizer configuration at a given location may actually impact the way the scales are unified in the resulting setup - for instance, if 5 quantizers in a single unified scale group have per-tensor quantization *selected*, then it is indeed possible to unify all of them and use a single actual quantizer module to execute the quantization. If, for the same 5 quantizers, there is an option to select per-channel variants, and the user selects 3 quantizers to be per-tensor and 2 to be per-channel, then the per-tensor subgroup will remain unified, and the remaining 2 will either be unified or not unified, based on the associated property `UNIFY_ONLY_PER_TENSOR` or `UNIFY_ALWAYS`. Note that in this case, when we have 3 per-tensor and 2 per-channel quantizers in the same group, we cannot make all of them unified.

AlexKoff88: Thanks for the explanation but I can't imagine how this can happen.

PR #652, opened 2021-04-13
Add compresion_lr_scale

a-ignatyev:


nncf/binarization/algo.py
```diff
@@ -71,7 +71,9 @@ def _binarize_weights_and_module_inputs(self, target_model: NNCFNetwork) -> List
                 op_weights = self.__create_binarize_module().to(device)
 
                 nncf_logger.info("Adding Activation binarizer in scope: {}".format(scope_str))
-                op_inputs = UpdateInputs(ActivationBinarizationScaleThreshold(module.weight.shape)).to(device)
+                compression_lr_scale = self.config.get("compression_lr_scale", None)
```
vshampor, line 74: Should call it `compression_lr_multiplier`, as we already have `scale` referring to the quantizer scale.

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,147 @@
...
+        if param.requires_grad:
+            if isinstance(param, CompressionParameter):
+                compression_params.append(param)
+            else:
+                regular_params.append(param)
+
+    return regular_params, compression_params
+
+
+def make_train_step(config):
```
vshampor, line 83: `get_model_parameters_after_single_optimizer_step`. Setting up type hints in the helper functions is also good for both understanding and debugging.

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,147 @@
...
+    config = get_quantization_config_without_range_init()
+    return config
+
+
+def get_sparsity_config():
+    config = get_basic_sparsity_config()
+    return config
+
+
+def modify_config(config, multiplier, multiplier_location, use_algo_list=False):
```
vshampor, line 46: `add_compression_multiplier_entry_to_config`

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+@pytest.mark.parametrize("get_config", [
+    get_quantization_config,
+    get_sparsity_config
+])
+@pytest.mark.parametrize("multiplier", [0, 1, 10])
+@pytest.mark.parametrize("multiplier_location", ["local", "global", "both"])
+@pytest.mark.parametrize("use_algo_list", [True, False])
+def test_gradients(get_config, multiplier, multiplier_location, use_algo_list):
```
vshampor, line 112: Also possible to set up type hints here.

vshampor: `test_compression_lr_multiplier_affects_compression_param_gradients`

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,147 @@
...
+
+        loss.backward()
+        optimizer.step()
+
+    return model.parameters()
+
+
+@pytest.mark.parametrize("get_config", [
+    get_quantization_config,
+    get_sparsity_config
```
vshampor, line 107: Should be extended for all algos and, if possible, combinations. See the `COMPRESSION_ALGORITHMS` registry for a complete list, but make sure to filter out the `NoCompressionAlgorithm` first. Something similar is being done in tests.test_algo_common.test_target_device_is_propagated_to_algos

vshampor (changes requested):

nncf/binarization/algo.py
a-ignatyev: Done.

nncf/sparsity/rb/layers.py
```diff
@@ -19,17 +19,18 @@
 from nncf.sparsity.layers import BinaryMask
 from .functions import calc_rb_binary_mask, binary_mask
 from ...functions import logit
-from ...layer_utils import COMPRESSION_MODULES
+from ...layer_utils import COMPRESSION_MODULES, CompressionParameter
```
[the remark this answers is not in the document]
a-ignatyev: Done.

nncf/config/schema.py
```diff
@@ -729,6 +732,7 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
         # Validation of each separate compression description schema occurs in a separate step.
         # This is required for better user feedback, since holistic schema validation is uninformative
         # if there is an error in one of the compression configs.
+        "compression_lr_multiplier": _NUMBER,
```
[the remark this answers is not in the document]
a-ignatyev: Done.

nncf/layer_utils.py
```diff
@@ -79,3 +79,14 @@ def forward(self, *args):
             if op_results is not None:
                 results = op_results
         return results
+
+
+class CompressionParameter(nn.Parameter):
+    def __new__(cls, data=None, requires_grad=True, compression_lr_multiplier=None):
```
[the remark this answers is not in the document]
a-ignatyev: Done.

tests/helpers.py
```diff
@@ -169,9 +170,17 @@ def get_grads(variables):
 def check_equal(test, reference, rtol=1e-4):
     for i, (x, y) in enumerate(zip(test, reference)):
         x = x.cpu().detach().numpy()
+        y = y.cpu().detach().numpy()
```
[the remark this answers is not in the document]
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,147 @@
...
+
+    if multiplier_location in ['global', 'both']:
+        config.update({
+            'compression_lr_multiplier': multiplier
+        })
+
+    return config
+
+
+def divide_params(params):
```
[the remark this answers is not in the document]
a-ignatyev: This function no longer exists.

tests/test_compression_lr_multiplier.py
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
a-ignatyev: Partially done. Now test configs could have multiple compression algorithms.

nncf/config/schema.py
```diff
@@ -729,6 +738,9 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
         # Validation of each separate compression description schema occurs in a separate step.
         # This is required for better user feedback, since holistic schema validation is uninformative
         # if there is an error in one of the compression configs.
+        "compression_lr_multiplier": with_attributes(_NUMBER,
+                                                     description="Used to increase/decrease gradients "
+                                                                 "for compression algorithms' parameters."),
```
vshampor, line 743: You duplicate the entry four times - what if we need to adjust the description? Extract the sub-dict `COMPRESSION_LR_MULTIPLIER_PROPERTY = {"compression_lr_multiplier": with_attributes(_NUMBER, ...)} ` and in each dict where the corresponding entry has to appear, do the `{..., **COMPRESSION_LR_MULTIPLIER_PROPERTY, ...}` expansion.

nncf/layer_utils.py
```diff
@@ -79,3 +79,21 @@ def forward(self, *args):
...
+    to increase/decrease gradients for compression algorithms' parameters.
+    '''
+
+    def __new__(cls, data=None, requires_grad=True, compression_lr_multiplier=None):
+        return super().__new__(cls, data, requires_grad=requires_grad)
+
+    def __init__(self, data=None, requires_grad=True, compression_lr_multiplier=None):
+        super().__init__()
+        if compression_lr_multiplier is not None and requires_grad:
+            self.register_hook(lambda grad: compression_lr_multiplier * grad)
```
vshampor, line 98: Set up type hints and add a docstring for `__init__`

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,268 @@
...
+
+    for algo in multipliers:
+        ref_grads = get_grads(ref_params[algo])
+        ref_grads = [multipliers[algo] * grad for grad in ref_grads]
+        target_grads = get_grads(target_params[algo])
+
+        check_equal(ref_grads, target_grads)
+
+
+def test_how_multipliers_affect_params_change(ref_and_target_configs: Tuple[NNCFConfig, NNCFConfig]):
```
vshampor, line 255: A good test name describes what is being tested. This is not always possible due to the constraint on the length of the function name, but in this case, the test case could be named `test_setting_multipliers_affects_model_parameters`

tests/test_compression_lr_multiplier.py
```diff
@@ -0,0 +1,268 @@
...
+def test_if_algorithms_add_params(ref_and_target_configs: Tuple[NNCFConfig, NNCFConfig]):
+    base_config, config_with_multiplier = ref_and_target_configs
+
+    algo_to_params = get_params_after_train_steps(config_with_multiplier)
+    algo_names = get_multipliers_from_config(base_config).keys()
+
+    assert sorted(algo_to_params.keys()) == sorted(list(algo_names) + ['regular'])
+
+
+def test_how_multipliers_affect_grads(ref_and_target_configs: Tuple[NNCFConfig, NNCFConfig]):
```
vshampor, line 239: `test_grad_values_are_multiplied`

vshampor (changes requested):

tests/test_compression_lr_multiplier.py
vshampor: @vanyalzr we need you as a tie-breaker here. Which of the two variants of the test naming sounds better to you?

tests/test_compression_lr_multiplier.py
vshampor: @vanyalzr same for this test.

tests/test_compression_lr_multiplier.py
vanyalzr: `test_setting_multipliers_affects_model_parameters` sounds better to me

tests/test_compression_lr_multiplier.py
vanyalzr: same, I'd go with `test_grad_values_are_multiplied`

nncf/config/schema.py
a-ignatyev: Done.

nncf/layer_utils.py
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
a-ignatyev: Done.

tests/test_compression_lr_multiplier.py
a-ignatyev: Done.

nncf/sparsity/rb/layers.py
```diff
@@ -19,17 +19,18 @@
...
 
 
 
 @COMPRESSION_MODULES.register()
 class RBSparsifyingWeight(BinaryMask):
-    def __init__(self, size, frozen=True, eps=1e-6):
+    def __init__(self, size, frozen=True, compression_lr_multiplier=None, eps=1e-6):
```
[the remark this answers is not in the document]
vshampor: Still no type hints here.

vshampor (changes requested):

vshampor:
@alexsu52 is it technically possible to do the effective compression LR multiplication on the NNCF-side for TF?

alexsu52:
> 
> 
> @alexsu52 is it technically possible to do the effective compression LR multiplication on the NNCF-side for TF?

@vshampor I don't see any concerns to support the effective compression LR multiplication on the NNCF-side for TF. It will not be easy, because we will be able to create custom gradient functions for NNCF operations (TF does not support hooks from the box). 

I have a question relative to this PR, did you consider the implementation of this feature via the separate learning rate scheduler for the compression parameters? It looks like this approach is more flexible. What are the pros and cons of the implemented approach?

vshampor:
@alexsu52 in this way it is possible to avoid additional entities that the user has to handle in order for this feature to work. A learning rate scheduler would have to be registered explicitly by the user in their training pipeline with the optimizer - this increases the amount of work that needs to be done in order to integrate NNCF into the training pipeline. The hooking way allows the existing integration code to use this feature without any code changes.

alexsu52:
> 
> 
> @alexsu52 in this way it is possible to avoid additional entities that the user has to handle in order for this feature to work. A learning rate scheduler would have to be registered explicitly by the user in their training pipeline with the optimizer - this increases the amount of work that needs to be done in order to integrate NNCF into the training pipeline. The hooking way allows the existing integration code to use this feature without any code changes.

Thanks for your reply!

nncf/layer_utils.py
```diff
@@ -79,3 +80,37 @@ def forward(self, *args):
...
+    This class utilize `compression_lr_multiplier` parameter from :class:`nncf.NNCFConfig`
+    to increase/decrease gradients for compression algorithms' parameters.
+
+    N.B. If you want to set `requires_grad` parameter, please only use `requires_grad_` method.
+    """
+
+    def __new__(cls, data: torch.Tensor = None, requires_grad: bool = True, compression_lr_multiplier: float = None):
+        return super().__new__(cls, data, requires_grad=requires_grad)
+
+    def __init__(self, data: torch.Tensor = None, requires_grad: bool = True, compression_lr_multiplier: float = None):
```
aleksei-kashapov, line 98:
```suggestion
    def __init__(self, data: torch.Tensor = None, requires_grad: bool = True, compression_lr_multiplier: float = 1.0):
```

a-ignatyev: If `compression_lr_multiplier` is None the hook will not be applied. But in the case when `compression_lr_multiplier`=1.0 there will floating-point multiplication which could lead to accumulating differences in calculation.

vshampor (approved):

PR #658, opened 2021-04-14
Add perchannel QDQ

a-ignatyev:


aleksei-kashapov (commented):
What if the version of torch < 1.8?

a-ignatyev:
> What if the version of torch < 1.8?

A user will get the following error: `ValueError: Unsupported ONNX opset version: 13`

vshampor:
@a-ignatyev @aleksei-kashapov I modifed this behavior to (hopefully) correctly work for both PT<1.8.0 and PT >1.8.0 cases, tell me what you think about it. TBD: tests

nncf/quantization/layers.py
```diff
@@ -236,6 +248,21 @@ def run_export_quantization(self, x: torch.Tensor):
                                                       input_high)
         if self._export_mode == QuantizerExportMode.ONNX_QUANTIZE_DEQUANTIZE_PAIRS:
             x, y_scale, y_zero_point, per_channel_idx = self._prepare_qdq_export_quantization(x)
+
+            if self.per_channel:
+                if get_torch_version_tuple() < (1, 8, 0):
```
a-ignatyev, line 253: I think the inner `if` could be replaced with `and` for better readability.

nncf/quantization/quantize_functions.py
```diff
@@ -162,17 +172,14 @@ def get_scale_zp_from_input_low_input_high(level_low, level_high, input_low, inp
...
+    _, per_channel_idx = get_channel_count_and_dim_idx(list(y_scale.shape))
 
-    y_scale = torch.squeeze(y_scale)
-    y_zero_point = torch.squeeze(y_zero_point)
-    return y_scale, y_zero_point, per_channel_idx
+    if len(y_scale.shape) > 1:
+        y_scale = torch.squeeze(y_scale)
+        y_zero_point = torch.squeeze(y_zero_point)
+        return y_scale, y_zero_point, per_channel_idx
+    else:
```
a-ignatyev, line 181: I think even without this branch of `if` it should work fine but there will be less code.

nncf/quantization/layers.py
vshampor: We still have to handle the implicit "else" branch of that inner if by doing `raise RuntimeError...`, and that "else" branch should only be called for specific PT versions.

nncf/quantization/quantize_functions.py
vshampor: There may be a subtle difference between `torch.Size([])` and `torch.Size([1])` in further processing, I added this code to avoid future potential problems.

nncf/quantization/layers.py
```diff
@@ -448,11 +471,7 @@ def _prepare_export_quantization(self, x: torch.Tensor):
             level_low = self.level_low
             level_high = self.level_high
             if self._half_range:
-                if self.scale_shape[0] > 1:
-                    for i in range(self.scale_shape[0]):
-                        x[i] = torch.clamp(x[i], min=input_low[i].item(), max=input_high[i].item())
-                else:
-                    x.data = torch.clamp(x, min=input_low.item(), max=input_high.item())
+                x = torch.min(torch.max(x, input_low), input_high)
```
vshampor, line 474: @aleksei-kashapov FYI, the tests are passing with this change in place. However, I remember you having some trouble with this before. Do you recall exactly why the min-max version of clamp could not be used originally?

aleksei-kashapov: Cause torch.clamp works only in per-tensor mode

nncf/quantization/layers.py
```diff
@@ -235,15 +247,26 @@ def run_export_quantization(self, x: torch.Tensor):
...
             if self.per_channel:
-                if torch.allclose(y_scale - y_scale[0], torch.zeros_like(y_scale)) and \
-                        torch.allclose(y_zero_point - y_zero_point[0], torch.zeros_like(y_zero_point)):
-                    y_scale, y_zero_point = y_scale[0], y_zero_point[0]
-                    return ExportQuantizeToONNXQuantDequant.apply(x, y_scale, y_zero_point)
-                raise RuntimeError("PyTorch export to ONNX using QuantizeLinear-DequantizeLinear "
-                                   "doesn't support per channel quantization")
-            return ExportQuantizeToONNXQuantDequant.apply(x, y_scale, y_zero_point)
+                if EXPORT_ONNX_OPSET_VERSION < 13:
+                    if torch.allclose(y_scale - y_scale[0], torch.zeros_like(y_scale)) and \
```
aleksei-kashapov, line 254: Do we still have such checkpoints that need this workaround?

nncf/quantization/layers.py
aleksei-kashapov: This is incorrect formula. You do clamp only in per-tensor mode, while it can be per-channel

nncf/quantization/layers.py
```diff
@@ -587,11 +606,7 @@ def _prepare_export_quantization(self, x: torch.Tensor):
             level_low = self.level_low
             level_high = self.level_high
             if self._half_range:
-                if self.scale_shape[0] > 1:
-                    for i in range(self.scale_shape[0]):
-                        x[i] = torch.clamp(x[i], min=input_low[i].item(), max=input_high[i].item())
-                else:
-                    x.data = torch.clamp(x, min=input_low.item(), max=input_high.item())
+                x = torch.min(torch.max(x, input_low), input_high)
```
aleksei-kashapov, line 609: Same. It works only in per-tensor mode

nncf/quantization/layers.py
aleksei-kashapov: Okay, I rechecked it again and figured out that torch.min and torch.max also work in per-channel mode. 
In that case this formula is correct and much better than I had before

aleksei-kashapov (approved):

nncf/quantization/layers.py
vshampor: Not really, but I thought it would be a waste to just delete this code while it's in perfect working order

vshampor:
I am moving this back to the draft state - we cannot merge it yet because opset 13 is not properly supported by the Model Optimizer and even the FP32 models exported to ONNX opset 13 will fail when imported to OpenVINO, which breaks our E2E. Will have to wait until MO starts supporting opset 13.

vshampor:
Obsoleted by another PR.

PR #659, opened 2021-04-14
Added MobilenetV3Small

l-bat:


alexsu52 (approved):
Looks good to me

beta/tests/tensorflow/test_models/mobilenet_v3.py
```diff
@@ -11,208 +11,8 @@
...
 """
 
-from tensorflow.python.keras import backend
-from tensorflow.python.keras import models
-from tensorflow.python.keras.applications import imagenet_utils
-from tensorflow.python.keras import layers
+from beta.examples.tensorflow.common.models import MobileNetV3Small as mobilenet_v3_small
 
 
 def MobileNetV3Small(input_shape=None):
```
andrey-churkin, line 17: I think we can remove this file. @alexsu52, do you agree with me?

l-bat: Test failed without it

alexsu52: @andrey-churkin essentially, we can remove this file and import the model from the example directly. But I hope TF will fix the issue with this model and we will add the correct implementation to the test. I would leave this file in the tests.

alexsu52 (approved):

andrey-churkin (approved):

alexsu52:
Please, update https://github.com/openvinotoolkit/nncf/blob/develop/README.md too

PR #660, opened 2021-04-14
Fixes for common mask propagation

AlexanderDokuchaev:
- returned old reference graphs (https://github.com/openvinotoolkit/nncf/pull/609)
- check node by mask propagation
- fix using unordering previous nodes to generate mask for concat layer

REF: 53100, 53144


nncf/common/graph/graph.py
```diff
@@ -178,6 +178,19 @@ def get_previous_nodes(self, node: NNCFNode) -> List[NNCFNode]:
         nx_node_keys = self._nx_graph.pred[self._node_id_to_key_dict[node.node_id]]
         return [self._nx_node_to_nncf_node(self._nx_graph.nodes[key]) for key in nx_node_keys]
 
+    def get_previous_nodes_sorted_by_in_port(self, node: NNCFNode) -> List[NNCFNode]:
+        """
+        Returns producer nodes of provided node sorted by 'in_port'.
+
+        :param node: Consumer node.
+        :return: List of producers nodes of provided node.
```
mkaglins, line 186: `:return: List of producers nodes of provided node.` -> `:return: List of producers nodes of provided node sorted by in_port.`

mkaglins (approved):

nncf/common/graph/graph.py
```diff
@@ -178,6 +178,19 @@ def get_previous_nodes(self, node: NNCFNode) -> List[NNCFNode]:
         nx_node_keys = self._nx_graph.pred[self._node_id_to_key_dict[node.node_id]]
         return [self._nx_node_to_nncf_node(self._nx_graph.nodes[key]) for key in nx_node_keys]
 
+    def get_previous_nodes_sorted_by_in_port(self, node: NNCFNode) -> List[NNCFNode]:
```
alexsu52, line 181: What reason do this function is added to the graph class? Can you sort nodes by in_port in the algorithm?

vshampor: @alexsu52 I think that this is a convenience function which may stay in the graph itself for the time being, since it introduces no new dependencies. May refactor that later; meanwhile the fix here is critical for the release, so I'll merge this - hope you don't mind.

alexsu52: @vshampor Could you open a ticket for this issue?

vshampor: Ref. 53414 opened.

PR #661, opened 2021-04-14
Freeze NNCFWrapper weights

negvet:
`NNCFOperation `is having a `trainable `property - it is possible to switch operation trainable weights off from training.
`NNCFWrapper `is updated to support freezing-unfreezing of the whole layer and freezing-unfreezing of the operation weights. If globally the wrapper is frozen, the operation weights can remain trainable and vice versa.


beta/nncf/tensorflow/layers/wrapper.py
```diff
@@ -65,14 +66,75 @@ def trainable(self):
...
+        else:
+            weights = trainable_weights_by_definition
+            for trainable_weight_on in trainable_ops_weights_on:
+                weights.remove(trainable_weight_on)
+            outputs = non_trainable_weights_by_definition + weights
+        return outputs
+
+    def classify_trainable_ops_weights(self):
+        """
+        Classifies operation trainable weights by trainable property of corresponding operation.
```
alexsu52, line 116:
```suggestion
        Classifies operation trainable weights by trainable property of the corresponding operation.
```

beta/nncf/tensorflow/layers/wrapper.py
```diff
@@ -65,14 +66,75 @@ def trainable(self):
...
+        Note: Operation with trainable weights can be switched off from training.
+              Operation trainable property can be adjusted
+              to be different from NNCFWrapper global trainable property.
+
+        :return trainable_ops_weights_on: list of trainable operation weights
```
alexsu52, line 121:
```suggestion
        :return trainable_ops_weights_on: List of trainable operation weights
```

beta/nncf/tensorflow/layers/wrapper.py
```diff
@@ -65,14 +66,75 @@ def trainable(self):
...
+                which is getting updated during training.
+        :return trainable_ops_weights_off: list of trainable operation weights
```
alexsu52, line 123:
```suggestion
        :return trainable_ops_weights_off: List of trainable operation weights
```

beta/nncf/tensorflow/layers/wrapper.py
```diff
@@ -65,14 +66,75 @@ def trainable(self):
...
+        trainable_ops_weights_off = []
+        for ops in self.weights_attr_ops.values():
+            for op_name, op in ops.items():
+                op_weights = self._ops_weights[op_name]
+                for weight_val in op_weights.values():
+                    if weight_val.trainable:
+                        if op.trainable:
+                            trainable_ops_weights_on.append(weight_val)
+                        else:
+                            trainable_ops_weights_off.append(weight_val)
```
alexsu52, line 136: IMHO: Removed one level of nesting

```suggestion
                    if not weight_val.trainable:
                        continue
                        
                    if op.trainable:
                        trainable_ops_weights_on.append(weight_val)
                    else:
                        trainable_ops_weights_off.append(weight_val)
```

alexsu52 (changes requested):

alexsu52:
Could you add a test for this?

beta/nncf/tensorflow/layers/operation.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, name):
         """
         self._call_pre_hooks = OrderedDict()
         self._name = name
+        self.trainable = True
```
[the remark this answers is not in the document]
negvet: Both comments are reasonable. Fixed

beta/nncf/tensorflow/layers/operation.py
```diff
@@ -93,3 +94,11 @@ def get_config(self):
     @classmethod
     def from_config(cls, config):
         return cls(**config)
+
+    @property
```
alexsu52, line 98: Usually, properties go after the init method.

beta/nncf/tensorflow/layers/wrapper.py
```diff
@@ -65,14 +66,77 @@ def trainable(self):
...
+        else:
+            trainable_ops_weights_on, trainable_ops_weights_off = [], []
+
+        if self.trainable:
+            weights = trainable_weights_by_definition
+            for trainable_weight_off in trainable_ops_weights_off:
+                weights.remove(trainable_weight_off)
+            outputs = weights
+        else:
+            outputs = trainable_ops_weights_on
```
daniil-lyakhov, line 92: I'm a bit confused that layer which has attribute `trainable` = `False` returns some weights by default. So, if I want
1) Only freeze wrapped layer weight I have to set `layer.trainable = False`
2) Only operation weights I have to set `op.trainable = False` for each operation I want to freeze
3) To freeze whole layer I have to both set `layer.trainable = False` and  `op.trainable = False` for each operation I want to freeze. 

If I would decided to freeze whole layer I would expect that I only need to specify  `layer.trainable = False. I think it would be better to freeze wrapped layer weight and operation weight by separate attributes (like `trainable_wrapped_layer` or something like that) and use `layer.trainable` as which freeze/unfreeze all layer weight. We should discuss this.

negvet: Let me clarify.
For now, the default behavior is as follows. If you set `layer.trainable` some value, `op.trainable` will be set the same value. 

Therefore:
1. _Freeze_ both wrapped layer weights and op weights -> `layer.trainable = False`
2. _Unfreeze_ both wrapped layer weights and op weights -> `layer.trainable = True`

Introducing `trainable `property to the operations, we have the following additional options:
3. _Frozen_ wrapped layer weights but _unfrozen_ op weights -> `layer.trainable = False; op.trainbale = True`
4. _Unfrozen_ wrapped layer weights but _frozen_ op weights -> `layer.trainable = True; op.trainbale = False`

Operations have the same `trainable` property as the whole layer because this property follows a similar logic.

daniil-lyakhov: Yes, I missed that `ops.trainable` have the same value as `layer.trainable`. That's actually make sense, thank you for explanation! By the way, I think keeping this instruction somewhere in docstring would be useful for `NNCFWrapper` users

daniil-lyakhov: But it still tricky for me because `trainable` attribute isn't just the state of the layer weights but contain its own state inside. For example 
```
layer.trainable = False
...
ops.trainable = True
``` 
leads to layer weight freezing and operation weights training. But
```
ops.trainable = False
...
layer.trainable = True
```
leads to complete freezing. You can see that the logic of trainable attribute become more complex than logic of indicator variable

negvet: Sure, I will create necessary docstrings

daniil-lyakhov:
```
layer.trainable = False
...
ops.trainable = True
``` 
leads to layer weight freezing and operation weights training. But
```
ops.trainable = False
...
layer.trainable = True
```
leads to complete unfreezing. 

Is it ok, @negvet @andrey-churkin, @alexsu52?

daniil-lyakhov:
Ok, if we interpret `layer.trainable` as a global switcher it makes sense

alexsu52 (approved):
@negvet Can I merge?

daniil-lyakhov (approved):

negvet:
@alexsu52 not yet

PR #665, opened 2021-04-15
Fix CPU HAWQ ref values

vshampor:


vshampor:
Jenkins please retry a build

PR #676, opened 2021-04-20
Update the signature of the export_model method

andrey-churkin:
Related to 53446

### DIFF
--- a/nncf/compression_method_api.py
+++ b/nncf/compression_method_api.py
@@ -142,12 +142,11 @@ class PTCompressionAlgorithmController(CompressionAlgorithmController):
                                                                 num_bn_forget_steps)
             bn_adaptation_runner.run(bn_adaptation_args.data_loader, num_bn_adaptation_steps)
 
-    # pylint: disable=keyword-arg-before-vararg
     def export_model(self,
                      save_path: str,
                      input_names: Optional[List[str]] = None,
                      output_names: Optional[List[str]] = None,
-                     *args, **kwargs) -> None:
+                     model_args = None) -> None:
         """
         Used to export the compressed model for inference into the ONNX format.
         Makes method-specific preparations of the model graph,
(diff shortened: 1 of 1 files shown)


alexsu52 (approved):

nncf/compression_method_api.py
```diff
@@ -157,9 +156,16 @@ def export_model(self,
...
             `input_names` - list of input tensors names (optional).
             `output_names` - list of output tensors names (optional).
-            *args, **kwargs - if the model's `forward` requires additional parameters
-            during export, specify these here.
+            `model_args` - tuple of additional positional and keyword arguments which are
+                required for the model's `forward` during export. Should be specified in
+                the following format:
+                    - (a, b, {'x': None, 'y': y}) for positional and keyword arguments
+                    - (a, b, {}) for positional arguments only
+                    - ({'x': None, 'y': y},) for keyword arguments only
```
vshampor, line 164: This probably can be expressed as type hints as well. Shouldn't remove this human-readable description, though

andrey-churkin: I think it's impossible. All we can write is `Tuple[Any, ...]`.

andrey-churkin: It might be better to use `model_args` like this
- `((a, b), {'x': None, 'y': y})` for positional and keyword arguments
- `((a, b), {})` for positional arguments only
- `((), {'x': None, 'y': y})` for keyword arguments only
This is more intuitive and in this case, we can specify the type hint `Tuple[Tuple[Any, ...], Dict[str, Any]]`.

vshampor: `Union[Tuple[Tuple[Any, ...], Dict[str, Any]], Tuple[Dict[str, Any]]`
We should exactly match the torch's way if we decided to be aligned with it.

aleksei-kashapov (approved):

vshampor:
@andrey-churkin will the third_party_integration patches break when this is merged?

andrey-churkin:
> @andrey-churkin will the third_party_integration patches break when this is merged?

I think patches won't be broken because `model_args` are not used in these patches.

PR #680, opened 2021-04-26
Fix abstract method styleguide

daniil-lyakhov:
Fix misleading example


daniil-lyakhov:
I've killed precommits to save ci resources. Please retry builds if it's necessary

alexsu52 (approved):

PR #683, opened 2021-04-27
Preserve namedtuples when objwalking

vshampor:


tests/test_sanity_sample.py
```diff
@@ -291,20 +295,34 @@ def test_pretrained_model_train(config, tmp_path, multiprocessing_distributed, c
...
+    full_test_case_id = test_case_id + ('-distributed' if current_multiprocessing_distributed else '-dataparallel')
+    primary_test_case_name = f'test_pretrained_model_train[{full_test_case_id}]'
+    depends(request, [primary_test_case_name])
 
+
+@pytest.mark.dependency()
 @pytest.mark.parametrize(
-    "multiprocessing_distributed", [
-        pytest.param(True, marks=pytest.mark.dependency(depends=["train_distributed"])),
-        pytest.param(False, marks=pytest.mark.dependency(depends=["train_dataparallel"]))],
```
vshampor, line 305: Note that this did not actually set up dependencies between the test cases corresponding to one and the same model-dataset-execution type configuration.

PR #701, opened 2021-05-06
Fix the rest of the compression training acc targets

vshampor:


vshampor:
Won't impact precommit.

PR #702, opened 2021-05-07
Update targets for updated unified-scale impacted models

vshampor:


vshampor:
Sporadic failure in NNCF precommit, unrelated to the current changes.

PR #704, opened 2021-05-13
Adjust requirements.txt

negvet:
`nncf/beta/requirements.txt` is getting split into three lists of requirements: NNCF TF, TF tests, and TF examples. 
The installation works in any order.

### DIFF
--- a/beta/tests/requirements.txt
+++ b/beta/tests/requirements.txt
@@ -1,5 +1,3 @@
-# TF Tests
-
 PyYAML
 tensorflow_addons==0.12.0
 pytest
@@ -7,7 +5,4 @@ pytest-mock
 pytest-dependency
 yattag>=1.14.0
 prettytable>=2.0.0
-
-
-tensorflow_hub # used both for tests and examples
-
+pydot
--- a/beta/tests/requirements.txt
+++ b/beta/tests/requirements.txt
@@ -1,1 +1,13 @@
+# TF Tests
+
 PyYAML
+tensorflow_addons==0.12.0
+pytest
+pytest-mock
+pytest-dependency
+yattag>=1.14.0
+prettytable>=2.0.0
+
+
+tensorflow_hub # used both for tests and examples
+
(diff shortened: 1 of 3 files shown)


beta/tests/requirements.txt
```diff
@@ -1 +1,8 @@
-PyYAML
\ No newline at end of file
+PyYAML
+tensorflow_addons==0.12.0
```
pfinashx, line 2: It's already in main requirements.txt, I think we can do not include it here.

pfinashx (changes requested):

alexsu52 (approved):

beta/tests/requirements.txt
negvet: Not sure how to better deal with `tensorflow_addons`. I did not include it in the main `nncf/beta/requirements.txt`, because it is used by
`beta/examples/tensorflow/common/optimizer.py`
`beta/examples/tensorflow/classification/main.py`
`beta/tests/tensorflow/sparsity/rb/test_integration.py`

In the scenario of installation `NNCF TF` + `TF examples` and usage of examples, `tensorflow_addons` can be used.
On the other side, in the case of the installation `NNCF TF` + `TF tests` and running just tests, `tensorflow_addons`  also can be used.

@as-suvorov, does `tensorflow_addons` approved for including into NNCF? If yes, we can include it in `NNCF TF` requirements to avoid duplicating.

pfinashx (approved):

beta/tests/requirements.txt
alexsu52: The current solution is ok for me. `tensorflow_addons` is not a dependency for NNCF.

pfinashx:
Jenkins please retry a build

pfinashx:
CI was changed according new requirements, please check PyLint warnings.

pfinashx:
Jenkins please retry a build

vshampor:
Jenkins please retry a build

PR #707, opened 2021-05-13
Detect NotImplemented result of operation

AlexanderDokuchaev:


nncf/dynamic_graph/wrappers.py
```diff
@@ -75,16 +76,18 @@ def wrapped(*args, **kwargs):
...
+            kwargs = processed_input.op_kwargs
+            result = operator(*args, **kwargs)
+
+            if isinstance(result, type(NotImplemented)):
+                nncf_logger.debug("Operation {} returned NotImplemented".format(op_name))
+            elif node is None:
                 node = ctx.maybe_add_node(processed_input, tensor_metas, ia_op_exec_context, module_attrs)
 
             if is_debug():
                 ctx.register_node_call(node)
```
vshampor, line 89: I suppose that this should also be moved under `if node is not None`

vshampor (changes requested):

nncf/dynamic_graph/wrappers.py
AlexanderDokuchaev: done

vshampor (approved):

PR #708, opened 2021-05-14
Fixed backward compatibility issues

ljaljushkin:
- Checkpoint without |OUTPUT and |INPUT suffixes wasn't loadable
- No way to load separate compression parameters to the model with unified ones.
- Updated squeezenet1_1_custom_cifar10_rb_sparsity_int8_ddp.pth to avoid issue with duplicated FQ and with per-tensor input quantization instead of per-channel.
- fixed backward-compatibility test:
registered data_loader for range init and set proper tolerance for accuracy comparison
- refactored KeyMatcher class

resolves 53579

### DIFF
--- a/nncf/checkpoint_loading.py
+++ b/nncf/checkpoint_loading.py
@@ -10,13 +10,12 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+import re
 import warnings
 from enum import Enum
 from typing import Dict
 from typing import List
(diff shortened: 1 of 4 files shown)


ljaljushkin:
backward compatibility build #10 is green

ljaljushkin:
Jenkins please retry a build

vshampor (approved):

PR #720, opened 2021-05-20
Put BN layers into evaluation mode during AutoQ search in OD sample

vanyalzr:
BN statistics were calculated batch-wise during evaluation in the OD sample, which technically gives incorrect metric values.

### DIFF
--- a/examples/object_detection/eval.py
+++ b/examples/object_detection/eval.py
@@ -349,13 +349,29 @@ def eval_net_loss(data_loader, device, net, criterion):
 def test_net(net, device, data_loader, distributed=False, loss_inference=False, criterion=None):
     """Test a Fast R-CNN network on an image database."""
 
+    # Put BN layers into evaluation mode
+    bn_modules_training_flags = {}
+
+    def put_bn_modules_in_eval_mode(module):
+        if isinstance(module, torch.nn.BatchNorm2d):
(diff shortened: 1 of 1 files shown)


vshampor (approved):

examples/object_detection/main.py
```diff
@@ -142,6 +142,11 @@ def criterion_fn(model_outputs, target, criterion):
             return loss_l + loss_c
 
         def autoq_test_fn(model, eval_loader):
+            # Put BN layers into evaludtion mode
```
ljaljushkin, line 145: why is it applicable for object detection only?
does it make sense to do it for all models inside autoq before calling autoq_test_fn?

vanyalzr: Will move eval mode switch for BNs inside `test_net` in this sample. For other samples, we already have `model.eval()` inside validation methods. I guess it's the responsibility of the user/training sample to do correct validation of the model.

ljaljushkin (changes requested):
After discussion with Ivan, we decided to refactor test_net of the OD sample to explicitly put BN in eval mode.
@vshampor, please don't merge

ljaljushkin:
@vuiseng9, it might affect AutoQ results for object detection, because mAP became a slightly different.

ljaljushkin (approved):

PR #721, opened 2021-05-20
Rename compression level to compression stage

vanyalzr:


nncf/api/compression.py
```diff
@@ -130,33 +130,33 @@ def get_state(self) -> Dict[str, object]:
...
     NONE = 0
     PARTIAL = 1
     FULL = 2
 
-    def __add__(self, other: 'CompressionLevel') -> 'CompressionLevel':
+    def __add__(self, other: 'CompressionStage') -> 'CompressionStage':
         """
-        Defines compression level of a composite compression controller, consist of
-        two algorithms, where `self` is the compression level of the first algorithm
-        and other - compression level of the second one.
```
ljaljushkin, line 146: These names made sense for me with compression level class.
full/partial compression level 

but with stages probably we should come up with other names, like
FINAL, INTERMEDIATE

vshampor: `to distinguish between uncompressed, partially compressed, and fully compressed models`

UNCOMPRESSED, PARTIALLY_COMPRESSED, FULLY_COMPRESSED

ljaljushkin: > `to distinguish between uncompressed, partially compressed, and fully compressed models`
> 
> UNCOMPRESSED, PARTIALLY_COMPRESSED, FULLY_COMPRESSED

yes, this one is better

alexsu52 (approved):

nncf/api/compression.py
vanyalzr: agreed, done

vshampor (approved):

PR #743, opened 2021-06-01
Mobilenet V3 rb sparsity configs

daniil-lyakhov:


daniil-lyakhov:
build sota_eval_tf 191

PR #747, opened 2021-06-02
Fixed test install for pip

l-bat:
cpu build: 10
gpu build:  9

### DIFF
--- a/tests/torch/install_checks.py
+++ b/tests/torch/install_checks.py
@@ -20,9 +20,10 @@ import sys
 
 import torch
 
-if len(sys.argv) != 2:
-    raise RuntimeError("Must be run with an execution type as argument (either 'cpu' or 'cuda')")
+if len(sys.argv) != 3:
+    raise RuntimeError("Must be run with an execution type as argument (either 'cpu' or 'cuda') and package type")
 execution_type = sys.argv[1]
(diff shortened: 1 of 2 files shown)


tests/torch/install_checks.py
```diff
@@ -32,8 +32,12 @@
...
 
 if execution_type == "cpu":
-    from nncf.torch.binarization.extensions import BinarizedFunctionsCPU
-    from nncf.torch.quantization.extensions import QuantizedFunctionsCPU
+    try:
+        from nncf.torch.binarization.extensions import BinarizedFunctionsCPU
+        from nncf.torch.quantization.extensions import QuantizedFunctionsCPU
+    except ImportError:
+        from nncf.binarization.extensions import BinarizedFunctionsCPU
+        from nncf.quantization.extensions import QuantizedFunctionsCPU
```
vshampor, line 40: Should only run the `try...except` version for the "pip" version of the install checks run - pass the type of the test via CLI to do that.

vshampor (changes requested):

tests/torch/install_checks.py
l-bat: done

vshampor (approved):

PR #749, opened 2021-06-02
Freeze networkx at 2.5.*

vshampor:
The 2.6-rc1 version uses RNG in a different way and disrupts our seed-dependent tests.


setup.py
```diff
@@ -44,7 +44,7 @@ def find_version(*file_paths):
                     "addict>=2.4.0",
                     "texttable>=1.6.3",
                     "scipy>=1.3.2",
-                    "networkx>=2.5",
+                    "networkx>=2.5, <2.6",
```
alexsu52, line 47:
```suggestion
                    "networkx~=2.5",
```

vshampor:
Some more investigation showed that the problem lies rather in the fact that NNCFGraph's `topological_sort` method generates the sorted nodes in a rather arbitrary topological sorting, while it is more suitable to generate a lexicographical toposort based on the node ID (which reflects the order of the operation call in PT). For the HAWQ/AutoQ reference graphs, the problem is similar but is caused by the networkx version-dependent process of iterating over all nodes in a graph using `for node in G.nodes`. Establishing a deterministic order in two of these places will serve as a better fix for the situation and enable to use both 2.5 and 2.6 versions of networkx, will open a PR shortly.

PR #753, opened 2021-06-03
Fix wrapping outputs in user forward.

asenina:
Partially fix 55333


nncf/torch/model_creation.py
```diff
@@ -138,7 +147,8 @@ def create_compressed_model(model: Module, config: NNCFConfig,
                                                                                 with_input_tracing=False,
                                                                                 with_output_tracing=False))
             else:
-                compressed_graph_builder = GraphBuilder(custom_forward_fn=dummy_forward_fn)
+                compressed_graph_builder = GraphBuilder(custom_forward_fn=
+                                                        create_wrapped_user_dummy_forward(dummy_forward_fn))
```
asenina, line 151: We discussed with @aleksei-kashapov  and decided to fix the bug with double compressed graph output in the way proposed in #748 . The custom `dummy_forward_fn`  wrapping here will be removed.

asenina:
@vshampor please merge

vshampor (approved):

PR #758, opened 2021-06-07
Fix link

andrey-churkin:


daniil-lyakhov:
But old link works too, how it possible? Old one: https://storage.openvinotoolkit.org/repositories/nncf/tensorflow/models/develop/tf1_mobilenet_v2_1.0_224_s0.35.tar.gz

PR #760, opened 2021-06-09
Update thresholds and targets for e2e

vshampor:
Resulting targets are still within 99% of FP32.


vshampor:
Jenkins please retry a build

PR #761, opened 2021-06-09
target_sparsity_level and target_pruning_level statistics were added

andrey-churkin:


daniil-lyakhov (approved):

nncf/common/sparsity/statistics.py
```diff
@@ -148,19 +159,20 @@ class RBSparsityStatistics(Statistics):
...
+                 target_sparsity_level: float,
                  mean_sparse_prob: float):
         """
         Initializes statistics of the RB-sparsity algorithm.
 
         :param model_statistics: Statistics of the sparsified model.
         :param masks_consistency: TODO
-        :param target_level: TODO
+        :param target_sparsity_level: Sparsity level which was applied
+            by the algorithm.
```
vshampor, line 170: Wording must be changed. A target level is a level which, in general, has not yet been completely applied by the algo; the "sparsity level which was applied by the algorithm" is a *current* level. 

Suggest: "A target level of sparsity for the algorithm to ultimately achieve". Same in other places.

vshampor (changes requested):

nncf/common/sparsity/statistics.py
andrey-churkin: I suggest a different formulation:
"A target level of the sparsity for the algorithm for the current epoch."

andrey-churkin: @alexsu52 @vshampor What do you think?

vshampor: Ok, it can be worded so.

andrey-churkin: I fixed it.

vshampor (approved):

PR #762, opened 2021-06-09
Introduce pattern fusing algo based on graphs

aleksei-kashapov:
- [x]  Add tests for reading custom patterns in configs (Deprecated: decided to remove this option from config)
- [x]   Compare the working time of the current pattern fusing algorithm and the PR version on Mask R-CNN, BERT, Densenet161


nncf/torch/graph/patterns.py
```diff
@@ -11,28 +11,105 @@
...
 
-RELU = N(VersionAgnosticNames.RELU) | N('hardtanh')
+NODE_COUNTER = 0
 
-BN = N('batch_norm') | N('batch_norm3d')
 
-POOLING = N('adaptive_avg_pool2d') | N('adaptive_avg_pool3d') | N('avg_pool2d') | N('avg_pool3d')
+class GraphPattern:
+    def __init__(self, pattern_name, types):
+        global NODE_COUNTER
```
vshampor, line 25: There is a thing in Python called "class attributes" (a.k.a. "static fields"), use it.

nncf/torch/graph/patterns.py
```diff
@@ -11,28 +11,105 @@
...
+        weakly_self_subgraphs = self.get_weakly_connected_subgraphs()
+        weakly_other_subgraphs = other.get_weakly_connected_subgraphs()
+        res_graph = nx.DiGraph()
+        for self_subgraph in weakly_self_subgraphs:
+            for other_subgraph in weakly_other_subgraphs:
+                # Create a copy of subgraph and add to a graph
+                subgraph_copy = _create_copy_of_subgraph(self_subgraph)
+                other_subgraph_copy = _create_copy_of_subgraph(other_subgraph)
+                subgraph_copy = _merge_one_subgraph_to_other(subgraph_copy, other_subgraph_copy)
+                _add_subgraph_to_graph(res_graph, subgraph_copy)
```
vshampor, line 73: A detailed explanation of how and why this works should be added to the final version of the function.

nncf/torch/graph/patterns.py
```diff
@@ -11,28 +11,105 @@
...
+import networkx.algorithms.isomorphism as ism
 
-RELU = N(VersionAgnosticNames.RELU) | N('hardtanh')
+NODE_COUNTER = 0
 
-BN = N('batch_norm') | N('batch_norm3d')
 
-POOLING = N('adaptive_avg_pool2d') | N('adaptive_avg_pool3d') | N('avg_pool2d') | N('avg_pool3d')
+class GraphPattern:
+    def __init__(self, pattern_name, types):
```
vshampor, line 24: Unnecessary to store `pattern_name` in the pattern object.

nncf/torch/graph/patterns.py
```diff
@@ -11,28 +11,105 @@
...
+import networkx as nx
+import networkx.algorithms.isomorphism as ism
 
-RELU = N(VersionAgnosticNames.RELU) | N('hardtanh')
+NODE_COUNTER = 0
 
-BN = N('batch_norm') | N('batch_norm3d')
 
-POOLING = N('adaptive_avg_pool2d') | N('adaptive_avg_pool3d') | N('avg_pool2d') | N('avg_pool3d')
+class GraphPattern:
```
vshampor, line 23: We expect from the class the following:
1) that simple patterns can be tersely specified in code; complex patterns - perhaps more verbosely, but still straightforwardly
2) that it defines a certain syntax for specifying a custom pattern in config and can parse the corresponding string representations of the pattern

nncf/torch/graph/patterns.py
aleksei-kashapov: Done

nncf/torch/graph/patterns.py
aleksei-kashapov: Done

nncf/torch/graph/patterns.py
aleksei-kashapov: Thx! Done

nncf/torch/graph/patterns.py
aleksei-kashapov: I introduce just three operations for patterns.
Let's say that A and B are two directed graphs.

 A: (a) (b)
 B: (c) (d)

1) A + B - for all weakly connected components will add nodes from B to A and append an edge from the last node at A to the first node at B
```
                      (a)    (a)     (b)    (b)
 A + B ---->           |      |       |      |
                      (c)    (d)     (c)    (d)
```
2) A | B - add graph B to graph A as a weakly subgraph

` A  |  B ----> (a) (b) (c) (d)`

3) A * B - connect all nodes of weakly subgraphs of A with all nodes of graph B
   
```
                     ---(a)---        ---(b)---
 A * B ---->         |       |        |       |
                    (c)    (d)       (c)     (d)

```


@vshampor Could we add the same syntax to our config file?

vshampor: @aleksei-kashapov this syntax is non-intuitive - even the existing syntax is better than this. It won't be accepted unless you describe some common cases in pattern specification where they become useful. Do you find the dot file-like syntax unsuitable for some reason? It seems to me that we would have to support both the dot-like syntax for specifying basic building blocks and compelx cases, and the existing + and | syntax for grouping patterns into semantic groups and for concatenating single-entry-single-exit blocks.

aleksei-kashapov: > @aleksei-kashapov this syntax is non-intuitive - even the existing syntax is better than this. It won't be accepted unless you describe some common cases in pattern specification where they become useful. Do you find the dot file-like syntax unsuitable for some reason? It seems to me that we would have to support both the dot-like syntax for specifying basic building blocks and compelx cases, and the existing + and | syntax for grouping patterns into semantic groups and for concatenating single-entry-single-exit blocks.

You can take a look at the implementation of Linear ops + Swish pattern in code:

```
FULL_PATTERN_GRAPH = LINEAR_OPS + ANY_BN_ACT_COMBO | ANY_BN_ACT_COMBO | \
                     ARITHMETIC + ANY_BN_ACT_COMBO | LINEAR_OPS_SWISH_ACTIVATION
```

aleksei-kashapov: dot-like syntax looks bulky. I think that it would be sufficient for all cases just use this one

aleksei-kashapov:
Jenkins please retry a build

aleksei-kashapov:
Jenkins please retry a build

tests/torch/graph/test_pattern_parse_from_config.py
```diff
@@ -0,0 +1,49 @@
...
+    graph = create_graph_pattern_from_pattern_view(pattern_view).graph
+    for node in graph.nodes:
+        graph.nodes[node]["type"] = graph.nodes[node]["type"][0]
+    return graph
+
+
+TEST_PATTERNS_CONFIG = [
+    [
+        [
+            "1 type=[A, B]",
```
ljaljushkin, line 19: Does it mean that multiple types are possible for the first node?
Would it be possible to use this syntax in config within `quantize_patterns` parameter?

aleksei-kashapov: I decided to remove this format until further discussion. If you think that it would be convenient to have it, please report me. Thanks!

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+                subgraph_copy = create_copy_of_subgraph(self_subgraph)
+                other_subgraph_copy = create_copy_of_subgraph(other_subgraph)
+                subgraph_copy = _add_second_subgraph_to_first_with_connected_edge(subgraph_copy, other_subgraph_copy)
+                add_subgraph_to_graph(final_graph, subgraph_copy)
+
+        final_pattern = copy.deepcopy(self)
+        final_pattern._graph = final_graph
+        return final_pattern
+
+    def __or__(self, other):
```
vshampor, line 114: Type hints

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Add other's DiGraph nodes to self's DiGraph.
+        Other's DiGraph will be placed as a new weakly connected components
+        """
+        new_pattern = copy.deepcopy(self)
+        other_copy = create_copy_of_subgraph(other.graph)
+        add_subgraph_to_graph(new_pattern._graph, other_copy)
+        return new_pattern
+
+    def __eq__(self, other):
```
vshampor, line 124: Type hints

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        return ism.is_isomorphic(self.graph, other.graph)
+
+    @property
+    def graph(self):
```
vshampor, line 128: Type hints

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._graph
+
+    def add_pattern_alternative(self, other: 'GraphPattern') -> None:
```
vshampor, line 131: Docstring

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        other_graph = other.graph
+        self_graph = self.graph
+        for node in other_graph.nodes:
+            if node in self_graph.nodes:
+                assert False
+            self_graph.add_node(node, type=other_graph.nodes[node]['type'])
+        for edge in other_graph.edges:
+            self_graph.add_edge(edge[0], edge[1])
+
+    def join_patterns(self, other: 'GraphPattern', edges: Optional[List[Tuple[Hashable, Hashable]]] = None) -> None:
```
vshampor, line 141: Docstring

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        return GraphPattern.NODE_COUNTER - 1
+
+    def add_edge(self, u_name, v_name) -> None:
+        self.graph.add_edge(u_name, v_name)
+
+    def get_weakly_connected_subgraphs(self) -> List[nx.DiGraph]:
+        return [self.graph.subgraph(c) for c in nx.weakly_connected_components(self.graph)]
+
+    def dump_graph(self, path: str) -> None:
+        nx.drawing.nx_pydot.write_dot(self.graph, path)
```
vshampor, line 167: Also add a function to better visualize the graph, so that the visualized graph has the attributes to be matched and their values visible.

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,131 @@
...
+
+def create_linear_h_swish_act() -> GraphPattern:
+    main_pattern = GraphPattern()
+
+    pattern = GraphPattern()
+    linear_ops_node = pattern.add_node(LINEAR_OPERATIONS)
+    bn_node = pattern.add_node(BATCH_NORMALIZATION_OPERATIONS)
+    __add__node = pattern.add_node(['__add__'])
+    hardtanh_node = pattern.add_node(['hardtanh'])
+    __truediv__node = pattern.add_node(['__truediv__'])
```
vshampor, line 61: Avoid double underscore in regular variable names.

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,131 @@
...
+
+    pattern.add_edge(linear_ops_node, __add__node)
+    pattern.add_edge(__add__node, hardtanh_node)
+    pattern.add_edge(hardtanh_node, __truediv__node)
+
+    return pattern
+
```
vshampor, line 103: Should separate the swish and h_swish into their own patterns, instead of always pairing them with linear or BN. We will need to fuse swish and h_swish even if they appear in the graph on their own.

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,131 @@
...
+
+@PATTERN_GRAPH.register('graph_pattern_factory')
```
vshampor, line 105: What is the purpose of the `PATTERN_GRAPH` registry? Does it store graph patterns? If so, why do you register a factory there?

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,131 @@
...
+    FULL_PATTERN_GRAPH = None
+
+    @staticmethod
+    def get_graph():
+        if GraphPatternFactory.FULL_PATTERN_GRAPH is None:
+            GraphPatternFactory.FULL_PATTERN_GRAPH = GraphPatternFactory._generate_full_pattern_graph()
+        return GraphPatternFactory.FULL_PATTERN_GRAPH
+
+    @staticmethod
+    def _generate_full_pattern_graph():
```
vshampor, line 116: Instead of creating a full pattern graph here, you should create subgraphs, register them, and then create a full pattern graph by a call on the registry object. For instance:

```python
ARITHMETIC = GraphPattern(ARITHMETIC_OPERATIONS)
GRAPH_PATTERNS.register(ARITHMETIC, name='ARITHMETIC', match=False)
...
GRAPH_PATTERNS.register(ARITHMETIC + ANY_BN_ACT_COMBO, name='ARITHMETIC + ANY_BN_ACT_COMBO', match=True)

full_pattern_graph = GRAPH_PATTERNS.get_full_pattern_graph()
GRAPH_PATTERNS.visualize_all(dot_file_dump_directory)
GRAPH_PATTERNS.visualize_matched(another_dot_file_dump_directory)
full_pattern_graph.visualize_graph(yet_another_dot_file_dump_directory)
```

nncf/torch/nncf_network.py
```diff
@@ -31,15 +29,14 @@
...
 from nncf.common.graph import NNCFGraph
 from nncf.common.graph import NNCFNode
-from nncf.common.graph import NNCFNodeExpression
 from nncf.common.graph import NNCFNodeName
-from nncf.common.graph.graph_matching import NodeExpression
 from nncf.common.graph.model_transformer import ModelTransformer
 from nncf.common.graph.transformations.commands import TargetType
 from nncf.common.graph.transformations.commands import TransformationPriority
-from nncf.common.hardware.config import HWConfig
+from nncf.torch.graph.patterns import GraphPattern
```
vshampor, line 36: The pattern matching should be made common since it will only operate on the nx graph underlying the common NNCFGraph representation.

nncf/torch/quantization/metrics.py
```diff
@@ -348,16 +350,13 @@ def _marking_edges(self, graph, node_key, queue, mark=True):
...
-
-        pattern = p.LINEAR_OPS + p.ANY_BN_ACT_COMBO | p.LINEAR_OPS + p.ELTWISE_UNIFORM_OPS
+        pattern = get_full_pattern_graph()
         # pylint: disable=protected-access
         matches = find_subgraphs_matching_expression(original_graph._nx_graph, pattern)
         merged_graph = deepcopy(original_graph._nx_graph)
         nx.set_node_attributes(merged_graph, False, self.IS_MERGED_GRAPH_ATTR)
         for match in matches:
-            if len(match) == 1:
+            if len(match) <= 1:
```
vshampor, line 359: Empty matches should not be returned at all

tests/torch/graph/test_graph_pattern.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_ops_combination_two_patterns():
+    first_type = ['a', 'b']
+    second_type = ['c', 'd']
+
+    first_pattern = GraphPattern(first_type)
+    second_pattern = GraphPattern(second_type)
+
+    ref_pattern = GraphPattern(first_type)
```
vshampor, line 14: The pattern matching tests should also be made common if possible.

tests/torch/test_compressed_graph.py
```diff
@@ -744,15 +744,17 @@ def test_output_quantization(_case_config):
...
 
     input_shape = [1, 3, 32, 32]
 
     config = get_basic_quantization_config(_case_config.quant_type, input_sample_sizes=input_shape)
 
     config["compression"].update({"quantize_outputs": False,
-                                  "quantizable_subgraph_patterns": [["sigmoid", "__mul__"],
-                                                                    ["__iadd__", "batch_norm"]]})
+                                  "quantize_patterns": [["1 sigmoid", "2 __mul__", "1 -> 2"],
+                                                        ["3 __iadd__", "4 batch_norm", "3 -> 4"]]})
```
vshampor, line 757: Better to name this `"fused_patterns"`. 
The node IDs should not be present if possible:

```
"quantize_patterns": [["sigmoid", "__mul__", "sigmoid -> __mul__"],
                                   ["__iadd__", "batch_norm", "__iadd__ -> batch_norm"]]
```
In this case you don't really need exact node IDs, just build each pattern separately. When building a pattern for such representation, if no match attributes are specified for the node, assume that the match should occur on the "type" attribute with a value equal to the node name. Document this as well.

Should also support the following syntax which has no extra lists, if possible:
```
"quantize_patterns": [
"sigmoid, __mul__, sigmoid -> __mul__",
"__iadd__, batch_norm, __iadd__ -> batch_norm"
]

**Review by vshampor — CHANGES_REQUESTED** [review]

File: nncf/common/graph/graph_matching.py [hunk 20]
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
+    patterns = []
+    for c in nx.weakly_connected_components(pattern_graph.graph):
+        patterns.append(pattern_graph.graph.subgraph(c))
+    # Get all patterns sorted by their lengths
+    # as we want match the longest patterns first
+    patterns = sorted(patterns, key=len, reverse=True)
+    for pattern in patterns:
+        matcher = ism.DiGraphMatcher(graph, pattern, node_match=are_nodes_matching)
+        for subgraph in matcher.subgraph_isomorphisms_iter():
+            subgraph = list(nx.topological_sort(graph.subgraph(subgraph)))
```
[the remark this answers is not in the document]
aleksei-kashapov: @vshampor why is this a case? Firstly, I am not sure that first in topological order nodes must have smaller node counter than the following ones. Secondly, in lexicographical order node with counter '111' < '12', which is incorrect. So, I suggest remaining as it is.

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
-                break
-        # If there is no pattern found, then skip this node
-        if longest_valid_match is None:
-            continue
-
-        for mn in longest_valid_match[0]:
-            subgraphs_nodes.add(mn)
+    subgraphs = []
+    visited_nodes = set()
+    patterns = []
```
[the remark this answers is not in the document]
aleksei-kashapov: Is this a typo or what typehints did you mention here?

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
+            if is_visited_node:
+                continue
+            if find_whether_subgraph_has_inner_outgoing_edges(graph, subgraph):
+                continue
+            visited_nodes.update(subgraph)
+            subgraphs.append(subgraph)
 
-        subgraphs.append(longest_valid_match[0])
-    return subgraphs
+    return subgraphs if subgraphs else [[]]
```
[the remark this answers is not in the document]
aleksei-kashapov: In that case, I will have to extend the output typehint. I prefer to stay only on one variant of the return value

nncf/torch/graph/patterns.py
aleksei-kashapov: I also was thinking about that. But in swish and h_swish we have a starting node from we calculate sigmoid and then __mul__ the result from sigmoid and the initial node. It doesn't seem solvable for me now

nncf/torch/graph/patterns.py
AlexKoff88: Support. No reason to have a registry for one entity. Can we split this into multiple graphs-patterns and iterate over them when we do the matching?

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,131 @@
...
+
+    pattern.add_edge(linear_ops_node, __add__node)
+    pattern.add_edge(__add__node, hardtanh_node)
+    pattern.add_edge(hardtanh_node, __truediv__node)
+
+    return pattern
+
+
+@PATTERN_GRAPH.register('graph_pattern_factory')
+class GraphPatternFactory:
```
AlexKoff88, line 106: It should be called appropriately, like IgnoredQuantizationPatterns.

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
vshampor: You may not leave it as it is because topological sorting does not give a unique result. If you need to implement special lexicographical portion of the sorting for your specific syntax of keys in the graph, use the corresponding interface of the `lexicographical_topological_sort` function.

nncf/torch/graph/patterns.py
AlexKoff88: An important note is that this functionality should be HW-dependant. For now, we can enable these patterns for any type of supported HW.

nncf/torch/graph/patterns.py
vshampor: Then you have to implement special "PatternInput" and "PatternOutput" nodes, and remove these from the composite graph pattern during the "+" or "join_patterns" operation.

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
vshampor: You won't have to extend the output typehint.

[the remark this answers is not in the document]
aleksei-kashapov: If I won't the output typehint will not be correct

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
aleksei-kashapov: thank you!

nncf/torch/quantization/metrics.py
aleksei-kashapov: As return value is [[]] we could have an empty match and this is the case why we need `<=`

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
vshampor: PyCharm does not raise any warnings for this function:
image

For this one, it does raise a warning:
image

It is OK for a `List`-hinted variable to be an empty list, at any level.

nncf/torch/quantization/metrics.py
vshampor: Using `[]` as return value will solve this.

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
+            predecessors = list(graph.pred[node_key].keys())
+            if any(successors_key not in subgraph for successors_key in successors):
+                return True
+            if any(predecessor not in subgraph for predecessor in predecessors):
                 return True
     return False
 
 
-def find_subgraphs_matching_expression(graph: nx.DiGraph, expression: Expression) -> List[List[str]]:
+def find_subgraphs_matching_expression(graph: nx.DiGraph, pattern_graph: GraphPattern) -> List[List[str]]:
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
-    nodes = nx.topological_sort(graph)
-    for node in nodes:
-        # If a node has already been added to any pattern skip this node
-        if node in subgraphs_nodes:
-            continue
 
-        all_matches = expression.all_matches([node], graph)
-        all_matches = sorted(all_matches, key=lambda x: len(x[0]), reverse=True)
+    def are_nodes_matching(node_1, node_2):
+        assert not isinstance(node_1['type'], list)
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
+        return node_1['type'] in node_2['type']
```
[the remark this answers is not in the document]
aleksei-kashapov: Add edge comparison function and extend current node comparison function. Please take a look

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
 
 
-def find_subgraphs_matching_expression(graph: nx.DiGraph, expression: Expression) -> List[List[str]]:
+def find_subgraphs_matching_expression(graph: nx.DiGraph, pattern_graph: GraphPattern) -> List[List[str]]:
     """
     Find a list of subgraphs for the particular graph that match the pattern expression.
     :param graph: The model graph.
-    :param expression: A pattern expression containing a logic of layer fusing.
+    :param pattern_graph: A graph consists of patterns for layer fusing logic.
     :return: A list of subgraphs for the particular graph, matching the pattern expression.
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/graph_matching.py
```diff
@@ -307,51 +45,57 @@ def find_whether_subgraph_has_inner_outgoing_edges(graph: nx.DiGraph, subgraph:
...
+    # as we want match the longest patterns first
+    patterns = sorted(patterns, key=len, reverse=True)
+    for pattern in patterns:
+        matcher = ism.DiGraphMatcher(graph, pattern, node_match=are_nodes_matching)
+        for subgraph in matcher.subgraph_isomorphisms_iter():
+            subgraph = list(nx.topological_sort(graph.subgraph(subgraph)))
+            is_visited_node = any(node in visited_nodes for node in subgraph)
+            if is_visited_node:
+                continue
+            if find_whether_subgraph_has_inner_outgoing_edges(graph, subgraph):
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/graph_matching.py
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def __add__(self, other):
+        """
+        Add DiGraph nodes of other to self and add edge between
+        last node of self's graph and first node of other's graph
+
+        The first and last nodes are found by nx.topological_sort()
+
+        Please, use this operand when the resulted pattern is quite
+        straightforward.
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+
+        def _add_second_subgraph_to_first_with_connected_edge(first_graph: nx.DiGraph,
+                                                              second_graph: nx.DiGraph) -> nx.DiGraph:
+            union_graph = nx.union(first_graph, second_graph)
+
+            first_graph_nodes = list(nx.topological_sort(first_graph))
+            last_node_first_graph = first_graph_nodes[-1]
+            assert first_graph.out_degree(last_node_first_graph) == 0
+            second_nodes = list(nx.topological_sort(second_graph))
```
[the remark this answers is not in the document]
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/torch/graph/patterns.py
aleksei-kashapov: Changed names

nncf/torch/quantization/metrics.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: For better visualization, I added a field "label", which Graphviz uses as the name of the node. So, the dumped pattern graph now looks like this:
image

nncf/torch/graph/patterns.py
aleksei-kashapov: Thanks for the idea! I added a node with a special type = 'INPUT_NODE'. This node implements the logic that you described. "+" operation changes this node on the last one from the other graph. I will add the same approach to "join_patterns"

nncf/torch/nncf_network.py
aleksei-kashapov: Done

tests/torch/graph/test_graph_pattern.py
aleksei-kashapov: Done

nncf/torch/graph/patterns.py
aleksei-kashapov: Maybe you mean that **we can not nable these patterns for any type of supported HW.** ?

nncf/common/graph/patterns.py
vshampor: The exact values that are being matched are still not being visualized, add a sort of a legend for the capitalized identifiers.

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,199 @@
...
+            assert self.graph.out_degree(last_node_self) == 0
+            edges = [[last_node_self, first_node_other]]
+        for edge in edges:
+            assert edge[0] in self.graph.nodes
+            assert edge[1] in other.graph.nodes
+        add_subgraph_to_graph(self.graph, other.graph)
+        for edge in edges:
+            self.graph.add_edge(edge[0], edge[1])
+
+    def add_node(self, label: str, node_type: Union[str, List[str]]) -> int:
```
vshampor, line 182: Node type should go first, label should go second and be optional

nncf/torch/graph/patterns.py
vshampor: `INPUT_NODE` is not a good name since we also have input nodes and output nodes in the NNCFGraph, and this name could be mistaken as to be referring to the model inputs and outputs. Use `PATTERN_INPUT`.

nncf/common/graph/patterns.py
alexsu52: There is such an opportunity to support 2 options for dumping the graph. The first variant as is. The second variant, dump each pattern in a separate dot file with the relevant name of the file.

nncf/torch/graph/patterns.py
```diff
@@ -11,27 +11,94 @@
...
+    pattern.add_edge(truediv_node, mul_node)
+    return pattern
 
-ELTWISE_UNIFORM_OPS = BN | RELU | ACTIVATIONS
 
-MATMUL = N('matmul') | N('bmm')
+def create_h_sigmoid_act() -> GraphPattern:
+    pattern = GraphPattern('H_SIGMOID')
+    input_pattern_node = pattern.add_node('*INPUT_NODE*', GraphPattern.INPUT_NODE_TYPE)
+    add_node = pattern.add_node('ADD', '__add__')
```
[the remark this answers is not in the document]
aleksei-kashapov: It could be done by extending `add_node()` by adding new attributes.

nncf/torch/graph/patterns.py
AlexKoff88: I mean the set of patterns can be HW specific. Some of them can be shared, some can be unique.

nncf/torch/graph/patterns.py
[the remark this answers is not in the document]
aleksei-kashapov: I updated `add_node()`. Now it takes an arbitrary dict and uses it as a node attributes

nncf/common/graph/patterns.py
aleksei-kashapov: Update to more arbitary variant -> `add_node(self, **attrs) -> int:`

tests/torch/test_compressed_graph.py
aleksei-kashapov: Depricated this config file option

nncf/torch/graph/patterns.py
aleksei-kashapov: Done, please, take a look

nncf/common/graph/patterns.py
aleksei-kashapov: I introduce a new class `QuantizationIgnorePatterns`, which stores all registered patterns. This class has 3 functions for visualizations:
1) Dump each pattern in a dir separately
2) Dump all matching patterns in a one file
3) Dump a specific pattern

nncf/torch/graph/patterns.py
aleksei-kashapov: Updated

nncf/torch/graph/patterns.py
aleksei-kashapov: Done; Implement QuantizationIgnorePatterns that stores all registered patterns

nncf/torch/graph/patterns.py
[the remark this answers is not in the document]
aleksei-kashapov: `add_node(self, **attrs)`

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    PATTERN_INPUT_NODE_TYPE = 'INPUT_NODE'
+
+    def __init__(self):
+        """
+        :param types: List or single string of backend operations names
+         that should be considered as one single node
+        """
+        self._graph = nx.DiGraph()
+        self.node_counter = 0
```
vshampor, line 87: Make protected.

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,235 @@
...
+
+import os
+import copy
+import itertools as it
+
+import networkx as nx
+import networkx.algorithms.isomorphism as ism
+
+
+class QuantizationIgnorePatterns:
```
vshampor, line 26: Rename to `HWFusedPatterns`

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,235 @@
...
+        if name in self._patterns_dict:
+            raise KeyError('{} is already registered'.format(name))
+        self._patterns_dict[name] = pattern
+        if match:
+            self.full_pattern_graph.add_pattern_alternative(pattern)
+
+    def get_full_pattern_graph(self) -> 'GraphPattern':
+        return self.full_pattern_graph
+
+    def visualize_all_matching_pattern(self, path: str) -> None:
```
vshampor, line 41: `visualize_full_pattern_graph`.

nncf/common/graph/patterns.py
```diff
@@ -0,0 +1,235 @@
...
+        self.full_pattern_graph = GraphPattern()
+
+    def register(self, pattern: 'GraphPattern', name: str, match: bool = True) -> None:
+        if name in self._patterns_dict:
+            raise KeyError('{} is already registered'.format(name))
+        self._patterns_dict[name] = pattern
+        if match:
+            self.full_pattern_graph.add_pattern_alternative(pattern)
+
+    def get_full_pattern_graph(self) -> 'GraphPattern':
```
vshampor, line 38: `get_full_matched_pattern_graph`

vshampor (changes requested):

nncf/torch/graph/patterns.py
AlexKoff88: > `INPUT_NODE` is not a good name since we also have input nodes and output nodes in the NNCFGraph, and this name could be mistaken as to be referring to the model inputs and outputs. Use `PATTERN_INPUT`.

@vshampor, if you introduce `GraphPattern.PATTERN_INPUT_NODE_TYPE` then `PATTERN_` is definitely a noise here which just makes the names loner.

FyodorK:
Jenkins please retry a build

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

nncf/common/graph/patterns.py
aleksei-kashapov: Done

aleksei-kashapov:
Jenkins please retry a build

aleksei-kashapov:
Jenkins please retry a build

nncf/torch/graph/pattern_operations.py
```diff
@@ -0,0 +1,49 @@
...
+# Copyright 2020-2021 Intel Corporation.
+#
+# LEGAL NOTICE: Your use of this software and any required dependent software
+# (the "Software Package") is subject to the terms and conditions of
+# the Intel(R) OpenVINO(TM) Distribution License for the Software Package,
+# which may also include notices, disclaimers, or license terms for
+# third party or open source software included in or with the Software Package,
+# and your use indicates your acceptance of all such terms. Please refer
+# to the "third-party-programs.txt" or other similarly-named text file
+# included with the Software Package for additional details.
```
alexsu52, line 11: @vshampor, @MaximProshin Is it correct license?

vshampor: Most likely not. @aleksei-kashapov use the Apache license as is being done in the majority of the other files.

aleksei-kashapov:
Measurements of algorithm time

| Model| Current algo (sec)| Proposed algo (sec)|
| ------------- | ------------- | ------------- |
| Mask R-CNN (Resnet-50) | 0.0588  | 0.68  |
| bert-large-uncased | 0.09  | 1.273509979  |

Model preparing time (NNCF wrapping model time) :
| Model| Current Model preparing time(sec)| Proposed Model preparing time(sec)|
| ------------- | ------------- | ------------- |
| Mask R-CNN (Resnet-50) | 129  | 130  |
| bert-large-uncased | 146  | 138 |

So in all cases, pattern matching algo takes less than 1% of all time for model preparing

nncf/torch/graph/pattern_operations.py
aleksei-kashapov: Updated

aleksei-kashapov:
@vshampor 
Finally, I suppose that this PR is ready to merge.

nncf/torch/graph/patterns.py
vshampor: Will be extended in a future PR, I think

vshampor (approved):

AlexKoff88 (approved):

alexsu52 (approved):

PR #765, opened 2021-06-10
Added execution model converting in venv

pfinashx:
Now Model Optimizer will be installed in separate venv

### DIFF
--- a/tests/tensorflow/test_sota_checkpoints.py
+++ b/tests/tensorflow/test_sota_checkpoints.py
@@ -25,7 +25,6 @@ import pytest
 import subprocess
 import re
 import shlex
-import sysconfig
 import yaml
 from prettytable import PrettyTable
 from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


pfinashx:
Please see E2E test build 151 vs regular build 152.
Also de-blocked execution Mask_rcnn models, but we need AC config refactor (maybe @daniil-lyakhov can take this).

tests/tensorflow/data/ac_configs/mask_rcnn_baseline.yml
```diff
@@ -0,0 +1,20 @@
...
+      - framework: dlsdk
+        device: CPU
+        adapter:
+          type: mask_rcnn
+          detection_out: reshape_do_2d
+          raw_masks_out: masks
+        inputs:
+          - name: image_info
+            type: CONST_INPUT
+            value: [[800, 1365, 1]]
```
alexsu52, line 14: @pfinashx, as I know OV does not support MaskRCNN. Do you really need an accuracy config for this model? Also, The AC config is incorrect, at least the input resolution.

pfinashx: As I know, support MaskRCNN is planned, so if we take last OV master, maybe it will support these models soon. Now on develop we don't even try to execute it in tests.
I know, that config is wrong, but I did not have time to debug it.
If you have information, that MaskRCNN will not be support, I will delete changes and config.

pfinashx: After discussion with @eaidova was found out, that implementation TF mask_rcnn models will last for a long time, so config deleted, skip this tests returned.

alexsu52: I guess OV will support it soon. Currently, I see no reason to add the incorrect AC config to the repository. I suggest adding AC config after MO converts the model to test, if it is possible. Will converting model to IR run without AC config?

pfinashx: Yes, converting will run, and IR will be collected in artifacts, if converting ends successfully.

alexsu52 (approved):

vshampor (approved):

PR #781, opened 2021-06-16
 Patch quantizer order in AutoQ RL agent traversal

vuiseng9:


nncf/torch/automl/environment/quantization_env.py
```diff
@@ -281,8 +287,21 @@ def _create_quantizer_table(self) -> pd.DataFrame:
         assert len(set(self.qconfig_space_map.keys()) - set(adjq_gid_map.keys())) == 0, \
             "both qconfig_space_map and adjq_gid_map must have exact keys."
 
+        # AutoQ requires quantizers in execution order
```
vshampor, line 290: Does it really "require" quantizers to be handled in a specific order, or is it merely that the reference graphs up to now have been built using the execution order of quantizers?

vuiseng9: @vshampor yes, it is by design to conform the Markov property of RL. It could possibly work without the order but given that we have verified the convergence in the current design, we should preserve the dynamics. Does it make sense?

nncf/torch/automl/environment/quantization_env.py
```diff
@@ -281,8 +287,21 @@ def _create_quantizer_table(self) -> pd.DataFrame:
...
+                    get_hook((qid, qmod), quantizers_in_exec_order)
+                )
+            )
+        self.qmodel.do_dummy_forward(force_eval=True)
+        for h in hooklist:
+            h.remove()
+
         d = OrderedDict()
-        for qid in self.qctrl.all_quantizations:
+        for qid, qmod in quantizers_in_exec_order.items():
```
ljaljushkin, line 304: do you really need qmod? 
just wonder why do you need OrderedDict, instead of simple List of qid?

vuiseng9: we dont. I've patched it :)

ljaljushkin (approved):

nncf/torch/automl/environment/quantization_env.py
vshampor: Please add this information to the comment and I'll merge this.

vuiseng9: @vshampor I've rebased and added the comment.

PR #783, opened 2021-06-16
Add target compression idx to layer attributes

vshampor:
In particular, this fixes the issue with ConvTranspose2d's being
per-channel quantized on a wrong dimension.

### DIFF
--- a/nncf/common/graph/layer_attributes.py
+++ b/nncf/common/graph/layer_attributes.py
@@ -45,8 +45,12 @@ class WeightedLayerAttributes(BaseLayerAttributes):
     def get_weight_shape(self) -> List[int]:
         pass
 
-    @abstractmethod
     def get_num_filters(self) -> int:
+        weight_shape = self.get_weight_shape()
+        return weight_shape[self.get_target_dim_for_compression()]
+
+    @abstractmethod
+    def get_target_dim_for_compression(self) -> int:
         pass
 
 
@@ -64,8 +68,8 @@ class GenericWeightedLayerAttributes(WeightedLayerAttributes):
     def get_weight_shape(self) -> List[int]:
         return self.weight_shape
 
-    def get_num_filters(self) -> int:
-        return self.weight_shape[0]
+    def get_target_dim_for_compression(self) -> int:
+        return 0
 
 
 class LinearLayerAttributes(WeightedLayerAttributes):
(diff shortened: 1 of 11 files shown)


vshampor:
Jenkins please retry a build

nncf/common/graph/layer_attributes.py
```diff
@@ -123,8 +127,11 @@ def get_weight_shape(self) -> List[int]:
             return [self.out_channels, self.in_channels // self.groups, *self.kernel_size]
         return [self.in_channels, self.out_channels // self.groups, *self.kernel_size]
 
-    def get_num_filters(self) -> int:
-        return self.out_channels
+    def get_target_dim_for_compression(self) -> int:
+        # Always quantize per each "out" channel
+        if self.transpose:
+            return 1
+        return 0
```
alexsu52, line 134: It looks like this is not applicable for TF models.

alexsu52 (changes requested):

vshampor:
E2E build 402 with updated checkpoints to quantize 1-st dim of ConvTranspose weights shows good accuracy.

nncf/common/graph/layer_attributes.py
vshampor: @alexsu52 how do you propose to resolve the fact that some common algos require exact definitions for what each weight tensor dimension means, and that at the same time the layouts of weight tensors are different between the backends?

alexsu52: I guess, the backend should provide this information.

alexsu52 (changes requested):

nncf/common/graph/layer_attributes.py
vshampor: Yes, and the backend should do so at the backend level of understanding, not at the layer attribute level which is common.

vshampor: I will not perpetuate the `__nncf_backend__` in any further PRs, and because proper resolution of this requires a separate effort, we will resolve this in a proper way once we get to integrating the quantization algo code with TF. Right now I need this fix to get correct performance with NNCF-produced PT UNet models.

vshampor:
TF E2E build 164

vshampor:
TF E2E build 164 shows no degradation vs nightly related to these changes.

nncf/common/graph/layer_attributes.py
vshampor: Created ticked 58678 to track the upcoming tensor layout backend abstraction.

PR #788, opened 2021-06-18
Introduce backend

l-bat:


vshampor:
IMO we should be moving away from the `__backend__` global variable entirely and rely on polymorphism for backend-specific effects. Otherwise we won't be able to support the external code where both TF and PT entities are used.

alexsu52:
> 
> 
> IMO we should be moving away from the `__backend__` global variable entirely and rely on polymorphism for backend-specific effects. Otherwise we won't be able to support the external code where both TF and PT entities are used.


Thanks for your comment. It depends on how we will implement the helper functions. Let's discuss offline.

PR #791, opened 2021-06-22
Add test of correctness resuming state with saturation fix applied

aleksei-kashapov:


aleksei-kashapov:
After a discussion with @ljaljushkin, we decided to add this test in https://github.com/openvinotoolkit/nncf/pull/682.

PR #804, opened 2021-06-28
Remove graphviz dependency

vshampor:
Not actually necessary to use graphviz package - we render to .dot instead.


vshampor:
NNCF precommit failure is the consequence of sporadic failures that have been fixed in #806

PR #806, opened 2021-06-29
Fix test

aleksei-kashapov:


tests/common/graph/test_graph_matching.py
```diff
@@ -75,4 +75,4 @@ def test_two_matches():
     ref_graph.add_edge('5', '6')
 
     matches = find_subgraphs_matching_pattern(ref_graph, pattern)
-    assert matches in [[['1', '2'], ['5', '6']]]
+    assert matches in [[['5', '6'], ['1', '2']], [['1', '2'], ['5', '6']]]
```
andrey-churkin: Seems like a hotfix. I think the `find_subgraphs_matching_pattern()` method should be deterministic.

PR #811, opened 2021-06-30
Remove tensorboard dependency

vanyalzr:


nncf/torch/accuracy_aware_training/runner.py
```diff
@@ -65,8 +70,7 @@ def initialize_training_loop_fns(self, train_epoch_fn, validate_fn, configure_op
             else 'runs'
         self._log_dir = configure_accuracy_aware_paths(self._log_dir)
         self._checkpoint_save_dir = self._log_dir
-        self._tensorboard_writer = self._tensorboard_writer
-        if self._tensorboard_writer is None:
+        if self._tensorboard_writer is None and TENSORBOARD_AVAILABLE:
             self._tensorboard_writer = SummaryWriter(self._log_dir)
```
vshampor, line 74: Should we warn the user if the TB installation is not found?

PR #812, opened 2021-06-30
Added requirements for setuptools

l-bat:
setuptools<43.0.0 could not find suitable distribution for 'tensorflow==2.4.0'

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,22 +19,23 @@ import sysconfig
 import codecs
 import os
 import re
-import setuptools
 from setuptools import setup, find_packages
-from pkg_resources import parse_version
 
 here = os.path.abspath(os.path.dirname(__file__))
 
 with open("{}/README.md".format(here), "r") as fh:
     long_description = fh.read()
 
-setuptools_version = parse_version(setuptools.__version__).base_version
-if setuptools_version < '43.0.0':
-    raise RuntimeError(
-        "To properly install NNCF, please install setuptools>=43.0.0, "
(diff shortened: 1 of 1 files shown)


alexsu52:
Jenkins please retry a build

alexsu52 (approved):

setup.py
```diff
@@ -19,13 +19,23 @@
...
 with open("{}/README.md".format(here), "r") as fh:
     long_description = fh.read()
 
+setuptools_version = parse_version(setuptools.__version__).base_version
+if setuptools_version < '43.0.0':
+    raise RuntimeError(
+        "To properly install NNCF, please install setuptools>=43.0.0, "
+        "while current setuptools version is {curr}".format(
+        curr=setuptools.__version__
+    ))
```
vshampor, line 37: Can we limit this to only the cases that currently fail on Python 3.6? I.e. check that Python is < 3.7 and that the tensorflow installation is requested. Otherwise we will fail in more cases than we should.

vshampor (changes requested):

setup.py
vshampor: Or does it really only depend on setuptools and not on Python (e.g. if, for example, the 3.7 distribution includes newer setuptools by default)? Even then we probably still should only fail for TF installation requests.

l-bat: This is an issue unrelated to the python version. On python3.8 and `setuptools<43.0.0` I get the same error.

vshampor (approved):

PR #821, opened 2021-07-05
Fix manual precision configs

vshampor:


ljaljushkin:
How did it pass CI on develop?

PR #827, opened 2021-07-06
Refactor test_prepare_checkpoints

daniil-lyakhov:
* Required checkpoint path in `prepare_checkpoints.py`
* Refactored test to load real checkpoints from sanity test_model_train to properly test  `prepare_checkpoints.py`, which required checkpoint path. (It was impossible to make new small model to pass it into prepare checkpoint because prepare checkpoint can build only yolo, retinanet and maskrcnn models.) 

Local check shows new version of the test small time consuming (~30 sec ~= not refactored test running time).

### DIFF
--- a/examples/tensorflow/common/argparser.py
+++ b/examples/tensorflow/common/argparser.py
@@ -34,29 +34,30 @@ def get_common_argument_parser(**flags):
             type=str,
             help='Name of metrics collecting .json file'))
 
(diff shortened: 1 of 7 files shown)


daniil-lyakhov:
#682 is being considered as more important thus should be merged first
CC @ljaljushkin

ljaljushkin (approved):

alexsu52 (approved):

PR #828, opened 2021-07-06
To add --all install option to setup.

FyodorK:
For now setup.py has two option --tf and --torch
Both intended to install tensorflow and pytorch parts of the product separately
But EXTRAS_REQUIRE also has 'all' label that suppose to install both tensorflow and torch at the same time.
The option '--all' covers using of this label.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -106,6 +106,10 @@ if "--tf" in sys.argv:
     INSTALL_REQUIRES.extend(EXTRAS_REQUIRE["tf"])
     sys.argv.remove("--tf")
 
+if "--all" in sys.argv:
+    INSTALL_REQUIRES.extend(EXTRAS_REQUIRE["all"])
+    sys.argv.remove("--all")
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -106,6 +106,10 @@ def find_version(*file_paths):
     INSTALL_REQUIRES.extend(EXTRAS_REQUIRE["tf"])
     sys.argv.remove("--tf")
 
+if "--all" in sys.argv:
```
alexsu52, line 109: @FyodorK @vshampor Do you really need it? I mean we can use tf and torch options at the same time.

FyodorK: 1. this option is needed for bdist installation
2. In case p.1 is questionable -> why this option has implemented in EXTRAS_REQUIRE dict but still not used at all?

l-bat (approved):

alexsu52 (approved):

PR #829, opened 2021-07-06
Accuracy validation doesn't depend on batch size and number of GPU

ljaljushkin:
Previously, last data point in all DataLoaders were dropped to prevent issue with BN (#379)
But it's actual for training only, on eval it shouldn't happen.


ljaljushkin:
sota eval 414 - WIP

ljaljushkin:
> sota eval 414 - WIP

eval is OK, except of binary models, but there're fixed in 415

vshampor (approved):

PR #835, opened 2021-07-08
Determine default num_classes for cifar dataset

l-bat:


examples/tensorflow/classification/main.py
```diff
@@ -322,6 +322,11 @@ def export(config):
 def main(argv):
     parser = get_argument_parser()
     config = get_config_from_argv(argv, parser)
+
+    if 'cifar' in config.dataset:
+        assert config.dataset[5:].isdigit()
+        config.num_classes = int(config.dataset[5:])
+
```
l-bat, line 329: Or we can check it explicitly
```suggestion
    if  config.dataset == 'cifar100':
        config.num_classes = 100
    elif config.dataset == 'cifar10':
        config.num_classes = 10
```

alexsu52: I think you should change the default value of the num_classes regarding to dataset here https://github.com/openvinotoolkit/nncf/blob/0938491f35585f8b58dfcfb3175341b5d88eec01/examples/tensorflow/classification/main.py#L150 I suggest doing it in the separate function `def get_num_classes()` in the main.py

alexsu52 (changes requested):

examples/tensorflow/classification/main.py
```diff
@@ -101,6 +101,18 @@ def get_dataset_builders(config, num_devices, one_hot=True):
...
 
+def get_num_classes(dataset):
+    if 'imagenet' in dataset:
+        num_classes = 1000
+    elif dataset == 'cifar100':
+        num_classes = 100
+    elif dataset == 'cifar10':
+        num_classes = 10
+    else:
+        raise RuntimeError('Could not determine number of classes by dataset name')
```
alexsu52, line 112: Please return the default value = 1000 and info message how many class classification problem is solved.

alexsu52 (changes requested):

alexsu52 (approved):

PR #837, opened 2021-07-08
The docker file was added for NNCF TF

andrey-churkin:
- The `cpu` and `gpu` folders for NNCF PT were moved from `nncf/docker` to `nncf/docker/torch`
- The docker files were updated for NNCF PT
    - Proxy env variables were removed
- The docker file was added for NNCF TF
- README.md was updated
    - Refactoring.
    - The additional settings  `--shm-size=1g` and `--ulimit memlock=-1` were added (see [link](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/troubleshooting.html#sharing-data))

Related to 59465


docker/tensorflow/gpu/Dockerfile
```diff
@@ -0,0 +1,47 @@
+FROM nvidia/cuda:11.0.3-cudnn8-runtime-ubuntu20.04
```
andrey-churkin, line 1: What should I really use?
`nvidia/cuda:11.0.3-cudnn8-runtime-ubuntu20.04` or `nvidia/cuda:11.0.3-cudnn8-devel-ubuntu20.04`
For more information please see _Overview of Images_ section [here](https://hub.docker.com/r/nvidia/cuda). My opinion: `runtime` is enough.

vshampor (approved):

alexsu52 (approved):

docker/README.md
```diff
@@ -1,30 +1,45 @@
...
  ```
 
-*You should not use `--runtime=nvidia` if you want to use `--cpu-only` mode.* 
+You should not use `--runtime=nvidia` if you want to use `--cpu-only` mode.
 
-*Use `--shm-size` to increase the size of the shared memory directory.*
+Use `--shm-size` to increase the size of the shared memory directory.
 
-Now you have a working container and you can run examples.
+For NNCF TensorFlow use `--shm-size=1g` and `--ulimit memlock=-1` settings in the command line to docker run.
```
AlexanderDokuchaev, line 42: It can be added to default command line, it's not break torch container.

docker/README.md
```diff
@@ -1,30 +1,45 @@
...
+```
+sudo docker image build \
+--network=host \
+--build-arg http_proxy=http://example.com:80 \
+--build-arg https_proxy=http://example.com:81 \
+--build-arg ftp_proxy=http://example.com:80 \
+<PATH_TO_DIR_WITH_DOCKERFILE>
+```
+
+Use `--http_proxy` , `--https_proxy`, `--ftp_proxy`, `--network` to duplicate the network settings of your localhost into context build.
```
AlexanderDokuchaev, line 23: Better way to setup proxy https://docs.docker.com/network/proxy/#configure-the-docker-client.

docker/tensorflow/gpu/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+        && pip3 install setuptools
+
+ENV PATH /usr/local/nvidia/bin:/usr/local/cuda/bin:${PATH}
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib:/usr/local/nvidia/lib64
+
+# nvidia-container-runtime
+ENV NVIDIA_VISIBLE_DEVICES all
+ENV NVIDIA_DRIVER_CAPABILITIES compute,utility
+
+ARG https_proxy
```
AlexanderDokuchaev, line 39: http_proxy missed.

P.S. Use ENV for proxy is bad practice, you should rebuild image to change proxy setting.

MaximProshin (approved):

docker/README.md
andrey-churkin: Thanks for your comment! I agree with you. I will try this approach.

docker/README.md
andrey-churkin: Done

docker/README.md
andrey-churkin: It works.

docker/tensorflow/gpu/Dockerfile
andrey-churkin: Thanks for your comment! I fixed it.

AlexanderDokuchaev (approved):

PR #838, opened 2021-07-08
Updated nncf_config with command line arguments

l-bat:


examples/tensorflow/common/sample_config.py
```diff
@@ -106,6 +106,11 @@ def create_sample_config(args, parser) -> SampleConfig:
 
     sample_config = SampleConfig.from_json(args.config)
     sample_config.update_from_args(args, parser)
+
+    for key, val in sample_config.items():
+        if key in nncf_config:
```
alexsu52, line 111: If batch_size is not in nncf_config you should add it into the nncf_config too.

alexsu52 (changes requested):

vshampor:
Why would we want to mix the sample configuration and the NNCF configuration again on a same hierarchy level? I separated the `SampleConfig` from `NNCFConfig` for this exact reason. Better to make "nncf_config" a separate field in all sample .json files.

l-bat:
> Why would we want to mix the sample configuration and the NNCF configuration again on a same hierarchy level? I separated the `SampleConfig` from `NNCFConfig` for this exact reason. Better to make "nncf_config" a separate field in all sample .json files.

`SampleConfig` can not be used because it does not pass validation using json schema. We can dump `NNCFConfig` during training and load it for testing

alexsu52:
> 
> 
> > Why would we want to mix the sample configuration and the NNCF configuration again on a same hierarchy level? I separated the `SampleConfig` from `NNCFConfig` for this exact reason. Better to make "nncf_config" a separate field in all sample .json files.
> 
> `SampleConfig` can not be used because it does not pass validation using json schema. We can dump `NNCFConfig` during training and load it for testing

@

alexsu52 (approved):

vshampor (approved):

PR #855, opened 2021-07-16
Attempt to fix dependency label

ljaljushkin:


ljaljushkin:
Jenkins please stop all builds

vshampor:
The fix doesn't seem to be working. Reverted the original in https://github.com/openvinotoolkit/nncf/pull/856. @generalova-kate please reopen the PR once you are sure that this is going to work without breaking our precommits.

PR #860, opened 2021-07-19
Add support quantization preset

asenina:


asenina:
@alexsu52 @AlexKoff88 @vshampor   What do you think about report  a warning (or runtime error) if `target_device: VPU` and `preset `parameters are set in the config at the same time?

nncf/tensorflow/quantization/algorithm.py
```diff
@@ -263,6 +264,12 @@ def _parse_range_init_params(self) -> TFRangeInitParams:
     def _parse_group_params(self, quant_config: Dict, quantizer_group: QuantizerGroup) -> None:
         group_name = quantizer_group.value
         params_dict = quant_config.get(group_name, {})
+        q_mode = params_dict.get('mode')
+        if q_mode is not None and 'preset' in quant_config:
+            raise RuntimeError("Global constraints cannot be parsed. Only quantize mode or preset must be set.")
```
alexsu52, line 269: I would expect the user to be able to override some of the preset options via config and not get a RuntimeError. I mean, If we will have a more complex preset which will be had more parameters not only mode. Should we raise a runtime error for all parameters in the preset?

alexsu52 (commented):
> 
> 
> @alexsu52 @AlexKoff88 @vshampor What do you think about report a warning (or runtime error) if `target_device: VPU` and `preset `parameters are set in the config at the same time?

My opinion is that we should not raise a runtime error, but reports warning for the target devices which has only one available quantization option will help users understand that the different presets lead to one quantization configuration.

asenina:
please merge after #865

nncf/config/schema.py
```diff
@@ -450,6 +455,9 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
         "quantize_outputs": with_attributes(_BOOLEAN,
                                             description="Whether the model outputs should be additionally quantized.",
                                             default=False),
+        "preset": with_attributes(QUANTIZATION_PRESETS_SCHEMA,
+                                  description="The preset defines the quantization schema "
+                                              "for weights and activations."),
```
vshampor, line 460: Should describe which presets are available and which does what.

vshampor (changes requested):

nncf/tensorflow/quantization/algorithm.py
```diff
@@ -263,6 +264,9 @@ def _parse_range_init_params(self) -> TFRangeInitParams:
     def _parse_group_params(self, quant_config: Dict, quantizer_group: QuantizerGroup) -> None:
         group_name = quantizer_group.value
         params_dict = quant_config.get(group_name, {})
+        if not params_dict and self._target_device != 'VPU':
+            preset = QuantizationPreset.from_str(quant_config.get('preset', 'performance'))
+            params_dict['mode'] = preset.get_quantization_mode(quantizer_group)
```
alexsu52, line 269: It looks like the preset parameters are a higher priority on the explicitly determined parameters by the user. Just to me, it is not intuitive. Maybe we need to revise this logic or add it to the documentation.

nncf/tensorflow/quantization/algorithm.py
```diff
@@ -263,6 +264,9 @@ def _parse_range_init_params(self) -> TFRangeInitParams:
     def _parse_group_params(self, quant_config: Dict, quantizer_group: QuantizerGroup) -> None:
         group_name = quantizer_group.value
         params_dict = quant_config.get(group_name, {})
+        if not params_dict and self._target_device != 'VPU':
```
alexsu52, line 267: The presets are not working for the VPU target. What do you think about adding some logs to inform users or raise an error that the VPU target does not support presets, currently?

nncf/tensorflow/quantization/algorithm.py
vshampor: Agreed, the explicit parameters should either override preset settings, or there should be an exception thrown prohibiting overrides in case presets are used.

tests/torch/quantization/test_algo_quantization.py
```diff
@@ -582,3 +582,30 @@ def test_shared_layers_are_weight_quantized_only_once():
...
+        'expected_activations_q': SymmetricQuantizer
+    },
+    {
+        'preset': 'mixed',
+        'expected_weights_q': SymmetricQuantizer,
+        'expected_activations_q': AsymmetricQuantizer
+    }]
+
+@pytest.mark.parametrize('data', TEST_QUANTIZATION_PRESET_STRUCT)
+def test_quantization_preset(data):
```
vshampor, line 600: Tests should be extended for various target devices and for the situation when there is both a preset and a specification of quantization for a given group/scope (via `scope_overrides`)

vshampor (changes requested):

tests/torch/quantization/test_algo_quantization.py
asenina: Done.

nncf/tensorflow/quantization/algorithm.py
asenina: Done.

nncf/tensorflow/quantization/algorithm.py
asenina: Done.

nncf/tensorflow/quantization/algorithm.py
```diff
@@ -451,11 +463,12 @@ def _get_quantizable_weighted_layer_nodes(self, nncf_graph: NNCFGraph) -> List[Q
 
             assert issubclass(metatype, TFLayerWithWeightsMetatype)
             nodes_with_weights.append(node)
+        scope_overrides_dict = self._get_algo_specific_config_section().get('scope_overrides', {})
         weighted_node_and_qconf_lists = assign_qconfig_lists_to_modules(nodes_with_weights,
                                                                         self.DEFAULT_QCONFIG,
                                                                         self.global_quantizer_constraints[
                                                                             QuantizerGroup.WEIGHTS],
-                                                                        scope_overrides_dict=None,
+                                                                        scope_overrides_dict,
```
asenina, line 471: Just notice, it looks like `scope_overrides` didn`t work for weights before in TF

alexsu52 (approved):

vshampor (approved):

PR #861, opened 2021-07-19
Fixed weekly tests with old-format checkpoints (issue 60538)

ljaljushkin:


ljaljushkin:
@vshampor weekly tests has started to pass locally.
Haven't checked all of them, but there was the same issue with loading.

vshampor (approved):

PR #865, opened 2021-07-21
Fix bug all potential qconfig were reference to the same object

asenina:


tests/torch/quantization/test_algo_quantization.py
```diff
@@ -544,10 +544,13 @@ def test_quantize_outputs_with_scope_overrides():
     register_bn_adaptation_init_args(config)
     model, ctrl = create_compressed_model_and_algo_for_test(model, config)
     output_quantizers =\
-        [q for qid, q in ctrl.all_quantizations.items() if isinstance(qid, NonWeightQuantizerId)][:-1]
-    for q in output_quantizers:
-        assert q.num_bits == 4
-        assert isinstance(q, AsymmetricQuantizer)
+        [q for qid, q in ctrl.all_quantizations.items() if isinstance(qid, NonWeightQuantizerId)]
+    for q in output_quantizers[1:]:
```
vshampor, line 548: Were the test premises incorrect, then?

asenina: Yes, the test was also incorrect.

PR #870, opened 2021-07-27
Fix pruning for case with disconnected NNCFGraph

AlexanderDokuchaev:
If model used unregistered module (like DeformConv2d from mmcv.ops) NNCFgraph can be disconnected.
As result some node have no input or output edges. 

* Added calculation of output shape for node without output edge
* Added check for result of `get_previous_conv`
* identity_mask_propagation propagate `none` if no input mask


nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -253,12 +256,36 @@ def _init_pruned_modules_params(self):
             self.pruning_quotas[cluster.id] = np.floor(self._modules_out_channels[cluster.elements[0].node_name] \
                                                        * self.pruning_quota)
 
+    def _calculate_output_shape(self, graph: NNCFGraph, node: NNCFNode) -> Tuple[int, ...]:
```
AlexKoff88, line 259: Can we have a test for this case?

I think we will have more such cases in the future. Is it possible to introduce a logic that we check the connectivity of the NNCF graph and raise an exception (or maybe a warning) if there are problems?

nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -253,12 +256,36 @@ def _init_pruned_modules_params(self):
...
+        """
+        Calculates output shape of convolution layer by input edge.
+        :param graph: the model graph
```
ljaljushkin, line 262: to be aligned with CodeStyle guidelines.

```suggestion
        Calculates output shape of convolution layer by input edge.
        
        :param graph: the model graph
```

ljaljushkin (changes requested):

nncf/common/pruning/mask_propagation.py
```diff
@@ -77,6 +77,9 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph):
     Propagates input mask through nx_node.
     """
     input_masks = get_input_masks(node, graph)
+    if not input_masks:
+        # In case for disconnected NNCFGraph
+        input_masks = [None]
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Before if  `input_masks == []` nncf crashed with assert, assert len(input_masks) == 1. 
This is not affect all supported models, only models with disconnected graph.

nncf/torch/pruning/filter_pruning/algo.py
AlexanderDokuchaev: Yes, will add tests.

I think add warning before use _calculate_output_shape, with message that node have output edge. ​
Exception was before on line `out_edge = graph.get_output_edges(node)[0]` with out of index.

> Is it possible to introduce a logic that we check the connectivity

It's possible, but we need discussion (what logic in nncf it can affect? what we can do with disconnected graph? list of test cases). I will create issue about it.

nncf/common/pruning/mask_propagation.py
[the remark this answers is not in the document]
ljaljushkin: How did you know? Did you check it manually? Don't think we should do it in that way further: it's time consuming and not reliable. 
IMHO, there should be automatic tests for `ModelPruner` and `MaskPropagationAlgorithm`, checking propagation of masks and covering corner cases like this.

nncf/torch/pruning/filter_pruning/algo.py
AlexKoff88: Besides creating an issue, can you also think about the proposal?

nncf/common/pruning/mask_propagation.py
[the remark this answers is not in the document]
AlexanderDokuchaev: About tests, yes, i will add it.

> How did you know? Did you check it manually? Don't think we should do it in that way further: manual testing is time consuming and not reliable.

On current moment models with disconnected graph and empty input masks can not be pruned because nncf crash on `assert len(input_masks) == 1`.  So we have no supported pruned models that affect by this changes.

ljaljushkin (approved):

AlexKoff88 (approved):

AlexKoff88:
@evgeniya-egupova, do you have any comments here?

nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -253,12 +256,38 @@ def _init_pruned_modules_params(self):
...
+                out_edge = graph.get_output_edges(node)[0]
+                out_shape = out_edge.tensor_shape[2:]
+            else:
+                # For disconnected NNCFGraph when convolution layers have no output edge
+                out_shape = self._calculate_output_shape(graph, node)
+                nncf_logger.error("Node %s have no output edge in NNCFGraph", node.node_name)
+            self._modules_out_shapes[node.node_name] = out_shape
 
         for node in graph.get_nodes_by_types([v.op_func_name for v in NNCF_LINEAR_MODULES_DICT]):
             out_edge = graph.get_output_edges(node)[0]
```
evgeniya-egupova, line 293: What if linear won't have output edge?

evgeniya-egupova:
Also, can convolutions be missing an input edge? For example with unregistered activation module?

nncf/torch/pruning/filter_pruning/algo.py
AlexanderDokuchaev: Exception "out of index".
I'll add same logic for linear.

AlexanderDokuchaev:
> Also, can convolutions be missing an input edge? For example with unregistered activation module?

In this case we can do nothing.

nncf/torch/pruning/filter_pruning/algo.py
AlexanderDokuchaev: done

AlexanderDokuchaev:
In case of missing an input edge, we can not get information about W and H

evgeniya-egupova:
> In case of missing an input edge, we can not get information about W and H

Actually, we can store input/output shapes as node attributes. This has been a topic of discussion before, and @vshampor insisted that we keep shapes in edges and find ways to fix the graph if problems occur. However, we now have at least two new cases where some nodes are missing input/output edges. I think we need to discuss this again.

vshampor:
> > In case of missing an input edge, we can not get information about W and H
> 
> Actually, we can store input/output shapes as node attributes. This has been a topic of discussion before, and @vshampor insisted that we keep shapes in edges and find ways to fix the graph if problems occur. However, we now have at least two new cases where some nodes are missing input/output edges. I think we need to discuss this again.

All of this is fixed if we have a convention to add "constant", or "unknown" nodes into the NNCFGraph wherever we know input shape, but not the originating operation. This is in itself a workaround, however, and must be explicitly marked as such. This is better fixed by making sure all PT ops are registered and wrapped.

nncf/common/pruning/mask_propagation.py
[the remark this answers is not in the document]
mkaglins: Agreed with @ljaljushkin, all cases (include valid graphs) should be covered by tests.

mkaglins (approved):

PR #875, opened 2021-07-29
Knowledge distillation TF

SKholkin:


nncf/tensorflow/nncf_network.py
```diff
@@ -0,0 +1,37 @@
+
+import tensorflow as tf
+
+
+class NNCFNetwork(tf.keras.Model):
+    def __init__(self, target_model):
+        super().__init__(target_model)
```
dupeljan, line 7: According to [docs](https://www.tensorflow.org/api_docs/python/tf/keras/Model) tf.kearas.Model `__init__` has 3 parameters: `inputs`, `outputs` and `name`. It's not convenient for me what happened when target model is passed there

nncf/tensorflow/nncf_network.py
```diff
@@ -0,0 +1,37 @@
...
+    def get_config(self):
+        return self.nncf_wrapped_model.get_config()
+
+    def compile(self, *args, **kwargs):
+        super(NNCFNetwork, self).compile(*args, **kwargs)
+        self.nncf_wrapped_model.compile(*args, **kwargs)
+
+    def call(self, inputs, training=None, mask=None):
+        compressed_output = self.nncf_wrapped_model(inputs)
+        if self.kd_original_model is not None:
```
daniil-lyakhov, line 25: You could manage the train/eval graph by training parameter. For example
```suggestion
        if self.kd_original_model is not None and training:
```
Now during evaluation tf build graph consisting only of target model

daniil-lyakhov:
It's look like `tf.keras.Model` subclassing isn't possible in the way introduced here. As the solution I can suggest to use target model and teacher model composition with attribute proxying by `__get_attribute__` to target model

nncf/tensorflow/nncf_network.py
```diff
@@ -0,0 +1,37 @@
...
+            original_model_outputs = self.kd_original_model(inputs)
```
SKholkin, line 26: Set training=False

PR #891, opened 2021-08-09
Fix embedding sparsity

vshampor:


tests/torch/sparsity/test_common.py
```diff
@@ -384,3 +386,30 @@ def test_sparsity_statistics_add_module(algo):
...
+        }
+    }
+    sparsity_init = 0.5
+    config['compression'] = {'algorithm': algo, 'sparsity_init': sparsity_init}
+    nncf_config = NNCFConfig.from_dict(config)
+    model = EmbeddingOnlyModel()
+    model, compression_ctrl = create_compressed_model_and_algo_for_test(model, nncf_config)
+
+    # Should pass
+    stats = compression_ctrl.statistics()
```
AlexKoff88, line 415: Should we check the actual sparsity level in this test?

vshampor: @AlexKoff88 this is a regression test against an error where the sparsified module has no `bias` attribute, such as the `Embedding`. The failure during missing attribute access happens way before the logic to calculate the sparsity level is invoked. In terms of the regular weight sparsity level calculation, the `Embedding` is no different than `Conv2d` or `Linear` modules, which, I presume, are already tested.

PR #898, opened 2021-08-17
Revert fix for init data loaders

mkaglins:
This reverts commit cd8a0f487bc1494b9ec567adad626947892ac7bd (PR https://github.com/openvinotoolkit/nncf/pull/377), that fix problems with init loaders in multiprocessing-distributed mode.

@vshampor Please, review this revert and acknowledge that all necessary revert changes are made.
@lzrvch Please, take a look.

### DIFF
--- a/examples/torch/semantic_segmentation/main.py
+++ b/examples/torch/semantic_segmentation/main.py
@@ -10,6 +10,7 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+from copy import deepcopy
 
 import functools
 import os
@@ -206,7 +207,7 @@ def load_dataset(dataset, config):
             shuffle=train_shuffle)
     # Loaders
     train_loader = create_train_data_loader(batch_size)
-    init_loader = train_loader
+    init_loader = deepcopy(train_loader)
     if config.batch_size_init:
         init_loader = create_train_data_loader(config.batch_size_init)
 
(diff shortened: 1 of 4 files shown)


examples/torch/semantic_segmentation/main.py
```diff
@@ -207,12 +206,9 @@ def create_train_data_loader(batch_size_):
             shuffle=train_shuffle)
     # Loaders
     train_loader = create_train_data_loader(batch_size)
+    init_loader = train_loader
     if config.batch_size_init:
         init_loader = create_train_data_loader(config.batch_size_init)
-    else:
-        init_loader = deepcopy(train_loader)
-    if config.distributed:
-        init_loader.num_workers = 0  # PyTorch multiprocessing dataloader issue WA
```
ljaljushkin, line 215: AFAIK, there's the same fix for init data loaders in tests. Why not to revert it there as well? 
That might be a prove that fix can be safely reverted.

mkaglins: Good idea, reverted this.

mkaglins:
Bugs with distributed mode are still not fixed in Pytorch so this PR with revert of workarounds for them can't be merged.
The problem with long initialization (that should be fixed by this PR) was fixed in https://github.com/openvinotoolkit/nncf/pull/904
So, this PR is not needed and can be closed.

PR #904, opened 2021-08-24
Fix for num_workers=0 for init_dataloader

mkaglins:
Fix for LeGR slowdown in multiprocessing-distributed mode due to` num_workers=0` for init loader (that was set to fix tests fails due to bug in Pytorch). 
Currently, num workers for init loader in examples will be set as needed for distributed mode. And only in tests `num_workers=0` will be used by redefining functions with data loaders creation.

### DIFF
--- a/examples/torch/classification/main.py
+++ b/examples/torch/classification/main.py
@@ -10,7 +10,6 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
-import os
 import os.path as osp
 import sys
 import time
--- a/examples/torch/classification/main.py
+++ b/examples/torch/classification/main.py
@@ -189,6 +189,7 @@ def main_worker(current_gpu, config: SampleConfig):
             validate_fn=lambda *x: validate_model_fn(*x)[::2],
             autoq_eval_fn=lambda *x: validate_model_fn(*x)[1],
             val_loader=val_loader,
+            train_loader=train_loader,
             model_eval_fn=model_eval_fn,
             device=config.device,
             execution_parameters=execution_params,
--- a/examples/torch/classification/main.py
+++ b/examples/torch/classification/main.py
@@ -462,10 +462,6 @@ def create_data_loaders(config, train_dataset, val_dataset):
         init_loader = create_train_data_loader(config.batch_size_init)
     else:
         init_loader = deepcopy(train_loader)
-
-    if config.distributed and 'NNCF_ZERO_DISTRIBUTED_WORKERS' in os.environ:
-        init_loader.num_workers = 0  # PyTorch multiprocessing dataloader issue WA
-
     return train_loader, train_sampler, val_loader, init_loader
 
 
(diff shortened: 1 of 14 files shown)


FyodorK:
Jenkins please retry a build

mkaglins:
Jenkins please retry a build

AlexanderDokuchaev (approved):
I think that it's should be described in changelog @vshampor.

alexsu52 (commented):
What do you think about providing this as a parameter instead of the env variable? Will you document this WA?

mkaglins:
> What do you think about providing this as a parameter instead of the env variable? Will you document this WA?

1) Env variable was chosen as change with minimal impact to interface and code (+ potential reverting of this changes when it will not be needed anymore). 
It was discussed and chosen with @vshampor offline. Also, you can find his opinion in the comment above. 

2) As for the documentation for this WA as @vshampor mentioned above it is just an internal change. I can add comments about it to code where this env variable is used.

alexsu52:
> 
> 
> > What do you think about providing this as a parameter instead of the env variable? Will you document this WA?
> 
>     1. Env variable was chosen as change with minimal impact to interface and code (+ potential reverting of this changes when it will not be needed anymore).
>        It was discussed and chosen with @vshampor offline. Also, you can find his opinion in the comment above.
> 
>     2. As for the documentation for this WA as @vshampor mentioned above it is just an internal change. I can add comments about it to code where this env variable is used.

It looks like that it needs only for testing. Thus, I would suggest writing a test without adding any functionality to the examples which need only for testing. For example, you can decorate `create_data_loaders` to set `num_workers`.

mkaglins:
> > > What do you think about providing this as a parameter instead of the env variable? Will you document this WA?
> > 
> > 
> > ```
> > 1. Env variable was chosen as change with minimal impact to interface and code (+ potential reverting of this changes when it will not be needed anymore).
> >    It was discussed and chosen with @vshampor offline. Also, you can find his opinion in the comment above.
> > 
> > 2. As for the documentation for this WA as @vshampor mentioned above it is just an internal change. I can add comments about it to code where this env variable is used.
> > ```
> 
> It looks like that it needs only for testing. Thus, I would suggest writing a test without adding any functionality to the examples which need only for testing. For example, you can decorate `create_data_loaders` to set `num_workers`.

Sounds reasonable, I will implement such fix.

FyodorK:
Jenkins please retry a build

FyodorK:
Jenkins please retry a build

FyodorK:
Jenkins please retry a build

mkaglins:
@alexsu52 Please, check discussed changes.

alexsu52 (approved):

mkaglins:
@alexsu52 Successful run of `test_sota_checkpoints`: https://adas-ci-icv.inn.intel.com/job/NNCF/job/nightly/job/ubuntu20_eval/514/

PR #848, opened 2021-07-14
Turning off pruning export in Pytorch

mkaglins:

PR #905, opened 2021-08-24
Pruning+quantization export

evgeniya-egupova:


nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -668,66 +666,13 @@ def _apply_binary_mask_to_module_weight_and_bias(module: torch.nn.Module,
...
-        parameters_count_after = model.get_parameters_count_in_model()
-        flops_after = model.get_MACs_in_model()
-        new_pruned_layers_stats = self.get_stats_for_pruned_modules()
-        stats = self.create_stats_table_for_pruning_export(pruned_layers_stats, new_pruned_layers_stats)
-
-        nncf_logger.info(stats.draw())
-        nncf_logger.info('Final Model Pruning Rate = %.3f', 1 - parameters_count_after / parameters_count_before)
-        nncf_logger.info('Total MAC pruning level = %.3f', 1 - flops_after / flops)
+        stats = self.statistics()
+        nncf_logger.info(stats.to_str())
```
alexsu52, line 675: Could you please provide some screenshots of how a sample output will be changed after these changes?

nncf/torch/pruning/export_helpers.py
```diff
@@ -481,38 +481,3 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
 
         node.data['input_masks'] = input_masks
         node.data['output_mask'] = None
-
-
-class ModelPruner(MaskPropagationAlgorithm):
```
alexsu52, line 486: @evgeniya-egupova, please align with @mkaglins about https://github.com/openvinotoolkit/nncf/pull/848.

AlexanderDokuchaev (approved):

FyodorK:
Jenkins please retry a build

nncf/torch/pruning/filter_pruning/algo.py
alexsu52: Thanks for your update!

I would suggest not log any statistics in the export_model function, but I think it will be useful for users to log all statistics by model before export. It can be done via changing examples. What do you think?

alexsu52 (approved):

PR #906, opened 2021-08-25
Fix bug with installation test

a-ignatyev:
Installation test tried to import modules that had dependencies installing during installation.

### DIFF
--- a/tests/common/helpers.py
+++ b/tests/common/helpers.py
@@ -10,116 +10,13 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+import numpy as np
 
 from abc import ABC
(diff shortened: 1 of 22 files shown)


l-bat:
What do you think about renaming the file with 'BaseTensorListComparator' to `base_tensor_list_comparator.py` or something else to reduce the number of modified files? I think the name `helpers_without_dependencies.py` may confuse users.

l-bat:
Could you please provide numbers of torch install and tf install jobs?

l-bat (changes requested):
Please add `numpy` for test requirements for PT and TF

l-bat:
Closed because the bug is related to the CI pipeline

PR #912, opened 2021-08-30
added conda recipe

### LINKED ISSUE #844: Add conda recipe
Thanks for this project! Could you please add conda recipe, so that it can be used with projects that have more complicated requirements?

sarthakpati:
Fixes #844

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,11 @@ Install NNCF and TensorFlow in one line:
 pip install nncf[tf]
 ```
 
+NNCF is also available via [conda](https://anaconda.org/conda-forge/nncf):
+```
+conda install -c conda-forge nncf
+```
+
 #### As a Docker image
 Use one of the Dockerfiles in the [docker](./docker) directory to build an image with an environment already set up and ready for running NNCF [sample scripts](#model-compression-samples).
 


l-bat:
Thank you for contribution!
PyTorch is installed by default, but it is not a part of `INSTALL_REQUIRES`
https://github.com/openvinotoolkit/nncf/blob/29ad62c664c1dd53b3c8c50fc001a1b36bd1e8ac/setup.py#L54-L71

l-bat:
Have you tried using nncf with GPU support?

sarthakpati:
Unsure if it would be an issue, because PyTorch is listed as a requirement in the recipe: https://github.com/conda-forge/nncf-feedstock/blob/master/recipe/meta.yaml#L40

😄

alexsu52:
> Unsure if it would be an issue, because PyTorch is listed as a requirement in the recipe:

Thanks for your contribution!

NNCF since v2.0 supports PyTorch and TensorFlow backends. Thus, the nncf package is currently framework agnostic and requires the user to install PyTorch or TensorFlow. I think the recipe requirements should be updated to avoid confusing users. @sarthakpati What do you think?

sarthakpati:
I agree. But this can only be done after https://github.com/conda-forge/tensorflow-feedstock/pull/117 is merged. For now, can we keep this only for PyTorch and then update the recipe after the new TF is made available?

sarthakpati:
Great, this is now merged, and pytorch is no longer listed as a separate dependency.

PR #914, opened 2021-09-01
Fix bug in __getattr__ of NNCFNetwork

kshpv:
Add a capability to change the bound functions of a wrapped module by passing NNCFNetwork instance as a first argument

### DIFF
--- a/nncf/torch/nncf_network.py
+++ b/nncf/torch/nncf_network.py
@@ -338,31 +338,34 @@ class NNCFNetwork(nn.Module, PostGraphBuildActing):
             raise RuntimeError("Unsupported insertion type: {}".format(point.insertion_type))
 
     def __getattr__(self, name):
+        class NotFound:
+            ...
+
         def get_NNCFNetwork_attr(self, name):
             if name in self.__dict__:
                 return self.__dict__[name]
-            return None
+            return NotFound
 
         def get_nncf_module_attr(self, name):
             if hasattr(self.__dict__['_modules'][MODEL_WRAPPED_BY_NNCF_ATTR_NAME], name):
                 attr = getattr(self.__dict__['_modules'][MODEL_WRAPPED_BY_NNCF_ATTR_NAME], name)
                 if hasattr(attr, '__self__'):  # If it is a bound function
                     from functools import partial
-                    attr = partial(attr, self)
+                    attr = partial(attr.__func__, self)
                     return attr
                 else:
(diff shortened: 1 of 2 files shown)


vshampor (changes requested):
Use `pass` instead of `...`.

kshpv:
> Use `pass` instead of `...`.

Done

vshampor (approved):

nncf/torch/nncf_network.py
```diff
@@ -338,10 +338,36 @@ def insert_at_point(self, point: PTInsertionPoint, fn_list: List[Callable]):
...
 
     def __getattr__(self, name):
-        wrapped_module = super().__getattr__(MODEL_WRAPPED_BY_NNCF_ATTR_NAME)
-        if hasattr(wrapped_module, name):
-            return getattr(wrapped_module, name)
-        return super().__getattr__(name)
+        class NotFound:
+            pass
+
+        def get_NNCFNetwork_attr(self, name):
```
andrey-churkin, line 344:
```suggestion
        def get_nncf_network_attr(self, name):
```

nncf/torch/nncf_network.py
```diff
@@ -338,10 +338,36 @@ def insert_at_point(self, point: PTInsertionPoint, fn_list: List[Callable]):
             raise RuntimeError("Unsupported insertion type: {}".format(point.insertion_type))
 
     def __getattr__(self, name):
-        wrapped_module = super().__getattr__(MODEL_WRAPPED_BY_NNCF_ATTR_NAME)
-        if hasattr(wrapped_module, name):
-            return getattr(wrapped_module, name)
-        return super().__getattr__(name)
+        class NotFound:
```
andrey-churkin, line 341: Why we can not use `None`?

```python
def get_nncf_network_attr(self, name):
    return self.__dict__.get(name)


attr = get_nncf_network_attr(self, name)

if attr is not None:
    return attr
```

kshpv: If we would return None we can't distinguish cases when the attribute exists and its value is None and there is no such attribute.

nncf/torch/nncf_network.py
kshpv: Done

alexsu52 (approved):
Looks good to me

andrey-churkin (approved):

AlexanderDokuchaev (approved):

PR #919, opened 2021-09-03
Update some configs in TF.(mobilenetv3_small, mobilenetv3_large, yolo)

asenina:


alexsu52 (approved):
It looks like this is the first step in applying a preset in a configuration without retraining the model.

PR #921, opened 2021-09-07
Fix for AA tensor board logging

negvet:
- Missed callbacks are added
- Proper logging into TB

### DIFF
--- a/nncf/common/accuracy_aware_training/training_loop.py
+++ b/nncf/common/accuracy_aware_training/training_loop.py
@@ -243,8 +243,8 @@ class AdaptiveCompressionTrainingLoop(TrainingLoop):
                                                    self.runner.cumulative_epoch_count)
 
             self.runner.train_epoch(model, self.adaptive_controller)
-            self.runner.dump_statistics(model, self.adaptive_controller)
             compressed_model_accuracy = self.runner.validate(model)
+            self.runner.dump_statistics(model, self.adaptive_controller)
             self.runner.accuracy_bugdet = compressed_model_accuracy - self.runner.minimal_tolerable_accuracy
             self.runner.add_tensorboard_scalar('val/accuracy_aware/accuracy_bugdet', self.runner.accuracy_bugdet,
                                                self.runner.cumulative_epoch_count)
--- a/nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
+++ b/nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
@@ -10,11 +10,13 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+import copy
 
 from tensorflow.python.keras.engine import training_utils
 from tensorflow.python.profiler import trace
 from tensorflow.python.eager import context
 from tensorflow.python.keras import callbacks as callbacks_module
+from tensorflow.python.keras.callbacks import History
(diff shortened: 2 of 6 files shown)


nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
```diff
@@ -62,6 +63,14 @@ def train_epoch_fn(compression_ctrl, model, epoch):
                 steps=data_handler.inferred_steps
             )
 
+        # Check if callbacks.on_train_begin() has been called before the first epoch.
```
negvet, line 66: The only callback which calls at `on_train_begin()` is `History`. It creates an `epoch` attribute, which is used at `on_epoch_end()`. 

Probably, should be a better way to handle it. Do not know how to call  `on_train_begin()` from outside of `train_epoch_fn()` in current design.
@alexsu52 do you have some options in mind?

kshpv (approved):

nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
```diff
@@ -86,6 +94,11 @@ def train_epoch_fn(compression_ctrl, model, epoch):
                     if model.stop_training:
                         break
 
+        if logs is None:
+            raise ValueError('Expect x to be a non-empty array or dataset.')
+        epoch_logs = copy.copy(logs)
+        callbacks_container.on_epoch_end(epoch, epoch_logs)
+
     if validation_data is None:
         validation_data = train_dataset
```
alexsu52, line 104: Please also add callback calls for `validate_fn`.

alexsu52 (changes requested):

negvet:
Now callback object is getting created just once in `keras_model_utils.py`. Which helped to properly call `callbacks.on_train_begin()/callbacks.on_train_end()`. This behaviour replicates the `model.fit()`.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -243,8 +243,8 @@ def run(self, model, train_epoch_fn, validate_fn,
                                                    self.runner.cumulative_epoch_count)
 
             self.runner.train_epoch(model, self.adaptive_controller)
-            self.runner.dump_statistics(model, self.adaptive_controller)
             compressed_model_accuracy = self.runner.validate(model)
+            self.runner.dump_statistics(model, self.adaptive_controller)
             self.runner.accuracy_bugdet = compressed_model_accuracy - self.runner.minimal_tolerable_accuracy
```
negvet, line 248: `dump_statistics()` requires the latest validation results to save the checkpoint - that is why `dump_statistics()` should go after `validate()` I guess

kshpv: It doesn't matter here. Because validation metrics are dumped in `validate()`, which is possible is not the right solution. Anyway, I suggest don't make this change

alexsu52 (approved):
It looks good to me.

vshampor (approved):

nncf/common/accuracy_aware_training/training_loop.py
kshpv: After discussion, it looks that these changes are necessary and fix a bug

kshpv (approved):

PR #927, opened 2021-09-13
Export ONNX with opset 13

vshampor:


vshampor:
E2E build 503

nncf/torch/exporter.py
```diff
@@ -80,7 +80,7 @@ def _export_to_onnx(self, save_path: str) -> None:
                               input_names=self._input_names,
                               output_names=self._output_names,
                               enable_onnx_checker=False,
-                              opset_version=10,
+                              opset_version=13,
```
ljaljushkin, line 83: Is opset13 backward compatible with previous versions? 
What if user wants a specific version which is not compatible with the latest one? Shouldn't we have a parameter to customize it?

PR #930, opened 2021-09-14
Act on input copies in NNCF PT

vshampor:


ljaljushkin (changes requested):
What does it fix? Do we have a simple reproducer for it? Test would be helpful.

PR #932, opened 2021-09-15
Fix view with size

vshampor:


kshpv (approved):

nncf/torch/hardware/fused_patterns.py
```diff
@@ -37,6 +42,8 @@ def _get_torch_hw_fused_patterns() -> HWFusedPatterns:
 
     retval.register(linear_ops + batch_norm_activations_permutation, 'LINEAR + BN_ACT_PERM',
                     match=True)
+    retval.register(matmul_ops + arithmetic_ops, 'MATMUL + ARITHMETIC',
```
kshpv, line 45: Just one comment - for better readability I suggest changing the name `retval` to something more clear like `hw_fused_patterns` or `fused_patterns`

PR #937, opened 2021-09-17
Pylint2.10 update

Noktyrn:
Made fixes to resolve pylint 2.10 warnings.


FyodorK:
Jenkins please retry a build

nncf/common/utils/os.py
```diff
@@ -14,7 +14,7 @@
 from contextlib import contextmanager
 from pathlib import Path
 
-
+# pylint: disable=W1514
```
ljaljushkin, line 17: why not using encoding='utf8'?

examples/torch/object_detection/layers/modules/ssd_head.py
```diff
@@ -106,6 +106,7 @@ def __init__(self, outputs, modules):
         super().__init__(*modules)
         self.outputs = [str(o) for o in outputs]
 
+    #pylint:disable=W0237
```
ljaljushkin, line 109: Looks like this warning happens when a method parameter has a different name than in the implemented interface or in an overridden method.
Can we just simply rename the parameter?

nncf/torch/layers.py
```diff
@@ -96,6 +96,7 @@ def _custom_forward_fn(self, input_):
         proxy_weight = self.weight
         return self._conv_forward(input_, proxy_weight, proxy_padding_value)
 
+    # pylint: disable=W0237
```
ljaljushkin, line 99: I think the `_conv_forward` function shouldn't have the same name as in the original implementation of Conv2d.
The warning should be resolved by simply renaming this method to something different, e.g. `_conv_forward_proxy`

ljaljushkin (changes requested):

nncf/common/utils/os.py
Noktyrn: I thought if somebody wants to open a file in the raw format, which is possible since this function accepts a variable number of arguments, it can lead to an error, if we specify the encoding.

examples/torch/object_detection/layers/modules/ssd_head.py
Noktyrn: This parameter was originally named input and if rename it, then there will be the warning W0622: Redefining built-in 'input'

nncf/torch/layers.py
Noktyrn: Ok, i'll change the name of this function

examples/torch/object_detection/layers/modules/ssd_head.py
ljaljushkin: ok, let's keep the original W0237 warning

FyodorK:
Jenkins please retry a build

ljaljushkin (approved):

PR #938, opened 2021-09-20
Support dimensions expand  / squeeze or transpose for pruning algo 

daniil-lyakhov:
It turns out we can't prune SE block because Reshape op is placed after global avg pool.
image

 This reshape just expand dimensions to emulate FC layer by conv2d and thus should have attribute `accept_pruned_input == True`. This unblock proper mobilenet_v3 pruning for tf and make one step forward to unblock mobilenet_v3 for PyTorch.
 
 Besides one can implement SE block using nn.Linear like [here](https://github.com/openvinotoolkit/deep-object-reid/blob/ote/torchreid/models/mobilenetv3.py#L33-L50) thus linear layer should be in prunable layers to provide reasonable pruning level. 
TODO:
- [x] Support reshape for PyTorch
- [x] Support reshape for TF
- [ ] Add tests
- [x] Support Linear pruning for PyTorch
- [ ] Support Linear pruning for TF


nncf/common/graph/layer_attributes.py
```diff
@@ -159,3 +159,11 @@ def get_weight_shape(self) -> List[int]:
 
     def get_target_dim_for_compression(self) -> int:
         return 0
+
+
+class ReshapeLayerAttributes(BaseLayerAttributes):
+    def __init__(self,
+                 input_shape,
+                 output_shape):
```
vshampor, line 167: Type hints

vshampor (changes requested):

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -94,6 +95,14 @@ def wrapped(*args, **kwargs):
                     if isinstance(curr_module, _NNCFModuleMixin):
                         ignored_algos = deepcopy(curr_module.ignored_algorithms)
 
+                if op_name in ["reshape", "view", "flatten", "squeeze", "unsqueeze"]:
+                    input_tensor = args[0]
+                    output_shape = args[1:]
+
+                    layer_attrs = ReshapeLayerAttributes(input_tensor.shape,
```
daniil-lyakhov, line 102: @vshampor I believe there is more obvious way to do this. Could you please help me to improve this?

nncf/torch/pruning/export_helpers.py
```diff
@@ -467,9 +492,26 @@ def input_prune(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph):
                              ' {}.'.format(node.data['key'], old_num_clannels, new_num_channels))
 
 
+@PT_PRUNING_OPERATOR_METATYPES.register('linear')
```
AlexKoff88, line 495: I think you should implement similar logic for `def input_prune(...)` and `def output_prune(...)` to what that conv operation has.

Tests for Linear layer when it is in SE block and in the end of the model are required.

nncf/torch/dynamic_graph/wrappers.py
daniil-lyakhov: I'll move this section into dynamic_graph -> graph converting stage, as we discussed

nncf/torch/pruning/export_helpers.py
daniil-lyakhov: Do we planning to make some tests for  `ModelPruner`? Now we have no any tests for it

nncf/torch/pruning/export_helpers.py
```diff
@@ -488,13 +571,19 @@ def __init__(self, model: NNCFNetwork, graph: NNCFGraph,
...
         """
         Applying propagated masks for all nodes in topological order:
         1. running input_prune method for this node
         2. running output_prune method for this node
         """
+        if not self._is_mask_propogated:
+            nncf_logger.warning('Mask propogation didn\'t called before'
+                                ' apply_mask call in ModelPruner. This can lead to'
+                                ' incorrect model pruning export.')
+
```
daniil-lyakhov, line 586: @vshampor, shall we just inform user or raise runtime error? I would preferer runtime error

nncf/torch/pruning/export_helpers.py
daniil-lyakhov: @AlexKoff88, @vshampor I suggest to remove ModelPruner and all `input_prune` and `output_prune` methods due to they are redundant now

vshampor: @daniil-lyakhov Agreed, but you should do it in a separate PR

nncf/torch/pruning/export_helpers.py
vshampor: `propogation` -> `propagation`. RuntimeError is more suitable here.

nncf/torch/pruning/export_helpers.py
daniil-lyakhov: Please watch #945

examples/tensorflow/common/models.py
```diff
@@ -199,24 +199,28 @@ def _depth(v, divisor=8, min_value=None):
...
         inputs)
-    if backend.image_data_format() == 'channels_first':
-        x = layers.Reshape((filters, 1, 1))(x)
-    else:
-        x = layers.Reshape((1, 1, filters))(x)
-    x = layers.Conv2D(
+    #if backend.image_data_format() == 'channels_first':
+    #    x = layers.Reshape((filters, 1, 1))(x)
+    #else:
+    #    x = layers.Reshape((1, 1, filters))(x)
```
ljaljushkin, line 205: Is it a temporary change? It seems we should handle it without commenting

daniil-lyakhov: Yes, it's only to test if nncf TF will work with se blocks boiled on dense layers instead of conv2d. PR is really raw by now

nncf/tensorflow/pruning/export_helpers.py
```diff
@@ -227,9 +228,46 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
...
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
+        output_mask = node.data.get('output_mask', None)
+        node.data['output_mask'] = output_mask
+
+
 @TF_PRUNING_OPERATOR_METATYPES.register('stop_propagation_ops')
 class TFStopMaskForwardOps(DefaultMetaOp):
-    additional_types = ['Dense', 'MatMul']
+    additional_types = ['MatMul']
```
ljaljushkin, line 270: Is it potentially break pruning of the models with Dense layers in the classifier producing vector with constant number of classes. I think we still can't prune such layers. Will it be handled in this PR?

daniil-lyakhov: Yes, I'll add tests to cover such cases. I think it should be the same logic as for the first convolution to prevent input image reduce but now it's about output classes

examples/tensorflow/classification/configs/pruning_quantization/mobilenet_v3_imagenet_pruning_geometric_median.json
```diff
@@ -0,0 +1,48 @@
...
+            "nesterov": true
+        }
+    },
+    "dataset": "imagenet2012",
+    "dataset_type": "tfds",
+
+    "compression": [
+        {
+            "algorithm": "filter_pruning",
+            "pruning_init": 0.9,
```
mkaglins, line 28: Is it just test changes for pruning rate?

daniil-lyakhov: Yes, it's only for testing

mkaglins: OK, I found here a lot of temporary changes, please, be sure you reverted them before merge PR.

daniil-lyakhov:
This implementation don't cover all cases

PR #940, opened 2021-09-22
Add matmul patterns for fusing

vshampor:
### Changes 
The patterns were added for fusing matmul and elementwise operations for PT quantization

### Reason for changes 
BERT exhibits this type of operations and currently an extra FQ is being inserted between matmul and add etc., which is unnecessary.

### Related tickets 
-

### Tests 
Examination of the resulting BERT-SQuAD model obtained with the transformers repo. SOTA PT eval build 539.

### DIFF
--- a/nncf/torch/hardware/fused_patterns.py
+++ b/nncf/torch/hardware/fused_patterns.py
@@ -5,6 +5,7 @@ from nncf.torch.graph.pattern_operations import ATOMIC_ACTIVATIONS_OPERATIONS
 from nncf.torch.graph.pattern_operations import BATCH_NORMALIZATION_OPERATIONS
 from nncf.torch.graph.pattern_operations import GROUP_NORMALIZATION_OPERATIONS
 from nncf.torch.graph.pattern_operations import LINEAR_OPERATIONS
+from nncf.torch.graph.pattern_operations import MATMUL_OPERATIONS
 from nncf.torch.graph.pattern_operations import RELU_OPERATIONS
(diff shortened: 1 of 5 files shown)


vshampor:
Eval build 539 shows no degradations.

nncf/torch/hardware/fused_patterns.py
```diff
@@ -37,6 +42,8 @@ def _get_torch_hw_fused_patterns() -> HWFusedPatterns:
 
     retval.register(linear_ops + batch_norm_activations_permutation, 'LINEAR + BN_ACT_PERM',
                     match=True)
+    retval.register(matmul_ops + arithmetic_ops, 'MATMUL + ARITHMETIC',
+                    match=True)
```
ljaljushkin, line 46: Could you please add a [synthetic model](https://github.com/openvinotoolkit/nncf/blob/5d60171af4b8a4bf840a9af0b5074c8b11d2d5a0/tests/torch/test_models/synthetic.py#L196) with this pattern and add a new test case here in the [test_synthetic_model_quantization](https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/test_compressed_graph.py#L718)?

ljaljushkin (changes requested):

nncf/torch/hardware/fused_patterns.py
vshampor: Done, see the .dot file visualization with the pattern fused:
Screenshot from 2021-09-30 18-01-56

ljaljushkin (approved):

PR #946, opened 2021-09-23
Fix view with size

vshampor:


nncf/torch/dynamic_graph/graph.py
```diff
@@ -109,9 +124,14 @@ def __init__(self,
...
-        return (self.op_address == other.op_address) and \
-               self.input_matcher(self.tensor_metas, other.tensor_metas, self.tm_comparators)
+    def __eq__(self, other):
+        return self.op_address == other.op_address and Counter(self.tensor_metas) == Counter(other.tensor_metas)
+
+    def matches_saved_inputs_from(self, other: 'OperationExecutionContext'):
+        # WARNING: not commutative
+        return self.op_address == other.op_address and self.input_matcher(other.tensor_metas,
+                                                                          self.tensor_metas,
+                                                                          self.tm_comparators)
```
ljaljushkin, line 134: In which situations commutative comparison is used?
Is it theoretically possible that inputs, let's say for the Add and Concat modules, would come in the different order from time to time?

nncf/torch/dynamic_graph/graph.py
```diff
@@ -70,19 +70,34 @@ def __call__(self, node_inputs: List[TensorMeta], real_inputs: List[TensorMeta],
 
 
 class DefaultInputsMatcher(InputsMatcher):
-    def __call__(self, node_inputs: List[TensorMeta], real_inputs: List[TensorMeta],
+    def __call__(self, saved_inputs: List[TensorMeta], actual_inputs: List[TensorMeta],
```
ljaljushkin, line 73: I like this names, it's more clear now 👍

nncf/torch/dynamic_graph/graph.py
vshampor: Commutative comparison is used for comparing `DynamicGraph` instances which hold the `OperationExecutionContext`.

PR #947, opened 2021-09-24
Migrate to torch 1.9.1

daniil-lyakhov:
Our project [deep-object-reid](https://github.com/openvinotoolkit/deep-object-reid) already have migrated to torch 1.9.1. This pr should check if it possible to just update torch version in nncf without efforts


pfinashx:
Jenkins please retry a build

ljaljushkin:
All tests failed because of minor change in the MobileNet-v2 model: 
image
`x.reshape(x.shape[0], -1)` is identical to `x=x.flatten(x,1)` according to the spec:

**Flatten**
Flattens input by reshaping it into a one-dimensional tensor. If start_dim or end_dim are passed, only dimensions starting with start_dim and ending with end_dim are flattened. The order of elements in input is unchanged.
https://pytorch.org/docs/stable/generated/torch.flatten.html

**Reshape**
Returns a tensor with the same data and number of elements as input, but with the specified shape. 
A single dimension may be -1, in which case it’s inferred from the remaining dimensions and the number of elements in input.
https://pytorch.org/docs/stable/generated/torch.reshape.html

The quantization setup is not affected and the same:
image
image

The fix in #953  - is just to regenerate reference graphs for mobilenet-v2 model.

daniil-lyakhov:
Closing due to duplicate PR

PR #953, opened 2021-09-27
Migrate to torch 1.9.1

ljaljushkin:
Fix for the failed tests in #947

All tests failed because of minor change in the MobileNet-v2 model: 
image
`x.reshape(x.shape[0], -1)` is identical to `x=x.flatten(x,1)` according to the spec:

**Flatten**
Flattens input by reshaping it into a one-dimensional tensor. If start_dim or end_dim are passed, only dimensions starting with start_dim and ending with end_dim are flattened. The order of elements in input is unchanged.
https://pytorch.org/docs/stable/generated/torch.flatten.html

**Reshape**
Returns a tensor with the same data and number of elements as input, but with the specified shape. 
A single dimension may be -1, in which case it’s inferred from the remaining dimensions and the number of elements in input.
https://pytorch.org/docs/stable/generated/torch.reshape.html

The quantization setup is not affected and the same:
image
image

The fix in #953  - is just to regenerate reference graphs for mobilenet-v2 model.


ljaljushkin:
Jenkins please retry a build

ljaljushkin:
compression training tests passed on CIFAR100 (build 122)

vshampor (approved):

daniil-lyakhov (approved):

daniil-lyakhov:
It's possible to models names have changed in new version of torchvision. It would be great if we will check each config before migrating (I guess sota checkpoint job can help to do this but we should check if torch is correct)

ljaljushkin:
> It's possible to models names have changed in new version of torchvision. It would be great if we will check each config before migrating (I guess sota checkpoint job can help to do this but we should check if torch is correc

I've launched sota checkpoint eval for all unique models in the test scope. Haven't experienced any issue with model creation, loading fp32 weights and evaluation on a single batch.

PR #954, opened 2021-09-28
Reduce batch size for LM training in sanity

vshampor:


ljaljushkin (approved):
Does it make sense to reduce batch size for all other cases?

vshampor:
Jenkins please retry a build

PR #958, opened 2021-10-01
Refactor export helpers

daniil-lyakhov:
* Have separated common parts of pruning export_helpers 
* Implement base realization of `mask_propogation` methods to have opportunity to use mask propagation algo without any specific framework(tf, torch) functions. 

The main point of this refactoring is mask propagation algo with framework agnostic masks support. Thus we will be allowed to use mask propogation algo at model analyzing stage to check state: "Every prunable layer have correspondent closing prunable layer". For this, numpy arrays will be propagated by base realization of `mask_propogation`

TODO:

- [ ] Implement test to cover common and algo specific mask propogation

### DIFF
--- a/nncf/common/pruning/export_helpers.py
+++ b/nncf/common/pruning/export_helpers.py
@@ -10,8 +10,19 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+
+import numpy as np
+
+from typing import Union
+
 from nncf.common.graph import NNCFGraph
 from nncf.common.graph import NNCFNode
+from nncf.common.pruning.utils import is_grouped_conv
+from nncf.common.pruning.utils import get_sources_of_node
+from nncf.common.pruning.utils import is_depthwise_conv
+from nncf.common.graph.layer_attributes import GroupNormLayerAttributes
+from nncf.common.pruning.mask_propagation import identity_mask_propagation
+from nncf.common.pruning.mask_propagation import get_input_masks
 
 
 class DefaultMetaOp:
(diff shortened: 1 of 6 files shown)


nncf/common/pruning/mask_propagation.py
```diff
@@ -40,7 +39,7 @@ def __init__(self, graph: NNCFGraph, pruning_operator_metatypes: PruningOperatio
         self._graph = graph
         self._pruning_operator_metatypes = pruning_operator_metatypes
 
-    def get_meta_operation_by_type_name(self, type_name: str) -> DefaultMetaOp:
+    def get_meta_operation_by_type_name(self, type_name: str) -> 'DefaultMetaOp':
```
daniil-lyakhov, line 42: Did this to prevent import cycling. Is that ok? @alexsu52 @shamp

vshampor: I'm against this, if this is avoidable. Commonly it is avoidable by adding more .py files and extracting the common dependencies/structs that themselves have little or no dependencies into these separate files. This also helps split the larger files into smaller, more human-friendly ones.

nncf/common/pruning/export_helpers.py
```diff
@@ -50,3 +61,192 @@ def get_all_op_aliases(cls):
             op_types.extend(subtype.get_all_aliases())
         op_types = list(set(op_types)) + cls.additional_types
         return op_types
+
+
+class OpInput(DefaultMetaOp):
```
vshampor, line 66: I think we really should introduce naming-level separation of these additional entities (that we use for mask propagation and operation marking for the pruning algo specifically) from the generic metatypes that we use during NNCF graph preparation. If possible, the word `Meta` should not be used in the `DefaultMetaOp` name and basically anywhere where we do not handle in actual `OperatorMetatype`s. Vice versa - everywhere we actually refer to `OperatorMetatype` names in the pruning context, such as when mapping the generic `OperatorMetatype` to the pruning-specific markers, we should use the names `meta`-something. The `subtypes` field in PT pruning markers, for instance, should be renamed to `associated_metatypes` or a similar, more specific identifier.

vshampor (changes requested):

nncf/common/pruning/mask_propagation.py
daniil-lyakhov: Ok, I'll refactor it

mkaglins:
@daniil-lyakhov Please, define what is **closing prunable layer** in PR description and add more details about why is it needed.

daniil-lyakhov:
> @daniil-lyakhov Please, define what is **closing prunable layer** in PR description and add more details about why is it needed.

Every prunable layer L should have **closing prunable layer** means next prunable layer after L  which input channels are pruning.

For example:  
image

Green group is group of prunable layers, which has same channel dimention. We are removing some output channels from them. And yellow  prunable layer is closing prunable layer which input channels we are removing. Hope this helped

nncf/common/pruning/export_helpers.py
```diff
@@ -50,3 +61,212 @@ def get_all_op_aliases(cls):
...
+
+        for input_node in graph.get_previous_nodes(node):
+            # If input has mask ->  it went from convolution (source of this node is a convolution)
+            if input_node.data.get('output_mask', None) is not None:
+                continue
+
+            source_nodes = get_sources_of_node(input_node, graph, cls.ConvolutionOp.get_all_op_aliases() +
+                                               cls.StopMaskForwardOp.get_all_op_aliases() +
+                                               cls.InputOp.get_all_op_aliases())
+            sources_types = [node.node_type for node in source_nodes] + [input_node.node_type]
```
daniil-lyakhov, line 190: Please be aware of two changes here:

```
if input_node.data.get('output_mask', None) is not None:
```
instead of 

```
if input_node.data.get('output_mask', None) is None:
```

and 

```
sources_types = [node.node_type for node in source_nodes] + [input_node.node_type]
```
instead of 
```
sources_types = [node.node_type for node in source_nodes]
```

daniil-lyakhov: @mkaglins

nncf/common/pruning/export_helpers.py
```diff
@@ -50,3 +61,212 @@ def get_all_op_aliases(cls):
...
+class OpGroupNorm(DefaultMetaOp):
+    @classmethod
+    def accept_pruned_input(cls, node: NNCFNode):
+        # For Instance Normalization
+        return isinstance(node.layer_attributes, GroupNormLayerAttributes) \
+               and node.layer_attributes.num_groups == node.layer_attributes.num_channels
+
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph):
+        identity_mask_propagation(node, graph)
```
daniil-lyakhov, line 160: @mkaglins, could you please tell me, is it valid point that even if `accept_pruned_intput==False` mask porpagated as is?

nncf/common/pruning/mask_propagation.py
daniil-lyakhov: Done

nncf/common/pruning/export_helpers.py
daniil-lyakhov: Done, I've  renamed all pruning operations to `TF/PTOpNamePruningOp`

nncf/common/pruning/export_helpers.py
```diff
@@ -10,43 +10,226 @@
...
+        if all(mask is None for mask in input_masks):
+            return None
+
+        filled_input_masks = []
+        for i, mask in enumerate(input_masks):
+            if mask is None:
+                concat_axis = node.layer_attributes.axis
+                concat_dim = input_edges[i].tensor_shape[concat_axis]
+                device = cls._get_masks_device(input_masks)
+                mask = cls._get_unit_mask(concat_dim, device)
```
daniil-lyakhov, line 193: After concat layer attributes implementation I realized one can just use shape of the first non empty input masks because all input masks **should** be the same and thus concat layer attributes is redundant. But current approach is a bit more straightforward. If you would like I'll remove concat axis attribute and everything will work ok. But maybe concat axis will help in the future

daniil-lyakhov:
Minor TODO:

- [ ] Support torch.stask axis attr

nncf/common/pruning/export_helpers.py
mkaglins: @daniil-lyakhov Please, describe how the second change (of sources_types) will impact how the algorithm works on known models with Concat.

And, please, test these changes carefully on these models.

nncf/common/pruning/export_helpers.py
mkaglins: I think Concat axis attribute can be helpful.

mkaglins: Discussed with @daniil-lyakhov offline that **axis** is sufficient since inputs can have different shapes on concatenated axis, for example: 
Concat([1, 2, 5, 6], [1, 3, 5, 6], [1, 4, 5, 6], axis=1) -> [1, 2+3+4, 5, 6]

daniil-lyakhov:
Jenkins please retry a build

PR #961, opened 2021-10-04
Change the default pruning schedule to exponential

evgeniya-egupova:


mkaglins (approved):

AlexKoff88:
"baseline" schedule still works incorrectly because it triggers the maximum compression rate right after the first epoch no matter how many steps we specified in the config. Another point, is why this is called baseline and whether it can be demanding at all? I would expect to have something simple instead, like evenly increasing compression rate from epoch 0 to pruning_step.

AlexKoff88:
But I am not against having "exponential" by default.

PR #962, opened 2021-10-04
Test compression training fix 

mkaglins:
Fixed test for compression training for LeGR algorithm. Ticket - 62532.
The number of training epochs and potential accuracy diff has been increased.

@ljaljushkin @vshampor


ljaljushkin:
@mkaglins how much is time increased for compression training tests? hope it takes not too long with 60 epoch now.

mkaglins:
> @mkaglins how much is time increased for compression training tests? hope it takes not too long with 60 epoch now.

It was a typo when epochs were set to 1. 
Of course, the time of training will increase (but only + 25% compared to the previous version of the config).

ljaljushkin (approved):

PR #964, opened 2021-10-04
fix flops calculation for grouped convolutions

evgeniya-egupova:


nncf/common/pruning/utils.py
```diff
@@ -306,9 +306,10 @@ def count_flops_and_weights_per_node(graph: NNCFGraph,
         name = node.node_name
         num_in_channels = input_channels.get(name, node.layer_attributes.in_channels)
         num_out_channels = output_channels.get(name, node.layer_attributes.out_channels)
+        filters_per_channel = num_out_channels // node.layer_attributes.groups
```
AlexKoff88, line 309: Test?

evgeniya-egupova: A part of TODO list for today

PR #969, opened 2021-10-07
WIP: Check concat in export_helper refactor pr by CI

daniil-lyakhov:
To validate export helpers changes on CI

### DIFF
--- a/nncf/common/pruning/export_helpers.py
+++ b/nncf/common/pruning/export_helpers.py
@@ -168,7 +168,7 @@ class ConcatPruningOp(DefaultPruningOp):
 
                 print(f'is_on_sources_equal = {is_on_sources_equal}')
                 print('behaviour changed!!!')
-            if decision_new:
+            if decision_old:
                 return False
         return True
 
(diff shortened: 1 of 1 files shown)


daniil-lyakhov:
Common tests 
`tests.common.pruning.test_export_helpers.test_stop_ops_elementwise_source_before_concat[False]`
`tests.common.pruning.test_export_helpers.test_stop_ops_elementwise_source_before_concat[True]`

only because old behavior was turned on. This is evidence `check_concat` function always returning True:
https://github.com/openvinotoolkit/nncf/blob/6b345842f7cd910842d841e939629f8199017718/nncf/common/pruning/export_helpers.py#L219

All models with concat was checked like this: [tf](https://github.com/openvinotoolkit/nncf/blob/6b345842f7cd910842d841e939629f8199017718/tests/tensorflow/pruning/test_concat.py), [torch](https://github.com/openvinotoolkit/nncf/blob/6b345842f7cd910842d841e939629f8199017718/tests/torch/pruning/test_concat.py)

PR #972, opened 2021-10-10
Mark and visualize integer edges in NNCF PT

vshampor:
### Changes

The integer edges are now being properly marked in NNCFGraph generated for the PyTorch backend, so that the solver can utilize this information. Furthermore the integer edges are now being displayed as dashed in `compressed_graph.dot` and `original_graph.dot` representations.

### Reason for changes

The models with integer tensors within their control flow graph will now be quantized properly - with less chance of a quantizer ending up quantizing an integer branch.

### Related tickets

N/A

### Tests

Added a test to build a graph on a toy model with integer tensors in the CFG, count the number of detected integer tensors and compare against the expected number.

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -14,6 +14,7 @@ import os
 from collections import defaultdict
 from copy import deepcopy
 from typing import Any, Callable, Dict, KeysView, List, Tuple, Type, ValuesView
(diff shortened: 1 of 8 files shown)


vshampor:
FYI this is how the integer paths in the model graph will look like:
integer_graph

PR #994, opened 2021-10-26
Pruning masks applying

daniil-lyakhov:
### Changes

##MKaglinslaya
- `FilterPruningBlock` changed to apply masks to weights, not only store mask
- Masks now applying only on the forward pass, all zeroing of original model weights (in apply_mask) was removed
- To apply masks to Batch/Group Norm layers, these layers are wrapped as NNCF modules and `FilterPruningBlock`  added to them. Masks propagation to these layers from convolutions.
- `UpdateWeightAndBiasPruning` pre-layer operation was added to consistently prune weights and bias  (@vshampor Please, take a look on these changes)
- `zero_grad` attribute was deleted as not needed anymore
@AlexanderDokuchaev @daniil-lyakhov Please,  take a look at these changes in pruning.
cc @evgeniya-egupova 

##DLyakhov
- Fix sparsity algo to prevent batchnorm layers sparsity
- Remove batchnorm wrapping in pruning algo when batchnorm pruning is disable

### Reason for changes

New masks applying mechanism allow re-pruning of the model since weights aren't pruning permanently.

### Related tickets

Ticket ID 53524

### Tests

A couple of tests in test_algo was changed to test the new mechanism (instead of zeroing weights zeroing of conv outputs are checked).


nncf/torch/pruning/base_algo.py
```diff
@@ -111,27 +112,59 @@ def _prune_weights(self, target_model: NNCFNetwork):
...
+        all_norm_layers = target_model_graph.get_nodes_by_types(types_to_apply_mask)
+        for node in all_norm_layers:
+            if node.data['output_mask'] is None:
+                # Skip elements that will not be pruned
+                continue
+
+            node_name = node.node_name
+            module = target_model.get_containing_module(node_name)
+
+            pruning_block = self.create_weight_pruning_operation(module, node_name)
```
ljaljushkin, line 153: I wonder why norm layers are handled separately from clustered groups of layers to be pruned?
Why not to handle them with all pruned layers in `create_pruning_groups` method?

ljaljushkin (approved):

daniil-lyakhov:
SOTA check
607, new checkpoints - all checked 
608, old checkpoints - all checked except pruning checkpoints, as expected
CC: @vshampor

vshampor (approved):

evgeniya-egupova (approved):

PR #999, opened 2021-11-01
mixed_min_max collector. Unifies range initialization between frameworks.

negvet:
### Changes

- A new collector (mixed_min_max) is introduced for TensorFlow and Torch which is similar to POT min_max.
- TensorStatistic and TensorStatisticCollector moved to the common part and are used by TensorFlow as well.
- All min_max collectors for TensorFlow are unified.
- min_max and mean_min_max are updated with abs_max instead of max for symmetric quant

### Reason for changes

Better initialization for some cases with mixed_min_max collector.
Now, part of the range init code is unified and used by both backends.

### Related tickets

68418, 58748

### Tests

Tests are updated. Test for a mixed_min_max collector is added.


negvet:
Jenkins please retry a build

nncf/torch/quantization/init_range.py
```diff
@@ -122,15 +133,19 @@ def generate_collectors_for_range_init_statistics_collection(target_model_graph:
...
         if num_samples_to_collect_override is not None:
             num_samples = num_samples_to_collect_override
         if init_config.init_type == "min_max":
-            return MinMaxStatisticCollector(reduction_shapes, num_samples)
+            return MinMaxStatisticCollector(reduction_shapes, rs_vs_params, num_samples)
+        if init_config.init_type == "mixed_min_max":
+            return MixedMinMaxStatisticCollector(is_weights, reduction_shapes, rs_vs_params, num_samples)
         if init_config.init_type == "mean_min_max":
-            return MeanMinMaxStatisticCollector(reduction_shapes, num_samples)
+            return MeanMinMaxStatisticCollector(reduction_shapes, rs_vs_params, num_samples)
```
ljaljushkin, line 148: `reduction_shapes` is not really needed here. You could iterate over values of `rs_vs_params` dictionary to get the same.
the redundancy can be solved by passing more general `rs_vs_params` for the base class.

ljaljushkin: or maybe by having a different constructor with params that are actually needed...

nncf/common/tensor_statistics/statistics.py
```diff
@@ -0,0 +1,53 @@
...
+class MinMaxTensorStatistic(TensorStatistic):
+    def __init__(self, min_values, max_values):
+        self.min_values = min_values
+        self.max_values = max_values
+
+    @classmethod
+    def from_stat(cls, statistic: TensorStatistic):
+        if isinstance(statistic, MinMaxTensorStatistic):
+            return cls(statistic.min_values, statistic.max_values)
+        if isinstance(statistic, MedianMADTensorStatistic):
```
ljaljushkin, line 30: that implementation looks strange in `MinMaxTensorStatistic` class. 
it shouldn't know about other subclasses like `MedianMADTensorStatistic`

nncf/torch/quantization/init_range.py
```diff
@@ -145,7 +160,7 @@ def generate_stat_collector_for_range_init_config(
 
     @staticmethod
     def get_all_scale_shapes(qp: QuantizationPointBase,
-                             target_nncf_graph: PTNNCFGraph) -> List[Tuple[int]]:
+                             target_nncf_graph: PTNNCFGraph) -> List[Tuple[Tuple[int], QuantizationMode, bool]]:
```
ljaljushkin, line 163: maybe returning of the class with properties would be helpful here. you could pass this structure further to collectors without parsing

ljaljushkin (changes requested):

nncf/common/tensor_statistics/statistics.py
negvet: Thanks for the comment. In this case, I reused the previous implementation. I guess that the main idea is that MinMaxTensorStatistic (as the basic one, which attributes are used for range initialization) can be created from another kind of TensorStatistic. For example, MinMaxTensorStatistic can be created from MedianMADTensorStatistic, which has different properties. That is why MinMaxTensorStatistic has to be aware of other TensorStatistic types and their properties.

nncf/torch/quantization/init_range.py
negvet: Thanks for the comment. It seems that in this case, the simplest scenario is just to create `rs_vs_params` dictionary object right within the `get_all_scale_shapes_with_params` and return it. This is simple enough and helps to avoid additional structures.

nncf/torch/quantization/init_range.py
negvet: Thanks for the comment. `reduction_shapes` are removed. Now all collectors accept only `rs_vs_params` and it is processed further accordingly. 
Only `min_max` based collectors require some parameters other than `reduction_shapes`. That is why I propose to extract the `reduction_shapes` and pass it to the base class, keeping other params at the level of the collectors.

vshampor (changes requested):
My big issue with this PR is the fact that the quantizer configuration specifics such as `per_channel` have leaked into the tensor statistic collectors. I think that doing specialized range initialization for quantizers with specific settings (such as `MixedMinMax`) could be achieved in the same manner as it is done now in PyTorch, mainly using `ReductionShape`s, and any collector specialization could be done at the level of the collector creation. This will also probably lead to more code reuse between PT and TF. Let me know if you think that the set of current TF initializers is fundamentally incompatible with this scheme.

nncf/torch/tensor_statistics/collectors.py
```diff
@@ -11,243 +11,209 @@
...
+from nncf.torch.tensor_statistics.statistics import PTMinMaxTensorStatistic
+from nncf.torch.tensor_statistics.statistics import PTMedianMADTensorStatistic
+from nncf.torch.tensor_statistics.statistics import PTPercentileTensorStatistic
+
+
+class PTOfflineTensorStatisticCollector(OfflineTensorStatisticCollector):
     def _register_input(self, x: torch.Tensor):
-        pass
+        with no_nncf_trace():
+            self._samples.append(x.detach().cpu().numpy())
```
alexsu52, line 34: What happens if I want to calculate several offline statistics for one tensor in the model? Will I get double memory consumption? cc' @vshampor

alexsu52 (changes requested):

nncf/tensorflow/quantization/initializers/collectors.py
```diff
@@ -11,132 +11,140 @@
...
+        if self._input_type == InputType.INPUTS and not self._per_channel and self._mode == 'asymmetric':
+            return tf.math.reduce_mean(tf.stack(self._all_min_values), axis=0)
+        return tf.math.reduce_min(tf.stack(self._all_min_values), axis=0)
 
-    def __call__(self, *args, **kwargs):
-        self.call(*args, **kwargs)
+    def _max_aggregate(self) -> tf.Tensor:
+        if self._input_type == InputType.INPUTS and not self._per_channel:
+            return tf.math.reduce_mean(tf.stack(self._all_max_values), axis=0)
+        return tf.math.reduce_max(tf.stack(self._all_max_values), axis=0)
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Fixed. CollectorParams is now responsible for the parametrization of collectors.

nncf/tensorflow/quantization/initializers/collectors.py
```diff
@@ -11,132 +11,140 @@
...
-        return (self._median + 3 * self.distribution_factor * self._mad).astype(np.float32)
 
-    def prepare_statistics(self):
+class MedianMADStatisticCollector(TFOfflineTensorStatisticCollector):
+    """
+    Collector uses three-sigma approach.
+    """
+
+    def __init__(self, channel_axes: Union[int, Tuple[int], List[int]],
+                 input_type: str, per_channel: bool = False, num_samples: int = None):
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Done.

nncf/torch/quantization/init_range.py
```diff
@@ -106,46 +106,50 @@ def generate_collectors_for_range_init_statistics_collection(target_model_graph:
...
-            return MeanPercentileStatisticCollector([min_percentile, max_percentile], reduction_shapes, num_samples)
+            return MeanPercentileStatisticCollector([min_percentile, max_percentile], rs_vs_params, num_samples)
         raise RuntimeError("Unknown range init type: {}".format(init_config.init_type))
 
     @staticmethod
-    def get_all_scale_shapes(qp: QuantizationPointBase,
-                             target_nncf_graph: PTNNCFGraph) -> List[Tuple[int]]:
+    def get_all_scale_shapes_with_params(qp: QuantizationPointBase,
+                                         target_nncf_graph: PTNNCFGraph) -> Dict[ReductionShape,
+                                                                                 Dict[str, Union[str, bool]]]:
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Fixed. Now `get_all_scale_shapes_with_params()` returns `Dict[ReductionShape, CollectorParams]`.

nncf/tensorflow/quantization/initializers/init_range.py
```diff
@@ -85,24 +87,28 @@ def __init__(self, range_init_params: TFRangeInitParams):
...
-    def generate_stat_collector(init_config, per_channel: bool, channel_axes: int, input_type: str):
+    def generate_stat_collector(init_config, channel_axes: int, input_type: str,
+                                mode: str = 'symmetric', per_channel: bool = False):
         range_type = init_config.init_type
+        num_samples = init_config.num_init_samples
         if range_type == 'min_max':
-            return MinMaxStatisticCollector(per_channel, channel_axes, input_type)
+            return MinMaxStatisticCollector(channel_axes, input_type, mode, per_channel, num_samples)
+        if range_type == 'mixed_min_max':
+            return MixedMinMaxStatisticCollector(channel_axes, input_type, mode, per_channel, num_samples)
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Done.

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,85 @@
...
+from abc import ABC
+from abc import abstractmethod
+from collections import deque
+from typing import Tuple, Set
+
+
+ReductionShape = Tuple[int]
+
+
+class TensorStatisticCollectorBase(ABC):
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Partially fixed. Will also add method descriptions.

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @abstractmethod
+    def _reset(self):
+        pass
+
+    def collected_samples(self) -> int:
+        return self._collected_samples
+
+
+class StatisticsNotCollectedError(Exception):
```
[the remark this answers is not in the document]
negvet: Thanks for the comment. Fixed.

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,85 @@
...
+    pass
+
+
+class OnlineTensorStatisticCollector(TensorStatisticCollectorBase, ABC):
+    pass
+
+
+class OfflineTensorStatisticCollector(TensorStatisticCollectorBase, ABC):
+    def __init__(self, reduction_shape: ReductionShape = None, num_samples: int = None, window_size: int = None):
```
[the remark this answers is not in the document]
negvet: As far as I can see it is not used, but I guess this parameter is reserved to handle low available memory issues.

nncf/tensorflow/quantization/initializers/collectors.py
```diff
@@ -11,132 +11,140 @@
...
-            for i, _ in enumerate(self._all_min_values):
-                self._all_min_values[i] = tf.reshape(self._all_min_values[i], shape=new_shape)
-                self._all_max_values[i] = tf.reshape(self._all_max_values[i], shape=new_shape)
 
-    def call(self, inputs: tf.Tensor):
-        # No need to store extra statistics in memory since weights won't change during range init
-        if not self._all_min_values or self._input_type == InputType.INPUTS:
-            ndims = len(inputs.shape)
-            axis = get_axes(ndims, self._per_channel, self._channel_axes)
+class MixedMinMaxStatisticCollector(MinMaxStatisticCollectorBase):
```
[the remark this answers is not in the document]
negvet: `MinMaxStatisticCollector` and `MeanMinMaxStatisticsCollector` indeed share `_register_input()`, but they have different aggregators. It is possible to use one collector and parametrize it accordingly to get certain behavior. But I think that it is better to encapsulate these two types of stat collections in two different collector classes.

nncf/torch/tensor_statistics/collectors.py
negvet: That is right. Currently, the memory consumption will be doubled.

negvet:
The last commit includes
- one collector per one reduction shape in torch.
- CollectorParams contractor is used to parametrize collectors in a low-level manner.
- Reduction shapes are unified between frameworks for all min_max types of collectors.
- All min_max types of collectors are moved to the common part.
- Backend-related functions for stat collection are registered in the related backend (can be used by collectors in the common part and in the backend).

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,280 @@
...
+        super().__init__(reduction_shape, num_samples)
+        self._samples = deque(maxlen=window_size)
+
+    def _reset(self):
+        self._samples.clear()
+
+
+class CollectorParams:
```
vshampor, line 96: The params below do not belong to collectors, but rather to quantizers. These concepts should not be present in the `common/tensor_statistics`, but rather at the `quantization/init_range` level or so; possibly separated between frameworks.

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def __init__(self, use_abs_max: bool, reduction_shape: ReductionShape,
+                 num_samples: int = None, nncf_backend = None):
+        super().__init__(reduction_shape, num_samples)
+        self._use_abs_max = use_abs_max
+
+        self._min_values = None
+        self._max_values = None
+
+        if nncf_backend == BackendType.TORCH:
```
vshampor, line 148: This probably should not be determined via an init parameter. Why not use inheritance and polymorphism, implementing corresponding aggregation functions in framework-specific inheriting classes?

nncf/tensorflow/layers/data_layout.py
```diff
@@ -60,3 +60,8 @@ def get_weight_channel_axis(layer, weight_attr):
                 return weight_def.channel_axes
 
     return -1
+
+def get_weight_shape(layer, weight_attr):
```
vshampor, line 64: Type hints missing

nncf/tensorflow/quantization/initializers/init_range.py
```diff
@@ -76,60 +82,128 @@ def get_init_config_for_scope_and_group(self, node_name: str, group: QuantizerGr
...
             min_percentile = init_config.init_type_specific_params.get('min_percentile', 0.1)
             max_percentile = init_config.init_type_specific_params.get('max_percentile', 99.9)
-            return MeanPercentileStatisticCollector(per_channel, channel_axes, input_type,
-                                                    min_percentile, max_percentile)
+            return MeanPercentileStatisticCollector(reduction_shape,
+                                                    [min_percentile, max_percentile],
+                                                    num_samples)
         raise ValueError(f'Range type {range_type} is not supported.')
 
+    def _register_layer_statistics(self, layer, layer_statistics, handles):
```
vshampor, line 151: Type hints missing

nncf/tensorflow/quantization/initializers/init_range.py
```diff
@@ -76,60 +82,128 @@ def get_init_config_for_scope_and_group(self, node_name: str, group: QuantizerGr
...
+
+        collector = RangeInitializer.generate_stat_collector(reduction_shape,
+                                                             collector_params,
+                                                             init_config,
+                                                             num_batches)
+        handles.append(layer.register_hook_pre_quantizer(collector.register_input))
+        layer.enabled = False
+        layer_statistics.append((layer, collector))
+
+    def _register_op_statistics(self, layer, op_statistics, handles):
```
vshampor, line 173: Same

nncf/tensorflow/quantization/initializers/utils.py
```diff
@@ -29,13 +32,27 @@ def get_per_channel_history(inputs_tensor: np.ndarray, axis: list) -> np.ndarray
     return np.transpose(np.reshape(inputs_tensor, (new_shape, -1)))
 
 
-def get_axes(ndims, per_channel, channel_axes, add_dim=False):
+def get_axes(ndims: int, per_channel: bool, channel_axes: Union[int, list, tuple]):
```
vshampor, line 35: Return value type hints missing

nncf/tensorflow/quantization/initializers/utils.py
```diff
@@ -29,13 +32,27 @@ def get_per_channel_history(inputs_tensor: np.ndarray, axis: list) -> np.ndarray
...
+    reduction_shape = get_axes(ndims, layer.per_channel, channel_axes_)
+    if use_per_sample_stats:
+        reduction_shape = reduction_shape[1:]
+    return tuple(reduction_shape)
+
+
+def get_reduction_shape_weights(layer, weight_attr, channel_axes, per_channel):
+    weight_shape = get_weight_shape(layer, weight_attr)
+    ndims = len(weight_shape)
+    channel_axes_ = channel_axes if isinstance(channel_axes, (list, tuple)) else [channel_axes]
```
vshampor, line 56: Type hints missing

nncf/common/tensor_statistics/statistics.py
```diff
@@ -0,0 +1,53 @@
...
+        self.median_values = median_values
+        self.mad_values = mad_values
+
+
+class PercentileTensorStatistic(TensorStatistic):
+    def __init__(self, percentile_vs_values_dict):
+        self.percentile_vs_values_dict = percentile_vs_values_dict
+
+
+def convert_stat_to_min_max_tensor_stat(statistic: TensorStatistic):
```
vshampor, line 38: Return value type hints are missing

nncf/torch/tensor_statistics/aggregation_functions.py
```diff
@@ -0,0 +1,97 @@
...
+
+def convert_shape(shape):
+    if all(dim == 1 for dim in shape[1:]):
+        return [1]
+    return [1] + shape[1:]
+
+
+@aggregator.register()
+def pt_list_to_extend_stat_history(x):
+    return [t.view(convert_shape(list(x.size()))) for t in torch.unbind(x)]
```
vshampor, line 97: Type hints everywhere (if this file is still required after the refactoring that introduces polymorphism-powered FW specific collector dispatch)

nncf/torch/quantization/init_range.py
```diff
@@ -89,12 +90,37 @@ def get_init_config_for_scope_and_group(self, qid: QuantizerId, group: Quantizer
                          "definition!".format(str(qid)))
 
 
+class PTCollectorParams(CollectorParams):
```
vshampor, line 93: See, your PTCollectorParams are at the `init_range` level. Perhaps these should be called `PTRangeInitCollectorParams` ; the common class for these should either be at the `nncf/common/quantization/init_range` level (and its naming aligned as well - for instance, `RangeInitCollectorParams` instead of just `CollectorParams`. Also it might be the case that there is no need to generalize this struct on the FW-common level - I leave this decision at your discretion.

vshampor (changes requested):

nncf/torch/quantization/init_range.py
```diff
@@ -190,15 +247,23 @@ def fwd_hook(module, input_, output):
...
         for name, data in self.modules_to_init.items():
-            quantizer_module, init_config = data  # type: BaseQuantizer, RangeInitConfig
+            quantizer_module, init_config, is_weights = data  # type: Tuple[BaseQuantizer, RangeInitConfig, bool]
             num_samples_override = None
             if self.batch_size is not None:
                 num_batches = np.ceil(init_config.num_init_samples / self.batch_size)
                 num_samples_override = num_batches
 
+            if isinstance(quantizer_module, SymmetricQuantizer):
+                mode = 'symmetric'
```
ljaljushkin, line 257: This comment is about further improvement, not critical for merging.
IMHO, the existing `QuantizationMode` class is better than strings. Potentially, this mode can be stored in quantized_module.

ljaljushkin (approved):
looks good to me, only minor comment

nncf/torch/quantization/init_range.py
negvet: I see and agree with the statement. I fixed it. Thanks!

nncf/torch/tensor_statistics/aggregation_functions.py
negvet: Thanks, fixed!

nncf/common/tensor_statistics/statistics.py
negvet: Thanks, fixed!

nncf/tensorflow/quantization/initializers/utils.py
negvet: Thanks, fixed!

nncf/tensorflow/quantization/initializers/utils.py
negvet: Thanks, fixed!

nncf/tensorflow/quantization/initializers/init_range.py
negvet: Thanks, fixed!

nncf/tensorflow/quantization/initializers/init_range.py
negvet: Thanks, fixed!

nncf/tensorflow/layers/data_layout.py
negvet: Thanks, fixed!

nncf/common/tensor_statistics/collectors.py
negvet: Thanks. I redefined it using inheritance and polymorphism.

nncf/common/tensor_statistics/collectors.py
negvet: Thanks, Fixed!

negvet:
Agreed to extend NNCFTensor and NNCFTensorProcessor functionality to handle tensor operations needed for statistic collection. As a result, the current approach with inheritance and polymorphism will be replaced by corresponding NNCFTensorProcessor functionality. Potentially this can reduce the code footprint and partially reuse NNCFTensorProcessor methods. Note, this approach brings more functional load to the NNCFTensor.

nncf/torch/quantization/init_range.py
negvet: Thanks for the comment, will update!

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,224 @@
...
+
+    def collected_samples(self) -> int:
+        return self._collected_samples
+
+
+class StatisticsNotCollectedError(Exception):
+    """Raised when the statistics are not collected but requested"""
+
+
+class OnlineTensorStatisticCollector(TensorStatisticCollectorBase):
```
daniil-lyakhov, line 85: Please add description for this class and bellow

nncf/tensorflow/quantization/initializers/collectors.py
```diff
@@ -11,268 +11,129 @@
...
-
-    @property
-    def min(self) -> tf.Tensor:
-        return tf.math.reduce_min(tf.stack(self._all_min_values), axis=0)
-
-    @property
-    def max(self) -> tf.Tensor:
-        return tf.math.reduce_max(tf.stack(self._all_max_values), axis=0)
+class TFMinMaxStatisticCollector(MinMaxStatisticCollector):
+    def _register_input(self, x: tf.Tensor):
```
daniil-lyakhov, line 33: let's discuss and choose a type for tf tensor. Can be or `tf.Tensor` or `tf.Variable`
```suggestion
    def _register_input(self, x: tf.Varible):
```

nncf/tensorflow/quantization/initializers/collectors.py
```diff
@@ -11,268 +11,129 @@
...
+        per_channel_median = []
+        per_channel_mad = []
+        for channel_history in per_channel_histories:
+            channel_history = discard_zeros(channel_history)
+            median = np.median(channel_history)
+            per_channel_median.append(median)
+            inputs_median_diff = abs(channel_history - median)
+            per_channel_mad.append(np.median(inputs_median_diff))
+        median = tf.squeeze(tf.convert_to_tensor(np.array(per_channel_median), dtype=tf.float32))
+        mad = tf.squeeze(tf.convert_to_tensor(np.array(per_channel_mad), dtype=tf.float32))
```
daniil-lyakhov, line 69: Can we move this to common by NNCFTensor?

nncf/common/tensor.py
```diff
@@ -47,6 +48,93 @@ class NNCFBaseTensorProcessor:
...
+        Computes the mean of elements across given dimensions of NNCFTensor.
+
+        :param x: NNCFTensor to reduce.
+        :param axis:
+        :return: Reduced NNCFTensor.
+        """
+
+    @classmethod
+    @abstractmethod
+    def stack(cls, x: Union[list, deque], axis: int = 0) -> NNCFTensor:
```
daniil-lyakhov, line 118: Please make typing more specific

nncf/tensorflow/quantization/initializers/collectors.py
negvet: Discussed and decided to use the tf.Tensor

nncf/common/tensor.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/collectors.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,226 @@
...
+    def collected_samples(self) -> int:
+        return self._collected_samples
+
+
+class StatisticsNotCollectedError(Exception):
+    """Raised when the statistics are not collected but requested."""
+
+
+class OnlineTensorStatisticCollector(TensorStatisticCollectorBase):
+    """Base class for collectors that collect statistics in online regime, without storing the data."""
```
daniil-lyakhov, line 83:
```suggestion
    """Base class for collectors that collects statistics in online regime, without storing the data."""
```

nncf/common/tensor_statistics/statistics.py
```diff
@@ -0,0 +1,53 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from abc import ABC
+
+
+class TensorStatistic(ABC):
```
daniil-lyakhov, line 17: Please don't forget to add docstring

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,226 @@
...
+class TensorStatisticCollectorBase(ABC):
+    """Collector estimate statistics at the quantization point based on the provided reduction shape."""
+
+    def __init__(self, reduction_shape: ReductionShape = None, num_samples: int = None):
+        self._reduction_shape = reduction_shape
+        self._enabled = True
+        self._collected_samples = 0
+        self._num_samples = num_samples
+
+    def register_input(self, x):
```
vshampor, line 34: Type hints

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,226 @@
...
+                self._num_samples is not None and self._collected_samples >= self._num_samples:
+            return x
+        if self._reduction_shape is None:
+            self._reduction_shape = {tuple(x.shape)}
+        self._register_input(x)
+        self._collected_samples += 1
+        return x
+
+    @abstractmethod
+    def _register_input(self, x):
```
vshampor, line 46: Type hints

nncf/tensorflow/layers/data_layout.py
```diff
@@ -60,3 +62,8 @@ def get_weight_channel_axis(layer, weight_attr):
                 return weight_def.channel_axes
 
     return -1
+
+def get_weight_shape(layer: tf.keras.layers.Layer, weight_attr: str):
```
vshampor, line 66: Return value type hint missing

nncf/torch/quantization/init_range.py
```diff
@@ -106,46 +133,67 @@ def generate_collectors_for_range_init_statistics_collection(target_model_graph:
...
+            reduction_shape: ReductionShape = None,
+            collector_params = None,
             num_samples_to_collect_override: int = None) -> TensorStatisticCollectorBase:
         num_samples = init_config.num_init_samples
         if num_samples_to_collect_override is not None:
             num_samples = num_samples_to_collect_override
         if init_config.init_type == "min_max":
-            return MinMaxStatisticCollector(reduction_shapes, num_samples)
+            use_abs_max = collector_params.get_low_level_params_for_collector()
+            reduction_shape_converted = collector_params.converte_reduction_shape()
```
vshampor, line 166: `converte` seems to be a typo

nncf/common/quantization/initialization/range.py
```diff
@@ -133,3 +134,42 @@ def __init__(self, init_range_data_loader: NNCFDataLoader,
...
+
+    def get_low_level_params_for_collector(self):
+        """Generate low-level parameters for collectors"""
+        if self._init_type == "min_max":
+            return self._get_params_for_min_max()
+        if self._init_type == "mixed_min_max":
+            return self._get_params_for_mixed_min_max()
+        if self._init_type == "mean_min_max":
+            return self._get_params_for_mean_min_max()
+        raise RuntimeError("Parameters not required or unknown range init type: {}".format(self._init_type))
```
vshampor, line 175: This could really stand to be refactored. At the very least the `_init_type` must be made a specific enum instead of an opaque str. A further improvement would be to avoid the functions for which you cannot write a meaningful return type hint such as this one, which essentially returns different object set in different situation. A more deep refactoring would have to be made for the latter to be fixed - you may postpone it by creating a ticket for this.

nncf/common/tensor.py
```diff
@@ -53,6 +53,93 @@ class NNCFBaseTensorProcessor:
...
+        :return: Absolute value of a NNCFTensor
+        """
+
+    @classmethod
+    @abstractmethod
+    def min(cls, x1: NNCFTensor, x2: NNCFTensor) -> NNCFTensor:
+        """
+        Returns the min of x1 and x2.
+
+        :param x1: NCFTensor to compare.
```
vshampor, line 94: Typo

nncf/common/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,226 @@
...
+        super().__init__(reduction_shape, num_samples)
+        self._use_abs_max = use_abs_max
+        self._tensor_processor = None
+
+        self._min_values = None
+        self._max_values = None
+
+    def _register_input_common(self, x: NNCFTensor):
+        if not self._tensor_processor:
+            self._tensor_processor = x.tensor_processor
```
vshampor, line 110: I don't really like this idiom. I guess I missed it during one of the previous reviews, but should each `NNCFTensor` really carry a processor with it? This situation corresponds to strong coupling, which is undesirable, and to additional possibility of system ending up in an incorrect state. I'd rather that the correct tensor processor was instantiated using polymorphism during the collector creation, rather than dynamically taken from the current tensor to be registered; and also each `NNCFTensor` should not retain a ref to the corresponding processor but instead all of the tensors in a given run under a specific FW should use a single tensor processor object obtained by separate, non-tensor-specific means. @daniil-lyakhov what do you think?

tests/torch/tensor_statistics/test_tensor_statistics.py
```diff
@@ -176,7 +208,7 @@ def test_collected_statistics(self, collector: Type[TensorStatisticCollectorBase
     @pytest.fixture(params=COLLECTORS)
     def collector_for_interface_test(self, request):
         collector_type = request.param
-        return collector_type(reduction_shapes={(1, )})
+        return collector_type(rs_vs_params={(1, ): {'mode': 'symmetric', 'per_channel': False}})
```
vshampor, line 211: `rs_vs_params` could also stand to be better typed. Basically we want to end up in a situation where the string literals such as `'mode'` or `'symmetric'` are only present in the raw config files, or at a very high level where the config is parsed. Everywhere else we should have typed structures and enums.

vshampor (changes requested):

nncf/common/quantization/initialization/range.py
negvet: As discused, `get_low_level_params_for_collector()` can be eliminated by crating public properties in `RangeInitCollectorParams`, which will be directly feed to instance collectors.

nncf/tensorflow/quantization/initializers/init_range.py
```diff
@@ -76,60 +82,129 @@ def get_init_config_for_scope_and_group(self, node_name: str, group: QuantizerGr
...
         self.range_init_params = range_init_params
         self.dataset = range_init_params.init_range_data_loader
         self.num_steps = range_init_params.get_max_num_init_steps()
+        self._collectors_with_per_sample_stat_collection = ['mixed_min_max', 'mean_min_max', 'mean_percentile']
 
         self.nncf_quantization_operation_classes = NNCF_QUANTIZATION_OPERATIONS.registry_dict.values()
 
     @staticmethod
-    def generate_stat_collector(init_config, per_channel: bool, channel_axes: int, input_type: str):
+    def generate_stat_collector(reduction_shape,
```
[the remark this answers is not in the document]
negvet: Fixed, thanks!

nncf/torch/quantization/init_range.py
negvet: Fixed, thanks!

nncf/common/quantization/initialization/range.py
negvet: Fixed, thanks!

nncf/common/tensor.py
negvet: Fixed, thanks!

nncf/tensorflow/layers/data_layout.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/statistics.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/collectors.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/collectors.py
negvet: Input here is either `torch.Tensor` or `tf.Tensor`. Conversion to FW-specific NNCFTensor is done within FW-redefined `_register_input()`. Do you propose to change it?

nncf/common/tensor_statistics/collectors.py
negvet: Input here is either `torch.Tensor` or `tf.Tensor`. Conversion to FW-specific NNCFTensor is done within FW-redefined `_register_input()`. Do you propose to change it?

nncf/common/tensor_statistics/collectors.py
daniil-lyakhov: The problem is in MaskPropagation algo. Running pruning algorithm, Symbolic mask propagation performed first. It uses framework specific metatypes to match node to correspondent purning operation and use SymbolicTensorProcessor to propagate masks. Then second mask propagation is performed, in this time with framework specific tensor processor. Now, mask propagation algo has no information about tensor processor. To set this information one need to keep it as an attribute of mask propagation and then forward to each pruning operation, because pruning operations *has no information about tensor nature*. In case you want to replace (ugly) forwarding to polymorphism instantiation, thus you will need to specify two kinds of pruning operations: pruning operations with framework specific tensor and with symbolic tensor, which is huge overkill in my opinion

daniil-lyakhov: The main idea behind such idiom is all common algorithm don't need any information about tensor nature. In case of statistic collectors, I think, it's more clear to specify tensor processor as attribute when initializing collector.

 But in case of pruning algo all logic of pruning should be refactored to use polymorphism

nncf/common/tensor_statistics/collectors.py
vshampor: In this cases we use `TypeVar` from `typing` as a stand-in for template type placeholder. Take a look at other uses of `TypeVar` throughout NNCF code and try to apply it here.

nncf/common/tensor_statistics/collectors.py
vshampor: As above

nncf/common/tensor_statistics/collectors.py
vshampor: Created ticket 71626 to address this out of the scope of the PR

nncf/common/tensor_statistics/collectors.py
negvet: Fixed, thanks!

nncf/common/tensor_statistics/collectors.py
negvet: Fixed, thanks!

tests/torch/tensor_statistics/test_tensor_statistics.py
negvet: Thanks! I got the idea. `rs_vs_params` are removed from the code at all after making forcing the collector to collect stats for one reduction shape. I update the code.

negvet:
Jenkins please retry a build

nncf/torch/quantization/init_range.py
```diff
@@ -89,12 +91,37 @@ def get_init_config_for_scope_and_group(self, qid: QuantizerId, group: Quantizer
                          "definition!".format(str(qid)))
 
 
+class PTRangeInitCollectorParams(RangeInitCollectorParams):
+    def __init__(self, is_weights: bool, mode: QuantizationMode, per_channel: bool, init_type: str, input_shape: tuple,
+                 channel_idx: int, collectors_with_per_sample_stat_collection: list):
+        super().__init__(is_weights, mode, per_channel, init_type)
+        self._input_shape = input_shape
+        self._channel_idx = channel_idx
+        self._collectors_with_per_sample_stat_collection = collectors_with_per_sample_stat_collection
```
vshampor, line 100: The `self._collectors_with_per_sample_stat_collection` looks strange as a parameter in the params struct. I propose you remove it and instead decide whether per-sample collection should be turned on for a given collector explicitly in the factory functions such as `StatCollectorGenerator.generate_stat_collector_for_range_init_config`

vshampor (changes requested):
Basically good now. In addition to the change I highlighted, please look over all of the impacted/added classes and functions and add docstrings and parameter description in accordance with the [style guide](https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#353-classes).

alexsu52 (approved):
LGTM. Could you run E2E tests for these changes?

nncf/common/pruning/mask_propagation.py
```diff
@@ -94,7 +98,7 @@ def symbolic_mask_propagation(self, prunable_layers_types: List[str],
                 node.data['output_mask'] = SymbolicMask(node.layer_attributes.out_channels, [node.node_id])
             # Propagate masks
             cls = self.get_meta_operation_by_type_name(node.node_type)
-            cls.mask_propagation(node, self._graph)
+            cls.mask_propagation(node, self._graph, self._tensor_processor)
```
daniil-lyakhov, line 101: Here only Symbolic mask processor can be used. I suggest to specify Symbolic mask processor here explicitly and make `self._tensor_processor` attribute optional in case only symbolic mask propagation is used

nncf/common/pruning/symbolic_mask.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, mask_producers: List[int] = None):
         super().__init__(-1, mask_producers)
 
 
-class SymbolicMaskProcessor(NNCFBaseTensorProcessor):
+class SymbolicMaskProcessor(NNCFPruningBaseTensorProcessor):
```
daniil-lyakhov, line 59: Why we need base class `NNCFPruningBaseTensorProcessor` which is base only for one class `SymbolicMaskProcessor`?

nncf/tensorflow/tensor_statistics/collectors.py
```diff
@@ -0,0 +1,148 @@
...
+        return [TFNNCFTensor(t) for t in tensor_list]
+
+
+class TFMinMaxStatisticCollector(MinMaxStatisticCollector):
+    @staticmethod
+    def _get_processor() -> NNCFCollectorTensorProcessor:
+        return TFNNCFCollectorTensorProcessor()
+
+    def _register_input(self, x: tf.Tensor):
+        self._register_input_common(TFNNCFTensor(x))
```
daniil-lyakhov, line 81: Can we unify this somehow (`TFMinMaxStatisticCollector`, `TFMixedMinMaxStatisticCollector`, `TFMeanMinMaxStatisticCollector`)? @vshampor @alexsu52

nncf/torch/tensor_statistics/collectors.py
```diff
@@ -11,243 +11,182 @@
...
 
+    @staticmethod
+    def _get_processor() -> NNCFCollectorTensorProcessor:
+        return PTNNCFCollectorTensorProcessor()
 
-class OfflineTensorStatisticCollector(TensorStatisticCollectorBase, ABC):
-    def __init__(self, reduction_shapes: Set[ReductionShape] = None, num_samples: int = None,
+    def _register_input(self, x: torch.Tensor):
+        with no_nncf_trace():
+            self._register_input_common(PTNNCFTensor(x))
```
daniil-lyakhov, line 88: Can we unify this (in `PTMinMaxStatisticCollector`, `PTMixedMinMaxStatisticCollector`, `PTMeanMinMaxStatisticCollector`) ?

nncf/torch/tensor_statistics/statistics.py
```diff
@@ -11,70 +11,45 @@
...
-        raise ValueError("Unknown statistic to generate min-max stat from!")
-
-    def __eq__(self, other: 'MinMaxTensorStatistic') -> bool:
-        return self.torch_tensor_eq(self.min_values, other.min_values) and \
-               self.torch_tensor_eq(self.max_values, other.max_values)
 
+class PTMedianMADTensorStatistic(MedianMADTensorStatistic):
+    @staticmethod
+    def tensor_eq(tensor1: torch.Tensor, tensor2: torch.Tensor, rtol=1e-6) -> bool:
+        return bool(torch.allclose(tensor1, tensor2, rtol=rtol))
```
daniil-lyakhov, line 31: Can we unify this?

negvet:
> LGTM. Could you run E2E tests for these changes?
Sure, I will run the e2e test.

nncf/torch/quantization/init_range.py
negvet: Updated, thenks!

nncf/common/pruning/mask_propagation.py
```diff
@@ -60,7 +65,7 @@ def mask_propagation(self):
         """
         for node in self._graph.topological_sort():
             cls = self.get_meta_operation_by_type_name(node.node_type)
-            cls.mask_propagation(node, self._graph)
+            cls.mask_propagation(node, self._graph, self._tensor_processor)
```
daniil-lyakhov, line 68:
```suggestion
            cls.mask_propagation(node, self._graph, SymbolicMaskProcessor)
```

nncf/common/pruning/mask_propagation.py
```diff
@@ -30,7 +32,9 @@ class MaskPropagationAlgorithm:
     for nodes that have masks already defined.
     """
 
-    def __init__(self, graph: NNCFGraph, pruning_operator_metatypes: PruningOperationsMetatypeRegistry):
+    def __init__(self, graph: NNCFGraph,
+                 pruning_operator_metatypes: PruningOperationsMetatypeRegistry,
+                 tensor_processor: Type[NNCFPruningBaseTensorProcessor] = SymbolicMaskProcessor):
```
daniil-lyakhov, line 37:
```suggestion
                 tensor_processor: Optional[Type[NNCFPruningBaseTensorProcessor]] = None):
```

nncf/common/pruning/mask_propagation.py
daniil-lyakhov:
```suggestion
            cls.mask_propagation(node, self._graph, SymbolicMaskProcessor)
```

nncf/tensorflow/tensor_statistics/collectors.py
vshampor: Not worth it IMO - would have to involve adding another base class or multiple inheritance.

vshampor (approved):

daniil-lyakhov (approved):

Issue #1001, opened 2021-11-03
Add support for LeakyReLU.


devansh20la:
Filter pruning skips all convolutions that are followed by leaky relu e.g
```
model = nn.Sequential(
            nn.Conv2d(3, 128, kernel_size=3),
            nn.LeakyReLU()
)
```

daniil-lyakhov:
Greetings, @devansh20la!

Thanks for your finding! Looks like leakyrelu operation is missed in torch pruning operation set. I'll fix it soon

PR #1005, opened 2021-11-08
Add Leaky Relu Metatype to Pruning Operations

daniil-lyakhov:
### Changes

Add LeakyRelu to PTIdentityMaskForwardPruningOp list

### Reason for changes

https://github.com/openvinotoolkit/nncf/issues/1001

### Related tickets

None

### Tests

None

### DIFF
--- a/nncf/torch/pruning/operations.py
+++ b/nncf/torch/pruning/operations.py
@@ -43,6 +43,7 @@ from nncf.torch.graph.operator_metatypes import (
     MulMetatype,
     PTOutputNoopMetatype,
     PRELUMetatype,
+    LeakyRELUMetatype,
     RELUMetatype,
     SigmoidMetatype,
     SILUMetatype,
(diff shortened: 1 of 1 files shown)


nncf/torch/pruning/operations.py
```diff
@@ -105,7 +106,7 @@ class PTOutputPruningOp(OutputPruningOp, PTPruner):
 @PT_PRUNING_OPERATOR_METATYPES.register('identity_mask_propagation')
 class PTIdentityMaskForwardPruningOp(IdentityMaskForwardPruningOp, PTPruner):
     subtypes = [HardTanhMetatype, TanhMetatype, RELUMetatype, PRELUMetatype, ELUMetatype, GELUMetatype, SigmoidMetatype,
-                SoftmaxMetatype, AvgPool2dMetatype, MaxPool2dMetatype, DropoutMetatype, SILUMetatype]
+                SoftmaxMetatype, AvgPool2dMetatype, MaxPool2dMetatype, DropoutMetatype, SILUMetatype, LeakyRELUMetatype]
```
ljaljushkin, line 109: This doesn't guarantee that pruning operation would be added for the model with LeakyReLU in a particular customer case. Ideally, there should a test with pruning a very basic synthetic model, e.g. https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/test_compressed_graph.py#L728
or probably, there's already a similar test in the pruning folder.

ljaljushkin: discussed internally. The proposed test would be redundant, because IdentityMask (based on ReLU) is properly tested.

ljaljushkin (approved):

PR #1019, opened 2021-11-18
Formulate PT OD configs in terms of epochs only

vshampor:
### Changes

The config values in PyTorch object detection samples are modified to use epochs instead of iterations; `iter_size` parameter is removed. 

### Reason for changes

The current object detection sample configs mix up epoch-based and iteration-based specifications for test/save intervals and sparsity application. This leads to confusion and in fact we currently cannot tell if the configs are even correct w.r.t. the scheduling and training length.

### Related tickets

71329

### Tests

test_sanity_sample


alexsu52 (approved):

vshampor:
Jenkins please retry a build

PR #1024, opened 2021-11-25
Activations only ignored scopes processing

SKholkin:
### Changes

Realization of activation only ignored scopes processing

### Reason for changes

There was no such logic before

### Related tickets

61860

### Tests

tests.torch.quantization.test_quantizer_propagation_solver.TestQuantizerPropagationSolver.test_activation_ignored_scope


tests/torch/quantization/test_quantizer_propagation_solver.py
```diff
@@ -1817,3 +1822,38 @@ def test_quantizers_are_not_set_up_for_integer_inputs(self, ip_graph_with_int_ed
...
                                                                                                       input_port_id=1)
+
+    @pytest.mark.parametrize('update_config_info, should_ignore_quantizers', [({}, []),
+                                                                              ({"ignored_scopes": ["LeNet/relu_1"]},
+                                                                               ['LeNet/relu_0']),
+                                                                              ({"activations": {
+                                                                                  "ignored_scopes": ["LeNet/relu_1"]}},
+                                                                               ['LeNet/relu_0'])])
+    def test_activation_ignored_scope(self, update_config_info, should_ignore_quantizers):
+        def compare_two_lists_permutation_invariant(list1, list2):
```
vshampor, line 1833: Will the `Counter` from `collections` be able to achieve the same effect?

nncf/torch/quantization/algo.py
```diff
@@ -335,7 +335,10 @@ def generate_setup(self) -> SingleConfigQuantizerSetup:
         if self._debug_interface:
             self._debug_interface.visualize_insertion_point_graph(insertion_point_graph)
         from nncf.common.quantization.quantizer_propagation.solver import QuantizerPropagationSolver
-        prop_graph_solver = QuantizerPropagationSolver(ignored_scopes=self.ignored_scopes,
+        ignored_scopes_for_solver = (lambda x: [x] if isinstance(x, str) else x)(self.ignored_scopes) + \
```
vshampor, line 338: Dangerous - it isn't obvious here that we are actually aiming to get either a list of `str` when starting from either a `str` or already a `list` of `str`. I suggest you make a separate named function for this, with asserts/runtime errors for the situation when the starting object is neither a `str` nor a `list` of `str` in which case we can do nothing but crash.

vshampor (changes requested):

nncf/torch/quantization/algo.py
SKholkin: A separate function was made but there is no asserts/runtime errors for situation when the starting object is neither a `str` nor a `list` of `str` because this function is used for ignored_scopes processing only which according to json schema 
can be only `str`, `List[str]` or `None` (this also valid for `target_scopes`)
`"ignored_scopes": with_attributes(make_object_or_array_of_objects_schema(_STRING), description=IGNORED_SCOPES_DESCRIPTION),
  "target_scopes": with_attributes(make_object_or_array_of_objects_schema(_STRING), description=TARGET_SCOPES_DESCRIPTION),`
Also None value processing in that particular case can have different logic from processing `None` value for `target_scopes` so this function isn't generalized and implemented as nested
@vshampor, @ljaljushkin  what do you think?

tests/torch/quantization/test_quantizer_propagation_solver.py
```diff
@@ -1817,3 +1822,34 @@ def test_quantizers_are_not_set_up_for_integer_inputs(self, ip_graph_with_int_ed
...
+
+    @pytest.mark.parametrize('update_config_info, should_ignore_quantizers', [({}, []),
+                                                                              ({"ignored_scopes": ["LeNet/relu_1"]},
+                                                                               ['LeNet/relu_0']),
+                                                                              ({"activations": {
+                                                                                  "ignored_scopes": ["LeNet/relu_1"]}},
+                                                                               ['LeNet/relu_0'])])
+    def test_activation_ignored_scope(self, update_config_info, should_ignore_quantizers):
+        model = LeNet()
+        all_quantization_names = ["LeNet/NNCFConv2d[conv1]/conv2d_0",
```
ljaljushkin, line 1834: please move it outside of test and make as a constant

tests/torch/quantization/test_quantizer_propagation_solver.py
```diff
@@ -1817,3 +1822,34 @@ def test_quantizers_are_not_set_up_for_integer_inputs(self, ip_graph_with_int_ed
...
+                                  "LeNet/NNCFLinear[fc3]/linear_0",
+                                  "/nncf_model_input_0",
+                                  "LeNet/relu_0",
+                                  "LeNet/relu_1",
+                                  "LeNet/relu_2",
+                                  "LeNet/relu_3"]
+        for item in should_ignore_quantizers:
+            try:
+                all_quantization_names.remove(item)
+            except ValueError:
```
ljaljushkin, line 1847: I'd make it more pythonic without exceptions:
```python
ref_quantization_names = list(filter(lambda x: x not in should_ignore_quantizers, all_quantization_names))
```

ljaljushkin (changes requested):

nncf/torch/quantization/algo.py
ljaljushkin: I'm ok with this function

vshampor (approved):

ljaljushkin (approved):

PR #1027, opened 2021-11-26
New "saturation_fix" config option

negvet:
### Changes

New config option "saturation_fix" was introduced. Which have choices for SF mode: `disable`, `enable`, `enable_for_first_conv_layer`. `enable` is the default option.
### Reason for changes

In order to short the tuning time and at the same time partially compensate for the saturation issue by using `enable_for_first_conv_layer` option.
### Related tickets

71400
### Tests

Tests are updated to validate all three options.

### DIFF
--- a/docs/compression_algorithms/Quantization.md
+++ b/docs/compression_algorithms/Quantization.md
@@ -133,7 +133,8 @@ This leads to accuracy degradation.
 
 To fix this issue inside NNCF, by default, all weight tensors are quantized in 8 bits but only 7 bits are effectively used.
 This regime is used when `"target_device": "CPU"` or `"target_device": "ANY"` set. This fix, potentially, requires longer fine-tuning.
-To apply the saturation issue fix only to the first layer use `"saturation_fix": "enable_for_first_conv_layer"`. To disable the saturation issue fix for all layers use `"saturation_fix": "disable"`.
+
+To control the application of saturation fix, `"saturation_fix"` config option is introduced. The default value is `"saturation_fix": "enable"`. To apply the saturation issue fix only to the first layer, use `"saturation_fix": "enable_for_first_conv_layer"`. To disable the saturation issue fix for all layers, use `"saturation_fix": "disable"`.
 
 ---
 
--- a/docs/compression_algorithms/Quantization.md
+++ b/docs/compression_algorithms/Quantization.md
@@ -129,12 +129,12 @@ For symmetric:
 
 There is a known issue with AVX2 and AVX512 CPU devices. The issue appears with 8-bit matrix calculations with tensors which elements are close to the maximum or saturated.
 AVX2 and AVX512 utilize a 16-bit register to store the result of operations on tensors. In case when tensors are saturated the buffer overflow happens.
-This leads to accuracy degradation.
+This leads to accuracy degradation. For more details of the overflow issue please refer [here](https://www.intel.com/content/www/us/en/developer/articles/technical/lower-numerical-precision-deep-learning-inference-and-training.html).
 
 To fix this issue inside NNCF, by default, all weight tensors are quantized in 8 bits but only 7 bits are effectively used.
 This regime is used when `"target_device": "CPU"` or `"target_device": "ANY"` set. This fix, potentially, requires longer fine-tuning.
 
-To control the application of saturation fix, `"saturation_fix"` config option is introduced. The default value is `"saturation_fix": "enable"`. To apply the saturation issue fix only to the first layer, use `"saturation_fix": "enable_for_first_conv_layer"`. To disable the saturation issue fix for all layers, use `"saturation_fix": "disable"`.
+To control the application of overflow fix, `"overflow_fix"` config option is introduced. The default value is `"overflow_fix": "enable"`. To apply the overflow issue fix only to the first layer, use `"overflow_fix": "first_layer_only"`. To disable the overflow issue fix for all layers, use `"overflow_fix": "disable"`.
 
 ---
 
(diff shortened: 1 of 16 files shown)


alexsu52 (commented):
I would like to propose another option how to expose the saturation fix option. The saturation fix option can be exposed as a choice option:
"saturation_fix": choices=["enable", "disable", "enable_for_first_conv_layer"]

@negvet @AlexKoff88 @MaximProshin @vshampor what do you think?

andrey-churkin:
> I would like to propose another option how to expose the saturation fix option. The saturation fix option can be exposed as a choice option: "saturation_fix": choices=["enable", "disable", "enable_for_first_conv_layer"]
> 
> @negvet @AlexKoff88 @MaximProshin @vshampor @andrey-churkin what do you think?

Looks good for me. I think it is better not to allow the increasing number of parameters in the config. This will avoid weird situations such as
```json
{
    "disable_saturation_fix": true,
    "apply_saturation_fix_only_to_first_layer": true
}
```

nncf/tensorflow/quantization/algorithm.py
```diff
@@ -464,8 +469,7 @@ def _get_quantizer_setup(self, model: tf.keras.Model) -> TFQuantizationSetup:
                         weight_def.weight_attr_name)
                     self._op_names.append(op_name)
 
-                    half_range = self._get_half_range(qconfig)
-                    applied_saturation_fix = applied_saturation_fix or half_range
+                    half_range, applied_saturation_fix = self._get_half_range(qconfig, applied_saturation_fix)
```
alexsu52, line 472: I'm not sure, but it seems that the quantizer setup is not required to be stored in the order of traversal of the graph. Could you propose how this can be done more reliably? cc' @vshampor

alexsu52 (changes requested):

nncf/tensorflow/quantization/algorithm.py
negvet: Fixed, thanks!

nncf/config/schema.py
```diff
@@ -543,14 +543,14 @@ def with_attributes(schema: Dict, **kwargs) -> Dict:
...
-                                                              "the fix will be applied. For a detailed information "
-                                                              ", please, take look at the docs"),
+        "saturation_fix": with_attributes(_STRING,
+                                          description="Option controls whether to apply the saturation "
+                                                      "issue fix for the appropriate NNCF config or not. "
+                                                      "If set to 'disable', the fix will not be applied. "
+                                                      "If set to 'enable' or 'enable_for_first_conv_layer', "
+                                                      "while appropriate target_devices are chosen, "
+                                                      "the fix will be applied to the all layers or to the first"
+                                                      "convolutional layer respectively."),
```
vshampor, line 553: Should document the default value.

nncf/torch/quantization/algo.py
```diff
@@ -1025,13 +1026,21 @@ def is_weights(ip: PTTargetPoint) -> bool:
...
+                                        'Now all weight quantizers will effectively use only 7 bits out of 8 bits. '
+                                        'This resolves the saturation issue problem on AVX2 and AVX-512 machines. '
+                                        'Please take a look at the documentation for a detailed information.')
+                elif self._saturation_fix == 'enable_for_first_conv_layer':
+                    if target_node in get_first_nodes_of_type(target_model_graph, ['conv2d']):
+                        half_range = True
+                        nncf_logger.warning('The saturation issue fix will be applied. Now first convolution '
+                                            'weight quantizers will effectively use only 7 bits out of 8 bits. '
+                                            'This resolves the saturation issue problem on AVX2 and AVX-512 machines. '
+                                            'Please take a look at the documentation for a detailed information.')
```
vshampor, line 1043: The texts are duplicated here and above. Could you extract the majority of text as a constant string and format it accordingly in-place?

vshampor (changes requested):

nncf/config/schema.py
negvet: Fixed, thanks!

nncf/torch/quantization/algo.py
negvet: Fixed, thanks!

negvet:
Jenkins please retry a build

vshampor (approved):

docs/compression_algorithms/Quantization.md
```diff
@@ -131,8 +131,10 @@ There is a known issue with AVX2 and AVX512 CPU devices. The issue appears with
 AVX2 and AVX512 utilize a 16-bit register to store the result of operations on tensors. In case when tensors are saturated the buffer overflow happens.
 This leads to accuracy degradation.
 
-To fix this issue inside NNCF, weight tensors are quantized in 8 bits but only 7 bits are effectively used.
-This regime is used when `"target_device": "CPU"` or `"target_device": "ANY"` set.
+To fix this issue inside NNCF, by default, all weight tensors are quantized in 8 bits but only 7 bits are effectively used.
+This regime is used when `"target_device": "CPU"` or `"target_device": "ANY"` set. This fix, potentially, requires longer fine-tuning.
+
+To control the application of saturation fix, `"saturation_fix"` config option is introduced. The default value is `"saturation_fix": "enable"`. To apply the saturation issue fix only to the first layer, use `"saturation_fix": "enable_for_first_conv_layer"`. To disable the saturation issue fix for all layers, use `"saturation_fix": "disable"`.
```
AlexKoff88, line 137: Several comments. In non-OV literature this problem is called overflow. I noticed that people could get it more easily. I think we could also provide a link to where users can read about this problem in details, e.g. [here ](https://www.intel.com/content/www/us/en/developer/articles/technical/lower-numerical-precision-deep-learning-inference-and-training.html).

Given that, I suggest rename it to overflow_fix. I also suggest renaming the second option to 
"first_layer_only".

AlexKoff88 (approved):

alexsu52 (approved):

PR #1040, opened 2021-12-13
Execute buffer-examining operations without ONNX tracing

vshampor:
### Changes

`is_enabled_quantization`, `signed_tensor` and other ops that get a constant value from a buffer now execute outside ONNX JIT trace, if currently tracing.

### Reason for changes

Allows to suppress warnings frequently seen during export:
```
/home/alexsu/work/projects/algo/nncf_tf/source/nncf-tf/github/nncf/nncf/torch/quantization/layers.py:148: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if not self.is_enabled_quantization():
```
, improving UX.

### Related tickets

N/A

### Tests

Manual export runs using samples.

### DIFF
--- a/nncf/torch/layers.py
+++ b/nncf/torch/layers.py
@@ -23,6 +23,7 @@ from torch.nn import init
 from torch.nn.utils.rnn import PackedSequence
 
 from nncf.torch.dynamic_graph.context import forward_nncf_trace
+from nncf.torch.utils import no_jit_trace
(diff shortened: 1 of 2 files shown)


alexsu52 (approved):

ljaljushkin (approved):

vshampor:
Jenkins please retry a build

PR #1053, opened 2021-12-22
Saturation issue onnx export quantized weights

SKholkin:
### Changes

Quantized weights are being exported to onnx **only** for half range case (saturation-overflow issue)

### Reason for changes

To avoid any quantization weights values shifts after half range to full range related changes. Also in general exporting quantized weights should lead to more stable export procedure.

### Related tickets

55602

### Tests

tests.torch.quantization.test_overflow_issue_export.test_overflow_quantization_export_with_middle_quants
with symmetric/asymmetric parametrization
Verifies is model with middle quants (very sensitive to small numerical changes) exports to onnx correctly with quantized weights. In particular onnx export is being tested by loading saved to *.onnx file model
Algo:
- create a simple model
- calculate middle quant points
- replace weights by middle quant point
- replace quantizer scale by middle quants related one
- export model to onnx file
- load model from onnx file
- compare reference weights with actual weights


tests/torch/quantization/test_overflow_issue_export.py
```diff
@@ -433,3 +434,77 @@ def test_is_overflow_fix_applied_model_resumed_correctly(tmp_path):
...
+        quantizer.input_range = nn.Parameter(torch.from_numpy(input_range.astype(np.single)))
+        return input_low, quant_len, levels
+
+    weights_size = list(model[0].weight.size())
+    compressed_model, compression_ctrl = create_compressed_model_and_algo_for_test(model, config)
+    compressed_model.eval()
+    weight_quantizer = list(compression_ctrl.weight_quantizers.values())[0].quantizer_module_ref
+
+    if quantization_mode == "symmetric":
+        input_low, quant_len, levels = set_scale_to_sym_quantizer_and_get_attrs(weight_quantizer, scale=1.)
```
ljaljushkin, line 484: please lightweight the test body
- move nested functions outside
- remove if for symmetric by parametrizing test case by type and set_scale_fn
- reuse get_config_for_export_mode
- extract ref weights generation to an outside function
- think about other improvements to make it more readable and simple

ljaljushkin (changes requested):

tests/torch/quantization/test_overflow_issue_export.py
SKholkin: Done, take a look at the changed test

tests/torch/quantization/test_overflow_issue_export.py
```diff
@@ -1,3 +1,4 @@
+from typing import List
```
daniil-lyakhov, line 1:
```suggestion
from typing import List

```

tests/torch/quantization/test_overflow_issue_export.py
```diff
@@ -433,3 +401,75 @@ def test_is_overflow_fix_applied_model_resumed_correctly(tmp_path):
         model, nncf_config, compression_state=compression_state)
     load_state(compressed_model, model_state_dict, is_resume=True)
     are_symmetric_fq_nodes_are_exported_correct_with_overflow_fix(tmp_path, compression_ctrl)
+
+
+def set_parameters_to_quantizer_and_get_attrs(quantizer, paramaters_to_set):
```
vshampor, line 406: Type hints missing, fix the typo with the second arg name

tests/torch/quantization/test_overflow_issue_export.py
```diff
@@ -433,3 +401,75 @@ def test_is_overflow_fix_applied_model_resumed_correctly(tmp_path):
...
+    if isinstance(quantizer, SymmetricQuantizer):
+        return set_scale_to_sym_quantizer_and_get_attrs(quantizer, **paramaters_to_set)
+    return set_input_low_and_input_range_to_asym_quantizer_and_get_attrs(quantizer, **paramaters_to_set)
+
+
+def set_scale_to_sym_quantizer_and_get_attrs(quantizer: SymmetricQuantizer, scale: float):
```
vshampor, line 412: Return type hints missing, same in all helper functions below

vshampor (changes requested):
Also need to discuss whether we really want to export quantized weights only in the half-range case, or for consistency/simplicity it is better to always export quantized weights.

tests/torch/quantization/test_overflow_issue_export.py
vshampor: Still no return type hints

vshampor (changes requested):

tests/torch/quantization/test_onnx_export.py
```diff
@@ -193,3 +196,26 @@ def test_target_compression_idx(tmp_path):
...
+def test_are_quantized_weights_exported_correct(tmp_path):
+    model = TwoConvTestModel()
+    nncf_config = get_config_for_export_mode(False)
+    nncf_config.update({"input_info": {
+        "sample_size": [1, 1, 20, 20]
+    }})
+    fill_params_of_model_by_normal(model)
+    _, compression_ctrl = create_compressed_model_and_algo_for_test(model, nncf_config)
+
+    quantizers = compression_ctrl.weight_quantizers.values()
```
daniil-lyakhov, line 210: Missing the part where quantized weights are filling by middle quants value. Could you please point out, where it happens

daniil-lyakhov: I see, it's in `test_overflow_issue_export.py`, but then I don't understand a point of this test. Is middle quant initialization isn't enough? It would be clearer for me if middle quant initialization was an option in such export test, so all kinds of an export test was in one place

SKholkin: Tests from `test_overflow_issue_export.py` and `test_onnx_export.onnx` have been squeezed into one

SKholkin:
Ready for review

nncf/torch/quantization/algo.py
```diff
@@ -1062,6 +1062,7 @@ def is_weights(ip: PTTargetPoint) -> bool:
                                             scale_shape=tuple(scale_shape),
                                             logarithm_scale=use_logarithm_scale,
                                             half_range=half_range,
+                                            export_quantized=is_weights(primary_ip),
```
ljaljushkin, line 1065: maybe `is_quantized_on_export` ? sounds more clear, but still requires some documentation. because even with this PR it was not obvious why this argument is needed.

ljaljushkin (changes requested):

tests/torch/quantization/test_unified_scales.py
```diff
@@ -781,6 +780,9 @@ def test_weight_and_act_quantizer_scale_unification(self, tmp_path):
 
         unified_fq_node_inputs = [resolve_constant_node_inputs_to_values(fq_node, onnx_model.graph)
                                   for fq_node in fq_nodes_with_expected_unified_scales]
+        # delete weights from input dict
+        for item in unified_fq_node_inputs:
+            item.pop(min(item.keys()))
```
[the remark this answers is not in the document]
SKholkin: Since quantization on export transforms weights from `Initializer `to `Constant `type of input to onnx node they are incorrectly (w.r.t. test logic which compares quantizer scales) being found by `resolve_constant_node_inputs_to_values` along with quantizer scales. So I decided to manually remove weights from constant node inputs dict to keep test running. Weight inputs to onnx nodes have the lowest ID of all other inputs, `min(item.keys())` gives ID of weighs input and allows to remove weights by ID, so they don't ruin test logic.

nncf/torch/quantization/algo.py
SKholkin: Name of param changed to `is_quantized_on_export` and docs were added to `PTQuantizerSpec`

daniil-lyakhov (approved):

AlexKoff88:
Can anybody confirm that this is accepted by MO at least?

SKholkin:
@AlexKoff88 
As nncf output we have onnx with fakequantized FP weights regardless of middle quant/not middle quants case. 
Logic of converting onnx with fakequantized weights model through MO and inferencing by accuracy checker is being checked by e2e tests https://adas-ci-icv.inn.intel.com/job/NNCF/job/nightly/job/TriggerBetta/696/ 
Logic of middle quants fakequantized onnx export is being checked by `tests.torch.quantization.test_onnx_export.test_export_quantized_weights_with_middle_quants` 
Checking combination of middle quants export to onnx and then converting it to MO doesn't have much sense because MO will just quantize already quantized onnx weights just like in case of regular model (again this logic is checked)
Also testing middle quants model with MO means that we are more trying to test MO quantization logic what we shouldn't do
Do you still need MO conversion to be checked specifically for middle quants case?

ljaljushkin (approved):

PR #1058, opened 2022-01-02
Fix weight norm applied to modules

vshampor:
### Changes

Upon construction (i.e. during module wrapping process), NNCF* modules now reapply the torch.nn.utils.weight_norm.WeightNorm pre-hook if it has been applied to the non-NNCF wrapped module.

### Reason for changes

The WeightNorm hook replaces the existing "weight" torch.nn.Parameter of the wrapped module with a raw Tensor of the same name and instead creates two internal torch.nn.Parameters from which the original "weight" attribute is then recomputed. In order for these changes to be propagated into the NNCF-wrapped module, the hook application must be done explicitly.

### Related tickets

N/A

### Tests
Added tests.torch.test_nncf_network.test_weight_normed_modules_are_replaced_correctly

### DIFF
--- a/nncf/torch/layers.py
+++ b/nncf/torch/layers.py
@@ -10,6 +10,8 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
+from typing import Dict
+
 import math
 import numbers
 from typing import Optional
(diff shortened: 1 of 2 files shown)


vshampor:
Jenkins please retry a build

PR #1065, opened 2022-01-11
Freeze pandas version to exclude the broken 1.4.0rc0 and higher

vshampor:
### Changes

As stated in the title

### Reason for changes

pip pulls the latest RC version (1.4.0rc0) which, surprisingly, does not even appear on pypi. The version actually breaks the installation - pandas cannot find the numpy module.

### Related tickets

75354

### Tests

Install tests should be working now.


vshampor:
Jenkins please retry a build

PR #1067, opened 2022-01-12
Fix bug in test_sanity_AA

kshpv:
### Changes

Change tmp_path location for each test

### Reason for changes

Fix bug

### Tests

No

### DIFF
--- a/tests/torch/test_sanity_sample.py
+++ b/tests/torch/test_sanity_sample.py
@@ -860,13 +860,14 @@ def accuracy_aware_config(request):
     ids=['distributed', 'dataparallel'])
 def test_accuracy_aware_training_pipeline(accuracy_aware_config, tmp_path, multiprocessing_distributed):
     config_factory = ConfigFactory(accuracy_aware_config['nncf_config'], tmp_path / 'config.json')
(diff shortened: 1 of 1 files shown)


tests/torch/test_sanity_sample.py
```diff
@@ -860,7 +860,8 @@ def accuracy_aware_config(request):
     ids=['distributed', 'dataparallel'])
 def test_accuracy_aware_training_pipeline(accuracy_aware_config, tmp_path, multiprocessing_distributed):
     config_factory = ConfigFactory(accuracy_aware_config['nncf_config'], tmp_path / 'config.json')
-
+    tmp_path = tmp_path / 'accuracy_aware'
```
ljaljushkin, line 863: IMHO, it's not a good practice to override input argument of the function, moreover fixture one.
why not to create a local variable for it?

ljaljushkin (changes requested):

tests/torch/test_sanity_sample.py
kshpv: Sure

ljaljushkin (approved):

PR #1070, opened 2022-01-13
Introduce ONNX GraphConverter to get NNCFGraph from ONNX

kshpv:
### Changes

As stated in the title

### Reason for changes

Implement ONNX Post-Training Quantization 

### Related tickets

75422

### Tests

Add test comparing NNCFGraph built for two ONNX synthetic models.


kshpv:
I would like to add tests on reference graphs of the real models (I suppose resnet-50 and mobilenetv2). To achieve that I have to store somewhere ONNX files of those models.
At the same time, I understand that storing the models is inefficient (the size is around several tens of MB).

In that case, it is possible to do precommit validation only on synthetic models, which are super simple, while in the E2E validation use real ones.

alexsu52 (changes requested):
I guess, we can treat onnx like a separate backend and in the future, we will install nncf with onnx support via `pip install nncf[onnx]`. You should create requirements.txt for onnx backend in the `nncf/experimental/onnx` at least. Perhaps a better solution would be to update `setup.py` in this PR. @vshampor, @AlexKoff88 What do you think?

Please, provide also the build with passed onnx tests.

kshpv:
> I guess, we can treat onnx like a separate backend and in the future, we will install nncf with onnx support via `pip install nncf[onnx]`. You should create requirements.txt for onnx backend in the `nncf/experimental/onnx` at least. Perhaps a better solution would be to update `setup.py` in this PR. @vshampor, @AlexKoff88 What do you think?
> 
> Please, provide also the build with passed onnx tests.

Currently, there is no build and all tests were passed locally. The task of creating CI precommit validation job for ONNX backend was created - ticket 75956.

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -0,0 +1,77 @@
+import onnx
+
+
+# pylint: disable=no-member
+
+class ONNXGraph:
```
vshampor, line 6: Make sure that every public method has a docstring and that every method has both argument and return type hints.

vshampor (changes requested):

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -0,0 +1,32 @@
...
+import pytest
+
+from nncf.experimental.onnx.graph.nncf_graph_builder import GraphConverter
+
+from tests.onnx.models import LinearModel
+from tests.onnx.models import MultiInputOutputModel
+
+import networkx as nx
+
+TEST_MODELS = [LinearModel(), MultiInputOutputModel()]
```
ljaljushkin, line 11: I'd use model_creator function, instead of instantiated model itself. 
Currently, all tests in the pre-commit scope will fail on test collection if there's error on creation of some model. 
With model_creator tests will be collected successfully and only one test case will fail, all others won't be affected.

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.parametrize("model", TEST_MODELS)
+def test_built_nncf_graphs(model):
```
ljaljushkin, line 15: looks similar to `check_nx_graph`. maybe alignment would be better than a new function with some duplication of code

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -0,0 +1,77 @@
+import onnx
+
+
+# pylint: disable=no-member
```
ljaljushkin, line 4: isn't it too much for the whole file?
the same for nncf_graph_builder.py
I'd add this for the particular place in the code.

ljaljushkin (changes requested):

tests/onnx/test_nncf_graph_builder.py
kshpv: Good remark!

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: The problem is that pylint abuses almost any getting attribute of onnx module.
Because there are many places where I  address such attributes I prefer to disable the pylint rule at the begging of the file.

tests/onnx/test_nncf_graph_builder.py
kshpv: I prefer to not import functions from other frameworks (TF or PT)

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -0,0 +1,32 @@
...
+    nncf_graph = GraphConverter.create_nncf_graph(model.onnx_model)
+
+    nx_graph = nncf_graph.get_graph_for_structure_analysis()
+
+    data_dir = os.path.join(os.path.dirname(__file__), 'data/reference_graphs')
```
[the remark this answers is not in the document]
kshpv: Done

tests/onnx/test_nncf_graph_builder.py
kshpv: Implement this idea

nncf/experimental/onnx/graph/metatypes/onnx_ops.py
```diff
@@ -0,0 +1,109 @@
+from typing import List
```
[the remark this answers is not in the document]
kshpv: Added

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Done

tests/onnx/test_nncf_graph_builder.py
ljaljushkin: you can create common utils, if similar functionality is used cross frameworks

kshpv: They are not the same

kshpv:
@alexsu52 I suggest skipping a part of adding official support of ONNX backend as it placed it currently to experimental. I suppose that adding info to docs + extending setup.py - is the latest step. 
Currently, I can add requirments.txt in `nncf/experimental/onnx` folder.

kshpv:
> I guess, we can treat onnx like a separate backend and in the future, we will install nncf with onnx support via `pip install nncf[onnx]`. You should create requirements.txt for onnx backend in the `nncf/experimental/onnx` at least. Perhaps a better solution would be to update `setup.py` in this PR. @vshampor, @AlexKoff88 What do you think?
> 
> Please, provide also the build with passed onnx tests.

@alexsu52 I suggest skipping a part of adding official support of ONNX backend as it placed it currently to experimental. I suppose that adding info to docs + extending setup.py - is the latest step.
Currently, I can add requirments.txt in nncf/experimental/onnx folder.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,136 @@
...
+        for output_node in nncf_graph.get_all_nodes():
+            output_node_id = output_node.node_id
+            outputs = onnx_graph.get_all_node_outputs(output_node.node_name)
+            for output in outputs:
+                nodes = onnx_graph.get_nodes_by_input(output)
+                shape = onnx_graph.get_node_output_shape(output)
+                for in_node in nodes:
+                    in_node_id = nncf_graph.get_node_by_name(in_node.name).node_id
+                    input_counter[in_node_id] = input_counter.get(in_node_id, -1) + 1
+                    output_counter[output_node_id] = input_counter.get(output_node_id, -1) + 1
```
vshampor, line 59: Use `defaultdict(int)` for your counter dicts and you won't have to keep this awkward code.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,136 @@
...
+        # Add Input Nodes
+        for i, _input in enumerate(onnx_graph.get_all_model_inputs()):
+            tensor_type = _input.type.tensor_type
+            input_shape = []
+            if tensor_type.HasField("shape"):
+                for d in tensor_type.shape.dim:
+                    if d.HasField("dim_value"):
+                        input_shape.append(int(d.dim_value))
+                    else:
+                        raise RuntimeError('There is no integer value in model Input dimension')
```
vshampor, line 77: Not clear from the message why is it that the model input dimension should have an *integer* value exactly. Can there ever be a non-integer value?

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,136 @@
...
+            else:
+                raise RuntimeError('There is no shape in model Inputs')
+            input_node = nncf_graph.add_nncf_node(node_name='input_node_' + str(i),
```
vshampor, line 80: Please follow the naming from the other frameworks: `nncf_input_node_N`, `nncf_output_node_N`

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,136 @@
...
+                                                  get_operator_metatype_by_op_name(
+                                                      NNCFGraphNodeType.INPUT_NODE),
+                                                  layer_attributes=None)
+            input_name = _input.name
+            to_nodes = onnx_graph.get_nodes_by_input(input_name)
+            for node in to_nodes:
+                in_node_id = input_node.node_id
+                to_node_id = nncf_graph.get_node_by_name(node.name).node_id
+                input_counter[in_node_id] = input_counter.get(input_node.node_id, -1) + 1
+                output_counter[to_node_id] = input_counter.get(to_node_id, -1) + 1
```
vshampor, line 92: Same about defaultdicts

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,136 @@
...
+                nncf_graph.add_edge_between_nncf_nodes(
+                    from_node_id=input_node.node_id,
+                    to_node_id=to_node_id,
+                    tensor_shape=input_shape,
+                    input_port_id=input_counter[in_node_id],
+                    output_port_id=output_counter[to_node_id],
+                    dtype=Dtype.FLOAT
```
vshampor, line 99: What about paths in the graph that are integer, such as picking a dimension value from the result of the `ShapeOf` operation and putting it into a `Gather` operation downstream? Remember Mask-RCNN.

vshampor (changes requested):

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: As I know the shape value could be int or symbolic ("N') or even None values. In my logic I take only integer values, raising RuntimeError in any other cases.

tests/onnx/test_nncf_graph_builder.py
kshpv: Created the ticket - 76508

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: Please, take a look at the new version

nncf/experimental/onnx/graph/nncf_graph_builder.py
vshampor: You are casting the value to an int, not checking that the value is actually an integer. These are not the same things to do.

nncf/experimental/onnx/graph/nncf_graph_builder.py
vshampor: Please extend tests to cover the new logic.

kshpv: I added one model with `Shape `node which output is int

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: You are right! Updated message in the Exception

kshpv:
The precommit job with ONNX tests run on this PR - http://213.221.44.203/job/nncf/job/seed_sandbox/job/precommit/job/manual_NNCF/2/console

kshpv:
@alexsu52 @vshampor @ljaljushkin. The ball is on your side.

alexsu52 (approved):

alexsu52:
> @alexsu52 @vshampor @ljaljushkin. The ball is on your side.

Please fix pylint issues

ljaljushkin (changes requested):

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        return list(self.onnx_model.graph.input)
+
+    def get_all_model_outputs(self) -> List[ValueInfoProto]:
+        """
+        Returns all model outputs.
+        """
+        return list(self.onnx_model.graph.output)
+
+    def get_all_node_inputs(self, node_name: str) -> List[ValueInfoProto]:
```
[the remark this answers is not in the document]
kshpv: Good one! I suggest to change the duplicated function to a single function with the signature - 
`get_all_node_edges(self, node_name: str) -> Dict[str, List[ValueInfoProto]]`.
 This function returns a dictionary with two keys: input, output. The values under these keys correspond to the input and output edges of the node.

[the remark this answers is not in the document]
kshpv: I removed duplicated code. Now it looks much better.

kshpv:
the precommit job on the latest commit - http://213.221.44.203/job/nncf/job/seed_sandbox/job/precommit/job/manual_NNCF/3/

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -0,0 +1,156 @@
...
+        return output
+
+    def get_node_edges(self, node_name: str) -> Dict[str, List[ValueInfoProto]]:
+        """
+        Returns node input and output edges.
+        """
+        graph = self.onnx_model.graph
+        for node in graph.node:
+            if node.name == node_name:
+                return {'input': list(node.input),
```
ljaljushkin, line 85: would be more cool with dataclass instead of this dict, but AFAIK we still have low python version to use it

kshpv: Unfortunately, we support python3.6+, while dataclasses are supported by python3.7+. We don't have a choice here.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -0,0 +1,138 @@
...
+                    input_port_id=input_counter[out_node_id],
+                    output_port_id=output_counter[to_node_id],
+                    dtype=nncf_dtype
+                )
+
+        return nncf_graph
+
+    @staticmethod
+    def convert_onnx_dtype_to_nncf_dtype(onnx_dtype: str) -> Dtype:
+        conversation_map = {
```
vshampor, line 132: `conversion`

vshampor (approved):

PR #1074, opened 2022-01-20
Launch BN adaptation for quantization only for non-zero sample values

vshampor:
### Changes

NNCFConfig will no longer require to have a BN adaptation data loader registered if the number of BN adaptation samples is set to 0 in the quantization algorithm (and if the only required compression is quantization).

### Reason for changes

This makes for easier debug/reproducer runs of quantization algo in a scratch file or a temporary Python console.

### Related tickets

N/A

### Tests

tests.torch.quantization.test_algo_quantization.test_quantization_can_be_run_with_no_data_loaders_if_zero_init_samples

### DIFF
--- a/nncf/common/compression.py
+++ b/nncf/common/compression.py
@@ -274,5 +274,5 @@ class BaseCompressionAlgorithmBuilder(CompressionAlgorithmBuilder):
             (dict, list, tuple, str, int, float, True, False, None) that represents state of the object.
         """
 
-    def _parse_bn_adapt_params(self) -> Dict:
+    def _parse_bn_adapt_params(self) -> Optional[Dict]:
         return extract_bn_adaptation_init_params(self.config, self.name)
--- a/nncf/common/compression.py
+++ b/nncf/common/compression.py
@@ -25,6 +25,7 @@ from nncf.common.utils.registry import Registry
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import infer_backend_from_model
 from nncf.config.extractors import extract_algo_specific_config
+from nncf.config.extractors import extract_bn_adaptation_init_params
 
 ModelType = TypeVar('ModelType')
 
@@ -272,3 +273,6 @@ class BaseCompressionAlgorithmBuilder(CompressionAlgorithmBuilder):
         :return: Returns a dictionary with Python data structures
             (dict, list, tuple, str, int, float, True, False, None) that represents state of the object.
         """
+
+    def _parse_bn_adapt_params(self) -> Dict:
+        return extract_bn_adaptation_init_params(self.config, self.name)
--- a/nncf/config/extractors.py
+++ b/nncf/config/extractors.py
@@ -128,7 +128,7 @@ def extract_range_init_params(config: NNCFConfig) -> Optional[Dict[str, object]]
     return params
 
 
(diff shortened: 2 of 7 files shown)


ljaljushkin (approved):

alexsu52 (changes requested):
Could you elaborate on what would change for the user? I mean what behavior was and will come for the following scenarios:
- BN adaptation parameters in the config and number of BN adaptation samples > 0.
- BN adaptation parameters in the config and number of BN adaptation samples == 0.
- There are no BN adaptation parameters in the config.

vshampor:
> Could you elaborate on what would change for the user? I mean what behavior was and will come for the following scenarios:
> 
> * BN adaptation parameters in the config and number of BN adaptation samples > 0.
> * BN adaptation parameters in the config and number of BN adaptation samples == 0.
> * There are no BN adaptation parameters in the config.

Before: 
* BN adaptation parameters in the config and number of BN adaptation samples > 0 - fails if the BN adaptation data loader is not provided.
* BN adaptation parameters in the config and number of BN adaptation samples == 0 - fails if the BN adaptation data loader is not provided.
* There are no BN adaptation parameters in the config - fails if the BN adaptation data loader is not provided since a default non-zero value BN adaptation samples is used in this case.

After:
* BN adaptation parameters in the config and number of BN adaptation samples > 0 - same as in the "Before" section.
* BN adaptation parameters in the config and number of BN adaptation samples == 0 - does not fail if the BN adapation data loader is not provided, calls to BN adaptation do nothing.
* There are no BN adaptation parameters in the config - same as in the "Before" section.

alexsu52:
> > Could you elaborate on what would change for the user? I mean what behavior was and will come for the following scenarios:
> > 
> > * BN adaptation parameters in the config and number of BN adaptation samples > 0.
> > * BN adaptation parameters in the config and number of BN adaptation samples == 0.
> > * There are no BN adaptation parameters in the config.
> 
> Before:
> 
>     * BN adaptation parameters in the config and number of BN adaptation samples > 0 - fails if the BN adaptation data loader is not provided.
> 
>     * BN adaptation parameters in the config and number of BN adaptation samples == 0 - fails if the BN adaptation data loader is not provided.
> 
>     * There are no BN adaptation parameters in the config - fails if the BN adaptation data loader is not provided since a default non-zero value BN adaptation samples is used in this case.
> 
> 
> After:
> 
>     * BN adaptation parameters in the config and number of BN adaptation samples > 0 - same as in the "Before" section.
> 
>     * BN adaptation parameters in the config and number of BN adaptation samples == 0 - does not fail if the BN adapation data loader is not provided, calls to BN adaptation do nothing.
> 
>     * There are no BN adaptation parameters in the config - same as in the "Before" section.
Currently, the BN adaptation algorithm is a mandatory initialization step of the quantization pipeline. It looks like you changed this behavior in the third case. I'm not sure that the stated reason is enough to approve changes.

vshampor:
The BN adaptation algorithm application is default, not mandatory. The override mechanism here - setting a config parameter to 0 - is explicit and cannot be set by the user by mistake, only deliberately. If the user gives no mention of the BN adaptation algo, the behaviour with this PR is the same as before the PR - the algo will fail if no BN adaptation data loader is specified, and if it is specified, will run a default number of BN adaptation samples (1000, if I recall correctly).

PR #1084, opened 2022-01-24
`disable-compression` command line option

alexsu52:
### Changes

Simplify scenario of evaluation pre-trained model.  I suggest using the `disable-compression` option instead of removing the compression section from the config file.

### Reason for changes

Customer request

### Related tickets

N/A

### Tests

N/A


ljaljushkin:
should we add it to PT for alignment with TF?

alexsu52:
> should we add it to PT for alignment with TF?

Sure, If It will be approved. It can be done as a separate PR.

vshampor (approved):

negvet (approved):

PR #1085, opened 2022-01-24
Fix ONNX export for quantizers on branches

vshampor:
### Changes

A bug was fixed wherein the quantizers that apply to different branches in the model were exported as chained one after the other, instead of each affecting only its own branch.

### Reason for changes

Bugfix

### Related tickets

76733

### Tests

tests.torch.quantization.test_onnx_export.test_branching_fqs_are_not_chained


vshampor:
Depends on #1074

AlexanderDokuchaev (approved):

PR #1086, opened 2022-01-26
Completely removed input masks for mask propagation

ljaljushkin:
### Changes

Removed adding `input_masks` attribute to node.data for mask propagation algorithm
**Still keep it for deprecated export utils** - the attribute pops up in the `input_prune` methods.

### Reason for changes

Mask propagation doesn't actually use `input_masks` anymore, It keeps `output_mask` only and calculate input masks based on output masks of previous nodes. It was done to avoid duplication and errors related to inconsistent input and output masks.
Without this change mask propagation algorithm may look misleading in terms of using input_masks

### Tests

just should be no regression in current test scope, new tests are not needed

### DIFF
--- a/nncf/torch/pruning/utils.py
+++ b/nncf/torch/pruning/utils.py
@@ -49,13 +49,12 @@ def get_bn_for_conv_node_by_name(target_model: NNCFNetwork, conv_node_name: NNCF
 
 def init_output_masks_in_graph(graph: NNCFGraph, nodes: List):
     """
-    Initialize masks in groph for mask propagation algorithm
+    Initialize masks in graph for mask propagation algorithm
 
     :param graph: NNCFNetwork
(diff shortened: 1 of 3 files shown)


nncf/torch/pruning/utils.py
```diff
@@ -49,13 +49,12 @@ def get_bn_for_conv_node_by_name(target_model: NNCFNetwork, conv_node_name: NNCF
 
 def init_output_masks_in_graph(graph: NNCFGraph, nodes: List):
     """
-    Initialize masks in groph for mask propagation algorithm
```
daniil-lyakhov, line 52: 👍

daniil-lyakhov (approved):

alexsu52 (approved):

AlexanderDokuchaev (approved):

PR #1092, opened 2022-01-30
Introduce experimental Post-Training API with ONNX implementation

kshpv:
### Changes

Add experimental Post-Training API; Add implementation for ONNX

### Reason for changes

65600

### Related tickets

76191

### Tests

To be discussed


examples/experimental/onnx/post_training_quantization.py
```diff
@@ -0,0 +1,21 @@
...
+dataloader = create_train_dataloader(dataset_dir=dataset_dir, input_shape=input_shape)
+# Step 2: Initialize the engine for metric calculation and statistics collection.
+engine = ONNXEngine()
+
+# Step 3: Create a pipeline of compression algorithms.
+builder = CompressionBuilder()
+quantization = ONNXPostTrainingQuantization(DEFAULT, engine, dataloader)
+builder.add_algorithm(quantization)
+# Step 4: Execute the pipeline.
+compressed_model = builder.compress_model(fp32_model_onnx_path)
```
AlexKoff88, line 21: I would rename `compress_model` to `apply`. To tell the truth we don't compress anything here. 

I also didn't find the model serialization step.

examples/experimental/onnx/post_training_quantization.py
```diff
@@ -0,0 +1,21 @@
...
+input_shape = []
+
+# Step 1: Initialize the data loader.
+dataloader = create_train_dataloader(dataset_dir=dataset_dir, input_shape=input_shape)
+# Step 2: Initialize the engine for metric calculation and statistics collection.
+engine = ONNXEngine()
+
+# Step 3: Create a pipeline of compression algorithms.
+builder = CompressionBuilder()
+quantization = ONNXPostTrainingQuantization(DEFAULT, engine, dataloader)
```
AlexKoff88, line 18: One of the requirements for one tool is to make optimization algorithms sharable across the supported backends (ONNX, Torch, TF).

I would make `engine` the last argument and let it be the default value.

nncf/common/utils/priority_queue.py
```diff
@@ -0,0 +1,67 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+from typing import List
+
+
+class PriorityQueue:
```
andrey-churkin, line 16: For my understanding:

Why we can not use the priority queue from the [heapq](https://docs.python.org/3/library/heapq.html) module?

examples/experimental/onnx/post_training_quantization.py
kshpv: Great comment! It is the thing that I would like to discuss - **One of the requirements for one tool is to make optimization algorithms sharable across the supported backends (ONNX, Torch, TF).**
The current issue with that - is that implementing this needs a refusing of the inheritance of `PostTrainingQuantization ` from `BaseCompressionAlgorithmBuilder`. We should define whether we need to based on current NNCF architecture.

nncf/common/utils/priority_queue.py
kshpv: `heqpq` module provides only min-heap, while I would like to use max-heap. The way how to adopt min-heap to max-heap is quite unfriendly.  I was surprised that standard python doesn't provide such an option.

Anyway, after some time I would agree that using the adopted `heapq` module is better. But it is a point for discussion

examples/experimental/onnx/post_training_quantization.py
kshpv: the serialization step will be added in the next commits.

examples/experimental/onnx/dataloader.py
```diff
@@ -0,0 +1,33 @@
+from nncf.experimental.post_training_api.dataloader import DataLoader
```
AlexKoff88, line 1: Are you sure you committed the `post_training_api.py`?

examples/experimental/onnx/dataloader.py
```diff
@@ -0,0 +1,33 @@
...
+
+from torch.utils.data.dataloader import DataLoader as TorchDataLoader
```
AlexKoff88, line 3: This related to the sample and should be moved out from the core part of ONNX backend

examples/experimental/onnx/post_training_quantization.py
```diff
@@ -0,0 +1,21 @@
+from examples.experimental.onnx.dataloader import create_train_dataloader
```
AlexKoff88, line 1: This also related to the particular sample and should be moved accordingly.

nncf/experimental/post_training_api/engine.py
```diff
@@ -0,0 +1,23 @@
...
+class Engine(ABC):
+    """
+    The basic class aims to provide the interface to infer the model.
+    """
+
+    def set_model(self, model: ModelType) -> None:
+        self.model = model
+
+    @abstractmethod
+    def infer_model(self, input_data: List[TensorType]) -> List[TensorType]:
```
AlexKoff88, line 20: Passing a list of input data is quite dangerous in terms of a potential out-of-memory error.

nncf/common/utils/priority_queue.py
andrey-churkin: I have one more question. Could you please explain why you need to use the priority queue? Why you can not use sorting by priority of algorithm inside the `apply()` method?

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,58 @@
+import os
```
ljaljushkin, line 1: I'd put it in the `examples/experimental/onnx` folder
otherwise experimental nas sample for torch will be mixed up with the onnx one

ljaljushkin (changes requested):

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+from nncf.experimental.post_training.quantization.algorithm import PostTrainingQuantization
+from nncf.experimental.post_training.utils import export
+
+from nncf.experimental.post_training.quantization.algorithm import PostTrainingQuantizationParameters
+from nncf.experimental.post_training.initialization.algorithm import InitializationAlgorithms
+from nncf.experimental.post_training.initialization.quantizer_range_finder import QuantizerRangeFinderParameters
+
+
+def create_train_dataset(dataset_dir):
+    image_size = [224, 224]
```
AlexKoff88, line 20: I think this should be revised using two possible options:
1. It can be moved to the constructor of DataLoader
2. It can be kept as is and some helper, like `from_torch_dataset`, is introduced instead of `MyDataLoader`

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+fp32_model = onnx.load(fp32_model_onnx_path)
+
+# Step 1: Wrap a model.
+compressed_model = CompressedModel(fp32_model)
+
+# Step 2: Initialize the data loader.
+train_dataset = create_train_dataset(dataset_dir)
+
+
+class MyDataLoader(DataLoader):
```
AlexKoff88, line 46: `MyDataLoader` should be replaced with something more solid, e.g. `ImageNetLoader` or `ImageNetDataLoader`

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+    transform = transforms.Compose([
+        transforms.Resize(size),
+        transforms.CenterCrop(image_size),
+        transforms.ToTensor(),
+        normalize,
+    ])
+    return torchvision.datasets.ImageFolder(os.path.join(dataset_dir, 'train'), transform)
+
+
+fp32_model_onnx_path = '/home/aleksei/nncf_work/onnx_quantization/mobilenet_v2.onnx'
```
AlexKoff88, line 33: If this is a sample can it look more mature, like receive and parse CLI parameters instead of hardcoding variables within the sample?

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+from nncf.experimental.post_training.compression_builder import CompressionBuilder
+from nncf.experimental.post_training.quantization.algorithm import PostTrainingQuantization
+from nncf.experimental.post_training.utils import export
+
+from nncf.experimental.post_training.quantization.algorithm import PostTrainingQuantizationParameters
+from nncf.experimental.post_training.initialization.algorithm import InitializationAlgorithms
+from nncf.experimental.post_training.initialization.quantizer_range_finder import QuantizerRangeFinderParameters
+
+
+def create_train_dataset(dataset_dir):
```
KodiaqQ, line 19: Not sure that we should define the method using the `train` in the naming. Post-training quantization doesn't require model re-training.

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+        return len(self.dataset)
+
+
+dataloader = MyDataLoader(train_dataset, batch_size=1, shuffle=True)
+
+# Step 3: Create a pipeline of compression algorithms.
+builder = CompressionBuilder()
+# Step 4: Create a quantiztion algorithm.
+quantization_parameters = PostTrainingQuantizationParameters(
+    iterations_number=10
```
KodiaqQ, line 64: What do `iterations` mean there? Is it a number of the samples for PTQ?

examples/experimental/onnx_ptq_classification.py
```diff
@@ -0,0 +1,71 @@
...
+)
+quantization = PostTrainingQuantization(quantization_parameters)
+builder.add_algorithm(quantization)
+# Step 4: Execute the pipeline.
+builder.apply(compressed_model, dataloader)
+# Step 5: Export the compressed model.
+export(compressed_model, int8_model_onnx_path)
```
AlexKoff88, line 71: It would be beneficial to have an accuracy validation at the end of the sample. This is the thing that users would like to see. The second is performance. But since it requires OpenVINO the benchmarking step can be disclosed in the corresponding README.

nncf/common/utils/priority_queue.py
AlexKoff88: Where does this code come from? Is it purely own implementation? Do we have tests for it?

nncf/experimental/onnx/graph/model_analyzer.py
```diff
@@ -0,0 +1,70 @@
...
+from nncf.common.quantization.structs import QuantizableWeightedLayerNode
+from nncf.common.quantization.structs import QuantizerConfig
+from nncf.common.insertion_point_graph import InsertionPointGraph
+
+from nncf.experimental.onnx.hardware.fused_patterns import ONNX_HW_FUSED_PATTERNS
+
+QUANTIZATION_LAYER_METATYPES = GENERAL_WEIGHT_LAYER_METATYPES
+
+
+class ONNXModelAnalyzer(ModelAnalyzer):
```
KodiaqQ, line 25: Why do we need this functionality?

nncf/experimental/post_training/quantization/algorithm.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class PostTrainingQuantizationParameters(PostTraniningAlgorithmParameters):
+    def __init__(self,
+                 preset: str = 'perfomance',
+                 iterations_number: int = 300,
+                 target_device: str = 'CPU',  # TODO: change to ENUM
+                 weight_bits: int = 8,
+                 activation_bits: int = 8,
+                 initialization_algorithms: List[InitializationAlgorithms] = None,
```
AlexKoff88, line 45: Why do we need to expose this to the user? `PostTrainingQuantizationAlgorithm` should uniquely define the set of steps to get accurate and performant models. This is our responsibility to define this step and not put this responsibility on the user. If at some point we come up with another algorithm (better or faster) we can make it default and deprecate this one or just make it as an alternative.

nncf/experimental/post_training/quantization/algorithm.py
```diff
@@ -0,0 +1,168 @@
...
+from nncf.experimental.post_training.initialization.algorithm import InitizalizationParameters
+from nncf.experimental.post_training.initialization.quantizer_range_finder import QuantizerRangeFinderParameters
+
+from nncf.experimental.post_training.utils import merge_two_dicts
+
+
+class PostTrainingQuantizationParameters(PostTraniningAlgorithmParameters):
+    def __init__(self,
+                 preset: str = 'perfomance',
+                 iterations_number: int = 300,
```
AlexKoff88, line 41: I think it is not about iterations but more about the number of samples used in the method. This also should correspond to the number of samples in the DataLoader (should be less or equal). `stat_subset_size` from POT is not the best name but we can call it like `num_samples`.

nncf/experimental/post_training/quantization/algorithm.py
```diff
@@ -0,0 +1,168 @@
...
+                 target_device: str = 'CPU',  # TODO: change to ENUM
+                 weight_bits: int = 8,
+                 activation_bits: int = 8,
+                 initialization_algorithms: List[InitializationAlgorithms] = None,
+                 initialization_algorithms_parameters: List[InitizalizationParameters] = None,
```
AlexKoff88, line 46: I would suggest taking parameters of DefaultQuantization from POT. See details [here](https://github.com/openvinotoolkit/openvino/blob/master/tools/pot/configs/default_quantization_spec.json). We can implement only basic things related to the initialization and aggregators. More advanced features can be implemented later.

andrey-churkin:
Could you provide a list of public methods/classes for users? It will be helpful to have such a list here to focus on it.

What I mean

**NNCF PTQ Public API**

Classes:
- `nncf.experimental.post_training.api.dataloader.DataLoader`
- ...

Methods:
- ...

I think it's even better to add this to the PR description.

kshpv:
The feature with the sample were merged in https://github.com/openvinotoolkit/nncf/pull/1143 and https://github.com/openvinotoolkit/nncf/pull/1123

PR #1096, opened 2022-02-03
Make NNCFNetwork deepcopy-able

vshampor:
### Changes
Fixed a return value for the internal NNCFNetwork function that populates one of the fields, and made the threading-related fields of `TracingContext` only instantiate when the context is entered.

### Reason for changes

The NNCFNetwork object is not deepcopy-able otherwise.

### Related tickets

76848

### Tests

test_deepcopy_nncf_network

### DIFF
--- a/nncf/torch/dynamic_graph/context.py
+++ b/nncf/torch/dynamic_graph/context.py
@@ -60,10 +60,11 @@ class TracingContext:
         self._pre_hooks = {}  # type: Dict[PreHookId, List[Callable]]
         self._num_nested_hooks = 0
 
-        self._thread_local = threading.local()
+        self._thread_local = None
+        self._cond = None
 
         self._n_instances_searching_graph = 0
-        self._cond = threading.Condition()
+
         self._is_tracing = True
         self._is_forwarding = False
         self._may_add_nodes = True
@@ -76,6 +77,7 @@ class TracingContext:
         global _CURRENT_CONTEXT
         self._save_context = _CURRENT_CONTEXT
         _CURRENT_CONTEXT = self
+        self._cond = threading.Condition()
         self._init_thread_local()
         if is_debug():
             self.reset_node_call_counters()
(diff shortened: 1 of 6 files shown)


nncf/torch/dynamic_graph/graph_tracer.py
```diff
@@ -98,8 +98,8 @@ def trace_graph(self, model: torch.nn.Module, context_to_use: Optional['TracingC
 
         context_to_use.enable_trace_dynamic_graph()
         from nncf.torch.utils import training_mode_switcher
-        context_to_use.base_module_thread_local_replica = model
         with context_to_use as _ctx:
+            _ctx.base_module_thread_local_replica = model
```
ljaljushkin, line 102: just for understanding, what was the main change fixing the deepcopy? I wonder whether all changes were required to fix it, or there is also some refactoring to handle thread_local replica more clear.

vshampor: My original fix seems to have been incorrect. I pushed another version, in which the main change fixing the deepcopy is separating the threading variables (which are non-deepcopyable) into a specific object with an overridden `__deepcopy__` method.

ljaljushkin: Does new test cover the case when previous implementation failed?

ljaljushkin:
Jenkins please retry a build

nncf/torch/dynamic_graph/context.py
```diff
@@ -319,18 +309,19 @@ def _init_thread_local(self):
         tl.operator_counters = {}
         tl.node_call_tracker = {}
 
+
```
alexsu52, line 312:
```suggestion
```

tests/torch/test_nncf_network.py
```diff
@@ -945,3 +945,10 @@ def get_ip_graph_for_test(nncf_graph: NNCFGraph,
                                    allowed_pre_hook_insertion_points=pre_hooks,
                                    allowed_post_hook_insertion_points=post_hooks)
     return ip_graph
+
```
alexsu52, line 948:
```suggestion


```

alexsu52 (approved):
Cool staff!

vshampor:
Jenkins please retry a build

nncf/torch/dynamic_graph/graph_tracer.py
vshampor: The failures of the previous implementation were caught by the rest of the tests.

ljaljushkin (approved):

Issue #1099, opened 2022-02-04
[Bug] input_low must be a CUDA tensor


dkurt:
**NNCF**: 2.1.0
**PyTorch**: 1.9.1+cu102
**Transformers**: 4.15.0

```
Traceback (most recent call last):
  File "run_audio_classification.py", line 372, in <module>
    main()
  File "run_audio_classification.py", line 346, in main
    train_result = trainer.train(resume_from_checkpoint=checkpoint)
  File "<string>", line 1364, in train
  File "<string>", line 1934, in training_step
  File "<string>", line 1969, in compute_loss
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/dynamic_graph/wrappers.py", line 156, in wrapped
    return module_call(self, *args, **kwargs)
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 1051, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/debug.py", line 69, in decorated
    retval = forward_func(self, *args, **kwargs)
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/nncf_network.py", line 247, in forward
    args, kwargs = self._wrap_inputs_fn(args, kwargs)
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/dynamic_graph/io_handling.py", line 106, in wrap_inputs
    bound_model_params.arguments[param_name] = nncf_model_input(bound_model_params.arguments[param_name])
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/dynamic_graph/wrappers.py", line 133, in wrapped
    result = ctx.execute_post_hooks(op_address, result)
  File "/home/dkurtaev/transformers/examples/pytorch/audio-classification/venv3/lib/python3.6/site-packages/nncf/torch/dynamic_graph/context.py", line 227, in execute_post_hooks
    outputs = hook(outputs)
[...]

PR #1101, opened 2022-02-04
Restore model device after ONNX export

dkurt:
When export ONNX every epoch, we need to restore it's device.

resolves https://github.com/openvinotoolkit/nncf/issues/1099

### DIFF
--- a/nncf/torch/exporter.py
+++ b/nncf/torch/exporter.py
@@ -20,7 +20,7 @@ from nncf.common.exporter import Exporter
 from nncf.torch.dynamic_graph.graph_tracer import create_dummy_forward_fn
 from nncf.torch.dynamic_graph.graph_tracer import create_mock_tensor
 from nncf.torch.nested_objects_traversal import objwalk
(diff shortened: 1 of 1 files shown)


nncf/torch/exporter.py
```diff
@@ -81,6 +81,7 @@ def _export_to_onnx(self, save_path: str) -> None:
 
         :param save_path: The path where the model will be saved.
         """
+        original_device = getattr(self._model, "device", None)
```
vshampor, line 84: Can we use `get_model_device` ([link](https://github.com/openvinotoolkit/nncf/blob/6b233273bcf6f8942d9570fa9ab3dbba89e7ec54/nncf/torch/utils.py#L347) instead?

vshampor (changes requested):

nncf/torch/exporter.py
dkurt: That works, thanks!

alexsu52 (approved):
LGTM

vshampor (approved):

PR #1110, opened 2022-02-15
New FakeQuantize kernel for PT

SKholkin:
### Changes

Kernel function for PT FakeQuantization was changed to more precise around zero one

### Reason for changes

Old FQ function had quantized zero point with errors for some scale/input_low-input_range parameters
for example 
`>>>from nncf.torch.quantization.quantize_functions import symmetric_quantize`
`>>>symmetric_quantize(torch.Tensor([0.]).to(torch.device('cuda')), 256, -128, 127,torch.Tensor([10.]).to(torch.device('cuda')),1e-06)`
`#tensor([9.5367e-07], device='cuda:0')`

### Related tickets

59006

### Tests

tests.torch.quantization.test_functions.test_mapping_to_zero


vshampor:
Jenkins please retry a build

tests/torch/quantization/test_functions.py
```diff
@@ -367,3 +368,34 @@ def test_quantize_asymmetric_backward(self, _seed, input_size, bits, use_cuda, i
...
+
+
+@pytest.mark.parametrize('quantization_mode', ['symmetric', 'asymmetric'])
+@pytest.mark.parametrize('device', ['cuda', 'cpu'])
+def test_mapping_to_zero(quantization_mode, device):
+    x_zero = torch.zeros([1]).to(torch.device(device))
+    levels = 256
+    eps = 1e-6
+    number_of_samples = 100
+    from torch.distributions.uniform import Uniform
```
vshampor, line 380: You forgot to set up an exact seed for this test, making it indeterministic, which is not convenient.

vshampor (changes requested):

vshampor (approved):

PR #1112, opened 2022-02-16
Improvements of the quantizers code

andrey-churkin:
### Changes

- Redundant code was removed from the `SymmetricQuantizer.quantize()` and `AsymmetricQuantizer.quantize()` methods
- Implementation of the `get_quantizer_config()`, `get_config()`, `from_config()` methods was moved from the `SymmetricQuantizer` and `AsymmetricQuantizer` classes to the `Quantizer` superclass.

### Reason for changes

Removing the code duplication increases the quality of code. The code becomes more readable and maintainable.

### Related tickets

N/A

### Tests

One test was changed.
Reason: new `qspec` parameter was added to the `Quantizer.__init__()` method.

Additional tests are not required.

### DIFF
--- a/nncf/tensorflow/quantization/quantizers.py
+++ b/nncf/tensorflow/quantization/quantizers.py
@@ -336,32 +336,16 @@ class SymmetricQuantizer(Quantizer):
         self._half_range = False
 
     def quantize(self, inputs, weights, _):
-        def _half_range_quantize():
-            return symmetric_quantize(
-                inputs,
-                weights['scale_var'],
-                weights['signed_var'],
-                num_bits=self.num_bits - 1,
-                per_channel=self.per_channel,
-                narrow_range=self.narrow_range,
-                eps=self._eps
-            )
-
-        def _default_quantize():
-            return symmetric_quantize(
-                inputs,
-                weights['scale_var'],
-                weights['signed_var'],
-                num_bits=self.num_bits,
-                per_channel=self.per_channel,
-                narrow_range=self.narrow_range,
-                eps=self._eps
-            )
-
-        if self._half_range:
-            return _half_range_quantize()
-
-        return _default_quantize()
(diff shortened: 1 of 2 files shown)


nncf/tensorflow/quantization/quantizers.py
```diff
@@ -100,6 +113,26 @@ def mode(self) -> str:
         """
         raise NotImplementedError
 
+    @property
+    def signedness_to_force(self) -> Optional[bool]:
```
alexsu52, line 117: Asymmetric quantizer does not have signedness_to_force attribute.

alexsu52 (changes requested):

nncf/tensorflow/quantization/quantizers.py
andrey-churkin: Thanks for the comment! I understand your point of view.

*My suggestion:*

Since all quantizers have the instance of the `TFQuantizerSpec` class which describes the specification of the quantizer I suggest adding the `quantizer_spec` or `spec` property (I'm open to suggestions about the property name) to the `Quantizer` superclass.

After these changes, when we want to get, for example, the `num_bits` parameter, we use following
```
q.quantizer_spec.num_bits
```
or
```
q.spec.num_bits
```
I like the last option more. I remember, that main reason for these is providing one (common for `SymmetricQuantizer` and `AsymmetricQuantizer` classes) implementation of the `get_quantizer_config()`, `get_config()`, `from_config()` methods.

@alexsu52 What do you think about it?

alexsu52: @andrey-churkin As I understand you want to create common implementation of the `get_quantizer_config()`, `get_config()`, `from_config()`. Could you make a base implementation of these methods in the base class and overdrive it in descendants if it's needed.

PR #1115, opened 2022-02-17
New FQ formula docs PT

SKholkin:
### Changes

Updated docs for new Fake Quantize formula (for pytorch) https://github.com/openvinotoolkit/nncf/pull/1110

### Reason for changes

New FQ formula -> new docs

### Related tickets

59006

### Tests

_


daniil-lyakhov:
How about use `ZP` instead of `quant_zero` to be aligned with asymmetric quantization formula?

vshampor (approved):

daniil-lyakhov:
> > How about to use `ZP` instead of `quant_zero` to be aligned with asymmetric quantization formula?
> 
> `quant_zero` and `ZP` come from different algorithms (Common quantization and Range tuning for Asymmetric Quantization) and historically have different names, also `ZP` isn't calculated in symmetric quantization algo (though basically the same `quant_zero` is). But as we can see it can be disturbing that one entity have two different names in two different algos so I suggest not to squish these two names into `quant_zero` or `ZP` but to add a note that they actually refer to the same entity.
> 
> * Add a note in Asymmetric Quantization part that previously in Common Quantization part `ZP` is referred as `quant_zero`
>   or
> * Add a note in Common Quantization part that later in Asymmetric Quantization part `quant_zero` is refered as `ZP`
> 
> Or do you think that we still have to squish those two names into one (`quant_zero` or `ZP`)? (maybe in code too)

According to the new quantization formula, zero point is calculated in each quantization case (symmetric, asymmetric). If `ZP` and `quant_zero` mean the same why giving them separate names?

vshampor:
> Through conversation with @daniil-lyakhov we decided to rename all `quant_zero` occurrences to `ZP` (including kernels code) so there are no more confusion related to different names of the same entity. @vshampor, @AlexKoff88 are you ok with that?

In code, we should have `zero_point` instead of a blank `ZP`. Make an association of `zero_point` with `ZP` of the documentation within a comment line.

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

PR #1120, opened 2022-02-25
Build NNCF extensions in separate folders per torch version

vshampor:
### Changes

As stated in the title.

### Reason for changes

NNCF uses JIT to build its extensions upon first usage; these are binary-locked to the current CUDA and torch versions. If the torch version changes, the built extension shared object files become incompatible with the new torch and a cryptic error is spawned. Using per-torch version folders to build extensions will solve the problem, since if a torch version changes, then the target directory for the extension also changes and triggers a rebuild of the extensions automatically.

### Related tickets

80323

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -19,7 +19,8 @@ disable = arguments-differ,
           wrong-import-order,
           attribute-defined-outside-init,
           import-outside-toplevel,
-          duplicate-code
+          duplicate-code,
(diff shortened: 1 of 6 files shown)


ljaljushkin (approved):

alexsu52 (approved):

vshampor:
Jenkins please retry a build

PR #1128, opened 2022-03-05
Adding saturation fix parameters to compression state for PT

SKholkin:
### Changes

`PTQuantizerSetup` entity was added to replace `SingleConfigQuantizerSetup` in some logic related to getting quantizer setup in `QuantizationBuilder`.

`PTQuantizerSetup` holds `Dict[QuantizationPointId, PTQuantizationPoint]`
`PTQuantizationPoint` is quantization point based on `PTQuantizerSpec` and represents final information to create a quantizer

`PTQuantizerSetup` has replaced `SingleConfigQuantizerSetup` in `QuantizationBuilder` saving state hence `PTQuantizationPoint` is saved instead of `SingleConfigQuantizationPoint`. `PTQuantizationPoint` stores qspec with saturation fix parameters hence they are saved in compression state along with other qspec parameters

`QuantizerBuilderStateVersion` entity was added to support backward compatibility for `QuantizationBuilder`

### Reason for changes

Saturation fix parameters are now being stored in compression_state and hence are being defined only once

### Related tickets

58312

### Tests
tests/torch/test_backward_compat/test_comp_state_without_qspec


nncf/torch/quantization/algo.py
```diff
@@ -679,7 +720,7 @@ def _get_quantizer_setup(self, target_model: NNCFNetwork) -> SingleConfigQuantiz
                                                                   self._precision_init_params,
                                                                   self._range_init_params,
                                                                   self._debug_interface)
-        single_config_quantizer_setup = setup_generator.generate_setup()
+        single_config_quantizer_setup = setup_generator.generate_setup(self._overflow_fix, self._use_logarithm_scale_per_group)
```
ljaljushkin, line 723: It would be more clear, if setup generator knew all required parameters on initialization. Instead of passing on `generate_setup` method. I doubt an instance of the generator can be used multiple times with different parameters.

nncf/torch/quantization/algo.py
```diff
@@ -332,7 +333,36 @@ def __init__(self, quant_config: Dict, target_model: NNCFNetwork,
...
+            narrow_range = True
+        else:
+            use_logarithm_scale = use_logarithm_scale_per_group[QuantizerGroup.ACTIVATIONS]
+            narrow_range = False
+
+        if self.hw_config and qp.is_weight_quantization_point():
+            target_model_graph = self._target_model.get_original_graph()
+            target_node = target_model_graph.get_node_by_name(qp.insertion_point.target_node_name)
+            if self.hw_config.target_device in ['CPU', 'ANY'] and qconfig.num_bits == 8:
+                if is_overflow_fix == 'enable':
```
ljaljushkin, line 350: Avoid usage of strings for the choices, use enum or class

nncf/torch/quantization/layers.py
```diff
@@ -52,6 +52,45 @@ def from_str(config_value: str) -> 'HWConfigType':
...
+    def from_qconfig(cls, qconfig: QuantizerConfig, half_range, narrow_range, logarithm_scale):
+        return cls(qconfig.num_bits,
+                   qconfig.mode,
+                   qconfig.signedness_to_force,
+                   qconfig.per_channel,
+                   half_range,
+                   narrow_range,
+                   logarithm_scale)
+
+    def get_state(self) -> Dict[str, Any]:
```
ljaljushkin, line 78: if you use `get_state`, there should be `from_state` or `load_state` on loading compression state.

nncf/torch/quantization/algo.py
```diff
@@ -384,6 +421,10 @@ def str_or_list_to_list(list_or_str: Union[List[str], str]) -> List:
                                                             strict=self.hw_config is not None)
         finalized_quantizer_setup = prop_graph_solver.get_final_quantizer_setup(finalized_proposal)
         finalized_quantizer_setup = self._handle_quantize_inputs_option(finalized_quantizer_setup)
+
+        for i, (qp_id, qp) in enumerate(finalized_quantizer_setup.quantization_points.items()):
+            qp.qconfig = self._get_ptqconfig(qp, qp.qconfig, is_overflow_fix, use_logarithm_scale_per_group)
```
ljaljushkin, line 426: why not having ptqconfig in quantization point instead of intermediate qconfig?

nncf/torch/quantization/layers.py
```diff
@@ -52,6 +52,45 @@ def from_str(config_value: str) -> 'HWConfigType':
         raise RuntimeError("Unknown quantizer ONNX export mode string")
 
 
+class PTQuantizerConfig(QuantizerConfig):
```
ljaljushkin, line 55: It would be a third class to describe parameters of quantization.

Please consider using `PTQuantizerSpec`

SKholkin: Actually `PTQuantizerConfig` can be replaced by `PTQuantizerSpec` but then `PTQuantizerSpec` has to implement `QuantizerConfig` methods: `is_a_bitwidth_variant` and `compatible_with_a_unified_scale_linked_qconfig`. This would lead to large `PTQuantizerSpec` that could be used as a `QuantizerConfig` (and may be inherited from such).
Also `QuantizerConfig.is_a_bitwidth_variant` and `QuantizerConfig.compatible_with_a_unified_scale_linked_qconfig` are used only in PT case so they could be removed from `QuantizerConfig` and be placed in `PTQuantizerSpec`
In total `PTQuantizerSpec` would additionally have:

- `per_channel `attr
- `is_a_bitwidth_variant` method (this method will be removed from `QuantizerConfig`)
- `compatible_with_a_unified_scale_linked_qconfig` method (this method will be removed from `QuantizerConfig`)
- `get_state` method

`per_channel` attr is needed for `is_a_bitwidth_variant` and `compatible_with_a_unified_scale_linked_qconfig` methods
`get_state` is needed for saving to compression_state

@vshampor Do you think this kind of changes to `PTQuantizerSpec` and `QuantizerConfig` would be appropriate?

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

ljaljushkin:
please fix CI and remove `draft` status of the PR

nncf/torch/quantization/algo.py
```diff
@@ -1057,12 +1075,23 @@ def is_weights(ip: PTTargetPoint) -> bool:
                                         'Please take a look at the documentation for a detailed information.'
                                         .format(quantizers_with_overflow_fix_str))
 
+        matching_qspec_info = self._qspec_comp_state_parameters.get(str(primary_qid), None)
```
ljaljushkin, line 1078: None is returned by default for `get`

nncf/torch/quantization/layers.py
```diff
@@ -90,11 +90,40 @@ def from_config(cls, qconfig: QuantizerConfig, narrow_range: bool,
                    is_quantized_on_export,
                    compression_lr_multiplier)
 
+    def get_save_parameters(self):
+        return PTQuantizerSpecParameters(self).get_state()
```
ljaljushkin, line 94: why do you need a subset of params?
why not save the spec itself?

nncf/torch/quantization/algo.py
ljaljushkin: State is not intended for direct access like this by design.
State should be passed to `load_state` or `from_state` methods only that know how to parse this state. You're breaking encapsulation here.

nncf/torch/quantization/layers.py
```diff
@@ -136,6 +165,9 @@ def close(self):
 
         self.load_listener = LoadStateListener(self)
 
+    def get_qspec(self):
```
ljaljushkin, line 168:
```suggestion
    def get_quantizer_spec(self):
```

nncf/torch/quantization/algo.py
```diff
@@ -1057,12 +1075,23 @@ def is_weights(ip: PTTargetPoint) -> bool:
...
+        if matching_qspec_info:
+            if not matching_qspec_info['half_range'] == half_range or \
+                    not matching_qspec_info['narrow_range'] == narrow_range or \
+                    not matching_qspec_info['logarithm_scale'] == use_logarithm_scale or \
+                    not matching_qspec_info['is_quantized_on_export'] == is_quantized_on_export or \
+                    not matching_qspec_info['compression_lr_multiplier'] == compression_lr_multiplier:
+                nncf_logger.warning("Some quantization parameters (such as half_range) loaded from compression state "
+                                    "conflict with ones provided from config."
+                                    " Choosing parameters corresponding to config.")
```
ljaljushkin, line 1087: implement __eq__ for QuantizerSpec, restore if from state and simply call ==

ljaljushkin (changes requested):

nncf/torch/quantization/layers.py
SKholkin: Part of the info from  `PTQuantizerSpec` is stored in `SingleConfigQuantizerSetup` so there is no need to repeat it.
`PTQuantizerSpecParameters` is not sufficient to restore `PTQuantizerSpec` but joint info from  `SingleConfigQuantizerSetup` and `PTQuantizerSpecParameters` is. So I suggest to leave `PTQuantizerSpecParameters` as it is for saving to CompressionState purposes but to restore `PTQuantizerSpec` from joint info. What do you think?

Also `scale_shape` needs to be added to `PTQuantizerSpecParameters` to restore `PTQuantizerSpec`

ljaljushkin: don't like these dependencies for restoring different objects.

quantizer_spec should be restored from quantizer_spec's state and nothing else. 
if you need some extra parameters, keep and restore them outside of spec class.

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

SKholkin:
Jenkins please retry a build

nncf/torch/checkpoint_loading.py
```diff
@@ -18,7 +18,60 @@
...
 
 from nncf.common.utils.logger import logger as nncf_logger
-from nncf.torch.utils import maybe_convert_legacy_names_in_model_state
+from nncf.common.compression import BaseCompressionAlgorithmController
+
+
+class PTCompressionStateVersion:
+    SAVE_NAME = 'version'
+    CURR_VERSION = '1.1'
+    COMPRESSION_STATE_STATE_DICT = '0.0'
```
ljaljushkin, line 27: STATE_DICT_VERSION

nncf/torch/checkpoint_loading.py
```diff
@@ -18,7 +18,60 @@
...
+            return True
+        return False
+
+    def __ne__(self, other):
+        if self.version != other.version:
+            return True
+        return False
+
+    def __lt__(self, other):
+        if self.version < other.version:
```
ljaljushkin, line 52: No need to define all operators, define just eq and lt 
https://docs.python.org/3/library/functools.html#functools.total_ordering

nncf/torch/checkpoint_loading.py
```diff
@@ -18,7 +18,60 @@
...
+            return True
+        return False
+
+    def __ge__(self, other):
+        if self.version >= other.version:
+            return True
+        return False
+
+    def is_state_dict(self):
+        if self == PTCompressionStateVersion(PTCompressionStateVersion.COMPRESSION_STATE_STATE_DICT):
```
ljaljushkin, line 72: Version class should be very basic without auxiliary methods. do it outside of this class

ljaljushkin (changes requested):

nncf/torch/model_creation.py
```diff
@@ -85,20 +86,22 @@ def create_compressed_model(model: Module,
     as an object of NNCFNetwork."""
 
     set_debug_log_dir(config.get("log_dir", "."))
+    is_legacy_model_state_dict = False
+    compression_state_version = None
+    if compression_state is not None:
+        compression_state_version = PTCompressionStateVersion.from_compression_state(compression_state)
+        is_legacy_model_state_dict = compression_state_version.is_state_dict()
```
ljaljushkin, line 93: PTCompressionStateVersion.MODEL_STATE_DICT_ONLY

nncf/torch/quantization/algo.py
```diff
@@ -467,16 +474,19 @@ def __init__(self, config, should_init: bool = True):
...
-    def _load_state_without_name(self, state_without_name: Dict[str, Any]):
+    def _load_state_without_name(self, state_without_name: Dict[str, Any], version: PTCompressionStateVersion = None):
         """
         Initializes object from the state.
 
         :param state_without_name: Output of `get_state()` method.
         """
         quantizer_setup_state = state_without_name[self._state_names.QUANTIZER_SETUP]
-        self._single_config_quantizer_setup = SingleConfigQuantizerSetup.from_state(quantizer_setup_state)
+        if version == PTCompressionStateVersion('0.1'):
```
ljaljushkin, line 485: version == PTCompressionStateVersion.SINGLE_QUANTIZER_SETUP

ljaljushkin (changes requested):

nncf/torch/quantization/algo.py
```diff
@@ -467,16 +474,19 @@ def __init__(self, config, should_init: bool = True):
 
         self._overflow_fix = self._algo_config.get('overflow_fix', 'enable')
         self._device_for_callable_obj_creation = 'cpu'
-        self._single_config_quantizer_setup = None  # type: Optional[SingleConfigQuantizerSetup]
 
-    def _load_state_without_name(self, state_without_name: Dict[str, Any]):
+    def _load_state_without_name(self, state_without_name: Dict[str, Any], version: PTCompressionStateVersion = None):
```
ljaljushkin, line 478:
```suggestion
    def _load_state_without_name(self, state_without_name: Dict[str, Any], version: Optional[PTCompressionStateVersion] = None):
```

nncf/torch/composite_compression.py
```diff
@@ -107,6 +109,16 @@ def initialize(self, model: ModelType) -> None:
     def _get_transformation_layout(self, target_model: NNCFNetwork) -> PTTransformationLayout:
         pass  # Higher-level get_transformation_layout is overridden, no need to define this
 
+    def load_state(self, state: Dict[str, Any], version: PTCompressionStateVersion = None) -> None:
```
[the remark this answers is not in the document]
SKholkin: Because of inheritance order `PTCompositeCompressionAlgorithmBuilder` prioritize `CompositeCompressionAlgorithmBuilder` methods which declare `load_state` method without version, so explicit override is needed.
Actually we could change order of inheritance order since none of other `PTCompositeCompressionAlgorithmBuilder` methods have conflicting implementations (so `load_state` won't have to be overloaded) but this would lead to explicit calling of `CompositeCompressionAlgorithmBuilder.__init__` instead of `super().__init__`
Same for `PTCompositeCompressionAlgorithmController`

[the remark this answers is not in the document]
ljaljushkin: What about introducing version in the common part? It would also align compression state with TF and reduce the need for tricky overriding and order of inheritance

nncf/torch/compression_method_api.py
```diff
@@ -176,7 +194,17 @@ def _get_state_without_name(self) -> Dict[str, Any]:
...
+        """
+        Initializes object from the state.
+
+        :param state: Output of `get_state()` method.
+        :param version: Version of compression state
+        """
+        if self.name in state:
+            self._load_state_without_name(state[self.name], version)
+
+    def _load_state_without_name(self, state_without_name: Dict[str, Any], version: PTCompressionStateVersion = None):
```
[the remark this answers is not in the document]
SKholkin: `PTCompressionAlgorithmBuilder` inherit `BaseCompressionAlgorithmBuilder` which have different signature of `_load_state_without_name` (no version) so to support base class signature version has been added as Optional argument But actually version HAS to be provided for loading `PTCompressionAlgorithmBuilder._load_state_without_name` so:
- would it be alright to keep version as not optional, regular argument 
or
-  would it be better to leave it as optional and to raise an error inside `_load_state_without_name`?

SKholkin:
Global `PTCompressionStateVersion` for PT Compression State was replaced by local `QuantizerBuilderStateVersion` since version is needed only for `QuantizationBuilder`

nncf/torch/compression_method_api.py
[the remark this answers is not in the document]
SKholkin: Related changes has been done

ljaljushkin (approved):

vshampor (approved):

PR #1130, opened 2022-03-10
Fix autocast

vshampor:
### Changes

FQ weights are now being autocast to FP16 if the torch.amp.autocast is enabled. To resolve issues during autocast-enabled initialization process, the statistics are being invariably cast to FP32 during final FQ init (since within the autocast context the tensor stats for activations are in FP16).

### Reason for changes

NNCF PT compatibility with torch.amp.autocast

### Related tickets

79587

### Tests

TBA


nncf/torch/quantization/layers.py
```diff
@@ -186,19 +186,30 @@ def reset_call_counter(self):
...
+                          min_values: torch.Tensor,
+                          max_values: torch.Tensor,
+                          log_module_name: str = None):
         """min_values and max_values must have the same shape as specified in self.scale_shape"""
         if self.initialized:
             nncf_logger.debug("Skipped initializing {} - loaded from checkpoint".format(log_module_name))
             return
+
+        if torch.all(torch.isinf(min_values)) or torch.all(torch.isinf(max_values)):
+            raise AttributeError('Statistics are not collected for {}'.format(log_module_name))
```
andrey-churkin, line 199: I think it is better to use `ValueError` here and below according to error [description](https://docs.python.org/3/library/exceptions.html#ValueError).

nncf/torch/quantization/algo.py
```diff
@@ -1066,8 +1067,15 @@ def is_weights(ip: PTTargetPoint) -> bool:
                                             compression_lr_multiplier=compression_lr_multiplier)
         quantizer = self.__create_quantize_module(qspec).to(self._device_for_callable_obj_creation)
         if range_init_minmax_values is not None:
-            quantizer.apply_minmax_init(min_values=range_init_minmax_values[0],
-                                        max_values=range_init_minmax_values[1],
+            # Need to cast to the model's current dtype since the statistics could have been gathered in an
+            # AMP autocast model (and therefore be FP16 since AMP autocast switches precision of activations
+            # at forward pass time)
+            own_type = get_model_dtype(target_model)
```
alexsu52, line 1073: The `get_model_dtype` function returns the type of the first model parameter, am I right?

vshampor: Correct, this is basically the same way that we used to determine model device historically, only separated in a specific function.

nncf/torch/quantization/layers.py
vshampor: Makes sense, done.

nncf/torch/quantization/algo.py
alexsu52: It looks like, It can be used for device determination. But I'm not sure that we can use the same way to determine the type of all model operations. Let's discuss it offline.

PR #1131, opened 2022-03-10
Add option to set onnx opeset version

AlexanderDokuchaev:
Add option to change onnx opset version by `save_format` argument.

Example: 
- `onnx_11` to export with opset 11
- `onnx` to export with default opset version (10)

### DIFF
--- a/nncf/torch/exporter.py
+++ b/nncf/torch/exporter.py
@@ -12,11 +12,13 @@
 """
 from typing import Any
 from typing import Optional
+from typing import Tuple
 from functools import partial
 from copy import copy
 import torch
 
 from nncf.common.exporter import Exporter
+from nncf.common.utils.logger import logger as nncf_logger
 from nncf.torch.dynamic_graph.graph_tracer import create_dummy_forward_fn
 from nncf.torch.dynamic_graph.graph_tracer import create_mock_tensor
 from nncf.torch.nested_objects_traversal import objwalk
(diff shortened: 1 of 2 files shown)


tests/torch/test_onnx_export.py
```diff
@@ -48,3 +51,27 @@ def test_io_nodes_naming_scheme(tmp_path):
...
+                         (
+                             ('onnx', ('onnx', {'opset_version': PTExporter._ONNX_DEFAULT_OPSET})),
+                             ('onnx_9', ('onnx', {'opset_version': 9})),
+                             ('onnx_10', ('onnx', {'opset_version': 10})),
+                             ('onnx_11', ('onnx', {'opset_version': 11})),
+                             ('onnx_0', ValueError),
+                             ('onnx_onnx', ValueError),
+                         )
+                        )
+def test_exporter_parser_format(save_format: str, refs: Any):
```
vshampor, line 67: Might want to add another test, or a stage of the test - wherein a basic, mock model is being exported using the PTExporter and the resulting ONNX has the desired opset.

nncf/torch/exporter.py
```diff
@@ -48,6 +50,41 @@ class PTExporter(Exporter):
...
+            str: short form of the save_format
+            dict: additional arguments for exporter
+        """
+        if save_format.startswith(PTExporter._ONNX_FORMAT):
+            splited_format = save_format.split('_')
+            opset = None
+
+            if len(splited_format) == 1:
+                opset = PTExporter._ONNX_DEFAULT_OPSET
+            elif len(splited_format) == 2:
```
vshampor, line 73: `splited` -> `split`, the 3rd form for the "split" is the same as the 1st.

vshampor (changes requested):

AlexanderDokuchaev:
Reopened https://github.com/openvinotoolkit/nncf/pull/1133

PR #1149, opened 2022-03-24
Fix pruning compression stage check

evgeniya-egupova:
### Changes

Filter pruning algorithm now uses proper calculation of current pruning level to determine compression stage.
Affects SC.

### Reason for changes

Compression stage of the model pruned by flops was calculated incorrectly. 
Affects Adaptive Compression training as best checkpoints were not saved.

### Related tickets
103420
-

### Tests

-


daniil-lyakhov:
@alexsu52, do you mind to close this pr?

daniil-lyakhov:
Moved to #2330

PR #1151, opened 2022-03-29
Update public PTQ API

KodiaqQ:
### Changes

Update public Post-Training Quantization API (and ONNX PTQ API)

### Reason for changes

Extend public PTQ API

### Related tickets

76191

### Tests


nncf/experimental/onnx/engine.py
```diff
@@ -38,15 +40,26 @@ def set_model(self, model: str) -> None:
...
+            model_outputs = self.sess.get_outputs()
+            for i, model_output in enumerate(model_outputs):
+                if model_output.name not in output:
+                    output[model_output.name] = []
+                # TODO (Nikita Malinin): Add backend-specific statistics aggregator usage
+                output[model_output.name].append(output_tensor[i])
         return output
+
+    def compute_metrics(self, metrics_per_sample=False):
+        pass
```
AlexKoff88, line 65: Exception?

nncf/experimental/post_training/api/metric.py
```diff
@@ -0,0 +1,54 @@
+"""
```
AlexKoff88, line 1: This is not related to this PR specifically but I noticed that NNCF copyrights format is different from what OpenVINO uses now.

KodiaqQ: Should we schedule a global copyright update? Create a ticket?

nncf/experimental/post_training/api/engine.py
```diff
@@ -19,30 +19,76 @@
...
+        """ Processes model input data based on the backend
+            :param inputs: list of input data
+            :return inputs: list of the input data
         """
-        Infer the model on the provided input.
-        Returns the model outputs and corresponding node names in the model.
+        return inputs
+
+    def compute_statistics(self, statistics_layout: Dict) -> Dict[str, TensorType]:
+        """ Performs model inference on specified dataset subset for statistics collection and input-based layers layout
```
kshpv, line 83: Should check whether the model is set

nncf/experimental/post_training/api/engine.py
```diff
@@ -19,30 +19,76 @@
...
+    def transform_input(self, inputs: TensorType) -> TensorType:
+        """ Processes model input data based on the backend
+            :param inputs: list of input data
+            :return inputs: list of the input data
         """
-        Infer the model on the provided input.
-        Returns the model outputs and corresponding node names in the model.
+        return inputs
+
+    def compute_statistics(self, statistics_layout: Dict) -> Dict[str, TensorType]:
```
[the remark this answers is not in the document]
KodiaqQ: The main point of the methods division is in the UX: in the general pipeline, users shouldn't worry about the implementation of the Samplers or data-iteration loops to collect statistics or calculate metrics based on the raw model output data. 
This is why we provide two different methods (or helpers): 

1. `compute_statistics`, more internal, for statistics calculation because the tool knows which statistics are needed for the current model and with the current quantization settings, which passed should be called before calculation and etc.; 
2. `compute_metrics`, the second helper, for metrics calculation that (will be updated soon) integrates metrics object usage that can be simply re-implemented based on the user's pipeline.

cc @AlexKoff88

nncf/experimental/post_training/api/engine.py
KodiaqQ: This is verified in the `ONNXEngine` implementation. There are no ideas yet on which default validation implementation will not overload the API and worsen the UX.

nncf/experimental/onnx/engine.py
```diff
@@ -38,15 +40,26 @@ def set_model(self, model: str) -> None:
...
 
-    def _infer(self, _input: np.ndarray) -> Dict[str, np.ndarray]:
-        """
-        Runs InferenceSession on the provided '_input'.
-        Returns the model outputs and corresponding node names in the model.
-        """
+    def transform_input(self, inputs):
+        return inputs.astype(np.float32)
+
+    def compute_statistics(self, statistics_layout) -> Dict[str, np.ndarray]:
```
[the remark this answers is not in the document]
KodiaqQ: Done, thanks.

nncf/experimental/post_training/api/engine.py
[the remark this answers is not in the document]
kshpv: But in the current design, the user also doesn't worry about Samplers or data-iteration loops to collect statistics or calculate metrics based on the raw model output data. The algorithm does it for them.

nncf/experimental/post_training/api/engine.py
kshpv: I think that the checking of the model status is the first essential step for all Engines

nncf/experimental/post_training/api/engine.py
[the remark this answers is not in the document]
kshpv: I would add that Sampler is even not a public class.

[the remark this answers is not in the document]
KodiaqQ: > I would add that `Sampler` is even not a public class. I mean it is not included in public API

It became public in the current PR. See - nncf/experimental/post_training/api/sampler.py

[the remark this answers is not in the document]
KodiaqQ: > But in the current design, the user also doesn't worry about Samplers or data-iteration loops to collect statistics or calculate metrics based on the raw model output data. The algorithm does it for them.

And the current implementation still provides the same UX that before. In which scenarios user can use the raw output model data?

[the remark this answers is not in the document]
kshpv: I see. I wonder why do we make it public? In what cases `RandomBatchSampler` or `BatchSampler` are not enough?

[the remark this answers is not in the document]
kshpv: > > But in the current design, the user also doesn't worry about Samplers or data-iteration loops to collect statistics or calculate metrics based on the raw model output data. The algorithm does it for them.
> 
> And the current implementation still provides the same UX that before. In which scenarios user can use the raw output model data?

To calculate their own metrcis for example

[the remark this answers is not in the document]
KodiaqQ: > To calculate their own metrics for example

The current implementation will provide the method `calculate_metrics` that will cal helpers from the `Metic` class like the `update`. This method will get raw outputs and targets for metric calculation.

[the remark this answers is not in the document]
kshpv: > > To calculate their own metrics for example
> 
> The current implementation will provide the method `calculate_metrics` that will cal helpers from the `Metic` class like the `update`. This method will get raw outputs and targets for metric calculation.

I thing that actually the question is somehow incorrect.
The user basically doesn't work with `Engine`.
The user can only implement own Engine for algorithms usage, right?
And the problem that I see that to implement own Engine in new design getting harder as I should add realisation of more methods, which is not intuitive. And I fell in real life there is no user who will do it.

nncf/experimental/post_training/api/engine.py
KodiaqQ: > I think that the checking of the model status is the first essential step for all Engines

Any suggestions to do this without overloading the API and worsening the UX?

kshpv: The solution could be as it is done right now for `Engine`:

```
    def infer(self, _input: TensorType) -> Dict[str, TensorType]:
        if not self.is_model_set():
            raise RuntimeError('The {} tried to infer the model, while the model was not set.'.format(self.__class__))
        return self._infer(_input)

    @abstractmethod
    def _infer(self, _input: TensorType) -> Dict[str, TensorType]:
        """
        Infer the model on the provided input.
        Returns the model outputs and corresponding node names in the model.
        """
```

KodiaqQ: In the case of `Engine` re-implementation, the user should implement a private method?

kshpv: I see no problem here

nncf/experimental/post_training/api/engine.py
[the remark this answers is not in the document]
KodiaqQ: In the general case of re-implementation of their own `Engine`, users update only methods for the data pre- and post-processing.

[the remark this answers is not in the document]
kshpv: I see. I think that I don't see the main point of the PR:
1) What is the reason for updating the current API? I mean what problems/disadvantages do we see and want to improve.
2) How does this PR solve these problems?

nncf/experimental/onnx/statistics/aggregator.py
```diff
@@ -45,12 +43,10 @@ def collect_statistics(self, model: onnx.ModelProto) -> None:
...
             max_number_samples = max(max_number_samples, v.num_samples)
 
-        with tempfile.NamedTemporaryFile() as temporary_model:
-            onnx.save(model_with_intermediate_outputs, temporary_model.name)
-            self.engine.set_model(temporary_model.name)
-            self.engine.sampler = create_onnx_sampler(self.dataloader, range(max_number_samples))
-            output = self.engine.compute_statistics(self.layers_statistics)
-            self._agregate_statistics(output, self.layers_statistics)
+        self.engine.set_model(model_with_intermediate_outputs)
+        self.engine.sampler = create_onnx_sampler(self.dataloader, range(max_number_samples))
```
kshpv, line 47: Why should you set Sampler for Engine when it is created inside `compute_statistics()`?

examples/experimental/onnx/onnx_ptq_classification.py
```diff
@@ -87,13 +87,13 @@ def run(onnx_model_path: str, output_model_path: str,
...
     onnx.save(quantized_model, output_model_path)
     print(f"The quantized model is saved on {output_model_path}")
 
     onnx.checker.check_model(output_model_path)
 
-    # Step 7: (Optional) Validate the quantized model.
+    # Step 6: (Optional) Validate the quantized model.
     if evaluate:
         print("Validating of the quantized model has started")
         metrics = builder.evaluate(quantized_model, metric, dataloader)
```
kshpv, line 99: You should mention that we have to use val part of the dataset

nncf/experimental/onnx/statistics/aggregator.py
KodiaqQ: This moment is need to be updated. Currently, it's WA and not a general solution, for the case when the sampler wasn't set before calculation. I see two possible ways to improve that: create `sampler` validation inside `compute_statistics()` and raise an exception or remove `sampler` creation outside the Engine methods. 
But currently, there is no way to get max_number_samples internally in the Engine. This is why `create_onnx_sampler()` calls twice here.

nncf/experimental/post_training/api/sampler.py
```diff
@@ -0,0 +1,38 @@
...
+        max_samples_len = min(len(sample_indices), data_loader_len) if sample_indices else data_loader_len
+        self.batch_indices = list(range(0, max_samples_len + 1, self.batch_size))
+
+    @abstractmethod
+    def __iter__(self):
+        pass
+
+    @abstractmethod
+    def __len__(self):
+        pass
```
andrey-churkin, line 38: If I'm not mistaken the `__len__()` method is not needed when we have the `__iter__()` method.

andrey-churkin (changes requested):

andrey-churkin:
General comments from my side:
- @KodiaqQ Could you please use another docstring format for the consistency with other code?
```python
"""
Processes model input data based on the backend.

:param inputs: List of input data.
:return inputs: List of the input data.
"""
```
instead of
```python
""" Processes model input data based on the backend
    :param inputs: list of input data
    :return inputs: list of the input data
"""
```

- @kshpv seems like several docstrings are missed for classes and methods inside of the `experimental.onnx` module. Are there any tickets? When do you plan to add docs?

kshpv:
> General comments from my side:
> 
> * @KodiaqQ Could you please use another docstring format for the consistency with other code?
> 
> ```python
> """
> Processes model input data based on the backend.
> 
> :param inputs: List of input data.
> :return inputs: List of the input data.
> """
> ```
> 
> instead of
> 
> ```python
> """ Processes model input data based on the backend
>     :param inputs: list of input data
>     :return inputs: list of the input data
> """
> ```
> 
> * @kshpv seems like several docstrings are missed for classes and methods inside of the `experimental.onnx` module. Are there any tickets? When do you plan to add docs?

Good point! I will create a ticket on that.

kshpv:
> General comments from my side:
> 
> * @KodiaqQ Could you please use another docstring format for the consistency with other code?
> 
> ```python
> """
> Processes model input data based on the backend.
> 
> :param inputs: List of input data.
> :return inputs: List of the input data.
> """
> ```
> 
> instead of
> 
> ```python
> """ Processes model input data based on the backend
>     :param inputs: list of input data
>     :return inputs: list of the input data
> """
> ```
> 
> * @kshpv seems like several docstrings are missed for classes and methods inside of the `experimental.onnx` module. Are there any tickets? When do you plan to add docs?

83473

examples/experimental/onnx/onnx_ptq_classification.py
```diff
@@ -20,22 +21,56 @@
 from nncf.experimental.post_training.compression_builder import CompressionBuilder
 from nncf.experimental.post_training.algorithms.quantization import PostTrainingQuantization
 from nncf.experimental.post_training.algorithms.quantization import PostTrainingQuantizationParameters
-from nncf.experimental.onnx.dataloaders.imagenet_dataloader import create_dataloader_from_imagenet_torch_dataset
+from nncf.experimental.onnx.dataloaders.imagenet_data_loader import create_dataloader_from_imagenet_torch_dataset
+from nncf.experimental.post_training.api.metric import Metric
+
+
+class Accuracy(Metric):
```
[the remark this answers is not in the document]
KodiaqQ: Done

examples/experimental/onnx/onnx_ptq_classification.py
KodiaqQ: Done

nncf/experimental/post_training/api/engine.py
KodiaqQ: These checks have been moved to `calculate_statistics()` and `calculate_metrics()` respectively.

nncf/experimental/onnx/engine.py
```diff
@@ -36,17 +41,49 @@ def set_model(self, model: str) -> None:
...
+            input_data, _ = sample
+            output_tensors, model_outputs = self._infer(input_data)
+            for out_id, model_output in enumerate(model_outputs):
+                if model_output.name not in output:
+                    output[model_output.name] = []
+                # TODO (Nikita Malinin): Add backend-specific statistics aggregator usage
+                output[model_output.name].append(output_tensors[out_id])
         return output
+
+    def compute_metrics(self, metrics_per_sample=False):
```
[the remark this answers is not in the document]
KodiaqQ: Done

nncf/experimental/onnx/engine.py
```diff
@@ -36,17 +41,49 @@ def set_model(self, model: str) -> None:
...
+        if not self.is_model_set():
+            raise RuntimeError('The {} tried to compute statistics, '
+                               'while the model was not set.'.format(self.__class__))
```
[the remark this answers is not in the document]
KodiaqQ: Checks has been moved

kshpv:
Jenkins please retry a build

nncf/experimental/post_training/api/engine.py
```diff
@@ -11,38 +11,128 @@
...
+    @property
+    def dataset(self):
+        return self._dataset
+
+    @dataset.setter
+    def dataset(self, dataset):
+        self._dataset = dataset
+
+    @property
+    def metrics(self):
```
kshpv, line 49: Let's use only one variant of attribute setters.
Now, we have `properties `and `set_model()`.
I suggest switching to aka `set_model()` variant as we don't suppose to have to get methods.

examples/experimental/onnx/onnx_ptq_classification.py
```diff
@@ -12,30 +12,76 @@
...
+    """
+
+    def __init__(self, top_k: int = 1):
+        super().__init__()
+        self._top_k = top_k
+        self._name = f'accuracy@top{self._top_k}'
+        self._matches = []
+
+    @property
+    def avg_value(self):
```
kshpv, line 46: Do you have tests on Metrics calulcation?

nncf/experimental/post_training/api/engine.py
```diff
@@ -11,38 +11,128 @@
...
+from typing import Callable
+from nncf.experimental.post_training.api.sampler import Sampler
 
 ModelType = TypeVar('ModelType')
 TensorType = TypeVar('TensorType')
+MetricType = TypeVar('MetricType')
 
 
-class Engine(ABC):
+class Engine:
```
kshpv, line 26: It's still ABC, because it doesn't make sense to create an instance of `Engine`, cause it doesn't have `infer()`

kshpv (changes requested):

nncf/experimental/post_training/api/sampler.py
vinnamkim: I guess `__len__()` is needed because we can use more efficient way to obtain the length rather than counting one-by-one using `__iter__()`.

https://stackoverflow.com/a/37192530

examples/experimental/onnx/onnx_ptq_classification.py
```diff
@@ -68,10 +122,11 @@ def run(onnx_model_path: str, output_model_path: str,
                              "(Please, take a look at torchvision.datasets.ImageFolder)",
                         type=str)
     parser.add_argument("--batch_size", help="Batch size for initialization", default=1)
-    parser.add_argument("--shuffle", help="Whether to shuffle dataset for initialization", default=True)
+    parser.add_argument("--shuffle", help="Whether to shuffle dataset for initialization", default=False)
```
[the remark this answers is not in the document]
vinnamkim: Reverted it to `default=True`

examples/experimental/onnx/onnx_ptq_classification.py
```diff
@@ -68,10 +122,11 @@ def run(onnx_model_path: str, output_model_path: str,
...
     parser.add_argument("--input_shape", help="Model's input shape", nargs="+", type=int, default=[1, 3, 224, 224])
     parser.add_argument("--init_samples", help="Number of initialization samples", type=int, default=300)
     parser.add_argument("--ignored_scopes", help="Ignored operations ot quantize", nargs="+", default=None)
+    parser.add_argument("--evaluate", help="Whether is a validation of the final model need", default=False)
```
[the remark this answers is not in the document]
vinnamkim: Edited it to "Run an evaluation step for the final quantized model"

nncf/experimental/post_training/api/engine.py
vinnamkim: Fixed to inherit from ABC

nncf/experimental/post_training/api/engine.py
vinnamkim: Switched setters to follow `set_*()`

examples/experimental/onnx/onnx_ptq_classification.py
vinnamkim: Added test to `tests/experimental/post_training/test_api.py`

PR #1155, opened 2022-03-30
Add functional test of ONNX PTQ of detection models

kshpv:
### Changes

- Add hardware fusing pattern for input preprocessing
- Add graph test for yolov2, tiny-yolov2. Currently, there are some issues with SSD models

      https://github.com/onnx/onnx/issues/4101
      https://github.com/onnx/onnx/issues/4102

### Reason for changes

Extend support of ONNX PTQ models

### Related tickets

83146

### Tests

As stated in the title


kshpv:
I tried to test SSD-300 model from torchvision repo, the model was successfully quantized and onnx assumes that the model is correct, while onnxruntime can't infer the model with the following error: 

`INVALID_ARGUMENT : Non-zero status code returned while running Conv node. Name:'Conv_123' Status Message: Invalid input shape: {0,0}`

I propose to include for now only yolov2 and tiny-yolov2 and focus on the domain - semantic segmentation and then proceed with SSD

kshpv:
Jenkins please retry a build

KodiaqQ:
@kshpv, did we plan to add a sample for the detection models in this PR or not? How did the models were tested in terms of accuracy?

kshpv:
> @kshpv, did we plan to add a sample for the detection models in this PR or not? How did the models were tested in terms of accuracy?

No, currently we only add functional support (by adding test which do quantization, but the dataset is dummy).
After that we will introduce a full sample with object detection models

KodiaqQ (approved):

vshampor (approved):

PR #1176, opened 2022-05-10
add type casting to symmetric_quantize

eugene123tw:


vinnamkim (approved):
LGTM.
But, don't we need to consider adaptive `eps` on `fp16` (https://github.com/pytorch/pytorch/issues/26218)?
I think this might also make a huge accuracy drop.

PR #1177, opened 2022-05-11
Fix a bug with functions_cuda_impl.cu

ZhangZhiPku:
### Changes

line 36 of file functions_cuda_impl.cu should be:
  else if (scale_dim >= 2 && input_range.size(1) > 1)

### Reason for changes

Compile error.

### Related tickets

None

### Tests

None


vinnamkim:
Hi, Thanks for your contribution!

C++ uses `and` as an alternative keyword for `&&`: https://en.cppreference.com/w/cpp/keyword/and
Did you have a compile error because of this?

ZhangZhiPku:
Yes they do, I don't konw when c++ starts supports using "and" instread of &&, however on my computer, the compilaton will never successed.

My Environment:
OS: Win11
CUDA: 11.3
Compiler: MSVC 19.29.30136
Ninja: latest

vinnamkim (approved):
It seems that MSVC requires adding an additional compiler option to support alternative keywords in C++.

https://developercommunity.visualstudio.com/t/alternative-keywords-in-c-is-not-supported/781344

LGTM.

PR #1164, opened 2022-04-05
ONNX Post-Training Quantization CamVid, Mapillary segmentation sample

kshpv:
### Changes

Introduce ONNX Post-Training Quantization CamVid, Mapillary segmentation sample

### Reason for changes

Extend support of PTQ of ONNX models

### Tests

Sanity test was added

PR #1184, opened 2022-05-20
Introduce ONNX Post-Training Quantization CamVid, Mapillary segmentation sample

vinnamkim:
### Changes
Introduce ONNX Post-Training Quantization CamVid, Mapillary segmentation sample

### Reason for changes
Extend support of PTQ of ONNX models

This PR makes up ***ONNX Post-Training Quantization CamVid, Mapillary segmentation sample #1164*** and replaces it.


wonjuleee (approved):
Looks great!

PR #1187, opened 2022-05-24
add setting the best checkpoint before training

sungchul2:
### Changes

Before training, set best checkpoint path to real path.

### Reason for changes

Best checkpoint path remains None when the model is not improved during training, so final checkpoint cannot be loaded.

### Related tickets

[CVS-84176](https://jira.devtools.intel.com/browse/CVS-84176)

### Tests

Set a validation function that returns gradually reduced accuracy and check if best checkpoint path is not None

### DIFF
--- a/tests/torch/accuracy_aware_training/test_training_loop.py
+++ b/tests/torch/accuracy_aware_training/test_training_loop.py
@@ -274,14 +274,11 @@ def test_early_exit_with_mock_validation(max_accuracy_degradation, exit_epoch_nu
 def test_early_exit_with_mock_validation_and_no_improvement(
     max_accuracy_degradation, maximal_total_epochs=5
 ):
-    epoch_counter = 0
-
     def mock_validate_fn(model, init_step=False, epoch=0):
         original_metric = 0.85
         if init_step:
             return original_metric
-        nonlocal epoch_counter
-        epoch_counter = epoch
+            
         return original_metric - 0.11 * (epoch+1)
 
     config = get_quantization_config_without_range_init(LeNet.INPUT_SIZE[-1])
--- a/tests/torch/accuracy_aware_training/test_training_loop.py
+++ b/tests/torch/accuracy_aware_training/test_training_loop.py
@@ -267,6 +267,66 @@ def test_early_exit_with_mock_validation(max_accuracy_degradation, exit_epoch_nu
     assert epoch_counter == exit_epoch_number
 
 
+@pytest.mark.parametrize(
+    ('max_accuracy_degradation'),
+    (({'maximal_absolute_accuracy_degradation': 0.1}),)
+)
+def test_early_exit_with_mock_validation_and_no_improvement(
+    max_accuracy_degradation, maximal_total_epochs=5
+):
+    epoch_counter = 0
+
+    def mock_validate_fn(model, init_step=False, epoch=0):
+        original_metric = 0.85
+        if init_step:
+            return original_metric
+        nonlocal epoch_counter
+        epoch_counter = epoch
+        return original_metric - 0.11 * (epoch+1)
+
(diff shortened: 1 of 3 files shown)


tests/torch/accuracy_aware_training/test_training_loop.py
```diff
@@ -267,6 +267,66 @@ def configure_optimizers_fn():
...
+        optimizer = SGD(model.parameters(), lr=1e-3)
+        return optimizer, None
+
+    early_stopping_training_loop = EarlyExitCompressionTrainingLoop(config, compression_ctrl,
+                                                                    dump_checkpoints=False)
+    assert early_stopping_training_loop.runner._best_checkpoint is None
+
+    model = early_stopping_training_loop.run(model,
+                                             train_epoch_fn=train_fn,
+                                             validate_fn=partial(mock_validate_fn),
```
vinnamkim, line 325: Is it better `validate_fn=partial(mock_validate_fn, init_step=False)` for better readability?

vinnamkim: or drop `partial()`

tests/torch/accuracy_aware_training/test_training_loop.py
```diff
@@ -267,6 +267,66 @@ def configure_optimizers_fn():
...
 
 
+@pytest.mark.parametrize(
+    ('max_accuracy_degradation'),
+    (({'maximal_absolute_accuracy_degradation': 0.1}),)
+)
+def test_early_exit_with_mock_validation_and_no_improvement(
+    max_accuracy_degradation, maximal_total_epochs=5
+):
+    epoch_counter = 0
```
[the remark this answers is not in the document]
sungchul2: Not required, removed.

tests/torch/accuracy_aware_training/test_training_loop.py
sungchul2: Done

sungchul2:
`AdaptiveCompressionTrainingLoop` usually finds and changes compression rates to improve the model.
However, when the test is conducted with the same setting of `EarlyExitCompressionTrainingLoop`, `compression_rate_target` is lower than `minimal_compression_rate` and the error occurs.
So, other setting is required to check how `AdaptiveCompressionTrainingLoop` works in a situation where accuracy is gradually decreasing.

sungchul2:
Because just set appropriate values to avoid [while](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/accuracy_aware_training/training_loop.py#L248-L249) condition, it doesn't matter.

If quantization cfgs is used like `EarlyExitCompressionTrainingLoop` test, `compression_controller.compression_stage()` is already set to `CompressionStage.FULLY_COMPRESSED`, so it doesn't matter if proceeding with the commits here.

However, `AdaptiveCompressionTrainingLoop` doesn't support quantization, so the test here uses sparsity cfgs.
So, if sparsity cfgs are used in `EarlyExitCompressionTrainingLoop` or `AdaptiveCompressionTrainingLoop`, `compression_controller.compression_stage()` is initially set to `CompressionStage.UNCOMPRESSED` because `current_sparsity_level`=0.
It means that best checkpoint path will not be set because it cannot pass [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/accuracy_aware_training/runner.py#L142-L144).
To update `current_sparsity_level` and `target_level` to non-zero values, it is necessary to pass [while](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/accuracy_aware_training/training_loop.py#L248-L249) and `was_compression_rate_changed` must be set to `True`.

sungchul2:
@vinnamkim 
fix: the error about loading the best checkpoint because `best_checkpoints` is empty at `AdaptiveCompressionTrainingLoop`.

vinnamkim (approved):
LGTM.

PR #1193, opened 2022-06-13
set ignored scope widely

sooahleex:
### Changes
set ignored_scope as op_addresses_to_skip too

### Reason for changes
pruning consistently tried to apply to ignored scope

### DIFF
--- a/nncf/common/pruning/model_analysis.py
+++ b/nncf/common/pruning/model_analysis.py
@@ -167,6 +167,9 @@ class ModelAnalyzer:
         for node in reversed_sorted_nodes:
             # Check all output nodes accept_pruned_input attribute
             out_nodes = self.graph.get_next_nodes(node)
+            if out_nodes == []:
+                self.can_prune[node.node_id] =False
+                continue
             outputs_accept_pruned_input = all(self.accept_pruned_input[node.node_id] for node in out_nodes)
 
             # Check all output nodes can_prune attribute
@@ -181,6 +184,8 @@ class ModelAnalyzer:
         """
         sorted_nodes = self.graph.topological_sort()
         for node in sorted_nodes:
+            if not self.can_prune[node.node_id]:
+                continue
             # Propagate attribute only in not conv case
             if self.node_propagate_can_prune_attr(node):
                 in_nodes = self.graph.get_previous_nodes(node)
--- a/nncf/common/pruning/model_analysis.py
+++ b/nncf/common/pruning/model_analysis.py
@@ -167,9 +167,6 @@ class ModelAnalyzer:
         for node in reversed_sorted_nodes:
             # Check all output nodes accept_pruned_input attribute
             out_nodes = self.graph.get_next_nodes(node)
-            if out_nodes == []:
-                self.can_prune[node.node_id] =False
-                continue
             outputs_accept_pruned_input = all(self.accept_pruned_input[node.node_id] for node in out_nodes)
 
             # Check all output nodes can_prune attribute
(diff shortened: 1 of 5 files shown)


nncf/common/pruning/model_analysis.py
```diff
@@ -167,6 +167,9 @@ def propagate_can_prune_attr_up(self):
         for node in reversed_sorted_nodes:
             # Check all output nodes accept_pruned_input attribute
             out_nodes = self.graph.get_next_nodes(node)
+            if out_nodes == []:
+                self.can_prune[node.node_id] =False
```
vinnamkim, line 171: Could you please auto-format with pylint?
It seems that this package uses https://github.com/openvinotoolkit/nncf/blob/develop/.pylintrc.
Not only for this line `=False`.

daniil-lyakhov:
Guys, looks like we were removed from nncf developers group. Please add me as an reviewer and don't merge this without our review!

vinnamkim:
> Guys, looks like we were removed from nncf developers group. Please add me as an reviewer and don't merge this without our review!
> 
> cc: @negved

Please request to join at this link: https://github.com/orgs/openvinotoolkit/teams/nncf_pytorch-maintainers/members

vinnamkim:
> > Guys, looks like we were removed from nncf developers group. Please add me as an reviewer and don't merge this without our review!
> > cc: @negved
> 
> Please request to join at this link: https://github.com/orgs/openvinotoolkit/teams/nncf_pytorch-maintainers/members

@negvet

daniil-lyakhov:
@sooahleex, we already have mechanism for ignoring layers, it's implemented here: https://github.com/openvinotoolkit/nncf/blob/160990cba0473452955c494b2087c4f4160241c7/nncf/common/pruning/node_selector.py#L279-L305

it used at the last step of node selection here: https://github.com/openvinotoolkit/nncf/blob/160990cba0473452955c494b2087c4f4160241c7/nncf/common/pruning/node_selector.py#L218

It's essential to generate masks for each target node even if it's in a ignored scope because of the algorithm implementation. Ignoring layers during mask propagation could lead to inconsistent pruning layout!

nncf/common/pruning/operations.py
```diff
@@ -204,6 +204,10 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
         if not not_empty_masks:
             return None
 
+        # if node is leaf node return None
+        if not graph.get_output_edges(node):
+            return None
+
```
daniil-lyakhov, line 210: Looks like problem is more fundamental: some of the input node has non empty pruning masks. The result of such return could lead to inconsistent pruning layout. @sooahleex, let's try to inspect this graph together

sooahleex:
> 



> @sooahleex, we already have mechanism for ignoring layers, it's implemented here:
> 
> https://github.com/openvinotoolkit/nncf/blob/160990cba0473452955c494b2087c4f4160241c7/nncf/common/pruning/node_selector.py#L279-L305
> 
> it used at the last step of node selection here:
> 
> https://github.com/openvinotoolkit/nncf/blob/160990cba0473452955c494b2087c4f4160241c7/nncf/common/pruning/node_selector.py#L218
> 
> It's essential to generate masks for each target node even if it's in a ignored scope because of the algorithm implementation. Ignoring layers during mask propagation could lead to inconsistent pruning layout!

I knew that part and I checked that works out for mask propagation. When I tried to do pruning, the error is occurred in `count_flops_and_weights_per_node` for the layer which I added in ignored scopes. For the function `count_flops_and_weights_per_node`, There is no part to check whether it is prunable. So I added `op_addresses_to_skip` to check whether the layer is prunable or not.

PR #1195, opened 2022-06-13
Disable TF32 presicison in tests

daniil-lyakhov:
### Changes

Tensor Float 32 precision was disabled in both TF and Torch tests

### Reason for changes

Precision of TF32 is not enough for NNCF tests. It's possible to use TF32 precision, but additional disscussion is needed

### Tests

Checked locally on machine with TF32 support (RTX 3090)

### DIFF
--- a/tests/tensorflow/conftest.py
+++ b/tests/tensorflow/conftest.py
@@ -18,6 +18,12 @@ except ImportError:
     tf = None
 
 
+@pytest.fixture(scope="session", autouse=True)
+def disable_tf32_precision():
+    if tf:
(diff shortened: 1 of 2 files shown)


daniil-lyakhov:
CC: @vinnamkim, @negvet

vinnamkim (approved):
LGTM. But, I'm curious which test are failed by TF32. If TF32 can do some serious damage to the NNCF algorithm, shouldn't we avoid using TF32 as the default or document this?

wonjuleee (approved):
LGTM. see https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/ for reference.

PR #1198, opened 2022-06-15
[CVS-86321] Fix PTQ API failure for Inception-v1-12 model

vinnamkim:
### Changes

1. Fix PTQ API failure for inception-v1-12 model
2. Update benchmark results in README.md
3. Add a test for the case where the Inception-v1-12 model had a problem

### Reason for changes

This is because PTQ API tries to quantize a weight tensor using initializer, but it fails if the weight tensor has no initializer. Inception-v1-12 model has a reshaped weight tensor which has no initializer since it is reshaped from the weight tensor.

### Related tickets

86321

### Tests

Ran tests locally because CI is down. All tests were passed.

image


daniil-lyakhov:
Please be aware of ticket 86349

wonjuleee (approved):
It looks good to me.

PR #1199, opened 2022-06-15
Limit the protobuf version - tensorflow requirement

negvet:
### Changes

When installing NNCF with TF the following error pop up
`error: protobuf 4.21.1 is installed but protobuf<3.20,>=3.9.2 is required by {'tensorboard'}`
Steps to reproduce:
```
pip install -U pip
python setup.py install --tf
```

Temporary patch proposed (upper limit of protobuf version). Not sure if adding this additional explicit requirement is safe.

### Reason for changes

Protobuf releases new version, which is getting installed by TensorFlow but not supported by TensorBoard. For reference please see https://github.com/tensorflow/tensorflow/issues/56077.
Update is needed be able to install NNCF with TF.

### Related tickets

[CVS-86310]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -135,6 +135,9 @@ if "--torch" in sys.argv:
     sys.argv.remove("--torch")
 
 if "--tf" in sys.argv:
+    # See also: https://github.com/tensorflow/tensorflow/issues/56077
+    # This is a temporary patch, that limites the protobuf version from above
+    INSTALL_REQUIRES.extend(["protobuf>=3.9.2,<3.20"])
     INSTALL_REQUIRES.extend(EXTRAS_REQUIRE["tf"])
     sys.argv.remove("--tf")
 


vinnamkim (commented):
Is this change working on `pip install nncf[tf]`?

negvet:
> Is this change working on `pip install nncf[tf]`?

Most probably it should.
Do you know how to check if? should I upload to testpypi just for a try?

Note: currenlty, the error does not reproduce with 
`pip install nncf[tf]`
I also tried with ` --no-cache-dir` option and updated `pip` - it works. Probably due to different installation order.

vinnamkim:
> > Is this change working on `pip install nncf[tf]`?
> 
> Thanks for the comment!
> 
> Most probably it should. Do you know how to check if? should I upload to testpypi just for a try?
> 
> Note: currenlty, the error does not reproduce with `pip install nncf[tf]` I also tried with ` --no-cache-dir` option and updated `pip` - it works. Probably due to different installation order.

Maybe we can test by
```
pip install -e .[tf]
```
https://stackoverflow.com/a/30239714

negvet:
> Maybe we can test by
> ```
> pip install -e .[tf]
> ```
> https://stackoverflow.com/a/30239714

Thanks for the suggestion. I tested. It works.

PR #1212, opened 2022-07-01
Add reorder weights for Linear layer

jpablomch:
Co-authored-by: Zheng, Yi <yi.zheng@intel.com>

### Changes

Add input_reorder and output_reorder for linear layer

### Reason for changes

Enable weights reordering for linear layer. Please see how other layers already have enabled weight reordering. This is used by BootstrapNAS for its [filter reordering algorithm](https://github.com/openvinotoolkit/nncf/blob/d5e9edf8f335fc6d97825ddc07cfbd03594defd2/nncf/experimental/torch/nas/bootstrapNAS/elasticity/filter_reorder.py#L50): 

### Tests

Added test in BootstrapNAS tests

### DIFF
--- a/nncf/torch/pruning/operations.py
+++ b/nncf/torch/pruning/operations.py
@@ -281,6 +281,33 @@ class PTTransposeConvolutionPruningOp(TransposeConvolutionPruningOp, PTPruner):
 class PTLinearPruningOp(LinearPruningOp, PTPruner):
     subtypes = [PTLinearMetatype, PTMatMulMetatype]
 
+    @classmethod
+    def input_reorder(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph):
+        input_masks = get_input_masks(node, graph)
+        reorder_indexes = input_masks[0]
+        if reorder_indexes is None:
+            return
+        reorder_indexes = reorder_indexes.tensor
+        reorder_indexes = reorder_indexes.int()
+        fc = model.get_containing_module(node.node_name)
+        fc.weight.data = torch.index_select(fc.weight.data, 1, reorder_indexes)
+        nncf_logger.debug(
+            'Reordered input channels (first 10 reorder indexes {}) of Linear: {} '.format(reorder_indexes[:10],
+                                                                                           node.data['key']))
+
+    @classmethod
+    def output_reorder(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph):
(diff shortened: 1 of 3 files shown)


nncf/torch/pruning/operations.py
```diff
@@ -281,6 +281,33 @@ def output_prune(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph) -> N
...
+    @classmethod
+    def output_reorder(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph):
+        reorder_indexes = node.data['output_mask']
+        if reorder_indexes is None:
+            return
+        fc = model.get_containing_module(node.node_name)
+        reorder_indexes = reorder_indexes.tensor
+        fc.weight.data = torch.index_select(fc.weight.data, 0, reorder_indexes)
+        if fc.bias is not None:
+            fc.bias.data = torch.index_select(fc.bias.data, 0, reorder_indexes)
```
vinnamkim, line 307: I wonder if those lines are also needed for `input_reorder`.

nncf/torch/pruning/operations.py
```diff
@@ -281,6 +281,33 @@ def output_prune(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph) -> N
...
     subtypes = [PTLinearMetatype, PTMatMulMetatype]
 
+    @classmethod
+    def input_reorder(cls, model: NNCFNetwork, node: NNCFNode, graph: NNCFGraph):
+        input_masks = get_input_masks(node, graph)
+        reorder_indexes = input_masks[0]
+        if reorder_indexes is None:
+            return
+        reorder_indexes = reorder_indexes.tensor
+        reorder_indexes = reorder_indexes.int()
```
vinnamkim, line 291: Why is `int()` only needed for `input_reorder`?

vinnamkim:
add to whitelist

nncf/torch/pruning/operations.py
jpablomch: @vinnamkim Thanks! We have removed int() since it is not required here.

nncf/torch/pruning/operations.py
jpablomch: @vinnamkim These lines are not needed for `input_reorder`. Thanks!

vinnamkim (approved):
LGTM.

PR #1214, opened 2022-07-04
Revise ONNX PTQ pipeline

vinnamkim:
### Changes

 - Add `NNCFData` data type which is a dictionary with string key and tensor values.
 - Revise ONNX PTQ pipeline to handle `NNCFData` as inputs and outputs.

### Reason for changes

 - ONNX PTQAPI fails for `yolov3-12` and `tiny-yolov3-11` models. This is because those models require to put `input_shape` tensor as their inputs. However, the current input pipeline cannot handle this case.

### Related tickets

86465


dongkwan-kim01:
I think it's good to be able to handle models that require input shape.

PR #1210, opened 2022-06-28
[Pruning algo] Filter counting fix for linear layers

daniil-lyakhov:
### Changes
THIS WILL CHANGE PRUNING LEVEL OF MODELS WHICH CONTAINS PRUNABLE LINEAR LAYERS, PRUNED BY FLOPS

1) Weights and flops calculation for pruned graphs with closing linear layer was fixed by `sparse_multiplier` param introduction
2) Groups with different amount of channels / output elements are filtering during `pruing_node_selector` call
3) Batched linear nodes are filtering during  `pruing_node_selector` call
4) NAS: Wrong weights / flops calculation for models with depthwise convs was fixed
5) NAS: `purning_node_selector` and shape pruning metatypes was aligned
6) Weights / flops calculation was refactored

### Reason for changes

1) Wrong flops/weights calculation in case of closing linear layer. `sparse_multiplier` is propagating through net and inside an input mask of a closing operation it's equal to amount of input channels of conv layer / input elements of linear layer corresponded to one output channel / output element of producer. 
For example in a case bellow conv_1 mask producer and conv_2 mask producer have `sparse_multiplier` equal to `H*W`
image

2)

3) Bathced linear is an linear which output edge with >= 2 dimentions. Example, in which linear will affect feature map of closing conv instead of channels in case of pruning, which is not right.
image

### Related tickets

86889

PR #1215, opened 2022-07-04
[Pruning] Input shapes param removal

daniil-lyakhov:
### Changes

Redundant input shape param was removed from `count_flops_and_weights_per_node` function.
`count_filters_num` now always takes into account pruned linear layers.

### Reason for changes

`collect_input_shapes` function isn't protected from case when input edges don't exist (unlike `collect_output_shapes`, which process nodes without output edges correctly). Turns out that output of `collect_input_shaped` used only in `count_flops_and_weights_per_node` function and could be easily replaced by linear layers `in_features` and `out_features` attributes

86889

### Tests

No need

### DIFF
--- a/nncf/common/pruning/utils.py
+++ b/nncf/common/pruning/utils.py
@@ -225,7 +225,6 @@ def get_cluster_next_nodes(graph: NNCFGraph, pruned_groups_info: Clusterization[
 
 
 def count_flops_and_weights(graph: NNCFGraph,
-                            input_shapes: Dict[NNCFNodeName, List[int]],
                             output_shapes: Dict[NNCFNodeName, List[int]],
                             conv_op_metatypes: List[Type[OperatorMetatype]],
                             linear_op_metatypes: List[Type[OperatorMetatype]],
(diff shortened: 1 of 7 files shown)


sooahleex (approved):
Look good to me

nncf/common/pruning/utils.py
```diff
@@ -327,10 +321,14 @@ def count_flops_and_weights_per_node(graph: NNCFGraph,
...
-        flops_numpy = 2 * np.prod(input_shapes[name]) * np.prod(output_shapes[name])
-        weights_numpy = np.prod(input_shapes[name]) * np.prod(output_shapes[name])
-        flops[name] = flops_numpy.astype(int).item()
-        weights[name] = weights_numpy.astype(int).item()
+
+        num_in_features = node.layer_attributes.in_features
+        num_out_features = node.layer_attributes.out_features
+
+        flops_numpy = 2 * num_in_features * num_out_features
+        weights_numpy = num_in_features * num_out_features
```
vinnamkim, line 329: Is matrix multiplication complexity `O(mnk)` for `(m, n)` and `(n, k)` matrices? I think that we should consider `input_shapes(m)` in addition to `num_in_features(n)` and `num_in_features(k)`. Am I right?

vinnamkim: Like this https://github.com/sovrasov/flops-counter.pytorch/blob/06793e8ec1b73b154b8078db571ac682125f73ce/ptflops/pytorch_ops.py#L31-L36

daniil-lyakhov: Agree, this will be fixed in #1210

PR #1219, opened 2022-07-06
Updated placement of exporting model

cavusmustafa:
### Changes

Based on HF patch, models used to be exported before QAT. Model exporting is moved further to export the model after QAT.

### Reason for changes

Previously, if the option "--to_onnx" is used with "--do_train", the model was exported before the training. So, we have to train the model, save the checkpoint, and export the model after when evaluating the model from the saved checkpoint. The reason of this changes is to enable using "--to_onnx" while training as well. So, with this change, the model model will not be exported right after it is compressed. It will be exported when the training is completed.

### Tests

Tested on question answering. Other scripts should be validated as well. I am not confident if the change break anything and it would be better if this can be tested by someone else who is more experienced with these examples.

### DIFF
--- a/third_party_integration/huggingface_transformers/0001-Modifications-for-NNCF-usage.patch
+++ b/third_party_integration/huggingface_transformers/0001-Modifications-for-NNCF-usage.patch
@@ -84,7 +84,7 @@ index 0c002deeb..f9e4f4171 100755
  
      # Preprocessing the datasets.
      # First we tokenize all the texts.
-@@ -444,6 +437,59 @@ def main():
+@@ -444,6 +437,51 @@ def main():
          if data_args.max_eval_samples is not None:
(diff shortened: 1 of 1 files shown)


alexsu52:
Hello @cavusmustafa, Thanks for your contribution. Could you take this PR out of the draft?

cavusmustafa:
Hi @alexsu52, it should be out of draft now. Thank you.

wonjuleee (approved):
It looks good to me.

ljaljushkin:
Looks good to me, but it would be great to make sure that there's no degradation in 3rd party sanity tests (tests/torch/test_sanity_third_party.py). These tests are not part of pre-commit and should be run separately.

At least, it should pass locally, ideally in Jenkins. But it seems like the corresponding Jenkins job is temporarily disabled.

@alexsu52 @wonjuleee

PR #1223, opened 2022-07-11
[ONNX PTQ API] Fix ssd-12

vinnamkim:
### Changes
- Support `opset` version < 13
   - Block per channel quantization for `opset` version < 13 models
   - Add `convert_opset_version` flag to `CompressionBuilder`
   - Add test cases for `opset` version < 13
- Update ssd-12 benchmark results

### Reason for changes
Support PTQ API for ssd-12 model which cannot be converted to `opset_version=13`

### Related tickets
87359

### DIFF
--- a/nncf/experimental/onnx/model_normalizer.py
+++ b/nncf/experimental/onnx/model_normalizer.py
@@ -11,7 +11,7 @@
  limitations under the License.
 """
 
-from collections import defaultdict
+from collections import Counter, defaultdict
 from copy import deepcopy
 import onnx
 from onnx.version_converter import convert_version, ConvertError  # pylint: disable=no-name-in-module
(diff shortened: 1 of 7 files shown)


nncf/experimental/onnx/model_normalizer.py
```diff
@@ -101,13 +102,33 @@ def add_const_value_infos_to_graph(graph: onnx.GraphProto):
...
+
+        for i, node in enumerate(model.graph.node):
+            name_counter[node.name] += 1
+
+        for name, cnt in name_counter.items():
+            if cnt > 1:
+                raise RuntimeError(
+                    f"Node {name} occurs more than once ({cnt} times). "
+                    "NNCF expects every node to have a unique name."
+                )
```
dongkwan-kim01, line 132: How about using counter like
```
if max(Counter(name for i, name in enumerate(model.graph.node)).values()) > 1:
    raise RuntimeError(
    ...
```

vinnamkim: Is there a way to avoid losing the duplicated node name if we use `Counter`?
I didn't see in your example.

dongkwan-kim01: Ah I didn't see the name, cnt in RuntimeError. We should save the counter or dict as original.
Or how about remove the for loop and list all the duplicated names like
```
        name_counter = Counter(name for i, name in enumerate(model.graph.node))
        if max(name_counter.values()) > 1:
            raise RuntimeError(
                f"Nodes {[(name, cnt) for name, cnt in name_counter.items() if cnt > 1]} (name, counts) occurred more than once ..."
```

vinnamkim: Fixed.

vinnamkim:
retest this please

PR #1227, opened 2022-07-19
Remove ONNX PTQ API performance results from README.md

vinnamkim:
### Changes
Remove ONNX PTQ API performance results from README.md

### Reason for changes

Please see the related ticket

### Related tickets
87917


vinnamkim:
Because there is no code change and it is an emergency, I merge it now.

PR #1233, opened 2022-07-20
Replace a saving ONNX temporal model on disk with model serialization

kshpv:
### Changes

As stated in the title

### Reason for changes

There is a problem on Windows machines with opening twice of the tempfile.
This is a comment from https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
`Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows).`

### Related tickets

None

### Tests

Nothing is needed for testing as our tests work on Ubuntu


vinnamkim (commented):
Hi @kshpv,
I think that it's not good to use a temporary file because it has OS dependency and can use a disk storage, not main memory. From https://onnxruntime.ai/docs/api/python/api_summary.html#inferencesession, `rt.InferenceSession` allow serialized ONNX model in a byte string beside the model path. We may be able to serialize `ONNXModelProto` using `model.SerializeToString()` like https://github.com/onnx/onnx/blob/529f7cab88eed143fc9f0126147e9732585e4c5e/onnx/__init__.py#L53-L70

vinnamkim (approved):
LGTM.

vinnamkim:
This will affect only ONNX part and our CI server is very busy now. So I cancelled TF and Torch pre-commit tests.

vinnamkim:
ONNX precommit tests are passed in `job/NNCF/job/precommit/job/ubuntu20_precommit_onnx/102/`

PR #1234, opened 2022-07-22
Implement FracBitsQuantizationBuilder and Controller

vinnamkim:
This should be merged after #1231.

### Changes
 - Implement Builder and Controller
 - Add and test ModelSizeCompressionLoss

### Reason for changes
 - To implement the initial version of FracBitsQuantizationBuilder, Controller, and Loss.

### Related tickets
87888, 87889

### Tests
Related unit tests added

PR #1237, opened 2022-07-27
Revert "Implement FracBitsQuantizers"

vinnamkim:
Reverts openvinotoolkit/nncf#1231


vinnamkim:
I merged this PR because we'll follow a feature branching strategy as mentioned in https://github.com/openvinotoolkit/nncf/pull/1234#issuecomment-1196268658.

PR #1244, opened 2022-08-05
Pruning support for chunk ops

sooahleex:
### Changes
- Add layer attribute and functions for chunk/split ops
- mask propagation for split/chunk ops

### Reason for changes
- To support pruning for chunk operation in HRNet
- Support both split and chunk operation

### Related tickets
86694

### Tests
Test for tensor processor
Test for pruning operations
Test for model pruning analysis (simple split/chunk block, multiple split/chunk-concat block, HRNet unit block)
Test for model converter
Test for graph building


sooahleex:
@daniil-lyakhov This is the draft version, and there are somethings to check
Need to check
- get chunk size in utils : In now, get chunks by len of output nodes, because the chunk result could be differ from the arguements
- test flow

TODO
- symbolic mask and test of it (in test_model_pruning_analysis)

daniil-lyakhov:
@sooahleex, thank you for your contribution! 

Please don't forget to update correspondent nncf tensorflow part as well

nncf/common/graph/layer_attributes.py
```diff
@@ -187,3 +187,15 @@ def __init__(self,
...
+    """
+    This class stores attributes of split/chunk modules/layers
+    that are useful for some algorithms.
+    """
+
+    def __init__(self,
+                 chunks: Union[int, List],
+                 dim: int=0):
+        self.chunks = chunks
+        self.dim = dim
```
daniil-lyakhov, line 201:
```suggestion
        self.dim = dim

```

nncf/common/graph/utils.py
```diff
@@ -68,3 +68,30 @@ def get_first_nodes_of_type(graph: NNCFGraph, op_types: List[str]) -> List[NNCFN
...
+    :returns: Split/Chunk axis of given split/chunk node.
+    """
+    axis = None
+    none_dim = None
+    for idx, (dim_in, dim_out) in enumerate(zip(input_shapes[0], output_shapes[0])):
+        if dim_in != dim_out:
+            axis = idx
+            break
+        if dim_in is None:
+            none_dim = idx
```
daniil-lyakhov, line 88: Could you please explain when this case happens?

nncf/common/pruning/operations.py
```diff
@@ -222,6 +223,42 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        chunk_size = node.layer_attributes.chunks
+        chunk_axis = node.layer_attributes.axis
+
+        result_masks = tensor_processor.split(input_mask, chunk_size, chunk_axis)
+        if len(set(output_shapes)) == 1:
+            for i, result_mask in enumerate(result_masks):
+                next_nodes[i].data['input_mask'] = result_mask
+        else:
+            idx = output_shapes.index(result_mask.shape[0])
+            next_nodes[idx].data['input_mask'] = result_mask
```
daniil-lyakhov, line 254: Looks like `result_mask` isn't defined here

nncf/common/pruning/symbolic_mask.py
```diff
@@ -109,3 +111,9 @@ def elementwise_mask_propagation(cls, input_masks: List[SymbolicMask]) -> Symbol
                 return AmbiguousSymbolicMask(producers)
 
         return SymbolicMask(input_masks[0].shape[0], producers)
+
+    @classmethod
+    def split(cls, tensor: SymbolicMask, chunks: int, axis: int) -> List[SymbolicMask]:
+        producer = tensor.mask_producers
+        tensor_shape = math.floor(tensor.shape[0]/chunks)
+        return [SymbolicMask(tensor_shape, producer) for _ in range(chunks)]
```
daniil-lyakhov, line 119:
```suggestion
        producers = tensor.mask_producers
        tensor_shape = np.floor(tensor.shape[0]/chunks)
        return [SymbolicMask(tensor_shape, producers) for _ in range(chunks)]
```

nncf/common/pruning/operations.py
```diff
@@ -222,6 +223,42 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        return result_masks
+
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
+                         tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> None:
+        result_mask = cls.generate_output_masks(node, graph, tensor_processor)       
+        node.data['output_mask'] = result_mask
```
daniil-lyakhov, line 261: `output_mask` become list of masks instead of mask, this could lead to bugs in code

nncf/common/pruning/operations.py
```diff
@@ -222,6 +223,42 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        next_nodes = [edge.to_node for edge in output_edges]
+        output_shapes = [get_input_channels(node) for node in next_nodes]
+
+        chunk_size = node.layer_attributes.chunks
+        chunk_axis = node.layer_attributes.axis
+
+        result_masks = tensor_processor.split(input_mask, chunk_size, chunk_axis)
+        if len(set(output_shapes)) == 1:
+            for i, result_mask in enumerate(result_masks):
+                next_nodes[i].data['input_mask'] = result_mask
```
daniil-lyakhov, line 251: `generate_output_masks` implicitly modifies input masks of nodes, I would prefer to avoid such things. 

Other things - `input_mask` value consistency. `input_mask` wasn't used before to prevent optional `input_mask`s and `output_mask` synchronization

nncf/common/pruning/operations.py
sooahleex: Fix this.

nncf/common/pruning/operations.py
sooahleex: It is already list, but to reduce confusion, modified it to add `s`.

nncf/common/pruning/operations.py
sooahleex: In the case of split, the output mask will show a list containing two masks, like `[[[Mask1], [Mask2]]]`. So, when I tried to check the next node of split, it have list as `input_mask`. I also think cover it in `mask_propagation.py` but, if the size of the result is different, it is difficult to check in here, so I added the `input_mask` value, and the `input_mask` will be included in the `node.data` only if there is multiple output such as split. So I put this in the `get_input_masks` function.

nncf/common/graph/utils.py
```diff
@@ -68,3 +68,28 @@ def get_first_nodes_of_type(graph: NNCFGraph, op_types: List[str]) -> List[NNCFN
...
+def get_split_axis(input_shapes: List[List[int]], output_shapes: List[List[int]]) -> int:
+    """
+    Returns split/chunk axis by given input and output shape of split/chunk node.
+
+    :param input_shapes:
+    :param output_shapes:
+    :returns: Split/Chunk axis of given split/chunk node.
+    """
+    axis = None
+    none_dim = None
```
daniil-lyakhov, line 82: `none_dim` looks redundant

nncf/common/pruning/operations.py
daniil-lyakhov: > In the case of split, the output mask will show a list containing two masks, like `[[[Mask1], [Mask2]]]`. So, when I tried to check the next node of split, it have list as `input_mask`. I also think cover it in `mask_propagation.py` but, if the size of the result is different, it is difficult to check in here, so I added the `input_mask` value, and the `input_mask` will be included in the `node.data` only if there is multiple output such as split. So I put this in the `get_input_masks` function.

Yes, there was no lists in input_masks before for nodes like conv and etc., but the solution with input_masks setting is broking logic of `propagate_mask` method: now it's not only set output masks for current node but set input masks for following nodes. I believe solution with only output masks, which has reference to following node is clearer (reference could be presented as output port or by nnncf_node idx) and will solve ambiguity in following case:

image

in approach you implemented, support of such pattern require a list of input masks maintenance but in "keep output masks only" this case will work well. 

As a conclusion:
output masks of node calculated and set only once during `propagate_masks` call due to direction of propagation (top->down), but `input_masks` depends on all node inputs and should be updated several times if node has several inputs.
I suggest to use "keep output masks only" strategy

nncf/common/graph/utils.py
sooahleex: In my thought, `none_dim` is necessary in the clause `if axis is None:`. Do you think this clause is unnecessary? If there are no exceptions at all, I think it would be better to remove this clause altogether. If `axis` can't figure it out above, it seems necessary to make an exception through this section.

nncf/common/pruning/operations.py
daniil-lyakhov: By the way, chunk op expand possible prunable cases, and case bellow will broke symbolic mask propagation due to assert which checks that prunable node has only one input mask. @sooahleex, please check this pattern and support it in symbolic mask propagation

sooahleex: I see. Then ㅑf the sizes of chunk results are not all the same, and one is different in the last case, then the mask sizes will be different and not all the same for the results, so how do we deal with this case? It seems that consideration is needed at the place where the symbolic mask is processed, like `symbolic_mask_propagation` function.

nncf/common/graph/utils.py
daniil-lyakhov: My concern is that `none_dim` is constant None, it isn't modified anywhere here

nncf/common/pruning/operations.py
daniil-lyakhov: I guess some mask merging procedure is needed in `symbolic_mask_propagation` as well as in framework specific `mask_propagation`

daniil-lyakhov: This logic is main "gothca" of chunk ops support

nncf/common/pruning/symbolic_mask.py
sooahleex: In this part, we need to use math instead of numpy. When we use numpy, the bool result in line 117 of mask_propagation.py could be np.bool_.

nncf/common/graph/utils.py
sooahleex: I remove for `none_dim` in here. Actually, I added it cause `concat` case consider `none_dim` case, but I think, `none_dim` is not needed in `chunk` case.

nncf/common/pruning/operations.py
sooahleex: I updated not to use `input_mask` and merge input masks together using union. I am little bit confused to union in here, not like elementwise.

nncf/common/pruning/operations.py
```diff
@@ -222,6 +222,34 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+    
+    @classmethod
+    def generate_output_masks(cls, node: NNCFNode, graph: NNCFGraph,
+                              tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> Optional[NNCFTensor]:
+        """
+        Generate output mask
+        """
+        input_edges = graph.get_input_edges(node)
+        previous_nodes = [edge.from_node for edge in input_edges]
+        input_mask = previous_nodes[0].data['output_mask']
```
daniil-lyakhov, line 238: Please use `get_input_masks` method here

tests/tensorflow/test_model_converter.py
```diff
@@ -11,6 +11,7 @@
  limitations under the License.
 """
 
+from curses import noecho
```
daniil-lyakhov, line 14:
```suggestion
```

tests/tensorflow/test_model_converter.py
```diff
@@ -22,14 +23,16 @@
 from nncf.common.graph import INPUT_NOOP_METATYPES
 from nncf.common.graph import OUTPUT_NOOP_METATYPES
 from nncf.common.graph.layer_attributes import MultipleInputLayerAttributes
-from nncf.tensorflow.graph.metatypes.common import RESHAPE_METATYPES
+from nncf.tensorflow.graph.metatypes.common import LAYER_METATYPES_AGNOSTIC_TO_DATA_PRECISION_WITH_MULTIPLE_CONCAT_INPUTS, RESHAPE_METATYPES
```
daniil-lyakhov, line 26:
```suggestion
from nncf.tensorflow.graph.metatypes.common import LAYER_METATYPES_AGNOSTIC_TO_DATA_PRECISION_WITH_MULTIPLE_CONCAT_INPUTS
from nncf.tensorflow.graph.metatypes.common import RESHAPE_METATYPES
```

daniil-lyakhov:
Please add tests to torch/pruning/filter_pruning/test_algo.py

sooahleex:
> Please add tests to torch/pruning/filter_pruning/test_algo.py

@daniil-lyakhov  I added more cases for split/chunk models in tests/torch/pruning/helpers.py.
ex. 
image

sooahleex:
run pytorch pre-commit tests

nncf/common/graph/utils.py
```diff
@@ -68,3 +68,24 @@ def get_first_nodes_of_type(graph: NNCFGraph, op_types: List[str]) -> List[NNCFN
...
         first_nodes_of_type.extend(graph.traverse_graph(root, partial_traverse_function))
     return first_nodes_of_type
+
+
+def get_split_axis(input_shapes: List[List[int]], output_shapes: List[List[int]]) -> int:
+    """
+    Returns split/chunk axis by given input and output shape of split/chunk node.
+
+    :param input_shapes:
+    :param output_shapes:
```
daniil-lyakhov, line 78: Please don't forget to add description

nncf/common/graph/utils.py
```diff
@@ -68,3 +68,24 @@ def get_first_nodes_of_type(graph: NNCFGraph, op_types: List[str]) -> List[NNCFN
...
+    """
+    axis = None
+    for idx, (dim_in, dim_out) in enumerate(zip(input_shapes[0], output_shapes[0])):
+        if dim_in != dim_out:
+            axis = idx
+            break
+
+    if axis is None:
+        axis = -1
+        logger.warning('Identity split/concat node detected')
```
daniil-lyakhov, line 89: 👍

nncf/common/pruning/utils.py
```diff
@@ -636,4 +645,44 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
...
+    else:
+        for split_mask in output_masks:
+            idx = output_shapes.index(split_mask.shape[0])
+            result_masks[next_nodes[idx].node_name] = split_mask
+            # update already matched
+            output_shapes[idx]=None
+
+    return list(result_masks.items())
+
+def find_input_mask_for_node(input_mask: List, node: NNCFNode) -> SymbolicMask:
```
daniil-lyakhov, line 679: Please specify type of input_mask elements
```suggestion
def find_input_mask_for_node(input_mask: List[ElemType], node: NNCFNode) -> SymbolicMask:
```

nncf/common/pruning/symbolic_mask.py
```diff
@@ -109,3 +109,8 @@ def elementwise_mask_propagation(cls, input_masks: List[SymbolicMask]) -> Symbol
                 return AmbiguousSymbolicMask(producers)
 
         return SymbolicMask(input_masks[0].shape[0], producers)
+
+    @classmethod
+    def split(cls, tensor: SymbolicMask, chunks: int, axis: int, output_shapes: List[int]=None) -> List[SymbolicMask]:
```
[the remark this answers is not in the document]
sooahleex: Fix

nncf/common/pruning/utils.py
sooahleex: Fix

nncf/common/pruning/mask_propagation.py
```diff
@@ -110,6 +111,10 @@ def symbolic_mask_propagation(self, prunable_layers_types: List[str],
                     assert len(input_masks) == 1
                     input_mask = input_masks[0] # type: SymbolicMask
 
+                    # find input_mask for node if multiple input masks existed
+                    if isinstance(input_mask, list):
+                        input_mask = find_input_mask_for_node(input_mask, node)
+
```
daniil-lyakhov, line 117: Can we resolve this inside `get_input_masks`?

nncf/common/pruning/utils.py
```diff
@@ -636,4 +645,44 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
...
+        for i, split_mask in enumerate(output_masks):
+            result_masks[next_nodes[i].node_name] = split_mask
+    else:
+        for split_mask in output_masks:
+            idx = output_shapes.index(split_mask.shape[0])
+            result_masks[next_nodes[idx].node_name] = split_mask
+            # update already matched
+            output_shapes[idx]=None
+
+    return list(result_masks.items())
```
daniil-lyakhov, line 677: I don't feel comfortable to match output masks to input nodes by shape, it could lead to mistakes in case shapes are equal and masks values are different on each branch. Can we keep index of input node for masks instead of mapping?

nncf/common/pruning/utils.py
```diff
@@ -636,4 +645,44 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
...
+
+def find_input_mask_for_node(input_mask: List[Tuple['str', SymbolicMask]], node: NNCFNode) -> SymbolicMask:
+    """
+    Find exact input mask for node among multiple input masks.
+
+    :param input_mask: Given input masks.
+    :return: Exact input mask for node.
+    """
+    node_name = node.node_name
+    input_mask = dict(input_mask)
```
daniil-lyakhov, line 687: This conversation from list of tuples to dict looks ambiguous, can we keep dictionary {node_name: SymbolicMask}?

daniil-lyakhov:
General comment:

Masks matching by shape is implicit method of matching, I suggest you to 
1) Have SymbolicMask (in case op has one ouput edge) or {Nodename: SymbolicMask} (in case op has several output edges) as value kept in node.data['output_mask']
2) use `get_input_masks`  everywhere to get input masks for node resolving all dependencies inside this util function

nncf/common/pruning/utils.py
sooahleex: The mapping process is absolutely necessary. After spliting the mask, we can not miss the process of finding the right mask for which node. It doesn't seem to matter which node the mask we found fits, whether it is index or node name. When we look at the shape of the mask, we don't look at the entire shape, we only look at the axis that we split.

nncf/common/pruning/mask_propagation.py
sooahleex: I also thought that, but in some case we use `get_input_masks` to get all input masks(multiple input case), literally, but other cases like this, we tried to get a input mask (singe input case). So If we resolve this inside `get_input_masks`, it could not cover both of those cases.

nncf/common/pruning/utils.py
sooahleex: Fix.

nncf/common/pruning/utils.py
```diff
@@ -594,6 +596,13 @@ def get_input_masks(node: NNCFNode, graph: NNCFGraph) -> List[Optional[NNCFTenso
     return input_masks
 
 
+def get_output_mask(node: NNCFNode) -> Optional[NNCFTensor]:
+    output_mask = node.data['output_mask']
+    if isinstance(output_mask, dict):
+        output_mask = find_input_mask_for_node(output_mask, node)
+    return output_mask
+
+
```
daniil-lyakhov, line 605: Please remove this function

nncf/common/pruning/mask_propagation.py
daniil-lyakhov: Can we do something like this?

```
def get_input_masks(node, graph) -> List[NNCFTensor]:
    retval = []
    input_masks = [input_node.data['output_mask'] for input_node in graph.get_previous_nodes(node)]
    for input_mask in input_masks:
        retval.append(input_mask[node.node_name] if isinstance(input_mask, dict) else input_mask)
    return retval
```

Len of retval equal to amount of input edges, so we can resolve input mask for each input edge

nncf/common/pruning/operations.py
```diff
@@ -223,6 +229,45 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+            return None
+
+        chunk_size = node.layer_attributes.chunks
+        chunk_axis = node.layer_attributes.axis
+
+        output_edges = graph.get_output_edges(node)
+        output_shapes = [edge.tensor_shape[chunk_axis] for edge in output_edges]
+
+        split_masks = tensor_processor.split(input_mask, chunk_size, chunk_axis, output_shapes)
+        result_masks = match_multiple_output_masks(split_masks, output_edges, chunk_axis)
```
daniil-lyakhov, line 260: `match_multiple_output_masks` is used only here, could you please make this function a classmethod for `SplitPruningOp`?

nncf/common/pruning/utils.py
```diff
@@ -636,4 +638,43 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
         # In case for disconnected NNCFGraph
         input_masks = [None]
     assert len(input_masks) == 1
-    node.data['output_mask'] = input_masks[0]
+    input_mask = input_masks[0]
+
+    if isinstance(input_mask, dict):
+        input_mask = find_input_mask_for_node(input_mask, node)
+    node.data['output_mask'] = input_mask
```
daniil-lyakhov, line 645: This will be redundant if `get_input_masks` will be implemented like here: https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r948793387

nncf/common/pruning/utils.py
```diff
@@ -636,4 +638,43 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
...
+
+def find_input_mask_for_node(input_mask: Dict['str', SymbolicMask], node: NNCFNode) -> SymbolicMask:
+    """
+    Find exact input mask for node among multiple input masks.
+
+    :param input_mask: Given input masks.
+    :return: Exact input mask for node.
+    """
+    node_name = node.node_name
+    return input_mask[node_name]
```
daniil-lyakhov, line 680: I suggest to use `[]` operator instead of this function

nncf/common/pruning/utils.py
```diff
@@ -636,4 +638,43 @@ def identity_mask_propagation(node: NNCFNode, graph: NNCFGraph) -> None:
...
+    :return: Matched output mask for each next node.
+    """
+    output_shapes = [edge.tensor_shape[chunk_axis] for edge in output_edges]
+    next_nodes = [edge.to_node for edge in output_edges]
+    result_masks = {node.node_name: None for node in next_nodes}
+
+    if len(set(output_shapes)) == 1:
+        for i, split_mask in enumerate(output_masks):
+            result_masks[next_nodes[i].node_name] = split_mask
+    else:
```
daniil-lyakhov, line 663: What if output_shapes will be like [1, 1, 2]?

nncf/common/pruning/mask_propagation.py
```diff
@@ -110,6 +111,10 @@ def symbolic_mask_propagation(self, prunable_layers_types: List[str],
                     assert len(input_masks) == 1
                     input_mask = input_masks[0] # type: SymbolicMask
 
+                    # find input_mask for node if multiple input masks existed
+                    if isinstance(input_mask, dict):
+                        input_mask = find_input_mask_for_node(input_mask, node)
+
```
daniil-lyakhov, line 117: This will be redundant if get_input_masks will be implemented like here: https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r948793387

nncf/common/pruning/operations.py
```diff
@@ -199,12 +201,16 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
...
+        input_masks = [find_input_mask_for_node(mask, node) if isinstance(mask, dict) else mask for mask in input_masks]
 
         not_empty_masks = [mask for mask in input_masks if mask is not None]  # type: List[NNCFTensor]
         if not not_empty_masks:
             return None
 
         first_non_empty_mask = not_empty_masks[0]
+        if isinstance(first_non_empty_mask, dict):
+            first_non_empty_mask = find_input_mask_for_node(first_non_empty_mask, node)
+
```
daniil-lyakhov, line 213: This will be redundant if get_input_masks will be implemented like here: https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r948793387

nncf/common/pruning/tensor_processor.py
```diff
@@ -74,3 +74,15 @@ def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTens
...
+
+    @classmethod
+    @abstractmethod
+    def split(cls, tensor: NNCFTensor, chunks: int, axis: int, output_shapes: List[int] = None) -> List[NNCFTensor]:
+        """
+        Split/chunk NNCFTensor into chunks along an exsiting dimension.
+
+        :param tensor: Given NNCFTensor.
+        :param chunks: The number of chunks
+        :param dims: The axis along which the tensors will be splited
```
daniil-lyakhov, line 86:
```suggestion
        :param axis: The axis along which the tensors will be splited
```

nncf/common/pruning/tensor_processor.py
```diff
@@ -74,3 +74,15 @@ def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTens
         :param input_masks: Given input masks.
         :return: Elementwise pruning operation output mask.
         """
+
+    @classmethod
+    @abstractmethod
+    def split(cls, tensor: NNCFTensor, chunks: int, axis: int, output_shapes: List[int] = None) -> List[NNCFTensor]:
```
daniil-lyakhov, line 80: Please remove `output_shapes` param and calculate in in `SymbolicMaskProcessor.Split` by `chunks` param

sooahleex: I hope so too, but for the symbolic mask processor, we could not know the result size of the split/chunk operation. The `output_shapes` must be received because the calculation result clearly differs from the chunk size input. In the `SymbolicMaskProcesor`, we could not use chunk or split operations here like `torch.chunk` or `tf.split`, so we need to receive the `output_shapes`.

daniil-lyakhov: Ok, can we than unify this function across frameworks + SymbolicMaskProcessor by using params
(`tesnor`, `output_shapes`, `axis`)? Looks like `chunks` is just `len(output_shapes)`

nncf/common/pruning/utils.py
sooahleex: I could not catch up yours. But in this case, I tested the case of like below.
Input of chunk operation : C=16 -> Output :C=6, C=6, C=4.
For output_shapes, I only consider the size of axis which the operation performed.
`output_shapes = [edge.tensor_shape[chunk_axis] for edge in output_edges]`
That's why I've made it possible to consider cases where there are multiple outputs of the same size.
Please tell me if you mean something other than this case.

nncf/tensorflow/pruning/tensor_processor.py
```diff
@@ -50,3 +50,10 @@ def repeat(cls, tensor: NNCFTensor, repeats: int) -> NNCFTensor:
     def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTensor:
         cls.assert_allclose(input_masks)
         return input_masks[0]
+
+    @classmethod
+    def split(cls, tensor: NNCFTensor, chunks: int, axis: int, output_shapes: List[int] = None) -> List[NNCFTensor]:
+        if axis >= len(tensor.tensor.shape):
+            axis = -1
```
daniil-lyakhov, line 57: When this happens? I preferer to have an exception instead of this if

sooahleex: This happened frequently. We usually set axis as -1 but during `get_split_axis`, it could be considered like axis=3. And in this case, there could be occurred the error. So I put this if clause to prevent this case.

nncf/common/pruning/tensor_processor.py
sooahleex: Yes, right. But some case, we did not put the `output_shapes` as input. So I set `output_shapes` usually None. I want to use `chunks` as much as possible and avoid putting `output_shape`.

nncf/common/pruning/mask_propagation.py
sooahleex: Fix.

nncf/common/pruning/operations.py
sooahleex: Fix.

nncf/common/pruning/utils.py
sooahleex: Fix.

nncf/common/pruning/utils.py
sooahleex: Fix.

nncf/common/pruning/operations.py
sooahleex: Fix.

nncf/common/pruning/tensor_processor.py
sooahleex: Fix.

nncf/common/pruning/mask_propagation.py
sooahleex: Fix.

nncf/common/pruning/utils.py
daniil-lyakhov: Ok, got it.


Can we do it without if statement in such manner?
```
result_mask = dict()
tmp_out_masks = out_masks.copy()
for edge in output_edges:
    node_name = edge.to_node.node_name
    idx = tmp_out_masks.index(edge.tesnsor_shape[chunk_axis])
    result_mask[node_name] = tmp_out_masks.pop(idx)

return result_mask
```

nncf/common/pruning/utils.py
daniil-lyakhov: This function could be removed

nncf/tensorflow/pruning/tensor_processor.py
daniil-lyakhov: Could you please present an example?

nncf/common/pruning/operations.py
```diff
@@ -199,12 +201,16 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
...
+        input_masks = [mask[node.node_name] if isinstance(mask, dict) else mask for mask in input_masks]
 
         not_empty_masks = [mask for mask in input_masks if mask is not None]  # type: List[NNCFTensor]
         if not not_empty_masks:
             return None
 
         first_non_empty_mask = not_empty_masks[0]
+        if isinstance(first_non_empty_mask, dict):
+            first_non_empty_mask = first_non_empty_mask[node.node_name]
+
```
daniil-lyakhov, line 213: This change could be removed completely, please check

nncf/common/pruning/operations.py
```diff
@@ -223,6 +229,70 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
 
+class SplitPruningOp(BasePruningOp):
+    @classmethod
+    def match_multiple_output_masks(cls, output_masks: List[SymbolicMask], output_edges: List[NNCFGraphEdge],
+                                    chunk_axis: int) -> Dict['str', SymbolicMask]:
+        """
+        Match multiple input mask to each next nodes.
+
+        :param output_masks: Given output masks.
+        :return: Matched output mask for each next node.
```
daniil-lyakhov, line 240: Please complete this doc

nncf/tensorflow/pruning/tensor_processor.py
sooahleex: For example, in the `SplitConcatModel` or `MultipleSplitConcatModel` ,  the models of `tests/torch/pruning.test_model_pruning_analysis.py`.  In this case, the axis is set as `1`, but the `len(tensor.tensor.shape)` is also 1, so the if we did not change axis as `-1`, the chunk/split operation could not be performed.

daniil-lyakhov: But we know here that tensor always has only one dimension, so we don't need `axis` param here

daniil-lyakhov: Let's make it similar to concat axis: `axis` param is required but always 0 is used as its value. We will use this param when NNCFTensor become more than 1D

nncf/common/pruning/tensor_processor.py
daniil-lyakhov: My concern is that we call function only in one place and always feed all parameters. And depending on realization (PT, TF or Symbolic) different subsets of params is used. `axis` param is needed (possibly) for non symbolic tensors and don't needed for symbolic tensors, but in case of params `chunks` and `output_shapes` - they are interchangeable. And as far as I understand realization just takes one of params which is more compostable for it's realization. This is logical overhead for me

nncf/common/pruning/utils.py
sooahleex: I fixed it following your advice. Please let me know if there is anything else to fix.

nncf/common/pruning/tensor_processor.py
sooahleex: I reduced the arguments of the split functions only `output_shapes`. As you said, the mask size is always one dimension, so I reduced the axis and use it as default value (`axis=0`). And remove the `chunks` to be replaced by `len(output_shapes)`.

nncf/tensorflow/pruning/tensor_processor.py
sooahleex: Remove this as you said, And I explained it in [conversation](https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r949806730).

nncf/common/pruning/operations.py
sooahleex: Fix.

nncf/common/pruning/operations.py
sooahleex: Fix it. Thank you for letting me know this.

nncf/common/pruning/operations.py
```diff
@@ -223,6 +227,69 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        """
+        Match multiple input mask to each next nodes.
+
+        :param output_masks: Given output masks.
+        :param output_edges: Given output edges of the node.
+        :param chunk_axis: Given the axis on which operation was performed.
+        :return: Matched output mask for each next node.
+        """
+        next_nodes = [edge.to_node for edge in output_edges]
+        result_masks = {node.node_name: None for node in next_nodes}
```
daniil-lyakhov, line 243:
```suggestion
```

nncf/common/pruning/operations.py
```diff
@@ -223,6 +227,69 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+
+    @classmethod
+    def generate_output_masks(cls, node: NNCFNode, graph: NNCFGraph,
+                              tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> Optional[NNCFTensor]:
+        """
+        Generate output mask from input masks for split/chunk operations.
+        If input mask is None return None
+
+        :param node: Node to determine it's sources.
+        :param graph: NNCF graph to work with
```
daniil-lyakhov, line 268:
```suggestion
        :param node: Node to determine its sources.
        :param graph: NNCF graph to work with.
```

nncf/common/pruning/operations.py
```diff
@@ -223,6 +227,69 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        :param tensor_processor: Interface with tensor processing methods.
+        :return: Filled input masks.
+        """
+        input_mask = get_input_masks(node, graph)[0]
+        if not input_mask:
+            return None
```
daniil-lyakhov, line 274: `get_input_masks` could return empty list

nncf/common/pruning/tensor_processor.py
```diff
@@ -74,3 +74,15 @@ def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTens
...
+    @classmethod
+    @abstractmethod
+    def split(cls, tensor: NNCFTensor, output_shapes: List[int]) -> List[NNCFTensor]:
+        """
+        Split/chunk NNCFTensor into chunks along an exsiting dimension.
+
+        :param tensor: Given NNCFTensor.
+        :param chunks: The number of chunks
+        :param axis: The axis along which the tensors will be splited
+        :returns: The list of NNCFTensor which is split
```
daniil-lyakhov, line 87: Please update doc

nncf/common/pruning/utils.py
```diff
@@ -33,6 +33,7 @@
 from nncf.common.pruning.clusterization import Cluster
 from nncf.common.pruning.clusterization import Clusterization
 from nncf.common.pruning.structs import PrunedLayerInfoBase
+from nncf.common.pruning.symbolic_mask import SymbolicMask
```
daniil-lyakhov, line 36:
```suggestion
```

nncf/common/pruning/operations.py
```diff
@@ -199,12 +201,14 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
         input_edges = graph.get_input_edges(node)
         previous_nodes = [edge.from_node for edge in input_edges]
         input_masks = [input_node.data['output_mask'] for input_node in previous_nodes]
+        input_masks = [mask[node.node_name] if isinstance(mask, dict) else mask for mask in input_masks]
```
daniil-lyakhov, line 204: Please employ `get_input_masks` here

nncf/common/pruning/operations.py
sooahleex: Fix.

nncf/common/pruning/operations.py
sooahleex: Fix. I could not think this part. Thanks.

nncf/common/pruning/operations.py
sooahleex: I see. Thank you for checking this.

nncf/common/pruning/operations.py
```diff
@@ -223,6 +225,63 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        """
+        Generate output mask from input masks for split/chunk operations.
+        If input mask is None return None
+
+        :param node: Node to determine its sources.
+        :param graph: NNCF graph to work with.
+        :param tensor_processor: Interface with tensor processing methods.
+        :return: Filled input masks.
+        """
+        input_mask = get_input_masks(node, graph)[0]
```
daniil-lyakhov, line 267: Last comment from my side:
let's check input masks exist before taking it
```suggestion
        input_masks = get_input_masks(node, graph)
        if not input_masks:
            return None
 
        input_mask = input_masks[0]
```

nncf/common/pruning/tensor_processor.py
```diff
@@ -74,3 +74,13 @@ def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTens
...
         """
+
+    @classmethod
+    @abstractmethod
+    def split(cls, tensor: NNCFTensor, output_shapes: List[int]) -> List[NNCFTensor]:
+        """
+        Split/chunk NNCFTensor into chunks along an exsiting dimension.
+
+        :param tensor: Given NNCFTensor.
+        :returns: The list of NNCFTensor which is split
```
daniil-lyakhov, line 85: Please describe `output_shapes` parameter

nncf/common/pruning/operations.py
```diff
@@ -197,14 +199,14 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
         :return: Filled input masks.
         """
         input_edges = graph.get_input_edges(node)
```
daniil-lyakhov, line 201:
```suggestion
```

sooahleex: It seems that `input_edges` is used in the line 215.

daniil-lyakhov: You are right, sorry

daniil-lyakhov (approved):
Good job, looks good to me 👍

ljaljushkin:
I wonder how the algorithm will work when chunk axis doesn't match channel axis, according to which pruning occurs.
What will happen with pruning mask and propagation algorithm?
I think it makes sense to cover in tests.

nncf/common/pruning/operations.py
```diff
@@ -223,6 +225,66 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+        :return: Filled input masks.
+        """
+        input_masks = get_input_masks(node, graph)
+        if not input_masks:
+            return None
+        input_mask = input_masks[0]
+        chunk_axis = node.layer_attributes.axis
+
+        output_edges = graph.get_output_edges(node)
+        output_shapes = [edge.tensor_shape[chunk_axis] for edge in output_edges]
```
ljaljushkin, line 274: Will it work as expected if axis =-1?
https://github.com/openvinotoolkit/nncf/pull/1244/files#diff-033daadd941b8d5ac07994ca370f200c90f8a8f5708edd49f0f20ed93d0756ebR88

ljaljushkin (changes requested):
Let me follow up on my 2 previous comments and add more clarifications.

Seems like this PR considers only one very common case, when there's Conv with shapes of weights [N, C, H, W], we prune channel dimension C and then apply Split of this dimension.
Current code always shrinks the pruning mask in accordance with split axis, which is not always correct.

I propose to consider other rare cases and cover them by tests. 
@daniil-lyakhov, please comment which tests should be extended.

1) weights have shape [N, C, C, W] and split dimension is not 1, but 2.
Mask propagation should fail because of inconsistency of number of channels (C) and length of the resulting mask (C/2).
2) weights have shape [N, C, 2C, W] and split dimension is not 1, but 2.
Mask propagation won't fail with the current code, but pruning will be invalid.
3) Identity split. In the current code, split axis = -1, it would lead to incorrect calculation of output shapes.

daniil-lyakhov:
@sooahleex, to test this please update `test_model_pruning_analysis.py` with model/models which have split operations described in previous comment

nncf/common/pruning/operations.py
```diff
@@ -265,9 +265,9 @@ def generate_output_masks(cls, node: NNCFNode, graph: NNCFGraph,
         :return: Filled input masks.
         """
         input_masks = get_input_masks(node, graph)
-        if not input_masks:
-            return None
```
daniil-lyakhov: I believe that it could be not safe because `get_input_masks` could return empty list

nncf/common/pruning/mask_propagation.py
```diff
@@ -112,7 +112,7 @@ def symbolic_mask_propagation(self, prunable_layers_types: List[str],
 
                     for producer in input_mask.mask_producers:
                         previously_dims_equal = True if can_prune_by_dim[producer] is None \
-                            else can_prune_by_dim[producer]
+                            else can_prune_by_dim[producer].decision
```
daniil-lyakhov, line 115: `PruningAnalysisDesicion` object implements `__bool__` method, so this change is ambiguous

tests/torch/pruning/helpers.py
```diff
@@ -597,13 +597,91 @@ def forward(self, x):
...
 
 
+class SplitModel(nn.Module):
+    #         (input)
+    #            |
+    #         (conv1)
+    #            |
+    #         (chunk)
+    #        /      \
+    #    (conv2)  (conv3)
```
daniil-lyakhov, line 607: Looks like you don't use this model anywhere in tests

tests/torch/pruning/helpers.py
```diff
@@ -597,13 +597,91 @@ def forward(self, x):
...
+        self.conv2 = create_conv(4, 8, 1, 1)
+        self.conv3 = create_conv(4, 8, 1, 1)
+
+    def forward(self, x):
+        x = self.conv1(x)
+        y1, y2 = torch.chunk(x, chunks=2, dim=2)
+
+        y1 = self.conv2(y1)
+        y2 = self.conv3(y2)
+        return y1, y2
```
daniil-lyakhov, line 648: Shape of the x after `conv1` is `(1, 4, 5, 5)`, so  `generate_output_masks` from mask with shape `4` generates two masks with dims `[2, 3]` (because 5 is splitted on two.
1) In such case split should not propagate any masks through. You need to add check inside `generate_output_masks`
2) This case isn't a case with `[N, C, C, W]` tensor, please correct the model

nncf/common/pruning/operations.py
sooahleex: I changed this to cover `[None]` case. I updated `generate_input_mask` function to cover empty list.

tests/torch/pruning/helpers.py
sooahleex: I understood it. I will change it like `self.conv1 = create_conv(1, 6, 3, 1)`. In this case, the split axis is `dim=2` and it seems like it could not proceed mask propagation more. Then in my thought, it seems necessary to make sure that the split axis fits with channel axis. What do you think?

daniil-lyakhov: I agree with you. In ` generate_output_masks` should be a check like

```
input_mask.shape[0] == sum(output_edges_shapes)
```

daniil-lyakhov: Oh, actually, case illustrated bellow will make pruning invalid because reshape mask propagation will just pass through existing masks, so pruning will be thinking that everything is ok 
image

sooahleex: I added this case in the `test_model_pruning_analysis` with `SplitReshapeModel` and cover it in `_is_not_mixing_dim` in common/pruning/operations.py.

sooahleex:
I fix the error which occurred due to [PR](https://github.com/openvinotoolkit/nncf/pull/1257).

nncf/common/pruning/operations.py
```diff
@@ -311,6 +313,22 @@ class ReshapePruningOp(BasePruningOp):
...
 
+    @staticmethod
+    def _is_not_mixing_dim(node: NNCFNode, graph) -> bool:
+        input_edges = graph.get_input_edges(node)
+        input_shape = input_edges[0].tensor_shape
+
+        output_edges = graph.get_output_edges(node)
+        if not output_edges:
+            return True
+        output_shape = output_edges[0].tensor_shape
```
daniil-lyakhov, line 324: please use `node.layer_attributes` here

nncf/common/pruning/operations.py
```diff
@@ -311,6 +313,22 @@ class ReshapePruningOp(BasePruningOp):
...
+
+        if len(input_shape)==len(output_shape) and set(input_shape)==set(output_shape):
+            input_shape_idx = np.argsort(input_shape)
+            output_shape_idx = np.argsort(output_shape)
+            return np.allclose(input_shape_idx, output_shape_idx)
```
daniil-lyakhov, line 329: Case (a, b) -> (b, a, 1) will return True here. 

Please implement check by dims alignment like this:

Positive case
image

Negative case
image

nncf/common/pruning/utils.py
```diff
@@ -595,7 +595,7 @@ def get_input_masks(node: NNCFNode, graph: NNCFGraph) -> List[Optional[NNCFTenso
     for input_mask in input_masks:
         retval.append(input_mask[node.node_name] if isinstance(input_mask, dict) else input_mask)
     if not retval:
-        return None
+        return [None]
```
daniil-lyakhov, line 598: Please just return retval in case `retval==[]`

nncf/common/pruning/operations.py
```diff
@@ -199,7 +199,10 @@ def generate_output_mask(cls, node: NNCFNode, graph: NNCFGraph,
         :return: Filled input masks.
         """
         input_edges = graph.get_input_edges(node)
-        input_masks = get_input_masks(node, graph)
+        previous_nodes = [edge.from_node for edge in input_edges]
+        input_masks = [input_node.data['output_mask'] for input_node in previous_nodes]
+        input_masks = [input_mask[node.node_name] if isinstance(input_mask, dict) else input_mask
+                       for input_mask in input_masks]
```
daniil-lyakhov, line 207: Please use `get_input_masks`

nncf/common/pruning/operations.py
sooahleex: To check all of the case in here is so complicated. Also, there are not so much case with split. I do not understand we need to cover all case in here which is not realistic.

nncf/common/pruning/operations.py
daniil-lyakhov: The logic of input/output shape collection is inside NNCFGraph building already. output_shape is required attribute of `ReshapeLayerAttributes`

sooahleex: I know. But as you seen, we need to check whether `output_edges` are existed or not. There is the case that `output_edges` are not existed but `output_shape` is existed. In this case, we need to return True.

daniil-lyakhov: You can check that layer attributes is existed, thus input_shape and output_shape are collected. `input_shapes` and `output_shapes` are required attributes of `ReshapeLayerAttributes`

sooahleex: In some case, `output_shapes` are existed but `output_edges` are not existed. `tests/common/pruning/test_pruning_operations.py::test_reshape_metatype_mask_prop` is the one of the case. If we did not check whether `output_edges` are existed or not, and just use `output_shapes` of `ReshapeLayerAttributes`, we will return False in here, and the test could be failed. So we need to check `output_edges` are existed or not in here. If you do not want to use `output_shapes` like this, I will change this like below. 
```
        output_edges = graph.get_output_edges(node)
        if not output_edges:
            return True
        output_shape = node.layer_attributes.output_shape
```

nncf/common/pruning/operations.py
ljaljushkin: Still can't see how the issue with **identity split** is resolved. @sooahleex, could you please comment on that?

In my understanding, we should do nothing with mask in case of identity split, but instead we take the last shape, since axis = -1 here: https://github.com/openvinotoolkit/nncf/pull/1244/files#diff-033daadd941b8d5ac07994ca370f200c90f8a8f5708edd49f0f20ed93d0756ebR88
The logic looks not valid for that case, doesn't it?

Seems like you need additional logic if chunk_axis == -1

ljaljushkin (changes requested):

sooahleex:
@ljaljushkin I added the `axis=-1` case through `SplitModel` in tests/torch/pruning/helpers.py. Also, in this case, axis will not be set as `-1` but be `3`. It is the same logic of the axis setting in the concat method. If you think the current model is not enough, please let me know how to check it more.

nncf/common/pruning/operations.py
sooahleex: I added the `axis=-1` case through `SplitModel` in tests/torch/pruning/helpers.py. Also, in this case, axis will not be set as `-1` but be `3`. It is the same logic of the axis setting in the concat method. If you think the current model is not enough, please let me know how to check it more.

sooahleex:
run pytorch pre-commit tests

nncf/common/pruning/operations.py
daniil-lyakhov: I think following logic is needed here:
```
if len(output_shapes) == 1: #if identity split detected
 return input_mask # propagate as is
...
```

nncf/common/pruning/operations.py
sooahleex: I reverted the use of `get_input_masks` in here. When I use `get_input_masks` in concat, tests/torch/nas/test_all_elasticity.py is failed. And I found this is due to the order of the result of `graph.get_previous_nodes(node)` and the above one is different. I think this is the reason that the current code in develop also not use `get_input_masks`. https://github.com/openvinotoolkit/nncf/blob/7b1b3b430cdc03e592aacc125282910de97297ee/nncf/common/pruning/operations.py#L199-L201

nncf/common/pruning/operations.py
sooahleex: I have no idea to cover (a, b) -> (b, a, 1). And is there any real case to do like this?

nncf/common/pruning/operations.py
ljaljushkin: I've run all tests in `test_groups` and none of them set axis to -1. 
Added PR (https://github.com/sooahleex/nncf/pull/1) with the test case that handles identity split. 

Actually, it didn't pass before introducing Daniil's suggestion: https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r956010338

sooahleex: [Splitmodel](https://github.com/sooahleex/nncf/blob/548e8cfe76f6b7f78c2dcf3785d743b39ddb97fe/tests/torch/pruning/helpers.py#L616)
I added the model to cover `dim=-1` here. If it is not proper, please add to what you want.

ljaljushkin: @sooahleex seems like there's a misunderstanding. 

I didn't mean `dim == -1` from this line: `y1, y2 = torch.chunk(x, chunks=2, dim=-1)`. 
I meant `axis == -1` from `get_split_axis` here: https://github.com/openvinotoolkit/nncf/pull/1244/files#diff-033daadd941b8d5ac07994ca370f200c90f8a8f5708edd49f0f20ed93d0756ebR88

Yes, both of them about chunk axis, but the meaning is different. 
-1 in the first line means last axis. 
-1 in the second will be only in the case of identity chunk when split do nothing and therefore there's no meaningful chunk axis. And logic with matching shapes was not valid if we take the last axis:
```python
output_shapes = [edge.tensor_shape[chunk_axis] for edge in output_edges]
if not input_mask.shape[0] == sum(output_shapes):
    return None
'''

It started to work with Daniil's suggestion only:
```
if len(output_shapes) == 1: #if identity split detected
 return input_mask # propagate as is
```

The model you mentioned doing ordinary chunk, not identity.
https://github.com/sooahleex/nncf/blob/548e8cfe76f6b7f78c2dcf3785d743b39ddb97fe/tests/torch/pruning/helpers.py#L616
That's why my PR makes sense to include.

**Comment by sooahleex** [comment]
run pytorch pre-commit tests

**Comment by sooahleex** [comment]
I added the case of Split->Reshape model and Identity split case as corner cases.
Split->Reshape model
splitreshapemodel
Identity Split model
identitysplitmodel

File: nncf/common/pruning/operations.py (hunk 45 above)
Comment (sooahleex) [reply nncf/common/pruning/operations.py:334]: As you mean, the case of (1, a, b) -> (1, b, a) is negative case. I already added the case of Split->Reshape model. It could cover (1, 8, 4, 8) -> (1, 4, 8, 8). If you mean I need to cover more case like (a, b*c, d, e) -> (a, b, c, d*e), I just thought the case of the below.
SplitReshapePositive
I'd appreciate it if you could give me your opinion on this model.

**Comment by sooahleex** [comment]
We will cover the reshape check in the different pr related to 90584 ticket. For that ticket and pr, some tests in this pr will be updated for `_is_not_mixing_dim`.

**Comment by daniil-lyakhov** [comment]
> We will cover the reshape check in the different pr related to 90584 ticket. For that ticket and pr, some tests in this pr will be updated for `_is_not_mixing_dim`.

Related to 90976 ticket, the 90584 ticket is not related to this PR

File: nncf/common/pruning/operations.py [hunk 48]
```diff
@@ -244,6 +319,21 @@ class ReshapePruningOp(BasePruningOp):
...
+
+        output_edges = graph.get_output_edges(node)
+        if not output_edges:
+            return True
+        output_shape = node.layer_attributes.output_shape
+
+        if len(input_shape)==len(output_shape) and set(input_shape)==set(output_shape):
+            input_shape_idx = np.argsort(input_shape)
+            output_shape_idx = np.argsort(output_shape)
+            return np.allclose(input_shape_idx, output_shape_idx)
```
daniil-lyakhov, line 334: This is equivalent to this:
```suggestion
if len(input_shape) != len(output_shape):
    return True
return input_shape == output_shape
```

nncf/common/pruning/operations.py
```diff
@@ -244,6 +319,21 @@ class ReshapePruningOp(BasePruningOp):
     def _is_flatten(node: NNCFNode) -> bool:
         return len(node.layer_attributes.output_shape) == 2
 
+    @staticmethod
+    def _is_not_mixing_dim(node: NNCFNode, graph) -> bool:
+        input_shape = node.layer_attributes.input_shape
+
+        output_edges = graph.get_output_edges(node)
+        if not output_edges:
+            return True
```
daniil-lyakhov, line 328: You don't really need this because `layer_attributes` is exist. This is checked in `accept_pruned_input`
```suggestion
```

daniil-lyakhov, line 328: You don't really need this because `layer_attributes` is exist. This is checked in `accept_pruned_input`
```suggestion
```

nncf/common/pruning/operations.py
```diff
@@ -244,6 +319,21 @@ class ReshapePruningOp(BasePruningOp):
...
+        output_shape = node.layer_attributes.output_shape
+
+        if len(input_shape)==len(output_shape) and set(input_shape)==set(output_shape):
+            input_shape_idx = np.argsort(input_shape)
+            output_shape_idx = np.argsort(output_shape)
+            return np.allclose(input_shape_idx, output_shape_idx)
+        return True
```
daniil-lyakhov, line 335: This is equivalent to this:
```suggestion
        if len(input_shape) != len(output_shape):
            return True
        return input_shape == output_shape
```

daniil-lyakhov: I don't understand why we need this set comparison. 
Do we really want to propagate through such things?
(a, c, b) -> (a * b, c, 1) and etc?

nncf/common/pruning/utils.py
```diff
@@ -590,8 +590,13 @@ def get_input_masks(node: NNCFNode, graph: NNCFGraph) -> List[Optional[NNCFTenso
     :param graph: Graph to work with.
     :return: Input masks.
     """
+    retval = []
     input_masks = [input_node.data['output_mask'] for input_node in graph.get_previous_nodes(node)]
-    return input_masks
+    for input_mask in input_masks:
+        retval.append(input_mask[node.node_name] if isinstance(input_mask, dict) else input_mask)
+    if not retval:
+        retval = []
```
daniil-lyakhov, line 598:
```suggestion
```

nncf/common/pruning/operations.py
sooahleex: As I mentioned before, when I removed this part, `tests/common/pruning/test_pruning_operations.py::test_reshape_metatype_mask_prop` has occurred the issue because it does not have output_edges.

nncf/common/pruning/operations.py
sooahleex: I tried to cover (b,a) -> (a, b) in here. The case of (a, c, b) -> (a*b, c, 1) is not covered in here.

nncf/common/pruning/operations.py
daniil-lyakhov: Looks like test should be adjusted but ok, we will fix it later.

nncf/common/pruning/utils.py
sooahleex: This part is needed to cover the empty case. We already discussed this in https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r955842604, https://github.com/openvinotoolkit/nncf/pull/1244#discussion_r952379487

daniil-lyakhov: not retval is always `[]` but ok, we will refactor it in separate PR

nncf/tensorflow/pruning/tensor_processor.py
```diff
@@ -50,3 +50,10 @@ def repeat(cls, tensor: NNCFTensor, repeats: int) -> NNCFTensor:
     def elementwise_mask_propagation(cls, input_masks: List[NNCFTensor]) -> NNCFTensor:
         cls.assert_allclose(input_masks)
         return input_masks[0]
+
+    @classmethod
+    def split(cls, tensor: NNCFTensor, output_shapes: List[int]) -> List[NNCFTensor]:
+        # pylint: disable=E1120
```
ljaljushkin, line 56: Just curious why the same function for PT does not lead to any pylint issue, but TF one needs that exception
https://github.com/openvinotoolkit/nncf/pull/1244/files#diff-a0b1a6b755a3b20734a5dba200a6c1a47f11e6d6fb54e12376971f586a24fd21R52-R56

ljaljushkin (approved):

nncf/tensorflow/pruning/tensor_processor.py
vshampor: E1120 is [no value for parameter](https://pylint.pycqa.org/en/latest/user_guide/messages/error/no-value-for-parameter.html), I actually fail to see where such a warning could be spawned in this code since it is a function declaration in this line, not a function call.

daniil-lyakhov: It could be pylint false positive, which appear due to c++ source code binding. There was such cases before

sooahleex:
sota eval tf 89 jobs for nightly validation passed.

daniil-lyakhov:
@ljaljushkin, can we merge it?

ljaljushkin:
> @ljaljushkin, can we merge it?

SOTA eval for TF and PT is OK, except some known issues that are not related to the PR.
Unfortunately, pre-commit for Pytorch is red due to some infrastructure issue. I'd not merge until we make sure it's passing.

daniil-lyakhov:
run pytorch pre-commit tests

ljaljushkin:
Pre-commit for PyTorch is not working due to infra issue. We've reported it to CI team. Have to wait until it's fixed and green

ljaljushkin:
run pytorch pre-commit tests

Issue #1161, opened 2022-04-01
AssertionError: Multiple post-hooks for a single NNCFGraph node are not supported!


luoyger:
Load tensorflow saved module exported by tf2, and run nncf compress code below,  error occured: AssertionError: Multiple post-hooks for a single NNCFGraph node are not supported!

What's wrong with my code or model ? Is there any suggestion ?

# my code to compress tf model:
```
import tensorflow as tf
from nncf import NNCFConfig
from nncf.tensorflow import create_compressed_model, register_default_init_args

model = tf.keras.models.load_model("../pctr3.2/")
nncf_config = NNCFConfig.from_json("pctr_int8.json")

representative_dataset = tf.data.TFRecordDataset("eval_processed_data.tfrecords")
nncf_config = register_default_init_args(nncf_config, representative_dataset, batch_size=1)

compression_ctrl, compressed_model = create_compressed_model(model, nncf_config)
compression_ctrl.export_model("compressed_model.pb", save_format='frozen_graph')
```

# nncf debug code:
```
 for post_hook_ip in allowed_post_hook_insertion_points:
            target_node_name_vs_post_hook_ips[post_hook_ip.target_node_name].add(post_hook_ip)
            if len(target_node_name_vs_post_hook_ips[post_hook_ip.target_node_name]) > 1:
                print("post_hook_ip.target_node_name:", post_hook_ip.target_node_name)
```

# logs:
these target_node_name have more than one post_hook_ip
```
post_hook_ip.target_node_name: dense_input
post_hook_ip.target_node_name: sparse_wgt_input
post_hook_ip.target_node_name: sparse_ids_input
post_hook_ip.target_node_name: seq_50_input
post_hook_ip.target_node_name: Identity
post_hook_ip.target_node_name: args_0
post_hook_ip.target_node_name: args_0_1
post_hook_ip.target_node_name: Identity
post_hook_ip.target_node_name: Identity_1
post_hook_ip.target_node_name: args_0
post_hook_ip.target_node_name: args_0_2
post_hook_ip.target_node_name: args_0_1
post_hook_ip.target_node_name: Identity
post_hook_ip.target_node_name: Identity_1
post_hook_ip.target_node_name: Identity_2
post_hook_ip.target_node_name: Identity_3
post_hook_ip.target_node_name: Identity_4
post_hook_ip.target_node_name: Identity_5
post_hook_ip.target_node_name: Identity_6
post_hook_ip.target_node_name: Identity_7
[...]

PR #1246, opened 2022-08-05
Handle KerasLayer in the experimental_quantization algorithm

andrey-churkin:
### Changes

We get an exception if we try to run tensorflow experimental quantization algorithm with model that contains `tensorflow_hub.KerasLayer`.

### Reason for changes

The tensorflow experimental quantization has not supported models with `tensorflow_hub.KerasLayer` layer yet. This layer contains already compiled subgraphs so we cannot wrap operations inside them. When we try to run the algorithm on such model we get strange error. Please look at this GitHub [issue](https://github.com/openvinotoolkit/nncf/issues/1161). Current changes resolves this problem. Now we get the message which explains the reason.

### Related tickets

Ref: 88614

### Tests

N/A

### DIFF
--- a/nncf/tensorflow/graph/utils.py
+++ b/nncf/tensorflow/graph/utils.py
@@ -44,6 +44,20 @@ def is_functional_model(model):
            and getattr(model, '_is_graph_network', False)
 
 
+def is_keras_layer_model(model: tf.keras.Model) -> bool:
+    """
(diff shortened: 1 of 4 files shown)


andrey-churkin:
@alexsu52, Could you look at this, please?

alexsu52 (approved):

alexsu52 (commented):
Thanks for your contribution. Please, add a test for this.

PR #1250, opened 2022-08-10
Extend ONNX metatypes, Add more graph tests 

kshpv:
### Changes

1. Extend graph tests with new detection and classification models:
     Detection:
         **SSD
         retinanet
         fcn-resnet50
         Mask RCNN**
     Classification:
        **mobilenet_v3_small
         mnasnet0_5
         densenet121
         squeezenet_1_0**
2. Add new ONNX metatypes

### Reason for changes

To increase the coverage of the supported models in pre commit tests - to increase the confidence of merged PRs.

### Related tickets

88882

### Tests

Some small changes to tests


nncf/experimental/onnx/model_normalizer.py
```diff
@@ -20,70 +20,71 @@
...
+                    if attr.type == onnx.AttributeProto.GRAPH:
+                        add_const_value_infos_to_graph(attr.g)
+                    if attr.type == onnx.AttributeProto.GRAPHS:
+                        for g in attr.graphs:
+                            add_const_value_infos_to_graph(g)
+
+        return add_const_value_infos_to_graph(model.graph)
 
+    @staticmethod
+    def convert_opset_version(model: onnx.ModelProto) -> onnx.ModelProto:
```
vshampor, line 86: Docstring missing

tests/onnx/quantization/common.py
```diff
@@ -37,22 +37,25 @@
 
 
 class DatasetForTest(Dataset):
-    def __init__(self, input_key, input_shape):
+    def __init__(self, input_key, input_shape, has_batch_dim):
```
vshampor, line 40: Type hints

tests/onnx/quantization/test_detection_models_graph.py
```diff
@@ -17,37 +17,79 @@
...
 INPUT_SHAPES = [
+    [1, 3, 1200, 1200],
     [1, 3, 416, 416],
     [1, 3, 416, 416],
+    [3, 30, 30],
+    [1, 3, 480, 640],
+    [1, 3, 480, 640],
+]
+
+DISSALOWED_OP_TYPES = [
```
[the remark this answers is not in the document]
kshpv: This model contains many layers that should be placed in ignored_scopes. To do that this is necessary to manually analyze the model graph, which is quite big. Because of that reason the ignored_types were added. 
It leads to not fully correct quantizers placements, but we could quantize the main computational part of the model - Convolutions.
I suggest to add this model in test with this option for now and mark it as not fully quantized and then when we have time to work on it.

tests/onnx/quantization/common.py
kshpv: Done

kshpv:
retest this please

vshampor (approved):

PR #1255, opened 2022-08-18
ONNX PTQ Introduce automatic adding nodes to ignored scopes 

kshpv:
### Changes

Add pre-quantization step in ONNXMinMaxQuantization to add nodes in ignored scope.

Fix incorrect default parameters creation for PostTrainingQuantization and ONNXMinMaxQuantization. Now all parameters needed for ONNXMinMaxQuantization are passed from PostTrainingQuantization to ONNXMinMaxQuantization directly.

### Reason for changes

Some of the models consist of post-processing part which should be disable for quantization, e.g. ssd_mobilent_v1 or Mask RCNN. This PR adds this functional.

### Related tickets

88757

### Tests

The test_add_ignored_scopes is added.


vshampor:
I think that this option should be disabled by default, and only be enabled via config parameter.

kshpv:
retest this please

kshpv:
This PR removes several quantizers in some models. This is expected behavior, it could lead to worse performance boost, but better accuracy. It would be interesting to research.

For the robust measurements we need to add perf mearsuments in our E2E CI, @MaximProshin

AlexKoff88:
> I think that this option should be disabled by default, and only be enabled via config parameter.

This is critical for accurate PTQ of models with post-processing embedded in the graph since no fine-tuning is available. But I agree that some testimonies that it works are required.

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -73,20 +77,72 @@ def generate_stat_collector(self, quantizer_config: QuantizerConfig) -> TensorSt
...
+        weight_nodes = self.nncf_graph.get_nodes_by_metatypes(WEIGHT_LAYER_METATYPES)
         quantizable_layer_nodes = [QuantizableWeightedLayerNode(weight_node, [QuantizerConfig()]) for weight_node in
                                    weight_nodes]
 
         hw_config_type = self.target_device
         hw_config_path = ONNXHWConfig.get_path_to_hw_config(hw_config_type)
         hw_config = ONNXHWConfig.from_json(hw_config_path)
 
+        self._add_ignored_scopes_after_last_weight_node()
+        self._add_ignored_scopes_after_topk_nms()
```
KodiaqQ, line 144: Why should we ignore nodes AFTER the last weight nodes and AFTER TopK & NMS layers separately? I thought we should ignore layers BETWEEN the last weight nodes and TopK/NMS.

nncf/experimental/post_training/algorithms/quantization/min_max_quantization.py
```diff
@@ -52,27 +52,43 @@ class MinMaxQuantizationParameters(AlgorithmParameters):
...
         self.ignored_scopes = [] if ignored_scopes is None else ignored_scopes
-        self.quantize_outputs = quatize_outputs
 
     def to_json(self) -> Dict[str, Union[str, float, int]]:
         """
         Serialize all MinMaxQuantization parameters to JSON.
         """
 
+    def _determine_weight_activation_modes(self, preset: Preset):
+        # TODO(kshpv): add support of presets
```
KodiaqQ, line 82: Why shouldn't do it right now instead of adding TODO's?

nncf/experimental/post_training/algorithms/quantization/algorithm.py
```diff
@@ -49,40 +47,25 @@ def __init__(self,
                  range_type: RangeType = RangeType.MEAN_MINMAX,
                  number_samples: int = 300,
                  target_device: HWConfigType = HWConfigType.CPU,
+                 quantize_outputs: bool = False,
```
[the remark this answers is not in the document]
kshpv: This parameter coms from the QAT NNCF quantization parameter, which means if the outputs of the Network should be quantized. I assume this is useful for some HW.

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: I reworked this part. Now this algorithm has only one function, which is called `_add_ignored_scopes_detection_output`.
Please, check it out

nncf/experimental/post_training/algorithms/quantization/min_max_quantization.py
kshpv: I've created a ticket on that - 93174

nncf/common/graph/graph.py
```diff
@@ -516,38 +516,46 @@ def visualize_graph(self, path: str):
 
     def get_graph_for_structure_analysis(self, extended: bool = False) -> nx.DiGraph:
         """
-        The graph to dump has certain node attributes omitted, compared to the graph stored
-        inside NNCFGraph.
+        Returns the nx.Digraph, which is built based on self._nx_graph.
+        The new graph has certain node attributes omitted, compared to the graph stored inside NNCFGraph.
+        If the node name consist of special reserved character, this character will be replaced.
```
alexsu52, line 521:
```suggestion
        If the node name consists of a special reserved character, this character will be replaced.
```

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -73,20 +80,79 @@ def generate_stat_collector(self, quantizer_config: QuantizerConfig) -> TensorSt
                                                     reduction_shape=axes, num_samples=self.number_samples)
         raise RuntimeError('This range type is not supported.')
 
+    def _add_ignored_scopes_detection_output(self) -> None:
```
alexsu52, line 83: Сan you make this algorithm generic?

alexsu52 (commented):
LGTM

nncf/experimental/post_training/algorithms/quantization/min_max_quantization.py
kshpv: The presets were added.

kshpv:
TODO: update ssd_mobilenet_v1 config

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
vshampor: Agreed. This should be done in the common `QuantizerPropagationSolver`, and also tested as common code.

vshampor (changes requested):

kshpv:
@alexsu52 @vshampor please take a look at the updated generic algorithm. Now this functionality is placed in `QuantizerPropagationSolver`

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -359,6 +373,153 @@ def run_on_ip_graph(self, ip_graph: InsertionPointGraph) -> QuantizationProposal
                                     quant_prop_graph=quant_prop_graph,
                                     quantization_point_id_vs_prop_quantizer=insertions_vs_associated_prop_quants)
 
+    def _is_node_in_quantizable_nodes(self, quantizable_layer_nodes_names, node_key) -> bool:
```
KodiaqQ, line 376: Docstrings, type hints.

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -359,6 +373,153 @@ def run_on_ip_graph(self, ip_graph: InsertionPointGraph) -> QuantizationProposal
...
+        if node_key.find('\n') != -1:
+            for key in node_key.split('\n'):
+                if key in quantizable_layer_nodes_names:
+                    return True
+        else:
+            if node_key in quantizable_layer_nodes_names:
+                return True
+        return False
+
+    def _update_traverse_path_flags(self, node_metatype: OperatorMetatype, traverse_path_flags: Dict[str, bool],
```
KodiaqQ, line 386: Docstrings, type hints.

tests/onnx/quantization/test_ignore_post_processing_nodes.py
```diff
@@ -0,0 +1,205 @@
...
+                      'Identity_2': ['NMS_1'], 'NMS_1': ['Identity_3'], 'Identity_3': ['Output_1'],
+                      'Identity_4': ['Identity_2', 'Identity_5'], 'Identity_5': ['NMS_1']}
+        super().__init__(node_names, node_op_types, node_edges)
+        self.reference_ignored_scopes = ['Identity_3', 'Identity_2', 'Identity_5', 'Identity_4', 'Identity_1']
+
+
+# pylint:disable=protected-access
+
+@pytest.mark.parametrize('model_to_test', [ModelToTest1(), ModelToTest2(), ModelToTest3(), ModelToTest4()])
+def test_add_ignoring_nodes_after_last_weight_node(model_to_test):
```
vshampor, line 187: Since you are testing common logic, this test is more suitable to be in the common test code. I've moved the rest of the solver tests to common code already. Please do so for this test as well.

tests/onnx/quantization/test_ignore_post_processing_nodes.py
```diff
@@ -0,0 +1,205 @@
...
+    quantizable_layer_nodes = [QuantizableWeightedLayerNode(weight_node, [QuantizerConfig()]) for weight_node in
+                               weight_nodes]
+
+    solver = QuantizerPropagationSolver(quantizable_layer_nodes=quantizable_layer_nodes,
+                                        post_processing_metatypes=[ONNXTopKMetatype,
+                                                                   ONNXNonMaxSuppressionMetatype])
+
+    quantization_proposal = solver.run_on_ip_graph(ip_graph)
+    ignored_node_names = [nncf_graph.get_node_by_key(ignored_node_key).node_name for ignored_node_key in
+                          quantization_proposal._quant_prop_graph.ignored_node_keys]
```
vshampor, line 204: You are accessing a protected attribute here, which should have led you to an understanding that you are doing something wrong.

Note that if you refactor your code as I suggested above, introducing a new `PostprocessingNodeLocator` that, for instance, takes a `QPSG` and returns a list of node keys to ignore, then you can test that object separately from the `Solver`, and also in the common code.

nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -157,6 +157,20 @@ def __init__(self, ip_graph: InsertionPointGraph,
         for barred_node_key in self.ignored_node_keys + iteration_scope_node_keys:
             self._add_barrier_after_node(barred_node_key)
 
+    def get_node_keys_by_metatypes(self, metatype_list: List[Type[OperatorMetatype]]) -> List[str]:
```
vshampor, line 160: Refactor so that it accepts a single metatype, and returns the list of node keys that correspond to this single metatype. If you need to get nodes that correspond to either of a set of metatypes, just call this function multiple times and merge the corresponding lists.

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -257,6 +262,10 @@ def __init__(self,
           trainable quantizer parameters.
         :param run_consistency_checks: Whether to run internal consistency checks at each propagataion step.
         :param quantize_outputs: Whether to insert additional quantizers right before each of the model outputs.
+        :param post_processing_metatypes: The framework specific TopK and NonMaximumSuppression NNCF Metatypes
+         that are used for automatic ignoring post-processing nodes, connected with TopK and NonMaximumSuppression.
+         The first element in the list should be TopK metatype, the second - NonMaximumSuppression.
+         If None automatic ignoring will be skipped.
```
vshampor, line 268: The description does not correspond to the logic. You are describing the specific case here while the logic is actually agnostic of the actual meaning of metatypes provided in this parameter. 
`The first element in the list should be TopK metatype, the second - NonMaximumSuppression.` - we'll have to rework this.

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -359,6 +374,152 @@ def run_on_ip_graph(self, ip_graph: InsertionPointGraph) -> QuantizationProposal
                                     quant_prop_graph=quant_prop_graph,
                                     quantization_point_id_vs_prop_quantizer=insertions_vs_associated_prop_quants)
 
+    def _is_node_in_quantizable_nodes(self, node_key: str, quant_prop_graph: QuantizerPropagationStateGraph,
```
vshampor, line 377: `_node_has_underlying_weights`

vshampor (changes requested):

nncf/common/quantization/quantizer_propagation/solver.py
kshpv: Actually, the logic of adding the nodes to the ignored behaves in two different ways depending on the visited metatypes. This logic is described in docstring:
```
If the path with the nodes has the node with NonMaxSuppressionMetatype,
all the nodes in this path will be added into ignored.
If the path with the nodes has the node with TopKMetatype,
all the nodes starting from TopKMetatype in this path will be added into ignored.
```

But you gave a good idea, with the new logic with forward traversing till any weight node, it could be replaced with a simpler one - If during backward traversing the algorithm faces any node from post-processing list then add all traversed nodes to ignored.

nncf/common/quantization/quantizer_propagation/graph.py
kshpv: Done

tests/onnx/quantization/test_ignore_post_processing_nodes.py
kshpv: Done

nncf/common/quantization/quantizer_propagation/solver.py
kshpv: Done

PR #1257, opened 2022-08-23
ignored_scopes has type `List[str]` but is used as type `None`.

luca-digrazia:
"filename": "examples/experimental/onnx/semantic_segmentation/onnx_ptq_segmentation.py"
"warning_type": "Incompatible variable type [9]",
"warning_message": " ignored_scopes is declared to have type `List[str]` but is used as type `None`.",
"warning_line": 28
"fix": None to []

### DIFF
--- a/examples/experimental/onnx/semantic_segmentation/onnx_ptq_segmentation.py
+++ b/examples/experimental/onnx/semantic_segmentation/onnx_ptq_segmentation.py
@@ -25,7 +25,7 @@ from nncf.experimental.onnx.datasets.segmentation_dataset import create_dataset_
 
 def run(onnx_model_path: str, output_model_path: str, dataset_name: str,
         dataset_path: str, num_init_samples: int,
-        input_shape: List[int], ignored_scopes: List[str] = []):
+        input_shape: List[int], ignored_scopes: Optional[List[str]] = None):
     print("Post-Training Quantization Parameters:")
     print("  number of samples: ", num_init_samples)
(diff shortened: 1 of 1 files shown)


examples/experimental/onnx/semantic_segmentation/onnx_ptq_segmentation.py
```diff
@@ -25,7 +25,7 @@
 
 def run(onnx_model_path: str, output_model_path: str, dataset_name: str,
         dataset_path: str, num_init_samples: int,
-        input_shape: List[int], ignored_scopes: List[str] = None):
+        input_shape: List[int], ignored_scopes: List[str] = []):
```
ljaljushkin, line 28:
```suggestion
        input_shape: List[int], ignored_scopes: Optional[List[str]] = None):
```
Please don't use mutable variable as a default value: https://docs.python-guide.org/writing/gotchas

It would be also helpful to fix all similar type hints issues if any in a single PR.

ljaljushkin (changes requested):

examples/experimental/onnx/semantic_segmentation/onnx_ptq_segmentation.py
luca-digrazia: Yes, good patch.

kshpv:
Let's finalize this PR, @luca-digrazia

ljaljushkin (approved):

PR #1262, opened 2022-08-30
Corrected ModelAnalyzer to prune not only conv, but linear

ljaljushkin:
### Changes

ModelAnalyzer was corrected to prune not only conv, but linear as well.
Removed is_prunable_depthwise_fn in ModelAnalyzer. The util function from common is used instead.

### Reason for changes

The bug was originally found and root-caused by @jpablomch. 
mobilenetv3_mean_linear
In the original code (https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/pruning/model_analysis.py
), only conv nodes can be pruned by input and output independently. As for the above figure, the blue part cannot be pruned(i.e., the output of the first linear layer and the input of the second linear layer) because node5(mean operator) cannot accept pruned input and the following linear layers inherit this attribute.
The modification made is to allow the linear layer can be pruned by input and output independently in ModelAnalyzer.

### Tests

pruning analysis tests


ljaljushkin:
@daniil-lyakhov

nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -110,8 +113,8 @@ def _is_pruned_module(self, module) -> bool:
         return isinstance(module, tuple(NNCF_PRUNING_MODULES_DICT.keys()))
 
     def get_op_types_of_pruned_modules(self) -> List[str]:
-        types = [v.op_func_name for v in NNCF_PRUNING_MODULES_DICT]
-        return types
+        return [op_name for meta_op in [PTConvolutionPruningOp, PTTransposeConvolutionPruningOp, PTLinearPruningOp]
+                for op_name in meta_op.get_all_op_aliases()]
```
daniil-lyakhov, line 117: What is a motivation of this change? Why we are keeping `NNCF_PRUNING_MODULES_DICT` in `_is_pruned_module` method?

ljaljushkin: It will be more aligned with TF. 
I'd remove NNCF_PRUNING_MODULES_DICT if it can be implemented via PruningOp's like in TF part.

ljaljushkin: Originally, I thought `NNCF_PRUNING_MODULES_DICT` is not enough, because it contains just `NNCFLinear` module and doesn't contain other subtypes (e.g. [addmm](https://pytorch.org/docs/stable/generated/torch.matmul.html), [matmul](https://pytorch.org/docs/stable/generated/torch.addmm.html)).
But it seems, we support only pruning of linear modules in PyTorch, not weight-less operations that have 2 input tensors. Therefore, it makes sense to revert this change. Thanks for highlighting!

daniil-lyakhov (approved):
Thank you for the fix, LGTM!

PR #1263, opened 2022-08-31
ONNX PTQ Fix quantizer insertion

kshpv:
### Changes

1) Update NNCFGraph dumping function with input_port_id and output_port_id attributes.
2) Improve code style of ONNXModelNormalizer, GraphConverter
3) Fix bug with assigning input_port_id and output_port_id in NNCFGraph. Now port_ids are identical to ONNX model.
4) Add new methods in ONNXGraph
5) Align ONNXMinMaxQuantization with the correct input_port_id
6) Add edge_name attrbiute to ONNXTargetPoint to associate input quantizers with the correscpoding ONNX model's edge.
7) Add ONNXIdentityMetatype

### Reason for changes

Fix bugs with associating Quantizers on node's input.

### Related tickets

No

### Tests

Add synthetic graph tests

Add test of building original NNCFGraph


nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -127,18 +129,26 @@ def get_quantization_target_points(self, model: onnx.ModelProto) -> List[ONNXTar
                 if NNCFGraphNodeType.INPUT_NODE not in quantization_point.insertion_point.target_node_name:
                     # If quantization of Input
                     if quantization_point.insertion_point.input_port_id is not None:
-                        # TODO (kshpv): need to be reconsidered:
-                        #  some operators such as Mul and Add could have activation input tensor on 0 or 1 indices
-                        # TODO (kshpv): input_port_id can be usefull in terms of quantizing only one edge.
-                        #  Some of the models could required this.
-                        outputs = onnx_graph.get_node_edges(node_name)['input'][0]
+                        # If there are several inputs to quantize
+                        if isinstance(quantization_point.insertion_point.input_port_id, list):
```
KodiaqQ, line 133: Instead of `input_port_id` type checking, can we quantize all inputs? Whether it's one input or more, doesn't matter.

nncf/experimental/onnx/graph/model_transformer.py
```diff
@@ -148,93 +149,117 @@ def select_model_inputs_outputs(model, outputs=None):
 
     def _apply_quantizer_insertion_transformations(self):
         # TODO: optimize
+        self.added_target_edges = Counter()
```
KodiaqQ, line 152: I think this variable should be initialized at `__init__` and then update if needed.

nncf/experimental/onnx/graph/model_transformer.py
```diff
@@ -148,93 +149,117 @@ def select_model_inputs_outputs(model, outputs=None):
...
+        assert quantizer.input[1] == dequantizer.input[1] and quantizer.input[2] == dequantizer.input[2]
+        scale_tensor_name = quantizer.input[1]
+        zero_point_tensor_name = quantizer.input[2]
+
+        onnx_scale = onnx.helper.make_tensor(scale_tensor_name, onnx.TensorProto.FLOAT, dims, scale)
+        onnx_zero_point = onnx.helper.make_tensor(zero_point_tensor_name, tensor_type, dims, zero_point)
+        return onnx_scale, onnx_zero_point
+
+    def _insert_quantizer_dequantizer(self, transformation: ONNXQuantizerInsertionCommand):
+        onnx_graph = ONNXGraph(self.transformed_model)
```
KodiaqQ, line 240: Can we create `ONNXGraph` once, outside of the loop of transformations?

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -117,6 +117,39 @@ def get_node_edges(self, node_name: str) -> Dict[str, List[ValueInfoProto]]:
                         'output': list(node.output)}
         raise RuntimeError('There is no node with the name {}'.format(node_name))
 
+    def get_input_port_id_for_nodes_after_input(self, input_name: str, to_node: NodeProto):
```
vshampor, line 120: Consider redesigning these functions in the same manner as described in the comment above *if* the `get` result not being found is a regular thing. Also, add return type hints.

nncf/experimental/onnx/model_normalizer.py
```diff
@@ -19,8 +19,15 @@
 
 
 class ONNXModelNormalizer:
+    """
+    The class, which helps to prepare the ONNX model for work with Post-Training Algorithms.
+    Implements methods for adding necessary information to the ONNX model.
+    """
+    DESIRED_OPSET_VERSION = 13
```
vshampor, line 26: `Desired` by whom specifically? Perhaps `TARGET` is a more fitting word?

nncf/experimental/onnx/model_normalizer.py
```diff
@@ -80,44 +87,64 @@ def add_const_value_infos_to_graph(graph: onnx.GraphProto):
...
-            modified_model.ir_version = 7  # Due to the 'Shufflenet-v1
-            modified_model = convert_version(modified_model, 13)
-
-            # ONNX shape inference
-            # https://github.com/onnx/onnx/blob/main/docs/proposals/SymbolicShapeInfProposal.md
-            modified_model = onnx.shape_inference.infer_shapes(modified_model)
-            ONNXModelNormalizer.add_input_from_initializer(modified_model)
-
+            if model_ir_version < ir_version:
+                # TODO(kshpv): is it legal to change ir_vesrsion?
```
vshampor, line 124: You should find this out before merging this PR, and to be sure about this, and leave a comment on whether it is legal (or at least doesn't break anything) or not to change `ir_version` and in what circumstances.

nncf/experimental/onnx/model_normalizer.py
```diff
@@ -80,44 +87,64 @@ def add_const_value_infos_to_graph(graph: onnx.GraphProto):
...
-            nncf_logger.info(
-                'Successfully converted the model to the opset = {}'.format(modified_model.opset_import[0].version))
+            nncf_logger.debug(
+                'The model was successfully converted  to the Opset Version = {}'.format(
+                    modified_model.opset_import[0].version))
         except (RuntimeError, ConvertError):
             modified_model = model
             nncf_logger.error(
-                "Couldn't convert target model to opset13. Use original model")
+                "Couldn't convert target model to opset13. Using the original model")
```
vshampor, line 139: Should probably use the constant defined above for the opset part of the log.

tests/onnx/data/reference_graphs/original_nncf_graph/vgg16.dot
```diff
@@ -0,0 +1,81 @@
...
+"33 Gemm_33" [id=33, type=Gemm];
+"34 Relu_34" [id=34, type=Relu];
+"35 Gemm_35" [id=35, type=Gemm];
+"36 Relu_36" [id=36, type=Relu];
+"37 Gemm_37" [id=37, type=Gemm];
+"38 nncf_model_input_0" [id=38, type=nncf_model_input];
+"39 nncf_model_output_0" [id=39, type=nncf_model_output];
+"0 Conv_0" -> "1 Relu_1"  [label="output_port_id = 0 \n shape = [1, 64, 224, 224] \n input_port_id = 0 \n ", style=solid];
+"1 Relu_1" -> "2 Conv_2"  [label="output_port_id = 0 \n shape = [1, 64, 224, 224] \n input_port_id = 0 \n ", style=solid];
+"2 Conv_2" -> "3 Relu_3"  [label="output_port_id = 0 \n shape = [1, 64, 224, 224] \n input_port_id = 0 \n ", style=solid];
```
vshampor, line 44: You should have noticed that something is wrong in your new code that you had to regenerate the entire set of NNCF reference graphs for this change. Hint: it's because you modified the function that generates graphs for the structure analysis, which you shouldn't have done.

tests/onnx/models.py
```diff
@@ -306,5 +338,304 @@ def __init__(self):
...
+        )
+
+        op = onnx.OperatorSetIdProto()
+        op.version = OPSET_VERSION
+        model = onnx.helper.make_model(graph_def, opset_imports=[op])
+        onnx.checker.check_model(model)
+        super().__init__(model, [input_shape], 'many_input_ports_quantizable_model.dot')
+
+
+ALL_MODELS = [ManyInputPortsQuantizableModel, OneInputPortQuantizableModel, WeightSharingModel, ReshapeWeightModel,
```
vshampor, line 640: Don't do registration in this way, otherwise one might forget to add a model to this list after defining the class. You may instead use a decorator to register each class at the same point where it is defined.

tests/onnx/quantization/common.py
```diff
@@ -95,8 +95,33 @@ def ptq_quantize_model(
     return quantized_model
 
 
+def check_nx_graph(nx_graph: nx.DiGraph, expected_graph: nx.DiGraph):
```
vshampor, line 98: Is this a copy-and-paste from the corresponding function in PT code? Extract common code.

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -14,53 +14,109 @@
...
-from tests.onnx.models import ModelWithIntEdges
+from tests.onnx.models import ALL_MODELS
+from tests.common.helpers import TEST_ROOT
+from tests.onnx.quantization.common import ModelToTest
+from tests.onnx.quantization.common import check_nx_graph
 
 import networkx as nx
 
-TEST_MODELS = [LinearModel, MultiInputOutputModel, ModelWithIntEdges]
+TEST_MODELS = ALL_MODELS
```
vshampor, line 31: Should have just used the original name and refactored the usages of `TEST_MODELS` instead.

vshampor (changes requested):
A general node - you need to walk through all of your code and make sure that `input_node` actually means one and the same thing everywhere, and is consistent with the rest of the frameworks. In the rest of the frameworks, I suppose, `input_node` refers to the node which is one of the *model* inputs, which is a special type of node and is handled separately from the main logic.

tests/onnx/data/reference_graphs/original_nncf_graph/vgg16.dot
kshpv: Of course, I know it. But I did it intentionally. I suppose that such graphs more useful in both ways:
1) For debugging purpose
2) For more accurate graph comparing
What do you think?

tests/onnx/quantization/common.py
kshpv: No, it is different

nncf/experimental/onnx/graph/model_transformer.py
```diff
@@ -148,93 +149,117 @@ def select_model_inputs_outputs(model, outputs=None):
...
+    def _get_target_edge_name(self, transformation: ONNXQuantizerInsertionCommand, onnx_graph: ONNXGraph) -> Optional[
+        str]:
+        target_edge_name = None
         if transformation.target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
             try:
-                target_edge_names.add(onnx_graph.get_weight_tensor_with_initializer(
-                    transformation.target_point.target_node_name))
+                target_edge_name = onnx_graph.get_weight_tensor_with_initializer(
+                    transformation.target_point.target_node_name)
             except RuntimeError as er:
```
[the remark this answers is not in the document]
kshpv: I partially agree with you. 
But look at the other point - if I use this function too much times in many lines of codes. In that case every time I have to check whether return value is None or not. 
At the same time, I know that if I use this function and it returns None, it is a problem in the algorithm and it can't be continued.
Assuming all of these I decided to use instead RuntimeError and if it is possible than there is no such tensor it could be interpreted by try-except
What do you think?

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: What if there is one of the inpurt_port_id connected to the constant attributes of the node? In that case we don't have to quantize it.

tests/onnx/data/reference_graphs/original_nncf_graph/vgg16.dot
vshampor: Both are addressed by a different type of graph, which is the graph received for *visualization*. This type of graph is used for basic structure comparison only and by automatic tests. The extension of the reference file format should be performed in a separate PR, if necessary, so that it is not mistakenly thought of as something that is required for PTQ.

tests/onnx/quantization/common.py
vshampor: There are already three functions to that name in NNCF test code, which all seem to do one and the same thing. Review all four of the functions, including your own, and extract common code so that, preferrably, one and the same logic is used in all places. If the same logic cannot be used everywhere, give a reason why, name different functions in a different way and describe the mechanism in the docstring for the function.

nncf/experimental/onnx/graph/model_transformer.py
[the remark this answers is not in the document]
vshampor: >  In that case every time I have to check whether return value is None or not. At the same time, I know that if I use this function and it returns None, it is a problem in the algorithm and it can't be continued.

You literally handle the exception in the next line here and continue the algorithm in some way. If this is a fatal error, why handle the exception with `try... except` at all?

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -104,6 +104,8 @@ def _get_quantizer_setup(self, model: onnx.ModelProto):
         return final_setup
 
     def get_quantization_target_points(self, model: onnx.ModelProto) -> List[ONNXTargetPoint]:
+        # pylint: disable=too-many-nested-blocks
+        # TODO(kshpv): to simple the function
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/engine.py
```diff
@@ -88,6 +88,8 @@ def _create_model_graphs(self, model):
         self.onnx_graph = ONNXGraph(model)
 
     def _register_statistics(self, outputs: NNCFData, statistic_points: StatisticPointsContainer) -> None:
+        # TODO(kshpv): remove many branches
+        # pylint: disable=too-many-branches
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/graph/model_transformer.py
```diff
@@ -148,93 +149,117 @@ def select_model_inputs_outputs(model, outputs=None):
...
             raise RuntimeError(
                 'Could not find the edge corresponding to node {}'.format(
                     transformation.target_point.target_node_name))
-        scale = transformation.quantizer_parameters.scale
-        zero_point = transformation.quantizer_parameters.zero_point
-        mode = transformation.quantizer_parameters.mode
+        self.added_target_edges[target_edge_name] += 1
+        return target_edge_name
 
+    def _get_quantize_dequantize_nodes(self, transformation, target_edge_name):
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/graph/model_transformer.py
```diff
@@ -148,93 +149,117 @@ def select_model_inputs_outputs(model, outputs=None):
...
-        #  The model with the possible issues is inception_v3.
-        # If several nodes on one edge
-        for target_edge_name in target_edge_names:
-            input_nodes = onnx_graph.get_nodes_by_input(target_edge_name)
-            if not input_nodes:
-                raise RuntimeError(
-                    f'Can not add the quantizer to the {target_edge_name} edge. This edge does not have end node.')
+        return quantizer, dequantizer
+
+    def _get_scale_zero_point_tensors(self, transformation, quantizer, dequantizer):
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/model_normalizer.py
kshpv: Вщту

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -32,139 +33,165 @@
...
-                    nncf_graph.add_edge_between_nncf_nodes(
-                        from_node_id=input_node.node_id,
-                        to_node_id=to_node_id,
-                        tensor_shape=input_shape,
-                        input_port_id=input_counter[in_node_id],
-                        output_port_id=output_counter[to_node_id],
-                        dtype=nncf_dtype
-                    )
+    @staticmethod
+    def _get_tensor_shape(onnx_graph: onnx.GraphProto, tensor: Union[str, onnx.ValueInfoProto]):
```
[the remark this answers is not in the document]

nncf/experimental/onnx/model_normalizer.py
kshpv: Done

nncf/experimental/onnx/model_normalizer.py
kshpv: Done

nncf/experimental/onnx/graph/model_transformer.py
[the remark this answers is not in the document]
kshpv: The handling was removed. It was added by mistake

nncf/common/graph/graph.py
```diff
@@ -543,7 +543,12 @@ def get_graph_for_structure_analysis(self, extended: bool = False) -> nx.DiGraph
                     style = 'dashed'
                 else:
                     style = 'solid'
-                out_graph.add_edge(u, v, label=edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR], style=style)
+
+                edge_label = f"output_port_id = {edge[NNCFGraph.OUTPUT_PORT_ID_EDGE_ATTR]} \\n " \
```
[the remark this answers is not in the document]
kshpv: Okay, could you tell the difference between the visalization graph and the graph for structure analysisi?
In my tests I prefer to use the most full graph.

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: The logic of this function the same as above

tests/onnx/models.py
kshpv: Done

tests/onnx/test_nncf_graph_builder.py
kshpv: Done

nncf/experimental/onnx/graph/model_transformer.py
kshpv: This counter needs only for quantizer insertion. I think that this place is good. Just need to make it private

nncf/experimental/onnx/graph/model_transformer.py
kshpv: Maybe we could. It could be done in separate PR

nncf/common/graph/graph.py
```diff
@@ -568,7 +573,11 @@ def _get_graph_for_visualization(self) -> nx.DiGraph:
                 style = 'dashed'
             else:
                 style = 'solid'
-            out_graph.add_edge(u, v, label=edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR], style=style)
+            edge_label = f"output_port_id = {edge[NNCFGraph.OUTPUT_PORT_ID_EDGE_ATTR]} \\n " \
```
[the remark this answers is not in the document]
kshpv: Rollback these changes

nncf/common/graph/graph.py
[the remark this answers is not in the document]
kshpv: Rollback these changes

tests/onnx/quantization/common.py
kshpv: I looked at them. The function very different. It doesn't make any sense to combine them. 
For exapmle, torch version doesn't check edge attributes. Also it removes special attribute 'scope', it works with the specific folder. 
The similar part if checking for nodes attributes, but torch version has special attributes working:
```
        if 'scope' in load_attrs:
            load_attrs.pop('scope')  # TODO: remove
        if attrs != load_attrs:
            assert attrs == load_attrs
```
So, it doesn't make any sense to combine them.

nncf/common/graph/graph.py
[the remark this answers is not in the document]
vshampor: You don't need to rollback the changes *here*. *Here* you should keep the new visualization code, I was only asking you to make the display shorter while still keeping the important information.

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -103,52 +102,51 @@ def _get_quantizer_setup(self, model: onnx.ModelProto):
         final_setup = solver.get_final_quantizer_setup(finalized_proposal)
         return final_setup
 
+    def _append_weight_quantization_target_point(self, quantization_point) -> None:
```
vshampor, line 105: Type hints missing

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -103,52 +102,51 @@ def _get_quantizer_setup(self, model: onnx.ModelProto):
...
+        weight_quantization_target_point = ONNXTargetPoint(TargetType.OPERATION_WITH_WEIGHTS,
+                                                           quantization_point.insertion_point.target_node_name)
+        # It prevents the duplicate weight quantizers from being added.
+        # It can happen when you have layers that share the identical weight tensor.
+        if weight_quantization_target_point in self._quantization_target_points:
+            return
+        self._quantization_target_points.append(weight_quantization_target_point)
+
+    def _append_activation_quantization_target_point(self, onnx_graph, quantization_point) -> None:
```
vshampor, line 114: Same

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
```diff
@@ -103,52 +102,51 @@ def _get_quantizer_setup(self, model: onnx.ModelProto):
         final_setup = solver.get_final_quantizer_setup(finalized_proposal)
         return final_setup
 
+    def _append_weight_quantization_target_point(self, quantization_point) -> None:
+        weight_quantization_target_point = ONNXTargetPoint(TargetType.OPERATION_WITH_WEIGHTS,
+                                                           quantization_point.insertion_point.target_node_name)
+        # It prevents the duplicate weight quantizers from being added.
+        # It can happen when you have layers that share the identical weight tensor.
+        if weight_quantization_target_point in self._quantization_target_points:
```
vshampor, line 110: This condition should be checked outside the function body. Otherwise the function name does not correspond to the function effect. It does not always "append" weight quantization target point. One of the solutions could be to add "maybe" to the function name, but see if you can refactor to have the functions do their thing without any "maybes" attached.

nncf/experimental/onnx/engine.py
```diff
@@ -83,11 +85,17 @@ def infer(self, input_data: NNCFData) -> NNCFData:
             for tensor, output in zip(output_tensors, model_outputs)
         }
 
-    def _create_model_graphs(self, model):
+    def _create_model_graphs(self, model) -> None:
```
vshampor, line 88: Type hints missing

nncf/experimental/onnx/graph/model_transformer.py
vshampor: I checked the `ONNXGraph`'s `__init__` method and it seems that the initialization does quite a lot of work - inferring shapes, populating the tensor names etc, doing at least O(N) work where N is the number of nodes in the graph. By building the ONNXGraph for every quantizer insertion you turn the work of inserting all quantizers into the graph from O(N) to O(N^2), which is unacceptable. Optimize your code in this PR utilizing @KodiaqQ's suggestion.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -32,139 +33,165 @@
...
 
     DEFAULT_TENSOR_SHAPE = [1]
 
-    # pylint: disable=too-many-statements
     @staticmethod
-    def create_nncf_graph(onnx_model: ModelProto) -> NNCFGraph:
+    def _is_valid_onnx_metatype(node: NodeProto) -> bool:
         """
-        Adds all ONNX nodes from 'onnx_model' and then adds thr special input_nodes and output_nodes.
+        Checks whether the node has the metatype which should be added to the NNCFGraph.a
```
vshampor, line 44: Typo

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -32,139 +33,165 @@
...
+                    dtype=nncf_dtype
+                )
+                input_port_id += 1
+
+    @staticmethod
+    def convert_onnx_dtype_to_nncf_dtype(onnx_dtype: str) -> Dtype:
+        """
+        Converts the primitive types from ONNX domain to NNCF domain.
+        """
+        conversation_map = {
```
vshampor, line 148: conversion

tests/onnx/quantization/common.py
vshampor: The 'scope' parameter is no longer used in the PT testing and this code can be safely removed. None of the torch reference graph files contain this attribute any more, except for the `gnmt_variable.dot` file which is associated with the test that was skipped for the last year or so and corresponds to a pretty much unused and non-working functionality of RNN compression. Remove this torch-specific part of code, along with the `scope=""` parts from the `gnmt_variable.dot` files.

For disabling edge attribute checks, add a `check_edge_attrs: bool = True` parameter into the common `check_nx_graph` function and use it in PT code correspondingly with `check_nx_graph(..., check_edge_attrs=False)`.

tests/onnx/models.py
vshampor: Consider reusing `nncf.common.utils.registry.Registry` which supplies a decorator already written for you.

vshampor (changes requested):
I took a closer look at your docstrings and the majority of them do not completely follow the [style guide](https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#352-functions-and-methods) - namely, they are missing the parameter and return value descriptions. Please go through your docstrings and add these `:param` and `:return` fields, the IDE should help you with that.

nncf/experimental/onnx/graph/model_transformer.py
kshpv: So because of that it has TODO: optimize. 
The reason why it couldn't easy to change beacuse every insertion of quantizer/dequantizer changes the model. For safe reason the graph recreates every time.

nncf/common/graph/graph.py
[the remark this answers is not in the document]
kshpv: okay, I see. In that case I suggest to do in the other PR, cause it is not relevant to this one.

tests/onnx/models.py
vshampor: In what sense it is an overhead for *you*? You should take into account overheads to overall code base, or to execution time, not a *personal* overhead. 😕 

You can use `*registry_name*.registry_dict.values()` to iterate over values. Or, better yet, since you noticed a good use case for iteration over registry values throughout the entire NNCF, just add the corresponding helper method or an iterator protocol to the `Registry` class.

nncf/common/graph/graph.py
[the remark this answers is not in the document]
vshampor: Post a link to the corresponding PR here.

nncf/experimental/onnx/graph/model_transformer.py
vshampor: What safety are you talking about exactly? Does the model change during an insertion of a quantizer make it impossible to insert other quantizers at the points determined before, do target edge addresses in ONNX graph change every time you insert a quantizer anywhere in the network? Please post a screenshot or a model file of a minimal example where adding one quantizer to a given place makes the address for placing the second quantizer invalid.

kshpv: The ONNX format demands that the nodes are topologically sorted. So when I put the new node I use the insert_index, which is obtained from ONNXGraph. If I don't update the model's indexes it can cause the error

```
        self.transformed_model.graph.initializer.extend([onnx_scale])
        self.transformed_model.graph.initializer.extend([onnx_zero_point])
        insert_index = onnx_graph.get_node_index(input_nodes[0].name)
        self.transformed_model.graph.node.insert(insert_index, quantizer)
        self.transformed_model.graph.node.insert(insert_index + 1, dequantizer)
```

tests/onnx/models.py
kshpv: After discussion - we made a conclusion that it makes sense to utilize `Registry`, while it needs to update this class.

tests/onnx/quantization/common.py
kshpv: So, beacuse of that I told you that it takes much time to do that. Because these functions have own specific which I don't know

nncf/experimental/onnx/graph/model_transformer.py
kshpv: After offline discussion we concluded to make a separate PR optimizing ONNXGraph creation. 
Which what I was going to do.

nncf/common/graph/graph.py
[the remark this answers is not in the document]
kshpv: [here](https://iwillcreateitlatter.kashapov/)

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: What If I make `add` instead `append` and transfrom the `_quantization_target_points` from `List` to `Set`

vshampor: That would be a good solution, provided that the target points are properly hashable.

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: Done

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: Done

nncf/experimental/onnx/engine.py
kshpv: Done

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: Done

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: Done

nncf/experimental/onnx/algorithms/quantization/min_max_quantization.py
kshpv: Done

kshpv:
run pytorch pre-commit tests

tests/onnx/quantization/common.py
vshampor: You still have not moved this function to common.

vshampor (changes requested):

tests/common/graph/nx_graph.py
```diff
@@ -0,0 +1,72 @@
...
+        return int(key)
+
+    sorted_content = sorted(content, key=partial(graph_key, offset=len(content)))
+    with open(path, 'w', encoding='utf8') as f:
+        f.write(start_line)
+        f.writelines(sorted_content)
+        f.write(end_line)
+
+
+def compare_nx_graph_with_reference(nx_graph: nx.DiGraph, path_to_dot: str,
```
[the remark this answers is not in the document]
kshpv: Wait, this is the reused `check_nx_graph`, which name was changed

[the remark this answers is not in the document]
kshpv: All backends use this `compare_nx_graph_with_reference()` or directly `check_nx_graph() `

tests/onnx/quantization/common.py
kshpv: Forgot to remove it from ONNX. Now it is done

nncf/common/graph/graph.py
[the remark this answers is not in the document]
kshpv: PR #1267

tests/torch/test_compressed_graph.py
```diff
@@ -91,64 +92,17 @@ def get_basic_quantization_config_with_hw_config_type(hw_config_type, input_samp
...
-        f.write(start_line)
-        f.writelines(sorted_content)
-        f.write(end_line)
-
-
 def check_graph(graph: PTNNCFGraph, path_to_dot, graph_dir, sort_dot_graph=True):
     # pylint:disable=protected-access
     nx_graph = graph.get_graph_for_structure_analysis()
-    check_nx_graph(nx_graph, path_to_dot, graph_dir, sort_dot_graph=sort_dot_graph)
+    check_nx_graph(nx_graph, path_to_dot, graph_dir, sort_dot_graph)
```
vshampor, line 98: This call should therefore reference the common `check_nx_graph` function. Provided you use `pathlib.Path` objects, it should be trivial for you to build a full path for the corresponding argument in the common `check_nx_graph`.

vshampor (changes requested):

tests/torch/test_compressed_graph.py
```diff
@@ -91,64 +92,17 @@ def get_basic_quantization_config_with_hw_config_type(hw_config_type, input_samp
...
 
 
 def check_nx_graph(nx_graph: nx.DiGraph, path_to_dot, graph_dir, sort_dot_graph=True):
```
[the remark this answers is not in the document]
kshpv: Common `check_nx_graph` is used inside `compare_nx_graph_with_reference`.
The problem here is that `compare_nx_graph_with_reference` needs an exact `path_to_dot`, so for torch I have to build the full path before the calling this function.
So, it is done here.

tests/torch/test_compressed_graph.py
kshpv: I can't use directly common `check_nx_graph` here cause I need to build the full path to the dot from `path_to_dot, graph_dir`, which is done in local `check_nx_graph`

kshpv: Done

tests/torch/test_compressed_graph.py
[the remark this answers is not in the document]
kshpv: Done

vshampor (approved):

PR #1267, opened 2022-09-14
Add output_port_id, input_port_id to the NNCFGraph visualization

kshpv:
### Changes

Add output_port_id, input_port_id to the NNCFGraph visualization.

image

### Reason for changes

To be able to see the ports connection between nodes in the graph visualization. To speed up graph debugging.

### Related tickets

The request from the comments in PR #1263 

### Tests

No tests

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -568,7 +568,9 @@ class NNCFGraph:
                 style = 'dashed'
             else:
                 style = 'solid'
-            out_graph.add_edge(u, v, label=edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR], style=style)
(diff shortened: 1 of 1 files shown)


vshampor (approved):

ljaljushkin:
@vshampor @kshpv is it really required for all dot files that we have? What do you think about making it optional for ONNX, for instance? Seems like it adds a noise in most of the cases.

vshampor:
@ljaljushkin  PyTorch has a basic notion of input and output ports as well. It is helpful to see at a glance which operands, for instance, of an `__add__`  operation are which in the graph, and we need input port information visualized for that.

ljaljushkin:
> @ljaljushkin PyTorch has a basic notion of input and output ports as well. It is helpful to see at a glance which operands, for instance, of an `__add__` operation are which in the graph, and we need input port information visualized for that.

ok, at least it's not introduced in the `get_graph_for_structure_analysis` function

PR #1270, opened 2022-09-20
Fixed incorrect skipping of automatically found building blocks

ljaljushkin:
### Changes

Automatic search of building block defines start node as executed node that produces cache tensor.

### Reason for changes

Blocks are skipped incorrectly, as format on searching and inference was not aligned.
On searching start node was supposed to be skipped, on inference start node is executed.

The block below in ShuffleNet was invalid and couldn't be skipped.

[START NODE: ShuffleNetV2/Sequential[layer1]/DownBlock[0]/ShuffleBlock[shuffle]/permute_0,

END_NODE: ShuffleNetV2/Sequential[layer1]/BasicBlock[1]/ShuffleBlock[shuffle]/view_0],

Start node is executed and it caches output, but permute changes output shape from [1,{*}2{*},24,16,16] to [1,24,{*}2{*},16,16] and it breaks skipping of this block.

The first node should be before 33 permute:  32 view
image

end node

image

### Related tickets

80679

### Tests

test_can_skip_trivial_block on 4 synthetic models


ljaljushkin:
@jpablomch

PR #1274, opened 2022-09-22
numpy version update for openvino compatibility

dongkwan-kim01:
### Changes

Numpy version update from `1.22.1` to `1.23.1`

### Reason for changes

Error in the openvino test in the CI server
image

### Tests

Testing with PR in CI


setup.py
```diff
@@ -115,7 +115,7 @@ def find_version(*file_paths):
     "docs": [],
     "tf": [
         "tensorflow~=2.8.2",
-        "numpy~=1.22.1"
+        "numpy~=1.23.1"
```
alexsu52, line 118: Could you remove the dependency on the **exact** version of numpy from NNCF? I suggest to include numpy as common dependency of NNCF without version specification.

cc' @vshampor, @andrey-churkin, @MaximProshin, @AlexKoff88

dongkwan-kim01: Good point, I'll move it to common dependency with range and test it on the CI and local server.

dongkwan-kim01:
jenkins retest this please

alexsu52:
retest this please

alexsu52 (approved):

PR #1279, opened 2022-09-29
Dl/fix fq cuda kernel fp16

daniil-lyakhov:
### Changes

Accumulator variable for FP16 quantization backward kernel is upcasted to FP32 precision.

### Reason for changes

FP16 FQ backward kernel was unstable, on some input data the kernel was producing not deterministic results with Nan present.

### Tests

Should update FQ tests to enable reference values comparison with


nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -352,20 +360,31 @@ std::vector<at::Tensor> q_single_scale_cuda_backward(at::Tensor grad_output,
...
     auto dev_last_block_counter_low = at::zeros({1},  at::device(grad_output.options().device()).dtype(at::kInt));
 
+    const auto grad_input_dtype = grad_output.scalar_type();
+    if (grad_input_dtype == c10::ScalarType::Half)
+    {
+        grad_input_range = grad_input_range.toType(c10::ScalarType::Float);
+        grad_input_low = grad_input_low.toType(c10::ScalarType::Float);
+        dev_tmp_range = dev_tmp_range.toType(c10::ScalarType::Float);
+        dev_tmp_low = dev_tmp_low.toType(c10::ScalarType::Float);
+    }
```
daniil-lyakhov, line 378: `toType` method is creating new tensor and is copying casted values from a source tensor.
 
Maybe if we will keep FP16 memory of `grad_input_range` and `grad_output_range` before casting and then copy result from copy FP32 values of `grad_input_range` and `grad_output_range` back to FP16 we will safe 2 calls of Tensor constructor/destructor.

daniil-lyakhov: CC: @vshampor, @alexsu52

nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -452,14 +494,25 @@ std::vector<at::Tensor> q_scale_per_activation_channel_cuda_backward(at::Tensor
     auto dev_last_block_counter_range = at::zeros({grid_size.x, 1},  at::device(grad_output.options().device()).dtype(at::kInt));
     auto dev_last_block_counter_low = at::zeros({grid_size.x, 1},  at::device(grad_output.options().device()).dtype(at::kInt));
 
+    const auto grad_input_dtype = grad_output.scalar_type();
+    if (grad_input_dtype == c10::ScalarType::Half)
+    {
+        grad_input_range = grad_input_range.toType(c10::ScalarType::Float);
+        grad_input_low = grad_input_low.toType(c10::ScalarType::Float);
+        dev_tmp_range = dev_tmp_range.toType(c10::ScalarType::Float);
+        dev_tmp_low = dev_tmp_low.toType(c10::ScalarType::Float);
```
alexsu52, line 503: Looks like casting is redundant here. I think we can create dev_tmp_range and dev_tmp_low tensors using accumulator type, but grad_input_low and grad_input_range tensors using output/input type.

alexsu52 (changes requested):

nncf/torch/extensions/include/common_cuda_funcs.cuh
```diff
@@ -6,9 +6,12 @@
 // which is rumoured to degrade performance.
 
 #include "common_cuda_defs.cuh"
+#define DISABLE_FP16(TYPE_NAME) std::enable_if_t< \
+                     std::is_same<float, TYPE_NAME>::value || \
+                     std::is_same<double, TYPE_NAME>::value, bool> = true
```
daniil-lyakhov, line 11: This macros allows to restrict possible types in functions templates. Wrapped TYPE_NAME could be only `float` or `double`, compiler doesn't build template version with FP16

nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
daniil-lyakhov: Agree with you, thank you for the suggestion! I implemented approach you described, please check

nncf/torch/extensions/src/binarization/cuda/functions_cuda_impl.cu
```diff
@@ -3,27 +3,27 @@
...
     for (int i = gtidx; i < total_elements_count; i += grid_size) {
         sum += abs(*(input + i));
     }
 
     sum /= total_elements_count;
 
-    __shared__ scalar_t sh_mem[CUDA_MAX_NUM_THREADS_PER_BLOCK];
-    reduce_with_shared_memory<scalar_t>(sh_mem, sum, tidx, bidx, dev_tmp, dev_last_block_counter, scale_output, gridDim.x);
+    __shared__ scalar_accum_t sh_mem[CUDA_MAX_NUM_THREADS_PER_BLOCK];
+    reduce_with_shared_memory<scalar_t, scalar_accum_t>(sh_mem, sum, tidx, bidx, dev_tmp, dev_last_block_counter, scale_output, gridDim.x);
```
daniil-lyakhov, line 26: @ljaljushkin, I moved all FP16 accumulator to FP32, I did the same for the binarization kernels. Do I need to check this changes in some specific way?

ljaljushkin: These questions is more to @vshampor 
I'd check sota eval for `resnet18_imagenet_binarization_xnor` and `resnet18_imagenet_binarization_dorefa` models.
Also, there's a `benchmark_binarize_layers.py`. Maybe it's worth checking the difference in performance

nncf/torch/extensions/include/common_cuda_funcs.cuh
vshampor: I guess you should use `at::ScalarType::Half`, i.e. use the correct namespace, then:
https://github.com/pytorch/pytorch/blob/a75edfa97c9d985a337fbea7b9c0f4061153aaf0/aten/src/ATen/Dispatch.h#L262

If it doesn't work, then you are probably just not including the right header. Search the ATen source code for the definition of these type aliases and you'll find the right header.

vshampor (changes requested):

nncf/torch/extensions/src/binarization/cuda/functions_cuda_impl.cu
```diff
@@ -263,38 +269,45 @@ std::vector<at::Tensor> ab_cuda_backward(
           );
     }));
 
+    auto accum_options = grad_output.options();
+    if (grad_output.scalar_type() == c10::ScalarType::Half) {
+        accum_options = accum_options.dtype(c10::ScalarType::Float);
+    }
```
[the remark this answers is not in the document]
daniil-lyakhov: In this place even `scalar_t` type is unavailable. And `scalar_accum_t` is a type and `c10::ScalarType` is an enum element. `scalar_accum_t` couldn't be used here as far as I understand

nncf/torch/extensions/include/common_cuda_funcs.cuh
daniil-lyakhov: `at::ScalarType::Float` and `at::ScalarType::Double` in comparison with `at::ScalarType::Half` are just elements of an enum. 
So, it's not possible to use them in template declaration

daniil-lyakhov:
Jenkins please retest this

daniil-lyakhov:
Benchmark results:
| max ratio forward fp32% | max ratio backward fp32% | max ratio forward fp16% | max ratio backward fp16%  |
|-------------------------|------------------------|-------------------------|-------------------------|
| 4.829080462             | 1.035223844            | 3.363204138             | 1.031571939             |

vshampor (approved):

tests/torch/quantization/test_functions.py
```diff
@@ -200,51 +314,92 @@ def test_quantize_symmetric_forward(self, _seed, is_signed, is_weights, is_fp16,
...
 
             ref_input_low = ref_scale * (level_low / level_high)
             ref_input_range = ref_scale - ref_input_low
 
+            # This is needed to prevent middle points in forward pass.
+            # Small deviation in computations could lead to completely different
+            # results: instead of quant a quant a + 1 will be returned.
+            # Different results in forward leads to high deviations on bacward pass.
+            # To prevent this, input values are put into [min_deviation, max_deviation]
+            # section of quant, so small deviation woun't change the quant on forward pass
```
ljaljushkin, line 387:
```suggestion
            # Different results in forward leads to high deviations on backward pass.
            # To prevent this, input values are put into [min_deviation, max_deviation]
            # section of quant, so small deviation won't change the quant on forward pass
```
really minor

tests/torch/quantization/test_functions.py
```diff
@@ -317,48 +474,101 @@ def test_quantize_asymmetric_forward(self, _seed, input_size, bits, use_cuda, is
...
+
+            for tensor_ in (test_input_low, test_input_range):
+                assert tensor_.dtype == torch.half if is_fp16 else torch.float
+
+            # This is needed to prevent middle points in forward pass.
+            # Small deviation in computations could lead to completely different
+            # results: instead of quant a quant a + 1 will be returned.
+            # Different results in forward leads to high deviations on bacward pass.
+            # To prevent this, input values are put into [min_deviation, max_deviation]
+            # section of quant, so small deviation woun't change the quant on forward pass
```
ljaljushkin, line 556:
```suggestion
            # Different results in forward leads to high deviations on backward pass.
            # To prevent this, input values are put into [min_deviation, max_deviation]
            # section of quant, so small deviation won't change the quant on forward pass
```
the same

tests/torch/quantization/test_functions.py
```diff
@@ -92,12 +93,93 @@ def idfn(val):
 
 @pytest.fixture
 def _seed():
-    np.random.seed(0)
-
+    np.random.seed(42)
```
ljaljushkin, line 96: Just out of curiosity, why did you change this?

daniil-lyakhov: Glad you ask! It's just for stability :) (as you know, seed 42 is the most stable)

tests/torch/quantization/test_functions.py
daniil-lyakhov: Thank you, fixed

tests/torch/quantization/test_functions.py
daniil-lyakhov: Fixed

alexsu52 (approved):
LGTM, Please fix minor comments.

nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -280,14 +280,14 @@ __global__ void q_scale_per_activation_channel_cuda_backward_kernel(
         fakeQuantize<scalar_t>(&output, (input + additional_offset), input_low, input_range, levels);
         calcGrad<scalar_t>((grad_input + additional_offset), &val_grad_input_low, &val_grad_input_range, (grad_output + additional_offset),
                  (input + additional_offset), &output, range_low, range_high, reverted_range, alpha);
-        per_thread_grad_sum_range += val_grad_input_range;
+        per_thread_grad_sum_range +=  val_grad_input_range;
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -353,19 +353,23 @@ std::vector<at::Tensor> q_single_scale_cuda_backward(at::Tensor grad_output,
     auto grad_input_range = at::empty({1}, grad_output.options());
     auto grad_input_low = at::empty({1}, grad_output.options());
 
+
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/torch/extensions/src/quantization/cuda/functions_cuda_impl.cu
```diff
@@ -378,6 +382,7 @@ std::vector<at::Tensor> q_single_scale_cuda_backward(at::Tensor grad_output,
           size);
     }));)
 
+
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

PR #1287, opened 2022-10-05
Retarget doc page build branch

vshampor:


.github/workflows/build_schema_page.yml
```diff
@@ -27,3 +27,4 @@ jobs:
         with:
           folder: schema_html_build # The folder the action should deploy.
           token: ${{ secrets.PUSH_TO_GH_PAGES_BRANCH }}
+          branch: doc_pages
```
ljaljushkin, line 30: Shouldn't it be develop branch?

vshampor: No, this branch is to contain only the html/web files from which the webpage is to be constructed. It can then be specified in the repo settings to be the source for the Github pages.

ljaljushkin (approved):

PR #1289, opened 2022-10-05
Do not ignore the entire fused pattern if non-primary nodes are in the ignored scopes

vshampor:
### Changes
Adjusted quantizer propagation common algo to only ignore the entire fused pattern if the primary node from the pattern ends up matching to an ignored scope.

### Reason for changes
Improves logical consistency of the "ignored_scopes" and its effect on the algo result.

### Related tickets

92302

### Tests
TBA


vshampor:
Merged as #1314

PR #1290, opened 2022-10-06
Make test quantizers FP32 instead of FP64/ fix refs

daniil-lyakhov:
### Changes

Test for quantization functions (CPU/GPU) was moved to FP32/FP16 precisions.
Reference quantization functions were fixed

### Related tickets

91663

### DIFF
--- a/tests/torch/quantization/test_functions.py
+++ b/tests/torch/quantization/test_functions.py
@@ -195,6 +195,17 @@ def skip_if_half_on_cpu(is_fp16, use_cuda):
 
 
 def check_quant_moved(test_val, ref_val, quant_len, rtol, atol = 1e-10):
+    """
+    Checks values in `test_val` and `ref_val` elementwise eather equal with given rtol/atol or
+    values differ by correspondent `quant_len` +- rtol.
+    
+    :param test_val: Given test value.
+    :param ref_val: Given reference value.
+    :param quant_len: Lenghts of quants in quantizers
+        (for each channel in case per channel quantization).
+    :param atol: Absolute tollerance.
+    :param rtol: Relative tollerance.
+    """
     if not isinstance(test_val, list):
         test_val, ref_val = [test_val], [ref_val]
     for t, r in zip(test_val, ref_val):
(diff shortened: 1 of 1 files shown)


tests/torch/quantization/test_functions.py
```diff
@@ -368,13 +503,16 @@ def test_quantize_asymmetric_backward(self, _seed, input_size, bits, use_cuda, i
             test_value.sum().backward()
```
daniil-lyakhov, line 503: This place is producing `inf` values when `is_fp16 == True`
@alexsu52

vshampor: Stands to reason. Probably will be fixed with your accumulator PR

tests/torch/quantization/test_functions.py
```diff
@@ -107,20 +180,41 @@ def get_test_data(data_list, is_cuda=False, is_backward=False, is_fp16=False):
...
 
 def skip_if_half_on_cpu(is_fp16, use_cuda):
     if is_fp16 and not use_cuda:
         pytest.skip("As of PyTorch 1.5, the 'abs' operation is not supported on CPU for half and therefore"
-                    "symmetric quantize fails. Remove this once this is fixed in PyTorch.")
-
+                    " symmetric quantize fails. Remove this once this is fixed in PyTorch.")
+
+
+def check_quant_moved(test_val, ref_val, quant_len, rtol, atol = 1e-10):
```
vshampor, line 197: Should add a short docstring here

vshampor (approved):

tests/torch/quantization/test_functions.py
daniil-lyakhov: Only thing we can do is to not use `.sum()` here, because `test_value` should be in FP16 precision

tests/torch/quantization/test_functions.py
daniil-lyakhov: Done, please check

tests/torch/quantization/test_functions.py
alexsu52: Could you generate and fix valid input values which do not produce inf?

daniil-lyakhov:
Jenkins retest this

daniil-lyakhov:
Moved to #1279

PR #1291, opened 2022-10-06
Extract common quantization tests

vshampor:
### Changes

Quantization solver-related tests were moved out to the common test directory.

### Reason for changes
Common code should be able to be tested without backend dependencies.

### Related tickets
N/A

### Tests
Existing tests should be passing.


alexsu52 (approved):
LGTM

PR #1295, opened 2022-10-10
PTQ for the OpenVINO backend

andrey-churkin:
### Changes

- The `nncf.data.Dataset` class was added. Also, available as `nncf.Dataset`. Please look at the documentation of this class to get more details.
- Added support of PTQ for the `openvino.runtime.Model`  via Post-training Optimization Tool API. The method `nncf.quantization.quantize()` was added. Also, available as `nncf.quantize()`. This method applies post-training quantization to the provided model.
- The `nncf.TargetDevice` enum was added. It uses to specify the `target_device` parameter in the `nncf.quantize()` method.
- The `nncf.QuantizationPreset` enum was added. It uses to specify the `preset` parameter in the `nncf.quantize()` method.

Public API:
- `nncf.Dataset`
- `nncf.quantize()`
- `nncf.TargetDevice`
- `nncf.QuantizationPreset`

### Reason for changes

The NNCF should support the PTQ for the OV model.

### Related tickets

Ref: 90642


nncf/data/dataloader.py
```diff
@@ -0,0 +1,103 @@
...
+    compression algorithms.
+
+    The `DataLoader` object contains the dataset and information
+    about how to transform the data item returned per iteration to
+    the model's expected input.
+    """
+
+    @property
+    @abstractmethod
+    def batch_size(self) -> int:
```
AlexKoff88, line 37: I see the batch is coming through the whole code. But I can't see how it is used. @andrey-churkin, can you please explain?

nncf/data/utils.py
```diff
@@ -0,0 +1,106 @@
...
+        import tensorflow
+    except ImportError:
+        tensorflow = None
+
+    batch_size = 1
+
+    if torch is not None and isinstance(data_source, torch.utils.data.DataLoader):
+        batch_size = data_source.batch_size
+
+    if tensorflow is not None and isinstance(data_source, tensorflow.data.Dataset):
```
AlexKoff88, line 75: Can we support HuggingFace `Dataset` object here? [This](https://github.com/huggingface/datasets/blob/6be722dbbc212d88eb5e550e56431cdcab180eee/src/datasets/arrow_dataset.py#L646) is the definition.

nncf/data/utils.py
```diff
@@ -0,0 +1,106 @@
...
+        batch_size = getattr(data_source, '_batch_size', batch_size)
+
+    return batch_size
```
AlexKoff88, line 78: This logic means that in case of OpenVINO or ONNX we will always have `batch_size = 1`.

nncf/openvino/engine.py
```diff
@@ -0,0 +1,188 @@
...
+    if not isinstance(dataloader, DataLoader):
+        raise Exception(f'Unexpected type of data loader: {type(dataloader)}')
+
+    if isinstance(sampler, pot.samplers.batch_sampler.BatchSampler):
+        return create_subset(dataloader, indices)
+
+    raise Exception(f'Unexpected type of sampler: {type(sampler)}')
+
+
+class CustomEngine(pot.IEEngine):
```
AlexKoff88, line 80: I would call it `OVEngine`, `NNCFOVEngine` or whatever. `Custom` is too generic.

nncf/openvino/quantization/helpers.py
```diff
@@ -0,0 +1,107 @@
...
+                  fast_error_correction: bool,
+                  model_type: Optional[str] = None) -> ov.Model:
+    """
+    Implementation of the `quantize()` method for the OpenVINO backend.
+    """
+    pot_model = _convert_openvino_model_to_compressed_model(model, target_device)
+
+    engine_config = {
+        'device': 'CPU',
+        'stat_requests_number': 2,
```
AlexKoff88, line 80: @alexsu52, @KodiaqQ and everybody, we need to decide how we can expose such backend-specific stuff through a common API that is being developed now.

nncf/openvino/engine.py
andrey-churkin: That's a good point. I renamed it.

nncf/data/utils.py
andrey-churkin: This code was removed. Not applicable now.

nncf/openvino/quantization/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+        'eval_requests_number': 2,
```
KodiaqQ, line 82: BTW, why is the `stat_requests_number` and `eval_requests_number` set as 2?

andrey-churkin: As said @AlexKoff88: "Based on the recent experience, I would recommend setting the number of requests to 2 at least".

andrey-churkin: @KodiaqQ Do you have any concerns?

KodiaqQ: For some types of networks (for example, GNA-specific), such parameters will affect the accuracy due to ReadValue-Assign pairs. But since the GNA isn't supported here, I don't know, should we care about this now.

examples/experimental/openvino/mobilenet_v2/main.py
```diff
@@ -0,0 +1,178 @@
...
+    ])
+    val_dataset = torchvision.datasets.ImageFolder(val_dir, preprocess)
+    val_dataloader = torch.utils.data.DataLoader(val_dataset,
+                                                 batch_size=1,
+                                                 shuffle=False,
+                                                 num_workers=4)
+    return val_dataloader
+
+
+# The `validate()` method was taken as is from the pythorch repository.
```
l-bat, line 130:
```suggestion
# The `validate()` method was taken as is from the pytorch repository.
```

examples/experimental/openvino/mobilenet_v2/main.py
```diff
@@ -0,0 +1,178 @@
...
+# You can find it [here](https://github.com/pytorch/examples/blob/main/imagenet/main.py).
+# Code regarding CUDA and training was removed.
+# Some code was changed and added.
+
+def validation_fn(model: ov.Model, data_loader: nncf.DataLoader) -> float:
+    compiled_model = ie.compile_model(model, device_name='CPU')
+    output_layer = next(iter(compiled_model.outputs))
```
l-bat, line 137: For me `output_layer = compiled_model.output(0)` is more readable

nncf/data/dataloader.py
```diff
@@ -0,0 +1,84 @@
+"""
```
alexsu52, line 1: Please take a look on my proposal. I tried to implement statements that we discussed offline. Please feel free to comment.
```
class Sampler(ABC):

    def __iter__(self) -> Iterator[int]:
        pass



class InferenceDataLoader:

    def __init__(data_source: Iterable, transform_fn: Callable[[LabledDataItem], UnlabledDataItem], sampler:Optional[Sampler] = None):
        self._data_source = data_source

    def __iter__(self) -> Iterator[UnlabledDataItem]:
        pass



class ValidationDataLoader:

    def __init__(data_source: Iterable, sampler: Optional[Sampler] = None):
        self._data_source = data_source

    def __iter__(self) -> Iterator[LabledDataItem]:
        pass



class Dataset:

    def __init__(data_source: Iterable, transform_fn: Callable[[LabledDataItem], UnlabledDataItem]):
        self._data_source = data_source
        self._transform_fn = transform_fn

    @property
    def data_source(self) -> Iterable:
        return self._data_source
    

    def get_inference_dataloader(self, sampler: Optional[Sampler] = None) -> InferenceDataLoader:
        pass

    def get_validation_dataloader(self, sampler: Optional[Sampler] = None) -> ValidationDataLoader:
        pass



def val_func(model: ModelType, val_dataloader: Iterable[LabledDataItem]) -> float:
    pass


def transform_fn(data_item: LabledDataItem) -> UnlabledDataItem:
    pass


def quantize(model: ModelType,
             calibration_dataset: Dataset,
             preset: QuantizationPreset = QuantizationPreset.PERFORMANCE,
             target_device: TargetDevice = TargetDevice.ANY,
             subset_size: int = 300,
             fast_error_correction: bool = True,
             model_type: Optional[str] = None) -> ModelType:


def quantize_with_accuracy_control(model: ModelType,
                                   calibration_dataset: Dataset,
                                   validation_dataset: Dataset,
                                   validation_fn: Callable[[ModelType, Iterable[LabledDataItem]], float],
                                   max_drop: float = 0.01,
                                   preset: QuantizationPreset = QuantizationPreset.PERFORMANCE,
                                   target_device: TargetDevice = TargetDevice.ANY,
                                   subset_size: int = 300,
                                   fast_error_correction: bool = True,
                                   model_type: Optional[str] = None) -> ModelType:
```

nncf/openvino/quantization/helpers.py
alexsu52: The ```class AdditionalQuantizationParamters``` will has the ```backend``` attribute where the user can assign instance of the ``OpenVINOParameters```, for example.  OpenVINOParameters will consist ```stat_requests_number``` attribute.

nncf/data/dataloader.py
andrey-churkin: Several points from my side:
- I think it's better to use the following naming:
  ```python
  class ValidationDataLoader --> class LabeledDataLoader:

  class InferenceDataLoader --> class UnlabeledDataLoader:

  get_inference_dataloader(...) --> get_unlabeled_dataloader()

  get_validation_dataloader(...) --> get_labeled_dataloader()
  ```
  I think these names are more usual.
- Why do we need the `data_source` property in the `Dataset` class?

alexsu52: > Several points from my side:
> 
>     * I think it's better to use the following naming:
>       ```python
>       class ValidationDataLoader --> class LabeledDataLoader:
>       
>       class InferenceDataLoader --> class UnlabeledDataLoader:
>       
>       get_inference_dataloader(...) --> get_unlabeled_dataloader()
>       
>       get_validation_dataloader(...) --> get_labeled_dataloader()
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       I think these names are more usual.
> 
>     * Why do we need the `data_source` property in the `Dataset` class?

Good question. Right now It is not needed.

alexsu52: For user perspective:
How to run quantization:
1. Create nncf.Dataset. Let a developer has DataLoader or Dataset from ```torch.utils.data``` and called it is as `data_source` and the developer wrote `transform_fn`
```
dataset = nncf.Dataset(dataloader, transform_fn)
```
2. call the quantization method
```
q_model = nncf.quantize(model, calibration_dataset: dataset)
```
How to test transform_fn or/and validation_fn on user side: 
```
inference_dataloader = dataset.get_inference_dataloader()
try:
    model(next(iter(inference_dataloader)))
except:
    pass
```
```
 validation_dataloader = dataset.get_validation_dataloader()
try:
    validation_fn(model, validation_dataloader)
except:
    pass
```

nncf/data/utils.py
```diff
@@ -0,0 +1,74 @@
...
+from nncf.data.types import DataSource
+from nncf.data.types import DataItem
+from nncf.data.types import ModelInput
+from nncf.data.dataloader import DataLoader
+from nncf.data.dataloader import DataLoaderImpl
+
+
+def create_dataloader(data_source: DataSource,
+                      transform_fn: Callable[[DataItem], ModelInput]) -> DataLoader:
+    """
```
eaidova, line 27: Is it possible to make transform function optional? In some cases dataloader can be defined specifically for quantization (e.g. for pipelined models, as example, we will not use dataloader for model part as loader for whole pipeline) and transformation logic can be part of __iter__ or __getitem__.

alexsu52: Thanks for you comment. We will make it possible to pass ```None``` as ```transform_fn```

nncf/openvino/quantization/helpers.py
```diff
@@ -0,0 +1,106 @@
...
+                'stat_subset_size': subset_size,
+                'use_fast_bias': fast_error_correction,
+                'model_type': model_type,
+            }
+        }
+    ]
+
+    pot_dataloader = POTDataLoader(calibration_dataset, calibration_dataset.transform_fn)
+    engine = OVEngine(engine_config, pot_dataloader, pot_dataloader)
+    pipeline = pot.create_pipeline(algorithms, engine)
```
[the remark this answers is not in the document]
alexsu52: @eaidova What do you mean by opportunity? Do you need some API like callback to get opportunity collect progress?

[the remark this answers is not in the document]
eaidova: @alexsu52 I just mean provide a flag `print_progress` to turn on it  when we call `ptq.quantization` , if I right remember pot IEEngine is able to log it:
https://github.com/openvinotoolkit/openvino/blob/master/tools/pot/openvino/tools/pot/engines/ie_engine.py#L71
or maybe allow to configure preferable log level

callback api maybe useful for advanced users, but too complex for understanding, so just print any logs instead of 10h of siliency in my case will be pretty well :)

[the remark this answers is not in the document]
alexsu52: What do you think about control it via logger and setting log level. I mean using nncf logger to control it. It wiil allow us to avoid additional input argument in the quantize method.

cc @KodiaqQ @l-bat @andrey-churkin @AlexKoff88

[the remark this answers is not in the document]
eaidova: could you please provide some code sample how it will looks?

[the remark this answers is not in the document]
alexsu52: > could you please provide some code sample how it will looks?

If we are talking about the output of information about the progress of quantization than the code sample will not changed. 

If you want to disable logging than you can do the following:
```
from nncf.common.uteils.logging import disable_logging
disable_logging()
```

If you want to change the log level, for example, you need debug logs:
```
import logging
from nncf.common.uteils.logging import set_log_level
set_log_level(logging.DEBUG)
```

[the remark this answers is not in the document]
eaidova: looks like now if user not set log level, it is disabled by default, right? or some changes requires to enable it?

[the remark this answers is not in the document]
alexsu52: If an user does not set log level, it is the INFO level enabled by default. It's mean that we print information about the progress of the quantization. The user can disable logger or change log level to get more information.

alexsu52 (commented):
LGTM. Please take this PR out of draft.

nncf/openvino/quantization/helpers.py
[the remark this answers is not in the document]
KodiaqQ: I think that @eaidova means forwarding the logging level from NNCF to POT at the same level. I mean, if the NNCF has log-level INFO, it should be the same in the POT. The same situation for DEBUG log level and etc.  The fact is that POT regulates AC logging in the same manner inside using these levels.
@eaidova am I correct?

examples/experimental/openvino/yolo_v5/main.py
```diff
@@ -0,0 +1,151 @@
...
+
+from yolov5.utils.general import check_dataset
+from yolov5.utils.general import download
+from yolov5.utils.dataloaders import create_dataloader
+from yolov5.val import run as validation_fn
+
+
+FILE = Path(__file__).resolve()
+# Relative path to the `yolo_v5` directory.
+ROOT = Path(os.path.relpath(FILE.parent, Path.cwd()))
```
vshampor, line 30: I'm pretty sure that everything can be done via `pathlib`, no need to involve `os.path`

examples/experimental/openvino/yolo_v5/main.py
```diff
@@ -0,0 +1,151 @@
...
+# Path to the directory where the original and quantized IR will be saved.
+MODEL_DIR = ROOT.joinpath('yolov5m_quantization')
+# Path to the dataset config from the `ultralytics/yolov5` repository.
+DATASET_CONFIG = ROOT.joinpath('yolov5', 'data', 'coco.yaml')
```
vshampor, line 34: Can be rewritten as `ROOT / 'yolov5' / 'data' / 'coco.yaml'`, consider doing this for the above too.

examples/experimental/openvino/yolo_v5/main.py
```diff
@@ -0,0 +1,151 @@
...
+    # expected input that can be used for the model inference.
+    def transform_fn(data_item):
+        images, *_ = data_item
+        images = images.float()
+        images = images / 255
+        images = images.cpu().detach().numpy()
+        return images
+
+    # Wrap framework-specific data source into the `nncf.Dataset` object.
+    calibration_dataet = nncf.Dataset(data_source, transform_fn)
```
vshampor, line 68: `calibration_dataset`

examples/experimental/openvino/yolo_v5/main.py
```diff
@@ -0,0 +1,151 @@
...
+    quantized_model = nncf.quantize(ov_model, calibration_dataet, preset=nncf.QuantizationPreset.MIXED)
```
vshampor, line 69: Unless we plan this function to be available with one and the same signature for all backends, it should be kept in the `nncf.openvino` namespace.

nncf/data/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+"""
+
+from typing import Iterable
+from typing import Any
+from typing import Callable
+from typing import Optional
+from typing import List
+
+
+DataItem = Any
```
vshampor, line 21: Use `TypeVar` instead of `Any`, because `Any` in different places within the same signature will match to objects of different type, while I think you need `DataItem` to represent one and the same type within one signature.

nncf/data/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+ModelInput = Any
+
+
+class Dataset:
```
vshampor, line 27: Consider subclassing from `Generic[DataItem, ModelInput]` to make it more obvious that this is template code.

nncf/data/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+        :param indices: The zero-based indices of data items that should be selected from
+            the data source. The indices should be sorted in ascending order. If indices are
+            not passed all data items are selected from the data source.
+        :return: The iterable object that contains selected data items from the data source, for which
+            the transformation function was applied.
+        """
+        return DataProvider(self._data_source, self._transform_func, indices)
+
+
+class DataProvider:
```
vshampor, line 83: Same

nncf/data/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+        if self._indices is None:
+            return map(self._transform_func, self._data_source)
+
+        if hasattr(self._data_source, '__getitem__'):
+            return DataProvider._get_iterator_for_map_style(self._data_source, self._transform_func, self._indices)
+
+        return DataProvider._get_iterator_for_iter(self._data_source, self._transform_func, self._indices)
+
+    @staticmethod
+    def _get_iterator_for_map_style(data_source, transform_func, indices):
```
vshampor, line 104: Type hints missing

nncf/data/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+        for index in indices:
+            yield transform_func(data_source[index])
+
+    @staticmethod
+    def _get_iterator_for_iter(data_source, transform_func, indices):
```
vshampor, line 109: Type hints missing

nncf/openvino/engine.py
```diff
@@ -0,0 +1,173 @@
...
+from openvino.tools import pot
+
+from nncf.api.compression import ModelType
+from nncf.data import Dataset
+
+
+logger = pot.utils.logger.get_logger(__name__)
+
+
+class DummyMetric(pot.Metric):
```
vshampor, line 30: Type hints missing throughout this class

nncf/openvino/engine.py
```diff
@@ -0,0 +1,173 @@
...
+        return attributes
+
+    def update(self, output, target):
+        raise NotImplementedError()
+
+    def reset(self):
+        self._avg_value = None
+
+
+class OVEngine(pot.IEEngine):
```
vshampor, line 68: Docs and typehints missing throughout the class

nncf/openvino/quantization/helpers.py
```diff
@@ -0,0 +1,106 @@
+"""
```
vshampor, line 1: Move functions from here to specialized files even if it means having one function to a file for now. The `_convert` functions may go to the `nncf/openvino/converters.py`, either as-is or consolidated into a `Converter` object or something like that, `quantize_impl` may go to `nncf/openvino/quantization/impl.py`

nncf/quantization/helpers.py
```diff
@@ -0,0 +1,60 @@
...
+from nncf.quantization.params import TargetDevice
+
+
+def quantize(model: ModelType,
+             calibration_dataset: Dataset,
+             preset: QuantizationPreset = QuantizationPreset.PERFORMANCE,
+             target_device: TargetDevice = TargetDevice.ANY,
+             subset_size: int = 300,
+             fast_error_correction: bool = True,
+             model_type: Optional[str] = None) -> ModelType:
```
vshampor, line 30: Need to think really hard on how we will expose this function and with which signature.

vshampor (changes requested):

examples/experimental/openvino/yolo_v5/main.py
andrey-churkin: Thanks for the comment. We plan to do that. So this method should be under the `nncf` namespace.

nncf/quantization/helpers.py
andrey-churkin: Thanks for the comment! It was discussed with @AlexKoff88. The proposal was implemented here. We would like to use this function as `nncf.quantize()`  with this list of parameters.

andrey-churkin: We will extend the list of parameters in the future. Probably, we should add the `advanced_parameters: AdvancedQuantizationParameters` optional argument to support features as ignored scope, some inference parameters and etc. We will discuss the structure of the `AdvancedQuantizationParameters` and introduce this in future PRs.

nncf/data/dataset.py
andrey-churkin: Done

nncf/data/dataset.py
andrey-churkin: Thanks for the comment! I changed it.

nncf/data/dataset.py
andrey-churkin: Done

examples/experimental/openvino/yolo_v5/main.py
andrey-churkin: Thanks for the comment! I fixed it.

examples/experimental/openvino/yolo_v5/main.py
andrey-churkin: Done

examples/experimental/openvino/yolo_v5/main.py
andrey-churkin: Done.

examples/experimental/openvino/mobilenet_v2/main.py
andrey-churkin: Thanks for the comment! I fixed it.

examples/experimental/openvino/mobilenet_v2/main.py
andrey-churkin: Done

nncf/openvino/quantization/helpers.py
andrey-churkin: The `_convert_openvino_model_to_compressed_model()` and `_convert_compressed_model_to_openvino_model` methods are used only inside the `nncf.openvino.quantization.helpers` module. Of course, it is better to create a separate module if these methods are needed in other places. But at the moment it's not like that. Probably these methods will be removed in the future.

nncf/data/dataloader.py
andrey-churkin: It was implemented.

nncf/data/utils.py
andrey-churkin: It was supported.

nncf/openvino/quantization/helpers.py
andrey-churkin: I think this case will be covered when we will have added the `advance_parameters: AdvancedQuantizationParameters` argument. So, we will be able to add some backend-specific parameters to be configured inside this class.

nncf/openvino/quantization/helpers.py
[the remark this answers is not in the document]
andrey-churkin: Done

l-bat (approved):

KodiaqQ (approved):

PR #1301, opened 2022-10-10
HWConfigType was renamed to TargetDevice

andrey-churkin:
### Changes

The `HWConfigType` class was renamed to the `TargetDevice` class.

### Reason for changes

There are a lot of places where we should pass the `target_device` parameter. One of them is the new `nncf.quantize()` method that was introduced in PR #1295. It would be great to have the understandable type alias to the `target_device` parameter.

### Related tickets

N/A

### Tests

N/A

Only renaming. No new features.

### DIFF
--- a/nncf/torch/quantization/layers.py
+++ b/nncf/torch/quantization/layers.py
@@ -49,7 +49,7 @@ class QuantizerExportMode(Enum):
     ONNX_QUANTIZE_DEQUANTIZE_PAIRS = "quantize_dequantize"
 
     @staticmethod
-    def from_str(config_value: str) -> 'HWConfigType':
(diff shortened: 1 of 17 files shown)


nncf/torch/quantization/layers.py
```diff
@@ -49,7 +49,7 @@ class QuantizerExportMode(Enum):
     ONNX_QUANTIZE_DEQUANTIZE_PAIRS = "quantize_dequantize"
 
     @staticmethod
-    def from_str(config_value: str) -> 'HWConfigType':
+    def from_str(config_value: str) -> 'QuantizerExportMode':
```
andrey-churkin, line 52: FYI, the bug was fixed.

PR #1309, opened 2022-10-10
Corrected extraction of layer attributes in case of weight_norm

### LINKED ISSUE #1308: RunTime Error: Frozen Layers are not allowed for elasticity
So we were trying to use TCN as input model for Bootstrap-Nas and we were continuously getting this error about frozen layers. 
It would be really great if we could get some guidance on how to overcome this issue. 
Link of repo which we used for training TCN model: [Link](https://github.com/locuslab/TCN/tree/master/TCN/mnist_pixel)

TCN-Frozen-Layers-Issue

ljaljushkin:
### Changes

Extract `requires_grad` attribute for weight_g and weight_v if there exist.

### Reason for changes

[torch.nn.utils.weight_norm](https://pytorch.org/docs/stable/_modules/torch/nn/utils/weight_norm.html#weight_norm)
“patches” ordinary module by removing weight and adding weight_g and weight_v.
The patched module looks as ordinary one. There's even a weight attribute, but it's just a tensor, not a learnable parameter anymore.
For conv modules, we always check `requires_grad` of weight attribute, and it's false after applying `weight_norm` utility. 
We should consider 'weight_g' or 'weight_v' if there exist.
The patched module looks as a frozen one, because weight is a tensor and requires_grad=False, but actually it's not because there's a modified functionality via weight_g and weight_v that are learnable.

### Related tickets

closes #1308

### Tests

random_multi_elasticity for TCN model


ljaljushkin:
@jpablomch @AmanGoyal99

vshampor (approved):

PR #1314, opened 2022-10-12
[Continue] Do not ignore the entire fused pattern if non-primary nodes are in the ignored scopes

ljaljushkin:
### Changes

Adjusted quantizer propagation common algo to only ignore the entire fused pattern if the primary node from the pattern ends up matching to an ignored scope.

### Reason for changes

Improves logical consistency of the "ignored_scopes" and its effect on the algo result.

### Related tickets

92302

### Tests

updated references + new test cases in test_activation_ignored_scope


ljaljushkin:
Some tests were failing in #1289:
1) `test_min_max_quantization_graph[model_to_test3]` (Maskrcnn-12)
Solution: made sure that the updated reference .dot file has new quantizers in the expected places.
Here's the proof:
maskrcnn_diff
maskrcnn_diff_fq onnx
maskrcnn_diff_graph
Add operation is ignored only. 
Matmul + Add is fused for ONNX: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/onnx/hardware/fused_patterns.py#L56
The entire fused pattern is not ignored, matmul is quantized, which is expected with the proposed changes #1289.

2) `quantization.test_algo_quantization.test_activation_ignored_scope[update_config_info1-should_ignore_quantizers1]`
`quantization.test_algo_quantization.test_activation_ignored_scope[update_config_info2-should_ignore_quantizers2]`
Solution: made sure that test produces quantizers in the expected places.
ReLU is ignored in Conv + ReLU pattern, but Conv's input is quantized, as expected.
lenet_test_case

vshampor (approved):

PR #1317, opened 2022-10-13
Elastic embedding

jpablomch:
### Changes

Enable elastic embedding for Neural Architecture Search

### Reason for changes

Used to create transformer-based super-networks

### Related tickets

N/A

### Tests

Tests included in PR.

### DIFF
--- a/nncf/common/pruning/utils.py
+++ b/nncf/common/pruning/utils.py
@@ -389,8 +389,8 @@ def get_input_channels(node: NNCFNode) -> int:
     :param node: Given prunable node.
     :return: Count of input channels of the given node.
     """
-    layer_attrs = node.layer_attributes # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes,
-                                        #             EmbeddingLayerAttributes]
+    layer_attrs = node.layer_attributes \
+        # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes, EmbeddingLayerAttributes]
     if isinstance(layer_attrs, ConvolutionLayerAttributes):
         return layer_attrs.in_channels
     if isinstance(layer_attrs, LinearLayerAttributes):
@@ -407,8 +407,8 @@ def get_output_channels(node: NNCFNode) -> int:
     :param node: Given prunable node.
     :return: Count of output channels of the given node.
     """
-    layer_attrs = node.layer_attributes # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes,
-                                        #             EmbeddingLayerAttributes]
+    layer_attrs = node.layer_attributes \
+        # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes, EmbeddingLayerAttributes]
     if isinstance(layer_attrs, ConvolutionLayerAttributes):
         return layer_attrs.out_channels
     if isinstance(layer_attrs, LinearLayerAttributes):
--- a/nncf/torch/dynamic_graph/wrappers.py
+++ b/nncf/torch/dynamic_graph/wrappers.py
@@ -217,6 +217,7 @@ def _collect_module_attrs_and_ignored_algorithms(ctx: TracingContext,
     return layer_attrs, ignored_algos
 
 
+# pylint:disable=too-many-return-statements
 def _get_layer_attributes(module: TorchModule, operator_name: str) -> BaseLayerAttributes:
     if operator_name == "group_norm":
         return GroupNormLayerAttributes(
--- a/tests/torch/nas/test_elastic_width.py
+++ b/tests/torch/nas/test_elastic_width.py
@@ -47,15 +47,6 @@ def _seed():
     manual_seed(0)
 
 
-@pytest.fixture(name='basic_model', params=(EmbeddingTwoSequentialFcLNTestModel, ConvTwoFcTestModel,
-                                            TwoConvAddConvTestModel, TwoSequentialConvBNTestModel))
-def fixture_basic_model(request):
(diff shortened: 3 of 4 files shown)


daniil-lyakhov:
Thank you for your contribution, @jpablomch!
Please rebase your PR on latest develop

nncf/common/pruning/operations.py
```diff
@@ -154,6 +154,18 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+class EmbeddingPruningOp(BasePruningOp):
+    @classmethod
+    def accept_pruned_input(cls, node: NNCFNode) -> bool:
+        return False
+
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
+                         tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> None:
+        output_mask = node.data.get('output_mask', None)
+        node.data['output_mask'] = output_mask
```
daniil-lyakhov, line 166: Should embedding layer have masks during pruning-aware training?

nncf/common/pruning/utils.py
```diff
@@ -387,11 +389,14 @@ def get_input_channels(node: NNCFNode) -> int:
     :param node: Given prunable node.
     :return: Count of input channels of the given node.
     """
-    layer_attrs = node.layer_attributes # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes]
+    layer_attrs = node.layer_attributes \
+        # type: Union[ConvolutionLayerAttributes, LinearLayerAttributes, EmbeddingLayerAttributes]
```
ljaljushkin, line 393: It seems like a useless type hint for me that just creates a noise.
@daniil-lyakhov do you remember why do we need it?

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -514,7 +530,7 @@ def __init__(self, target_model: NNCFNetwork,
         self._add_dynamic_inputs = add_dynamic_inputs
 
         graph = self._target_model.get_original_graph()
-        prunable_types = [NNCFConv2d.op_func_name, NNCFLinear.op_func_name]
+        prunable_types = [NNCFConv2d.op_func_name, NNCFLinear.op_func_name, NNCFEmbedding.op_func_name]
```
ljaljushkin, line 533: "Prunable" types are defined in 2 different places. Potentially, it could lead to inconsistency and increasing burden for developers to make it the same. I'd refactor it by creating `ShapePruningProcessor` in builder and by passing it to handler.

ljaljushkin (approved):
I have only 2 comments. 
The first one is the minor, let's see what @daniil-lyakhov would say about it.
The second is about cleaning some “technical debts” that can be done is a separate PR.
I'm ok to merge it.

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
ljaljushkin: Should be resolved in a separate PR
https://github.com/openvinotoolkit/nncf/issues/1326

daniil-lyakhov:
Jenkins please retest this

nncf/common/pruning/operations.py
```diff
@@ -154,6 +154,18 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+
+
```
daniil-lyakhov, line 168: Is it used somewhere? The logic here allows to produce masks, but it's never used in the pruning algo.

ljaljushkin: It's used here: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L625
Embedding became a prunable type and part of `PT_PRUNING_OPERATOR_METATYPES`, we set elastic width encoded in the `output_mask` and propagate them. 
```python
algo = MaskPropagationAlgorithm(
            self._propagation_graph, PT_PRUNING_OPERATOR_METATYPES, PTNNCFPruningTensorProcessor)
        algo.mask_propagation()
```

nncf/common/pruning/utils.py
ljaljushkin: `node.layer_attributes` is not an input parameter. But it can be mentioned in the description of `graph`.

nncf/common/pruning/operations.py
daniil-lyakhov: I'm only concerned that filter pruning user doesn't expect embedding layer could possibly produce masks. It works now and doesn't introduce any side effects for FilterPruning as far as I understand(because embedding layer isn't updated with preop, and model analysis doesn't consider embedding as prunable), but in general this PruningOp looks unfinished from Filter Pruning algorithm side.
This change looks implicit, to enable elastic embedding for NAS algo but shown as common. I would be concerned if I would see this

ljaljushkin (changes requested):
The comment is valid and should be resolved prior to merge.
https://github.com/openvinotoolkit/nncf/pull/1317#discussion_r999592363
If Embedding is introduced in a common part, it should be supported by common filter pruning algorithm.
The alternative option to register a NAS specific EmbeddingOp that doesn't affect common Filter Pruning algorithm.

ljaljushkin:
Hi @jpablomch! This PR looks inactive.
If it is still relevant, please rebase and address comments, or close otherwise.

PR #1322, opened 2022-10-14
Use reference quantization on CPU if compiled extensions are unavailable

vshampor:
### Changes
As stated in the title. Also changed the way of loading extensions to the lazy approach, so that the extensions are only loaded if associated quantization or binarization functionality is invoked, meaning that non-INT8 workflows will have less points of failure for the user.

### Reason for changes
UX improvement due to not failing where we may keep working functionally.

### Related tickets
94348

### Tests
test_extension_unavailable

### DIFF
--- a/nncf/torch/binarization/binarize_functions.py
+++ b/nncf/torch/binarization/binarize_functions.py
@@ -17,8 +17,8 @@ import torch
 
 from nncf.common.logging import nncf_logger
 from nncf.torch.utils import add_domain
+from nncf.torch.binarization.extensions import BinarizedFunctionsCUDA
 
-from .extensions import BinarizedFunctionsCUDA
 from torch import _C  # pylint:disable=protected-access
 
 
@@ -68,10 +68,9 @@ class XNORBinarizeFn(torch.autograd.Function):
     @staticmethod
     def forward(ctx, x):
         if x.is_cuda:
-            output = BinarizedFunctionsCUDA.WeightBinarize_forward(x, True)
+            output = BinarizedFunctionsCUDA.get("WeightBinarize_forward")(x, True)
         else:
             # Current CPU kernel implementations do not improve performance
-            # output = BinarizedFunctionsCPU.WeightBinarize_forward(x, True)
             norm = x.abs().mean([1, 2, 3], keepdim=True)
             sign = ((x > 0).type(x.dtype) * 2 - 1)
             output = sign * norm
@@ -100,10 +99,9 @@ class DOREFABinarizeFn(torch.autograd.Function):
     @staticmethod
     def forward(ctx, x):
         if x.is_cuda:
-            output = BinarizedFunctionsCUDA.WeightBinarize_forward(x, False)
+            output = BinarizedFunctionsCUDA.get("WeightBinarize_forward")(x, False)
         else:
             # Current CPU kernel implementations do not improve performance
-            # output = BinarizedFunctionsCPU.WeightBinarize_forward(x, False)
             norm = x.abs().mean()
             sign = ((x > 0).type(x.dtype) * 2 - 1)
             output_flat = sign * norm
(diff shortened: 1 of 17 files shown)


tests/shared/helpers.py
```diff
@@ -28,6 +28,10 @@
 from tests.shared.paths import PROJECT_ROOT
 
 TensorType = TypeVar('TensorType')
+TEST_ROOT = Path(__file__).absolute().parents[1]  # https://github.com/PyCQA/pylint/issues/5783
+PROJECT_ROOT = TEST_ROOT.parent.absolute()  # pylint:disable=no-member
+EXAMPLES_DIR = PROJECT_ROOT / 'examples'
+GITHUB_REPO_URL = 'https://github.com/openvinotoolkit/nncf/'
```
ljaljushkin, line 34: Is it used somewhere?

tests/torch/quantization/test_functions.py
```diff
@@ -96,39 +47,51 @@ def _seed():
...
+                               bits: int,
+                               get_deviation: Callable[[], int],
+                               min_deviation: float):
+    if np.prod(input_low.shape) == 1:
+        input_low_ch, input_range_ch = \
+            map(lambda x: x.squeeze(), [input_low, input_range])
     else:
-        min_adj_ch, input_low_ch, input_range_ch, quant_len_ch =\
-            map(lambda x: x[tuple(ch_idx)].squeeze(), [min_adj, input_low, input_range, quant_len])
+        input_low_ch, input_range_ch = \
```
ljaljushkin, line 64: Is it equivalent without `min_adj_ch` and `quant_len`? Why was it needed before?

tests/torch/quantization/test_functions.py
```diff
@@ -242,7 +207,7 @@ def to_tensor(a):
 
 
 def check_outputs_for_quantization_functions(test_val: torch.Tensor, ref_val: np.ndarray,
-                                             is_fp16: bool, rtol=1e-4, atol=1e-10):
```
ljaljushkin, line 245: Was `is_fp16` unused before?

ljaljushkin (changes requested):

nncf/torch/binarization/binarize_functions.py
```diff
@@ -17,6 +17,7 @@
 
 from nncf.common.logging import nncf_logger
 from nncf.torch.utils import add_domain
+from nncf.torch.binarization.extensions import BinarizedFunctionsCUDA
 
 from .extensions import BinarizedFunctionsCUDA
```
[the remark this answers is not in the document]
vshampor: Thanks, will remove

tests/shared/helpers.py
vshampor: Indeed, in `tests.shared.helpers.create_venv_with_nncf` when we try to install from a commit on github.

tests/torch/quantization/test_functions.py
vshampor: No idea, the original code was by @daniil-lyakhov. I reviewed the usage of these variables carefully and saw no reason to keep them, they are equivalent to some arguments already in the scope. Maybe @daniil-lyakhov just forgot to remove them.

tests/torch/quantization/test_functions.py
vshampor: No, even my IDE marked this as an unused parameter.

alexsu52 (approved):
LGTM

tests/shared/helpers.py
ljaljushkin: I can't see the usage of `GITHUB_REPO_URL` in your PR. 
The same 4 lines are defined in the tests/shared/paths.py
Why not reuse them from this file?

nncf/torch/binarization/binarize_functions.py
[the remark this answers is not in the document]
vshampor: Done

tests/shared/helpers.py
vshampor: I get it now, that was a rebase artifact. Removed.

ljaljushkin (approved):

tests/torch/quantization/test_functions.py
daniil-lyakhov: Yeah, for previous version of the quantize function this `min_adj_ch` was not equal to `-inp_low`. So, this changes are totally valid

daniil-lyakhov (approved):

PR #1327, opened 2022-10-18
NNCF ONNX PTQ Add support of quantization of weights for elementwise operations

kshpv:
### Reason for changes

To get the correct quantization schema of Mask RCNN model.

### Related tickets

92537

### Tests

The test on algorithm for weight nodes detection is added


kshpv:
run onnx pre-commit tests

PR #1319, opened 2022-10-13
Joint Pruning, Quantization and Distillation for Transformers

vuiseng9:
### Changes

Add movement pruning feature, enabling end-to-end compression for transformer inference acceleration via OpenVINO

### Tests

Some basic tests are included, team will continue to refactor, add tests to this pull request.

PR #1328, opened 2022-10-18
Updated patch for NNCF integration to the latest transformers release

ljaljushkin:
### Changes

Updated patch for NNCF integration to the latest transformers release (v4.23.1)

### Reason for changes

for smoother integration of JPQD #1319

### Related tickets

94449

### Tests

3rd party sanity

### DIFF
--- a/tests/torch/test_sanity_third_party.py
+++ b/tests/torch/test_sanity_third_party.py
@@ -21,7 +21,7 @@ from nncf.torch import BKC_TORCH_VERSION
 from tests.common.helpers import PROJECT_ROOT
 from tests.torch.helpers import Command
 
-TRANSFORMERS_COMMIT = "bff1c71e84e392af9625c345f9ea71f7b6d75fb3"
+TRANSFORMERS_COMMIT = "bd469c40659ce76c81f69c7726759d249b4aef49"
 INSTALL_PATH = PROJECT_ROOT.parent
 DATASET_PATH = os.path.join(PROJECT_ROOT, "tests", "torch", "data", "mock_datasets")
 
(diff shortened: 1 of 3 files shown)


tests/torch/test_sanity_third_party.py
```diff
@@ -204,7 +204,8 @@ def test_glue_distilbert_eval(self, temp_folder):
 
     @pytest.mark.dependency(depends=['install_trans'], name='lm_train')
     def test_lm_train(self, temp_folder):
-        com_line = "examples/pytorch/language-modeling/run_clm.py --model_name_or_path gpt2" \
+        # GPT2 is loaded via torch.frombuffer which is not available in torch==1.9.1 yet
+        com_line = "examples/pytorch/language-modeling/run_clm.py --model_name_or_path distilgpt2" \
```
ljaljushkin, line 208: GPT2 is loaded via `safetensor` functionality that uses `torch.frombuffer`. This function appears since torch1.10, but we support torch==1.9.1 only.

tests/torch/test_sanity_third_party.py
```diff
@@ -21,7 +21,7 @@
 from tests.common.helpers import PROJECT_ROOT
 from tests.torch.helpers import Command
 
-TRANSFORMERS_COMMIT = "bff1c71e84e392af9625c345f9ea71f7b6d75fb3"
+TRANSFORMERS_COMMIT = "bd469c40659ce76c81f69c7726759d249b4aef49"
```
ljaljushkin, line 24: v4.23.1

ljaljushkin:
@vshampor 3rd party sanity tests are green (Build 23)
the only degradation is `gpt2` model, which can't be loaded with pytorch 1.9.1.
Sanity tests use 'distilgpt2' instead, which doesn't have an issue with this version of pytorch.

vshampor (approved):

PR #1330, opened 2022-10-19
Added the ignored scope parameter to the quantize function signature

alexsu52:
### Changes
Added the ignored scope parameter to the quantize function signature. Support the ignored scope in the OpenVINO backend.

### Reason for changes

Feature alignment with POT OpenVINO

### Related tickets

N/A

### Tests

[test_helpers.py](https://github.com/openvinotoolkit/nncf/blob/1f09ad3b4cb5c3b6805bece2abb3a4e29f564cc5/tests/openvino/quantization/test_helpers.py)

### DIFF
--- a/nncf/openvino/quantization/helpers.py
+++ b/nncf/openvino/quantization/helpers.py
@@ -11,21 +11,21 @@
  limitations under the License.
 """
 
+import logging
 import tempfile
 from pathlib import Path
-from typing import Optional
-import logging
+from typing import Dict, Optional
 
 import openvino.runtime as ov
 from openvino.tools import pot
 
-from nncf.data import Dataset
-from nncf.openvino.utils import POTDataLoader
-from nncf.openvino.engine import OVEngine
-from nncf.parameters import IgnoredScope
-from nncf.parameters import TargetDevice
 from nncf.common.quantization.structs import QuantizationPreset
 from nncf.common.utils.logger import logger as nncf_logger
+from nncf.data import Dataset
+from nncf.openvino.engine import OVEngine
+from nncf.openvino.utils import POTDataLoader
+from nncf.parameters import IgnoredScope
+from nncf.parameters import TargetDevice
 
 
 def _convert_openvino_model_to_compressed_model(model: ov.Model,
@@ -67,7 +67,7 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
     return ov_model
 
 
-def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> dict:
+def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
     """
     Creates POT ignored scope configuration from the ignored scope that is
     defined using `IgnoredScope` class.
@@ -84,7 +84,7 @@ def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> dict:
         if isinstance(node_names, str):
             node_names = [node_names]
         ignored['scope'] = node_names
-    if ignored_scope.node_name_regexps is not None:
+    if ignored_scope.node_name_patterns is not None:
         raise RuntimeError('Quantization algorithm form the OpenVINO backend '
                            'does not support node name regular expressions '
                            'in the ignored scopes yet')
--- a/nncf/openvino/quantization/helpers.py
+++ b/nncf/openvino/quantization/helpers.py
@@ -78,20 +78,20 @@ def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
         return {}
 
     ignored = {}
-    if ignored_scope.node_names is not None:
-        node_names = ignored_scope.node_names
-        if isinstance(node_names, str):
-            node_names = [node_names]
-        ignored['scope'] = node_names
-    if ignored_scope.node_name_patterns is not None:
+    if ignored_scope.names is not None:
+        names = ignored_scope.names
+        if isinstance(names, str):
+            names = [names]
+        ignored['scope'] = names
+    if ignored_scope.patterns is not None:
         raise RuntimeError('Quantization algorithm form the OpenVINO backend '
-                           'does not support node name regular expressions '
-                           'in the ignored scopes yet')
-    if ignored_scope.node_types is not None:
-        node_types = ignored_scope.node_types
-        if isinstance(node_types, str):
-            node_types = [node_types]
-        ignored['operations'] = [{'type': node_type} for node_type in node_types]
+                           'does not support regular expressions in the ignored '
+                           'scopes yet')
+    if ignored_scope.types is not None:
+        types = ignored_scope.types
+        if isinstance(types, str):
+            types = [types]
+        ignored['operations'] = [{'type': type} for type in types]
     return ignored
 
 
--- a/nncf/openvino/quantization/helpers.py
+++ b/nncf/openvino/quantization/helpers.py
@@ -69,8 +69,7 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
 
 def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
     """
-    Creates POT ignored scope configuration from the ignored scope that is
-    defined using `IgnoredScope` class.
+    Maps the content of `IgnoredScope` class to the `ignored` section of POT config.
 
     :param ignored_scope: The ignored scope
     :return: A POT ignored scope configuration as dict
--- a/nncf/openvino/quantization/helpers.py
+++ b/nncf/openvino/quantization/helpers.py
@@ -67,7 +67,7 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
     return ov_model
 
 
-def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
+def _create_ignored_scope_config(ignored_scope: Optional[IgnoredScope]) -> Dict:
     """
     Maps the content of `IgnoredScope` class to the `ignored` section of POT config.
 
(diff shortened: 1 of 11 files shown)


nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,41 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
...
+    defined using `IgnoredScope` class.
+
+    :param ignored_scope: The ignored scope
+    :return: A POT ignored scope configuration as dict
+    """
+    ignored = {}
+    if ignored_scope.node_names is not None:
+        ignored['scope'] = ignored_scope.node_names
+    if ignored_scope.regexps is not None:
+        raise RuntimeError('Quantization algorithm form the OpenVino backend '
```
AlexKoff88, line 82:
```suggestion
        raise RuntimeError('Quantization algorithm form the OpenVINO backend '
```

nncf/openvino/quantization/helpers.py
```diff
@@ -96,7 +125,8 @@ def quantize_impl(model: ov.Model,
                 'stat_subset_size': subset_size,
                 'use_fast_bias': fast_error_correction,
                 'model_type': model_type,
-            }
+            },
+            'ignored': _create_ignored_scope_config(ignored_scope)
```
AlexKoff88, line 129: some tests are required for sure maybe without running anything but checking the result of `_create_ignored_scope_config` functions

alexsu52: Done

nncf/parameters.py
```diff
@@ -0,0 +1,53 @@
...
+
+class IgnoredScope:
+    """
+    Dataclass that contains description of the ignored scope
+    """
+
+    def __init__(self,
+                 node_names: List[str] = None,
+                 node_name_regexps: List[str] = None,
+                 node_types: List[str] = None):
```
andrey-churkin, line 44:
```suggestion
                 node_names: Optional[List[str]] = None,
                 node_name_regexps: Optional[List[str]] = None,
                 node_types: Optional[List[str]] = None):
```

nncf/parameters.py
```diff
@@ -0,0 +1,53 @@
...
+    ANY = 'ANY'
+    CPU = 'CPU'
+    GPU = 'GPU'
+    VPU = 'VPU'
+
+
+class IgnoredScope:
+    """
+    Dataclass that contains description of the ignored scope
+    """
```
andrey-churkin, line 39: It would be great to describe here what ignored scope means

nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,43 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
     return ov_model
 
 
+def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> dict:
```
KodiaqQ, line 70: `-> Dict:`?

nncf/parameters.py
```diff
@@ -0,0 +1,53 @@
+"""
```
andrey-churkin, line 1: I think it is better to move this file to a separate subpackage `nncf.parameters`

alexsu52: Could you briefly explain why?

nncf/parameters.py
alexsu52: done

KodiaqQ (approved):

andrey-churkin (approved):

nncf/parameters.py
andrey-churkin: Probably it does not make sense to do this at this stage.

nncf/parameters.py
```diff
@@ -0,0 +1,82 @@
...
+
+    ANY = 'ANY'
+    CPU = 'CPU'
+    GPU = 'GPU'
+    VPU = 'VPU'
+
+
+class IgnoredScope:
+    """
+    Dataclass that contains description of the ignored scope.
```
vshampor, line 38: This is not a `dataclass` in a Python sense, at least not yet.

nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,43 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
     return ov_model
 
 
+def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
+    """
+    Creates POT ignored scope configuration from the ignored scope that is
```
AlexKoff88, line 72:
```suggestion
    Maps the content of `IgnoredScope` class to `ignored` section of POT config.
```

nncf/parameters.py
```diff
@@ -0,0 +1,82 @@
...
+
+    The ignored scope specify model sub-graphs which should be excluded from
+    the optimization process such as quantization, pruning and etc.
+
+    For example, you want to exclude some model nodes from the optimization
```
AlexKoff88, line 43: You can go different way here, like:
```
Examples:
```
Then the place the following code with comments:
```
import nncf

# Exclude by node name:
node_names = ['node_1', 'node_2', 'node_3']
ignored_scope = nncf.IgnoredScope(node_names=node_names)

# Exclude using regexp:
pattern = ['node_\d']
ignored_scope = nncf.IgnoredScope(node_name_patterns=pattern)

# Exclude by operation type (for OpenVINO model only)
node_types = ['Multiply', 'GroupConvolution', 'Interpolate']
ignored_scope = nncf.IgnoredScope(node_types=pattern)
```

nncf/parameters.py
```diff
@@ -0,0 +1,82 @@
...
+    process by node name:
+
+    ```
+    import nncf
+    node_names = ['node_1', 'node_2', 'node_3']
+    ignored_scope = nncf.IgnoredScope(node_names=node_names)
```
AlexKoff88, line 49: `node_` looks excessive really. Can we go without it, like `names`, `patterns`, `types`.

nncf/parameters.py
alexsu52: Looks like `dataclass`, but it's not python `dataclass` yet :)

nncf/parameters.py
alexsu52: Thanks for the comment! To be honest, I did as you suggested at first. 

Of course, if we do this, then the attribute names will become shorter, but won't the readability of the code suffer?

IgnoredScope is impersonal by definition. If we change `node_names` -> `names`, the user may have a question: the name of what? Then the user have to go to the documentation... @AlexKoff88, what do you think of the above?

@KodiaqQ @andrey-churkin @vshampor please share your opinion.

andrey-churkin: I think in both cases user should read the documentation about this class before writing some code because the `IgnoredScope` is an advanced feature. The `names`, `patterns`, and `types` look better for me.

KodiaqQ: Based on POT's experience, users understand this field in their own way: one will take this field as it should and `insert the names of the original layers` there, and the other will take it differently and `insert the names of the quantizers (or fake quantize nodes)` there. And both of them may be wrong, because, the names from the NNСFGraph were implied, for example.
My comment is that whatever field name we set, users will perceive it in their own way (and they will be right), so it's better just to write good documentation. We will not be able to come up with a concise and understandable field name for everyone, in this case, it seems to me.

nncf/parameters.py
alexsu52: Please take a look.

nncf/parameters.py
alexsu52: Please take a look.

Please do not hesitate if need to do some changes or rollback changes.

AlexKoff88 (approved):

nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,42 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
     return ov_model
 
 
+def _create_ignored_scope_config(ignored_scope: IgnoredScope) -> Dict:
```
andrey-churkin, line 70:
```suggestion
def _create_ignored_scope_config(ignored_scope: Optional[IgnoredScope] = None) -> Dict:
```

nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,42 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
...
+    :return: A POT ignored scope configuration as dict
+    """
+    if ignored_scope is None:
+        return {}
+
+    ignored = {}
+    if ignored_scope.names is not None:
+        names = ignored_scope.names
+        if isinstance(names, str):
+            names = [names]
```
andrey-churkin, line 84: The `IgnoredScope.names` attribute cannot have the `str` type.

nncf/openvino/quantization/helpers.py
```diff
@@ -66,13 +67,42 @@ def _convert_compressed_model_to_openvino_model(model: pot.graph.nx_model.Compre
...
+        ignored['scope'] = names
+    if ignored_scope.patterns is not None:
+        raise RuntimeError('Quantization algorithm form the OpenVINO backend '
+                           'does not support regular expressions in the ignored '
+                           'scopes yet')
+    if ignored_scope.types is not None:
+        types = ignored_scope.types
+        if isinstance(types, str):
+            types = [types]
```
andrey-churkin, line 93: The same as above.

andrey-churkin:
LGTM, please take a look at the comments above. Thanks.

andrey-churkin (approved):

nncf/openvino/quantization/helpers.py
alexsu52: It's just sugar... avoids common mistakes when defining an ignored scope

nncf/openvino/quantization/helpers.py
alexsu52: It's just sugar... avoids common mistakes when defining an ignored scope

vshampor (approved):

eaidova (approved):

PR #1339, opened 2022-10-25
Add telemetry to NNCF

vshampor:
### Changes

Implemented basic scope of telemetry events, using `openvino-telemetry` as a backend. Currently directing events to a test GA dashboard.

### Reason for changes
Usage analytics

### Related tickets
test_telemetry* in common and all backends

### Tests
TBA

### DIFF
--- a/nncf/common/composite_compression.py
+++ b/nncf/common/composite_compression.py
@@ -322,7 +322,10 @@ class CompositeCompressionAlgorithmController(CompressionAlgorithmController):
             assert backend is BackendType.TORCH
             from nncf.torch.exporter import PTExporter #pylint: disable=cyclic-import
             exporter = PTExporter(self.model, input_names, output_names, model_args)
-        exporter.export_model(save_path, save_format)
+        if save_format is not None:
+            exporter.export_model(save_path, save_format)
+        else:
+            exporter.export_model(save_path)
 
     def disable_scheduler(self) -> None:
         self._scheduler = CompositeCompressionScheduler()
--- a/nncf/definitions.py
+++ b/nncf/definitions.py
@@ -19,4 +19,4 @@ HW_CONFIG_RELATIVE_DIR = "common/hardware/configs"
 # so that certain actions within NNCF proper, such as telemetry event collection or
 # debug dumps, are performed or not performed
 NNCF_CI_ENV_VAR_NAME = "NNCF_CI"  # Must be set in CI environments
-NNCF_DEV_ENV_VAR_NAME = "NNCF_DEV" # Must be set in environments of the NNCF dev team machines
+NNCF_DEV_ENV_VAR_NAME = "NNCF_DEV"  # Must be set in environments of the NNCF dev team machines
--- a/nncf/experimental/quantization/compression_builder.py
+++ b/nncf/experimental/quantization/compression_builder.py
@@ -21,10 +21,8 @@ from nncf.experimental.quantization.telemetry_extractors import CompressionStart
 
 from nncf.quantization.algorithms.algorithm import Algorithm
 from nncf.common.tensor_statistics.aggregator import StatisticsAggregator
-from nncf.telemetry import NNCFTelemetry
 from nncf.telemetry import tracked_function
 from nncf.telemetry.events import NNCF_ONNX_CATEGORY
-from nncf.telemetry.events import NNCF_TF_CATEGORY
 
 TModel = TypeVar('TModel')
 
@@ -69,7 +67,7 @@ class CompressionBuilder:
 
         return None
 
-    @tracked_function(NNCF_TF_CATEGORY, [CompressionStartedFromBuilder(argname="self"), ])
+    @tracked_function(NNCF_ONNX_CATEGORY, [CompressionStartedFromBuilder(argname="self"), ])
     def apply(self, model: TModel, dataset: Dataset) -> TModel:
         """
         Apply compression algorithms to the 'model'.
@@ -82,7 +80,6 @@ class CompressionBuilder:
         5) Collect all statistics.
         6) Apply algorithms.
         """
-        algo_names = sorted([x.__class__.__name__ for x in self.algorithms])
 
         if not self.algorithms:
             nncf_logger.info('There are no algorithms added. The original model will be returned.')
--- a/nncf/experimental/quantization/compression_builder.py
+++ b/nncf/experimental/quantization/compression_builder.py
@@ -17,11 +17,14 @@ from nncf import Dataset
 from nncf.common.utils.logger import logger as nncf_logger
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
+from nncf.experimental.quantization.telemetry_extractors import CompressionStartedFromBuilder
 
 from nncf.quantization.algorithms.algorithm import Algorithm
 from nncf.common.tensor_statistics.aggregator import StatisticsAggregator
 from nncf.telemetry import NNCFTelemetry
+from nncf.telemetry import tracked_function
 from nncf.telemetry.events import NNCF_ONNX_CATEGORY
+from nncf.telemetry.events import NNCF_TF_CATEGORY
 
 TModel = TypeVar('TModel')
 
@@ -66,6 +69,7 @@ class CompressionBuilder:
 
         return None
 
+    @tracked_function(NNCF_TF_CATEGORY, [CompressionStartedFromBuilder(argname="self"), ])
     def apply(self, model: TModel, dataset: Dataset) -> TModel:
         """
         Apply compression algorithms to the 'model'.
@@ -78,11 +82,7 @@ class CompressionBuilder:
         5) Collect all statistics.
         6) Apply algorithms.
         """
-        NNCFTelemetry.start_session(NNCF_ONNX_CATEGORY)
         algo_names = sorted([x.__class__.__name__ for x in self.algorithms])
-        NNCFTelemetry.send_event(event_category=NNCF_ONNX_CATEGORY,
-                                 event_action='compression_started',
-                                 event_label=','.join(algo_names))
 
         if not self.algorithms:
             nncf_logger.info('There are no algorithms added. The original model will be returned.')
@@ -105,5 +105,4 @@ class CompressionBuilder:
 
         for algorithm in self.algorithms:
             modified_model = algorithm.apply(modified_model, statistics_aggregator.statistic_points)
-        NNCFTelemetry.end_session(NNCF_ONNX_CATEGORY)
         return modified_model
--- a/nncf/experimental/quantization/compression_builder.py
+++ b/nncf/experimental/quantization/compression_builder.py
@@ -15,13 +15,13 @@ from typing import TypeVar
 
 from nncf import Dataset
 from nncf.common.utils.logger import logger as nncf_logger
-from nncf.common.graph.model_transformer import ModelTransformer
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
 
-from nncf.common.engine import Engine
 from nncf.quantization.algorithms.algorithm import Algorithm
 from nncf.common.tensor_statistics.aggregator import StatisticsAggregator
+from nncf.telemetry_wrapper.telemetry import NNCFTelemetry
+from nncf.telemetry_wrapper.telemetry import NNCF_ONNX_CATEGORY
 
 TModel = TypeVar('TModel')
 
@@ -78,6 +78,11 @@ class CompressionBuilder:
         5) Collect all statistics.
         6) Apply algorithms.
         """
+        NNCFTelemetry.start_session(NNCF_ONNX_CATEGORY)
+        algo_names = sorted([x.__class__.__name__ for x in self.algorithms])
+        NNCFTelemetry.send_event(event_category=NNCF_ONNX_CATEGORY,
+                                 event_action='compression_started',
+                                 event_label=','.join(algo_names))
 
         if not self.algorithms:
             nncf_logger.info('There are no algorithms added. The original model will be returned.')
@@ -91,6 +96,7 @@ class CompressionBuilder:
         modified_model = self._get_prepared_model_for_compression(model, backend)
 
         statistics_aggregator = self._create_statistics_aggregator(dataset, backend)
+
         for algorithm in self.algorithms:
             statistic_points = algorithm.get_statistic_points(modified_model)
             statistics_aggregator.register_stastistic_points(statistic_points)
@@ -99,4 +105,5 @@ class CompressionBuilder:
 
         for algorithm in self.algorithms:
             modified_model = algorithm.apply(modified_model, statistics_aggregator.statistic_points)
+        NNCFTelemetry.end_session(NNCF_ONNX_CATEGORY)
         return modified_model
--- a/nncf/experimental/quantization/compression_builder.py
+++ b/nncf/experimental/quantization/compression_builder.py
@@ -20,8 +20,8 @@ from nncf.common.utils.backend import get_backend
 
 from nncf.quantization.algorithms.algorithm import Algorithm
 from nncf.common.tensor_statistics.aggregator import StatisticsAggregator
-from nncf.telemetry_wrapper.telemetry import NNCFTelemetry
-from nncf.telemetry_wrapper.telemetry import NNCF_ONNX_CATEGORY
+from nncf.telemetry import NNCFTelemetry
+from nncf.telemetry.events import NNCF_ONNX_CATEGORY
 
 TModel = TypeVar('TModel')
 
--- a/nncf/openvino/quantization/quantize.py
+++ b/nncf/openvino/quantization/quantize.py
@@ -31,6 +31,8 @@ from nncf.common.quantization.structs import QuantizationPreset
 from nncf.common.utils.logger import logger as nncf_logger
 from nncf.openvino.engine import OVEngine
 from nncf.openvino.quantization.accuracy_aware import NMSEBasedAccuracyAware
+from nncf.telemetry_wrapper.telemetry import NNCFTelemetry
+from nncf.telemetry_wrapper.telemetry import NNCF_OV_CATEGORY
 
 
 def _convert_openvino_model_to_compressed_model(model: ov.Model,
@@ -105,16 +107,11 @@ def quantize_impl(model: ov.Model,
     """
     Implementation of the `quantize()` method for the OpenVINO backend.
     """
+
+    NNCFTelemetry.start_session(NNCF_OV_CATEGORY)
     pot.utils.logger.init_logger(
         level=logging.getLevelName(nncf_logger.getEffectiveLevel())
     )
-    pot_model = _convert_openvino_model_to_compressed_model(model, target_device)
-
-    engine_config = {
-        'device': 'CPU',
-        'stat_requests_number': 2,
-        'eval_requests_number': 2,
-    }
 
     algorithms = [
         {
@@ -130,12 +127,45 @@ def quantize_impl(model: ov.Model,
         }
     ]
 
+    algo_names = sorted([x['name'] for x in algorithms])
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='compression_started',
+                             event_label=','.join(algo_names))
+
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='target_device',
+                             event_label=target_device.value)
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='preset',
+                             event_label=preset.value)
+
+    pot_model = _convert_openvino_model_to_compressed_model(model, target_device)
+
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='ov_model_converted_to_compressed',
+                             event_label='')
+    engine_config = {
+        'device': 'CPU',
+        'stat_requests_number': 2,
+        'eval_requests_number': 2,
+    }
+
     engine = OVEngine(engine_config, calibration_dataset, calibration_dataset)
     pipeline = pot.create_pipeline(algorithms, engine)
     compressed_model = pipeline.run(pot_model)
+
     pot.compress_model_weights(compressed_model)
 
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='model_compressed',
+                             event_label='')
+
     quantized_model = _convert_compressed_model_to_openvino_model(compressed_model)
+
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
+                             event_action='compressed_model_converted_to_ov',
+                             event_label='')
+    NNCFTelemetry.end_session(NNCF_OV_CATEGORY)
 
     return quantized_model
 
@@ -198,6 +228,8 @@ def quantize_with_accuracy_control_impl(model: ov.Model,
     compressed_model = pipeline.run(pot_model)
     pot.compress_model_weights(compressed_model)
 
+
     quantized_model = _convert_compressed_model_to_openvino_model(compressed_model)
+
 
     return quantized_model
--- a/nncf/openvino/quantization/quantize.py
+++ b/nncf/openvino/quantization/quantize.py
@@ -31,8 +31,8 @@ from nncf.common.quantization.structs import QuantizationPreset
 from nncf.common.utils.logger import logger as nncf_logger
 from nncf.openvino.engine import OVEngine
 from nncf.openvino.quantization.accuracy_aware import NMSEBasedAccuracyAware
-from nncf.telemetry_wrapper.telemetry import NNCFTelemetry
-from nncf.telemetry_wrapper.telemetry import NNCF_OV_CATEGORY
+from nncf.telemetry import NNCFTelemetry
+from nncf.telemetry.events import NNCF_OV_CATEGORY
 
 
 def _convert_openvino_model_to_compressed_model(model: ov.Model,
--- a/nncf/openvino/quantization/quantize.py
+++ b/nncf/openvino/quantization/quantize.py
@@ -109,7 +109,6 @@ def quantize_impl(model: ov.Model,
     """
     Implementation of the `quantize()` method for the OpenVINO backend.
     """
-
     pot.utils.logger.init_logger(
         level=logging.getLevelName(nncf_logger.getEffectiveLevel())
     )
@@ -202,8 +201,6 @@ def quantize_with_accuracy_control_impl(model: ov.Model,
     compressed_model = pipeline.run(pot_model)
     pot.compress_model_weights(compressed_model)
 
-
     quantized_model = _convert_compressed_model_to_openvino_model(compressed_model)
-
 
     return quantized_model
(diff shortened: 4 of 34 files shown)


aiarkinx:
run openvino pre-commit tests

nncf/telemetry_wrapper/telemetry.py
```diff
@@ -0,0 +1,49 @@
...
+# https://support.google.com/analytics/answer/1033068
+try:
+    from openvino_telemetry import Telemetry
+    NNCFTelemetry = Telemetry(app_name='nncf', app_version=__version__, tid=MEASUREMENT_ID)
+except ImportError:
+    nncf_logger.debug("openvino_telemetry package not found. No telemetry will be sent.")
+    NNCFTelemetry = None
+
+if os.getenv(NNCF_CI_ENV_VAR_NAME) or os.getenv(NNCF_DEV_ENV_VAR_NAME) or NNCFTelemetry is None:
+    from unittest.mock import MagicMock
```
ljaljushkin, line 36: Seems like we need to specify this dependency for nncf

ljaljushkin (changes requested):

nncf/telemetry_wrapper/telemetry.py
vshampor: It's a builtin since 3.3.
https://docs.python.org/3/library/unittest.mock.html

nncf/telemetry_wrapper/telemetry.py
```diff
@@ -0,0 +1,49 @@
...
+# For recommendations on proper usage of categories, actions, labels and values, see:
+# https://support.google.com/analytics/answer/1033068
+try:
+    from openvino_telemetry import Telemetry
+    NNCFTelemetry = Telemetry(app_name='nncf', app_version=__version__, tid=MEASUREMENT_ID)
+except ImportError:
+    nncf_logger.debug("openvino_telemetry package not found. No telemetry will be sent.")
+    NNCFTelemetry = None
+
+if os.getenv(NNCF_CI_ENV_VAR_NAME) or os.getenv(NNCF_DEV_ENV_VAR_NAME) or NNCFTelemetry is None:
```
andrey-churkin, line 35: Do we really need both `NNCF_CI_ENV_VAR_NAME` and `NNCF_DEV_ENV_VAR_NAME`?

nncf/definitions.py
```diff
@@ -14,3 +14,5 @@
 import os
 NNCF_PACKAGE_ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 HW_CONFIG_RELATIVE_DIR = "common/hardware/configs"
+NNCF_CI_ENV_VAR_NAME = "NNCF_CI"
+NNCF_DEV_ENV_VAR_NAME = "NNCF_DEV"
```
andrey-churkin, line 18: It would be great to add a description for these environment variables.

nncf/telemetry_wrapper/telemetry.py
vshampor: Yes, because for the CI we might want not to enable certain features (e.g. debug dumps) that we would want for the development.

nncf/tensorflow/helpers/model_creation.py
```diff
@@ -94,13 +98,19 @@ def create_compressed_model(model: tf.keras.Model,
             original_model_accuracy = evaluation_args.eval_fn(model)
 
     builder = create_compression_algorithm_builder(config, should_init=not compression_state)
+
+    NNCFTelemetry.send_event(event_category=NNCF_TF_CATEGORY,
+                             event_action='compression_started',
```
ljaljushkin, line 103: Will it be duplicated by the same action on `apply` for experimental quantization?
https://github.com/openvinotoolkit/nncf/pull/1339/files#diff-37d0c21a327711cf378c2ddbbbbbb63bed3343b92fba312427a9e840bd42fc9eR113

ljaljushkin (approved):

nncf/definitions.py
vshampor: Done

nncf/definitions.py
```diff
@@ -14,3 +14,9 @@
 import os
 NNCF_PACKAGE_ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 HW_CONFIG_RELATIVE_DIR = "common/hardware/configs"
+
+# Environment variables below, if set, mark the execution environment
+# so that certain actions within NNCF proper, such as telemetry event collection or
+# debug dumps, are performed or not performed
+NNCF_CI_ENV_VAR_NAME = "NNCF_CI"  # Must be set in CI environments
+NNCF_DEV_ENV_VAR_NAME = "NNCF_DEV" # Must be set in environments of the NNCF dev team machines
```
AlexKoff88, line 22: AFAIK, the analytics itself should filter out intel IPs. But if we are going to go with these variables it would be better to handle this in the setup.py (like set it up in the Python env when `python setup.py develop` is used).

nncf/telemetry_wrapper/telemetry.py
```diff
@@ -0,0 +1,49 @@
+"""
+ Copyright (c) 2022 Intel Corporation
```
AlexKoff88, line 2: I don't like the name `_wrapper` in the name of the module. You can call it just `telemetry/telemetry.py` and define the module content in the `__init__.py` so that the import will look like `from nncf.telemetry import NNCFTelemetry`.

nncf/openvino/quantization/quantize.py
```diff
@@ -130,13 +127,46 @@ def quantize_impl(model: ov.Model,
         }
     ]
 
+    algo_names = sorted([x['name'] for x in algorithms])
```
AlexKoff88, line 130: Now half of the code of this method is about sending telemetry. I wonder if there is a more elegant way to do that, maybe with `with` operator.

nncf/experimental/quantization/compression_builder.py
```diff
@@ -78,6 +78,11 @@ def apply(self, model: TModel, dataset: Dataset) -> TModel:
         5) Collect all statistics.
         6) Apply algorithms.
         """
+        NNCFTelemetry.start_session(NNCF_ONNX_CATEGORY)
+        algo_names = sorted([x.__class__.__name__ for x in self.algorithms])
+        NNCFTelemetry.send_event(event_category=NNCF_ONNX_CATEGORY,
+                                 event_action='compression_started',
+                                 event_label=','.join(algo_names))
```
KodiaqQ, line 85: Do we also need to send the algorithm configurations used for quantization for further analysis?

nncf/telemetry_wrapper/telemetry.py
vshampor: We do not expect users to utilize this. Nevertheless, I refactored the package names so that the effect is the same.

nncf/openvino/quantization/quantize.py
vshampor: It can save space on session starting and ending code, but not on the intermediate event sending.

nncf/definitions.py
vshampor: These variables are for internal usage only. If someone installs NNCF in the editable mode (i.e. `python setup.py develop` or `pip install -e .`, this does not necessarily mean that this is an NNCF team member - that might be a user who just likes to install their packages in editable mode.

AlexKoff88 (approved):

MaximProshin (approved):

KodiaqQ (approved):

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Callable
+from typing import List, Union
+
+from nncf.telemetry.events import telemetry_category
+from nncf.telemetry.wrapper import NNCFTelemetry
+from nncf.telemetry.extractors import TelemetryExtractor
+from nncf.telemetry.extractors import VerbatimTelemetryExtractor
+
+
+class tracked_function:
```
alexsu52, line 24: Please align the class name with the style guide.

nncf/openvino/quantization/quantize.py
```diff
@@ -130,13 +129,26 @@ def quantize_impl(model: ov.Model,
         }
     ]
 
+    NNCFTelemetry.send_event(event_category=NNCF_OV_CATEGORY,
```
alexsu52, line 132: ?

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,70 @@
...
+    will in general result in a telemetry session being created and a set of events being sent before
+    function execution. The category of the session and events will be determined by parameters to the decorator.
+    """
+    def __init__(self, category: str = None, collectors: List[Union[str, TelemetryExtractor]] = None):
+        """
+        :param category: A category to be attributed to the events. If set to None, no events will be sent.
+        :param collectors: Add argument names in this list as string values to send an event with an "action" equal
+        to the argument name and "label" equal to the argument value before the function start. The argument must be
+        either a string or a dictionary of strings. If that is not the case, instead of argument names you can specify
+        an object of a customized `TelemetryExtractor` subclass; use the same approach for more complex event reporting.
```
alexsu52, line 36:
```suggestion
        :param collectors: Add argument names in this list as string values to send an event with an "action" equal
            to the argument name and "label" equal to the argument value before the function start. The argument must be
            either a string or a dictionary of strings. If that is not the case, instead of argument names you can specify
            an object of a customized `TelemetryExtractor` subclass; use the same approach for more complex event reporting.
```

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,70 @@
...
+            events = []  # type: CollectedEvent
+            for collector in self._collectors:
+                argname = collector.argname
+                argvalue = bound_args.arguments[argname] if argname is not None else None
+                event = collector.extract(argvalue)
+                events.append(event)
+
+            with telemetry_category(self._category) as category:
+                if category is not None:
+                    NNCFTelemetry.start_session(self._category)
```
alexsu52, line 56: What happens if we have nested function calls with the same category? Will each function start its own session?

alexsu52 (dismissed):
It looks better, but the body of some functions has been changed.

nncf/telemetry/decorator.py
vshampor: Does not apply to decorators.

nncf/experimental/quantization/compression_builder.py
```diff
@@ -78,6 +82,7 @@ def apply(self, model: TModel, dataset: Dataset) -> TModel:
         5) Collect all statistics.
         6) Apply algorithms.
         """
+        algo_names = sorted([x.__class__.__name__ for x in self.algorithms])
```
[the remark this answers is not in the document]
vshampor: Removed

nncf/common/composite_compression.py
```diff
@@ -322,7 +322,10 @@ def export_model(self,
             assert backend is BackendType.TORCH
             from nncf.torch.exporter import PTExporter #pylint: disable=cyclic-import
             exporter = PTExporter(self.model, input_names, output_names, model_args)
-        exporter.export_model(save_path, save_format)
+        if save_format is not None:
```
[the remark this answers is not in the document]
vshampor: The `export_model` functions were poorly designed so that passing `None` as `save_format` triggered usage of the default parameter. This means, among other things, that we cannot gather this parameter explicitly. We probably don't want to see `None` in the telemetry readings as the values for the "save_format" event. I then went ahead and set the argument of the `save_format` to specify correct default value, uncovering the places in code where `None` was passed to the function on purpose to sidestep the wrong design by more wrong design. I corrected the code that used the function along with the function signatures.

nncf/openvino/quantization/quantize.py
vshampor: Removed

nncf/telemetry/decorator.py
vshampor: Done

nncf/telemetry/decorator.py
vshampor: Will fix

vshampor: Fixed

kshpv (changes requested):
I don't see the telemetry added to nncf/onnx/quantization/quantize.py.
This will be the main executable function for quantization for ONNX backend. 
Also, I don't see telemetry in the same functions for torch and TF

nncf/telemetry/extractors.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Union
+
+SerializableData = Union[str, Dict]
+
+
+@dataclass
+class CollectedEvent:
+    name: str
+    data: SerializableData = None  # GA limitations
+    int_data: int = None
```
andrey-churkin, line 28: What should I put here?

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,74 @@
...
+            previous_category = get_current_category()
+            with telemetry_category(self._category) as category:
+                if category is not None:
+                    if category != previous_category:
+                        NNCFTelemetry.start_session(self._category)
+                    for event in events:
+                        NNCFTelemetry.send_event(event_category=category,
+                                                 event_action=event.name,
+                                                 event_label=event.data,
+                                                 event_value=event.int_data)
```
andrey-churkin, line 65: Do we really need to send several events? I mean we send one event per method's argument in this approach. I thought we should send only one event per method call.

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,74 @@
...
+    def __call__(self, fn: Callable) -> Callable:
+        fn_signature = inspect.signature(fn)
+
+        def wrapped(*args, **kwargs):
+            bound_args = fn_signature.bind(*args, **kwargs)
+            bound_args.apply_defaults()
+            events = []  # type: CollectedEvent
+            for collector in self._collectors:
+                argname = collector.argname
+                argvalue = bound_args.arguments[argname] if argname is not None else None
```
andrey-churkin, line 52: Could you please explain the case when `argname` is None?

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        :param category: A category to be attributed to the events. If set to None, no events will be sent.
+        :param collectors: Add argument names in this list as string values to send an event with an "action" equal to
+            the argument name and "label" equal to the argument value before the function start.
+            The argument must be either a string or a dictionary of strings. If that is not the case, instead of
+            argument names you can specify an object of a customized `TelemetryExtractor` subclass; use the same
+            approach for more complex event reporting.
+        """
+        self._category = category
+        self._collectors = [VerbatimTelemetryExtractor(x) if isinstance(x, str) else x for x in collectors]
```
andrey-churkin, line 41: It does not work when `collectors` is None. What should we do when `None` is passed to the `__init__()` method? Should we support this?

nncf/telemetry/events.py
```diff
@@ -0,0 +1,50 @@
...
+
+@contextmanager
+def telemetry_category(category: str) -> str:
+    previous_category = get_current_category()
+    _set_current_category(category)
+    yield category
+    _set_current_category(previous_category)
+
+
+def get_algo_names_from_builder(builder: CompressionAlgorithmBuilder) -> List[str]:
```
andrey-churkin, line 45: This method is not used.

andrey-churkin (changes requested):

nncf/telemetry/events.py
vshampor: Removed

nncf/telemetry/decorator.py
vshampor: We probably should if we want to just start and end the session without additional events. Fixed to work when `collectors is None`

nncf/telemetry/decorator.py
vshampor: For when you don't need to collect arguments, but need to send a specific event. See, for instance, `nncf.openvino.quantization.telemetry_extractors.CompressionStarted`

nncf/telemetry/decorator.py
vshampor: While it seems to be possible to send a dict of something instead of separate key:value events, I'm not sure that it will be easy to analyze in the telemetry collection base. We'll see if it has a serious enough impact so that it needs to be optimized.

andrey-churkin:
Some comments from my side:
- I didn't add comments regarding the code style at this stage. So the final review is required.
- If I want to send just the message (for example, `"The {method_name} was started"`) I should create subclasses of
    `TelemetryExtractor` for these methods. It looks strange to me.
- I'm not sure that an approach where we have a lot of `telemetry_extractors` modules in the project is good. Is it possible
    to put them in one place?

nncf/telemetry/extractors.py
vshampor: Added a docstring

nncf/torch/exporter.py
```diff
@@ -83,10 +88,11 @@ def parse_format(save_format: str) -> Tuple[str, dict]:
                     )
                 )
 
-            return PTExporter._ONNX_FORMAT, {'opset_version': opset}
+            return PTExportFormat.ONNX, {'opset_version': opset}
         return save_format, {}
 
-    def export_model(self, save_path: str, save_format: Optional[str] = None) -> None:
+    @tracked_function(NNCF_PT_CATEGORY, ["save_format"])
+    def export_model(self, save_path: str, save_format: Optional[str] = PTExportFormat.ONNX) -> None:
```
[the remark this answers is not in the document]
vshampor: Done

vshampor:
> I don't see the telemetry added to nncf/onnx/quantization/quantize.py. This will be the main executable function for quantization for ONNX backend. Also, I don't see telemetry in the same functions for torch and TF

The telemetry is added instead in the underlying functions - `create_compressed_model` and `CompressionBuilder.apply` - to the same effect.

andrey-churkin:
By the way, since we collect telemetry by wrapping methods, the information about which method was called is required (if not we should not decorate it=). I mean that we should send some event messages to identify them. What do you think about adding the `event_mesage` argument to the `tracked_function.__init__()` method? This argument has the default `None` value. If `event_mesage` is `None` the following default message is used: `The {method_name} was called`.

vshampor:
> * If I want to send just the message (for example, `"The {method_name} was started"`) I should create subclasses of
>   `TelemetryExtractor` for these methods. It looks strange to me.
This is not a logging or instrumentation task, so we don't need to wrap every function call. The method names are also subject to renaming, while fixed-name events are not. Telemetry events are rare enough and are worth the effort - it's harder to identify the good event than it is to write the extraction code.

> * I'm not sure that an approach where we have a lot of `telemetry_extractors` modules in the project is good. Is it possible
>   to put them in one place?

I plan for the number of these files to decrease, actually - the more common our API is, the less backend specializations will be needed. I separated these right now because the telemetry collection surface is currently at different levels in the different backends and potentially has to unwrap arguments of backend-specific nature, so putting all collectors in one place risks having circular imports at one point in time.

kshpv:
> > I don't see the telemetry added to nncf/onnx/quantization/quantize.py. This will be the main executable function for quantization for ONNX backend. Also, I don't see telemetry in the same functions for torch and TF
> 
> The telemetry is added instead in the underlying functions - `create_compressed_model` and `CompressionBuilder.apply` - to the same effect.

There is a possibility to do quantization without explicitly utilizing `CompressionBuilder`. This is how it used in the sample `examples/experimental/mobilenet_v2/main.py`. Also, the OpenVINO backend now doesn't use `CompressionBuilder` at all.
So, I suggest to add the telemetry call to each `quantization_impl()` or even to the main function `nncf.quantize()`

nncf/experimental/quantization/compression_builder.py
```diff
@@ -66,6 +67,7 @@ def _get_prepared_model_for_compression(self, model: TModel, backend: BackendTyp
 
         return None
 
+    @tracked_function(NNCF_ONNX_CATEGORY, [CompressionStartedFromBuilder(argname="self"), ])
```
andrey-churkin, line 70: We are going to remove the `CompressionBuilder` class in future PRs. So I think it is better to decorate the `quantize_impl()` methods directly.

vshampor: Added the decorator to the `quantize_impl` as well.

nncf/common/composite_compression.py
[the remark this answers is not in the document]
alexsu52: Thanks for the detailed answer. The changes make sense.

nncf/telemetry/decorator.py
alexsu52: need prove:)

nncf/openvino/quantization/quantize.py
```diff
@@ -105,16 +109,10 @@ def quantize_impl(model: ov.Model,
     """
     Implementation of the `quantize()` method for the OpenVINO backend.
     """
+
```
alexsu52, line 112: please, remove extra blank line

nncf/openvino/quantization/quantize.py
```diff
@@ -198,6 +202,8 @@ def quantize_with_accuracy_control_impl(model: ov.Model,
     compressed_model = pipeline.run(pot_model)
     pot.compress_model_weights(compressed_model)
 
+
```
alexsu52, line 205: please, remove extra blank line

nncf/openvino/quantization/quantize.py
```diff
@@ -198,6 +202,8 @@ def quantize_with_accuracy_control_impl(model: ov.Model,
...
     quantized_model = _convert_compressed_model_to_openvino_model(compressed_model)
 
+
```
alexsu52, line 208: please, remove extra blank line

nncf/quantization/telemetry_extractors.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any
```
alexsu52, line 1: Please add licenses.

nncf/telemetry/decorator.py
```diff
@@ -0,0 +1,77 @@
...
+from nncf.telemetry.extractors import VerbatimTelemetryExtractor
+
+
+class tracked_function:
+    """
+    Decorator to add telemetry events to a given function execution. The call to a decorated function
+    will in general result in a telemetry session being created and a set of events being sent before
+    function execution. The category of the session and events will be determined by parameters to the decorator.
+    """
+    def __init__(self, category: str = None, collectors: List[Union[str, TelemetryExtractor]] = None):
```
alexsu52, line 31: What category should be provided for multifunctional functions such as `quantize()`?

vshampor: I guess we can set up a category for that later (i.e. "NNCF_COMMON_CATEGORY"), once we identify at least one event in common code that we would like to collect. I think we are mostly interested in events attributed to specific backend right now.

nncf/telemetry/decorator.py
vshampor: Surprisingly there are no solid recommendations for that, but all built-in decorators (and external ones too that I've seen) use lower-case naming almost exclusively.

Some discussion on the topic can be found here:
https://stackoverflow.com/questions/10212556/naming-convention-for-descriptors

tests/onnx/test_telemetry.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+from tests.onnx.quantization.common import min_max_quantize_model
+from tests.onnx.models import LinearModel
+from tests.shared.helpers import telemetry_send_event_test_driver
+
+def test_telemetry_is_sent(mocker):
+    def use_nncf_fn():
+        model_to_test = LinearModel()
+        _ = min_max_quantize_model(model_to_test.input_shape[0], model_to_test.onnx_model)
```
kshpv, line 21: Can we add two cases for test: via Compression builder, which is already here and through nncf.quantize?

kshpv (changes requested):

kshpv (approved):

nncf/telemetry/extractors.py
```diff
@@ -0,0 +1,57 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+from abc import ABC
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import Any
+from typing import Optional
+
+SerializableData = str
```
andrey-churkin, line 19: The `target_device` parameter is `Enum`.  You pass value from this enum directly. Does it work?

nncf/quantization/quantize.py
```diff
@@ -26,6 +26,7 @@
 from nncf.parameters import TargetDevice
 
 
+
```
andrey-churkin, line 29:
```suggestion
```

vshampor: Done

nncf/telemetry/extractors.py
vshampor: Good catch, extended the default behaviour to automatically serialize `Enum`s by their `.value`.

nncf/quantization/telemetry_extractors.py
```diff
@@ -0,0 +1,23 @@
...
+
+from nncf.telemetry.extractors import CollectedEvent
+from nncf.telemetry.extractors import TelemetryExtractor
+
+
+class CompressionStartedWithQuantizeApi(TelemetryExtractor):
+    def extract(self, _: Any) -> CollectedEvent:
+        return CollectedEvent(name="compression_started",
+                              data="quantize_api")
```
andrey-churkin, line 23: I don't think that using `TelemetryExtractor` is the correct way to pass the event messages. This class should be used only to extract necessary information from an argument that has a difficult structure.


I have a question. Let's suppose we have 2 methods `func_1(`) and `func_2()` and these methods determine the different stages of our pipeline. These methods have a parameter which is called `some_param`. We decorate these methods and want to track this parameter. How do we understand where it comes from?

nncf/telemetry/decorator.py
andrey-churkin: I got it. Seems like some investigation is needed here to understand the usability of such approach.

nncf/telemetry/decorator.py
andrey-churkin: It is related to this https://github.com/openvinotoolkit/nncf/pull/1339#discussion_r1036942625

nncf/quantization/telemetry_extractors.py
vshampor: You guessed it - you would have to set up custom extractors in both cases and differentiate the events by the event name. The event name is thus in general decoupled from both the function name and the argument name. I think that this decoupling is desirable because we can do refactoring on function names and argument names without having to submit conversion rules to whoever analyzes the telemetry. 

We are not tracking general parameter usage across functions, we are tracking specific events that give us valuable, analyzeable information. The alternative, as always, is to use direct `NNCFTelemetry.send_event` calls. If there is another idea that will allow us to flexibly collect events and their serializable data from function argument, please let me know.

andrey-churkin: > The event name is thus in general decoupled from both the function name and the argument name. I think that this decoupling is desirable because we can do refactoring on function names and argument names without having to submit conversion rules to whoever analyzes the telemetry.

Yes, I also think so.

I have the following thoughts:

- Specify all events that we are interested in.

```python
class EventId(Enum):
     PTQ_API = 'PTQ_API'  # quatize(), quatize_impl()
     QAT_API = 'QAT_API '  # create_compressed_model()
     MODEL_EXPORT = 'MODEL_EXPORT'  # export_model()
     # Something else...
```

- Add `event_id` parameter to the `tracked_function` decorator.

```python
@tracked_function(
    NNCF_ONNX_CATEGORY,
    event_id=EventId.PTQ_API,
    extractors=['target_device', 'preset']
)
def quantize_impl(...):
    ...
```

- We send the following events:
    - `CollectedEvent(name=event_id.value)` - to identify the stage of the pipeline.
    - `CollectedEvent(name=f{event_id.value}:{argname})` - to track argument's value

vshampor:
run pytorch pre-commit tests

nncf/telemetry/decorator.py
alexsu52: Honestly, I kept in my mind something like that:
```
tracked_function = _TrackedFunction
``

PR #1342, opened 2022-10-27
CVS-94288

nikita-savelyevv:


nikita-savelyevv:
https://github.com/openvinotoolkit/nncf/pull/1343

PR #1349, opened 2022-11-02
openvino extra to setup.py

alexsu52:
### Changes
```
python setup.py install --openvino
```
### Reason for changes

adding openvino backend

### Related tickets
N/A

### Tests

[test_install.py](https://github.com/openvinotoolkit/nncf/compare/develop...alexsu52:as/dependencies?expand=1#diff-53910d318d1940f81165dcb97a5800355f0c0e7ed56b73f247be55171a9dc3f0)

### DIFF
--- a/nncf/__init__.py
+++ b/nncf/__init__.py
@@ -30,6 +30,18 @@ try:
 except ImportError:
     tf = None
 
-if torch is None and tf is None:
+try:
+    import onnx
+except ImportError:
+    onnx = None
+
+try:
+    import openvino.runtime as ov_runtime
+    import openvino.tools.pot as ov_pot
+except ImportError:
+    ov_runtime = None
+    ov_pot = None
+
+if torch is None and tf is None and onnx is None and ov_runtime is None and ov_pot is None:
     import warnings
-    warnings.warn("None of PyTorch or TensorFlow have been found. Please, install one of the frameworks")
+    warnings.warn("None of PyTorch, TensorFlow, ONNX, OpenVINO have been found. Please, install one of the frameworks")
--- a/setup.py
+++ b/setup.py
@@ -131,12 +131,17 @@ EXTRAS_REQUIRE = {
         "protobuf==3.20.1",
         "onnxruntime-openvino==1.11.0",
     ],
+    "openvino": [
+        "openvino-dev"        
+    ]
 }
 
 EXTRAS_REQUIRE["all"] = [
     EXTRAS_REQUIRE["tf"],
     EXTRAS_REQUIRE["torch"]
 ]
+
+SETUP_REQUIRES = []
 
 if "--torch" in sys.argv:
     INSTALL_REQUIRES.extend(EXTRAS_REQUIRE["torch"])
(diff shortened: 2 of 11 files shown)


setup.py
```diff
@@ -131,6 +131,10 @@ def find_version(*file_paths):
         "protobuf==3.20.1",
         "onnxruntime-openvino==1.11.0",
     ],
+    "openvino": [
+        "openvino",
```
andrey-churkin, line 135: > While installing OpenVINO Development Tools, OpenVINO Runtime will also be installed as a dependency, so you don’t need to install OpenVINO Runtime separately.
https://docs.openvino.ai/latest/openvino_docs_install_guides_install_dev_tools.html

tests/openvino/install_checks.py
```diff
@@ -0,0 +1,51 @@
...
+          raise RuntimeError(f'Could not import {name}')
+
+
+for loader, module_name, is_pkg in pkgutil.walk_packages(nncf.__path__,
+                                                         nncf.__name__ + '.',
+                                                         onerror):
+     if module_name in sys.modules or excluded_module(module_name):
+          continue
+     if not is_pkg:
+          loader.find_module(module_name).load_module(module_name)
```
vshampor, line 51: Just what does this do?

alexsu52: Module loading. More information you can find here https://docs.python.org/3/library/pkgutil.html

tests/openvino/test_install.py
```diff
@@ -0,0 +1,33 @@
...
+
+@pytest.fixture(name="package_type",
+                params=["install", "develop", "sdist", "bdist_wheel",
+                        "pip_pypi", "pip_local", "pip_e_local", "pip_git_develop"])
+def package_type_(request):
+    return request.param
+
+
+def test_install(tmp_venv_with_nncf, tmp_path, package_type):
+    run_install_checks(tmp_venv_with_nncf, tmp_path, package_type, test_dir='openvino')
```
AlexKoff88, line 33: please check this function since it has some PT and TF-specific logic that should be extended as well.

tests/openvino/install_checks.py
vshampor: Whatever it is, an explanation should be given in the comments because the code is obscure.

How is loading every module and submodule justified in this test in comparison to doing quick, smoke-like imports as in the TF and PT tests?

nncf/__init__.py
```diff
@@ -30,6 +30,16 @@
...
     tf = None
 
-if torch is None and tf is None:
+try:
+    import onnx
+except ImportError:
+    onnx = None
+
+try:
+    import openvino
```
l-bat, line 39: To ensure that openvino is properly installed, we should check `import openvino.runtime` and `import openvino.tools.pot` will not raise `ImportError`

tests/openvino/install_checks.py
alexsu52: Sorry, but I don't see something light like smoke there https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/install_checks.py

This looks easy:
```
import nncf
print(nncf.__version__)
```

tests/openvino/test_install.py
alexsu52: Thanks! Done!

andrey-churkin (approved):

l-bat (approved):

AlexKoff88 (approved):

andrey-churkin:
@vshampor Can you please look at this?

vshampor (approved):

PR #1323, opened 2022-10-17
Use torch 1.12

vshampor:
### Changes

Bumped torch and torchvision version and adapted NNCF sources and tests for them.
Better to review by commits

### Reason for changes

For HuggingFace Optimum integration, we need to have NNCF working with the latest PyTorch 1.12.1 version

### Related tickets

92561

### Tests

- [x] precommit
- [x] third-party sanity
- [x] pytorch sota eval 
- [ ] e2e ?

PR #1352, opened 2022-11-04
[Draft] Fixed per-channel quantized models into Q/DQ

AlexKoff88:
- Check of PyTorch version should be added


ljaljushkin:
After merge https://github.com/openvinotoolkit/nncf/pull/1323 there's no need in checking torch version, because currently we support 1.12 only.

ljaljushkin:
The related fixes were made a in separate PR: https://github.com/openvinotoolkit/nncf/pull/1384 and https://github.com/openvinotoolkit/nncf/pull/1386

PR #1355, opened 2022-11-07
Actualized some reference .dot files

ljaljushkin:
### Changes

Updated .dot files which are used as reference graphs in tests

### Reason for changes

Real changes in the graph are hard to review when files are not actual

### Related tickets

#1323 

### Tests

- test_compressed_graph
- test_autoq_precision_init
- test_hawq_precision_init 
- test_get_ip_graph_with_merged_operations


ljaljushkin:
> If the reference files are wrong, why are the tests passing?
References were valid before, but there just were in a obsolete format.

`test_get_ip_graph_with_merged_operations` just compares node names from loaded graph.
```python
write_dot_graph(merged_ip_graph, str(path_to_dot_file))
load_graph = read_dot_graph(str(path_to_dot_file))
sanitized_loaded_keys = [key.replace('\\n', '\n') for key in load_graph.nodes.keys()]
```
For this test meta data of node is not important.

Tests for the graphs compares the loaded graph. So the difference in node and edge order is not important.

vshampor:
Maybe we could set up a special function, or a conditional path determined by a boolean argument in the functions, that would only dump the nodes and edges w/o metadata, and rewrite these impacted tests to use these functions instead? No sense in keeping data in reference files which is not checked for anyway.

nncf/common/utils/dot_file_rw.py
```diff
@@ -10,5 +10,19 @@ def write_dot_graph(G: nx.DiGraph, path: pathlib.Path):
     nx.nx_pydot.write_dot(G, str(path))
 
 
+def get_graph_without_data(G: nx.DiGraph) -> nx.DiGraph:
```
ljaljushkin, line 13: @vshampor please take a look

vshampor (approved):

PR #1356, opened 2022-11-07
Restricted ninja version to be not higher than 1.11

ljaljushkin:
Should fix the following error on test_force_cuda_build[venv-develop-GPU]:
`RuntimeError: Ninja is required to load C++ extensions`

### Changes

Restricted ninja version to be not higher than 1.11

### Reason for changes

`RuntimeError: Ninja is required to load C++ extensions`

### Related tickets

n/a

### Tests

test_force_cuda_build

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ def find_version(*file_paths):
     return version_value
 
 
-INSTALL_REQUIRES = ["ninja>=1.10.0.post2",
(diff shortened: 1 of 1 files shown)


ljaljushkin:
all pytorch job started to fail with latest release of ninja==1.11. 
for some reason it's not visible for building torch extension. ninja==1.10 works fine

PR #1358, opened 2022-11-08
Remove multiprocessing_distributed from the configs

vshampor:
### Changes
As stated in the title.

### Reason for changes
This config construction conflicts with setting a `--gpu-id N` option via CLI, and there's no way to "disable" the multiprocessing option if it is already specified in the sample config.

### Related tickets
N/A

### Tests
TBR


vshampor:
SOTA eval build 94 shows no eval time degradations (build 88 as reference)

PR #1362, opened 2022-11-10
Update building NNCFGraph for ONNX

kshpv:
### Changes

Make building NNFGraph through traversing of the node of ONNX model.

### Reason for changes

New logic eliminates all Constant subgraphs in NNCFGraph. This obviates of adding activation Quantizers/Dequantiers node in Constant subgraphs.

### Related tickets

96177

### Tests

All graph tests were updated


nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -182,14 +184,27 @@ def create_nncf_graph(onnx_model: ModelProto) -> NNCFGraph:
...
+
+        while nodes_queue:
+            current_node = nodes_queue.popleft()
+            if current_node in visited_nodes:
+                continue
+            nodes_queue.extend(onnx_graph.get_children(current_node))
+            if GraphConverter._is_valid_onnx_metatype(current_node):
+                metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(current_node.op_type)
+                layer_attributes = ONNXExtendedLayerAttributes(current_node.input, current_node.output)
+                nncf_graph.add_nncf_node(node_name=current_node.name,
```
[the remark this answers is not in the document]
kshpv: @vshampor I added `is_shared` and `layer_name` attribtes to the NNCFNodes. After that I expected to get only one weight quantizer for the sharable weight, but I still get an one weight quantizer per sharable weight node. Did I do something wrong?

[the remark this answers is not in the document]
kshpv: After offline discussion with @vshampor we figured out that QPS returns quantizer per node even it they share a weight. So, backend-specific part should filter shared quantizers if there are exist.

kshpv:
run build

kshpv:
run build

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -12,6 +12,8 @@
 """
 from typing import Union, List
 
+from collections import deque
```
KodiaqQ, line 15: Not a necessary import.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -179,13 +179,40 @@ def _get_stat_collector(self, quantizer_config: QuantizerConfig) -> TensorStatis
                                                                         num_samples=self._parameters.number_samples)
         raise RuntimeError('This range type is not supported!')
 
+    def _filter_nncf_graph(self, nncf_graph: NNCFGraph) -> NNCFGraph:
```
KodiaqQ, line 182: Why should this method be a part of the `MinMaxQuantization` algorithm?

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -18,9 +18,12 @@
 import numpy as np
 
 from nncf.experimental.onnx.model_normalizer import ONNXModelNormalizer
+from nncf.experimental.onnx.graph.metatypes.onnx_metatypes import ONNX_OPERATION_METATYPES
+from nncf.experimental.onnx.graph.metatypes.onnx_metatypes import WEIGHT_LAYER_METATYPES
 
 
 # pylint: disable=no-member
```
KodiaqQ, line 25: Just add here `too-many-public-methods`

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -191,17 +194,20 @@ def get_nodes_by_type(self, node_type: str) -> List[onnx.NodeProto]:
...
         """
-        node_inputs = self.get_node_edge_names(node_name)['input']
-        weight_tensor = node_inputs[1]
-        return weight_tensor
+        metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node.op_type)
+        if metatype in WEIGHT_LAYER_METATYPES:
+            node_inputs = self.get_node_edge_names(node.name)['input']
+            weight_tensor = node_inputs[1]
+            return weight_tensor
+        return None
```
KodiaqQ, line 210: In this case, we should check that `None` wouldn't be called further.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -179,13 +179,40 @@ def _get_stat_collector(self, quantizer_config: QuantizerConfig) -> TensorStatis
...
+        """
+        Removes all Constant nodes from NNCFGraph, making it inference graph.
```
l-bat, line 184: What is the problem with building the inference graph in `NNCFGraphFactory.create(model)`? It looks like method removes Constant nodes, but not Constant subgraphs.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -301,8 +328,7 @@ def _apply(self,
             node = nncf_graph.get_node_by_name(target_node_name)
             if quantization_target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
                 try:
-                    input_tensor_names, _ = self._backend_entity.get_tensor_names(node)
-                    weight_initializer_name = input_tensor_names[1]
+                    weight_initializer_name = self._backend_entity.get_weight_tensor_name(model, node.node_name)
                     weight_tensor = self._backend_entity.get_initializer_value(model, weight_initializer_name)
```
l-bat, line 332: To make the logic common to all backends, it would be better to use `port_id` instead of `weight_initializer_name`.

nncf/quantization/algorithms/min_max/algorithm.py
kshpv: I didn't find better place to put this method. If you have any suggestions - welcome!
So `MinMaxQuantization` should use filtered `NNCFGraph` without Constant subgraphs to avoid incorrect q/dq placements. At the same time "full" `NNCFGraph` could be useful for other algorithms. Also, it has more nice visualization with Constant subgraphs, e.g. weight quantizers.

nncf/quantization/algorithms/min_max/algorithm.py
kshpv: `MinMaxQuantization` should use filtered `NNCFGraph` without Constant subgraphs to avoid incorrect q/dq placements. At the same time "full" `NNCFGraph` could be useful for other algorithms. Also, it has more nice visualization with Constant subgraphs, e.g. weight quantizers.

nncf/quantization/algorithms/min_max/algorithm.py
kshpv: We need edge name to filter some quantizers which put in the shared weight. In that case they have common edge name but the different node names. The only way to do filtration is to keep edge names. 
Do you have other idea?

kshpv: Yes, it should be. Good suggestion!

nncf/quantization/algorithms/min_max/algorithm.py
andrey-churkin: @kshpv Could you please look at my comment above?

nncf/quantization/algorithms/min_max/algorithm.py
KodiaqQ: @kshpv, I disagree. `MinMaxQuantization` is the algorithm and should do nothing except that needs to execute its own pipeline. `Filtering NNCFGraph` is not part of any algorithm. It should be placed into one of the model-transformer entities or as a pass of the model preparation for the further quantization pipeline.

KodiaqQ: @kshpv, additionally, if there are any `incorrect q/dq placements` that is the problem of the algorithm that creates the so-called `quantizer_setup`. As I remember, it's the main purpose of the `QuantizerPropagationSolver`, isn't it?

kshpv: @KodiaqQ, yes. I agree with that point. Model-transformer works only on backend-specific model, so it seems there is no good way to put this logic there.

KodiaqQ: @kshpv, it doesn't matter. The quantization algorithm should not care about the model preparations. Since there are not possible to place this functionality into the `ModelTransformer` or somewhere else, it means that `NNCFGraph` builds wrong again.

kshpv: @KodiaqQ there is a good point in general. 
In the signature of our algorithms we pass Model, while, e.g. MinMax doe not utilize it at all, only to build NNCFGraph. Do we need the model in FBC?
I mean maybe we just need to pass NNCFGraph for all algorithms. Is it possible?

KodiaqQ: @kshpv it would be a great next step to use the same general entity like `NNCFGraph` in the algorithms, but currently, it's not possible because `NNCFGraph` did not contain constants and some kind of necessary information. @alexsu52, could you please add details about the usage of the general model representation in the algorithms?

kshpv: @vshampor @alexsu52

KodiaqQ: To my point, some kind of `filtration` logic correlates with the `ignoring pattern` functionality. And it should be implemented as the additional logic near it or as one of the steps for the `NNCFGraph` building.

alexsu52: Hi folks, 

Obviously, we have some dual understanding of NNCFGraph. We have to align here and plan improvement steps. I would suggest to schedule offline discussion.

In the context of this PR, the _filter_nncf_grap is common pass of NNCFGraph preparation for `QuantizationPropagationSolver` and I would recomend to call it in the `QuantizationPropagationSolver` implementation.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -301,9 +328,8 @@ def _apply(self,
...
             if quantization_target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
                 try:
-                    input_tensor_names, _ = self._backend_entity.get_tensor_names(node)
-                    weight_initializer_name = input_tensor_names[1]
-                    weight_tensor = self._backend_entity.get_initializer_value(model, weight_initializer_name)
+                    weight_initializer_name, weight_tensor = self._backend_entity.get_weight_tensor(model,
+                                                                                                    node.node_name)
                     # If the nodes share one weight tensor, we should have only one quantizer on that
                     if weight_initializer_name in weight_initializer_names:
                         continue
```
l-bat, line 335: Is it possible to use `ONNXGraph.is_node_shared(onnx_node)` method instead of `weight_initializer_names` to create a `quantizer_insertion_command` once for shared nodes?

nncf/common/graph/graph.py
```diff
@@ -673,3 +673,18 @@ def get_all_edges(self) -> Generator[NNCFGraphEdge, None, None]:
         for nx_edge in self._nx_graph.edges:
             yield self.get_edge(self.get_node_by_key(nx_edge[0]),
                                 self.get_node_by_key(nx_edge[1]))
+
+    def remove_nodes(self, nodes: List[NNCFNode]) -> None:
```
alexsu52, line 677: The graph topology of NNCFGraph is immutable. Could you shortly explain why do you need this method.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -48,7 +51,8 @@ def _is_valid_onnx_metatype(node: NodeProto) -> bool:
         node_type = node.op_type
         metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node_type)
         if metatype == ONNXConstantMetatype:  # We don't need to quantize Constants
-            nncf_logger.debug('The metatype is ONNXConstantMetatype, which in not quantizable. Skipping this node.')
+            nncf_logger.debug('The metatype is ONNXConstantMetatype, which means that the node is Constant.'
+                              'All constant nodes are not added to NNCFGraph. The node is skipped.')
```
alexsu52, line 55: Why do you need this debug message?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -131,7 +130,7 @@ def _apply(self,
             if not self._is_node_with_bias(node):
                 nncf_logger.debug('Skipping node {} because there is no bias'.format(node_name))
                 continue
-            if not self._is_quantized_weights(node, nncf_graph):
```
alexsu52, line 134: It looks like you can revert this changes.

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/quantization/algorithms/min_max/algorithm.py
kshpv: Done

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Removed returning None

nncf/common/graph/graph.py
kshpv: This was redundant code. Removed this method

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: For debugging purpose. I feel it could be helpful. What do you think?

kshpv:
run onnx pre-commit tests

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -186,10 +187,16 @@ def create_nncf_graph(onnx_model: ModelProto) -> NNCFGraph:
         for node in filter(GraphConverter._is_valid_onnx_metatype, onnx_graph.get_all_nodes()):
             metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node.op_type)
             layer_attributes = ONNXExtendedLayerAttributes(node.input, node.output)
+            is_shared, layer_name = None, None
+            if metatype in WEIGHT_LAYER_METATYPES:
+                is_shared = onnx_graph.is_node_shared(node)
```
l-bat, line 192: Does this mean that we can use `is_shared` instead of `weight_initializer_names = set()` in MinMax algorithm?

kshpv: What's the purpose of checking that node is shared? I mean if I know that node is shared I also need to know whether the corresponding quantizer was added.
Or you have a suggestion?

kshpv:
> It would be nice if you update code below
> 
> https://github.com/openvinotoolkit/nncf/blob/3470a2b4335cfaa42d068ff566a79b883051d58c/nncf/quantization/algorithms/min_max/utils.py#L83-L86

Done. I hope it will not affect on models metrics. I will check it with E2E

kshpv:
run onnx pre-commit tests

nncf/experimental/onnx/graph/nncf_graph_builder.py
l-bat: I suggest to separate `non_unified_scales_quantization_points` and ` unified_scales_quantization_points` like in TF part
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/tensorflow/quantization/algorithm.py#L351-L385. 

We should probably do it in another PR.

nncf/common/insertion_point_graph.py
```diff
@@ -243,6 +245,95 @@ def _get_default_post_hook_ip_list(nncf_graph: NNCFGraph) -> List[PostHookInsert
...
+    def get_next_nodes(self, node_key: str) -> List[str]:
+        """
+        Returns consumer nodes of provided node with key 'node_key'.
+
+        :param node_key: Producer node key.
+        :return: List of consumer nodes of provided node.
+        """
+        return self.succ[node_key]
+
+    def traverse_graph(self,
```
KodiaqQ, line 313: Why did you define this method and not reuse it from the `NNCFGraph` that does the same?

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +405,39 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
...
+            # Skip for tests where there is no input node.
+            return ip_graph
+        weight_nodes = []
+        if quantizable_layer_node_keys is not None:
+            weight_nodes = [ip_graph.get_merged_node_from_single_node_key(weight_node) for weight_node in
+                            quantizable_layer_node_keys]
+        visited_nodes = []
+        partial_traverse_function = partial(traverse_function, visited_nodes=visited_nodes)
+        traversed_node_keys = []
+        for star_node in chain(input_nodes, weight_nodes):
```
KodiaqQ, line 438: `star_node` or `start_node`?

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +405,39 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
...
+        def traverse_function(curr_node, output, visited_nodes):
+            if curr_node in visited_nodes:
+                return True, output
+            output.append(curr_node)
+            visited_nodes.append(curr_node)
+            return False, output
+
+        input_nodes = ip_graph.get_input_nodes()
+        if not input_nodes:
+            # Skip for tests where there is no input node.
```
KodiaqQ, line 429: Why we should add the functionality in the common part of the framework that uses only in tests?

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -29,25 +28,51 @@ def get_all_aliases(cls) -> List[str]:
...
+    weight_definition = None  # type: OpWeightDef
+
+
 @ONNX_OPERATION_METATYPES.register()
-class ONNXConvolutionMetatype(ONNXOpMetatype):
+class ONNXConvolutionMetatype(ONNXOpWithWeightsMetatype):
     name = 'ConvOp'
     op_names = ['Conv']
     hw_config_names = [HWConfigOpName.CONVOLUTION]
+    weight_definitions = OpWeightDef(weight_port_id=1, channel_axis=0, bias_port_id=2)
```
KodiaqQ, line 58: Are you sure that the channel axis in the Convolution is 0? Is it the same `channel_axis` that uses in the `FastBiasCorrection`?

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +205,83 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
-                tensor_value = self.get_initializers_value(node.input[0])
-                reshaped_tensor_value = tensor_value.reshape(shape)
-                return tensor_name, reshaped_tensor_value
-            raise RuntimeError('Could not find the weight value of the node') from e
+        except RuntimeError as exc:
+            weight_port_id = self.get_weight_tensor_port_id(node)
+            parent_node_on_weight_port = self.get_parents(node)[weight_port_id]
+            nodes = deque([parent_node_on_weight_port])
+            while nodes:
+                try:
```
KodiaqQ, line 217: Two levels of `try-except`? Really?

nncf/quantization/algorithms/min_max/backend.py
```diff
@@ -148,13 +148,23 @@ def get_weight_tensor(model: TModel, node: NNCFNode) -> Tuple[str, np.ndarray]:
         """
 
     @staticmethod
-    def get_weight_tensor_port_id(model: TModel, node: NNCFNode) -> Optional[int]:
+    def get_weight_tensor_port_id(model: TModel, node: NNCFNode) -> int:
```
KodiaqQ, line 151: `get_weight_tensor_port_id` here and now `get_weight_tensor_port_id` in the `ONNXGraph`?

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +205,83 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+    def get_bias_tensor_port_id(self, node: onnx.NodeProto) -> int:
+        """
+        Returns input port id, where a bias tensor should output.
+
+        :param node: Node, for which input port id is returned,
+        :return: input port id, where a weight bias should output.
+        """
+        return self._get_param_from_weight_definitions(node, 'bias_port_id')
+
+    def _get_weight_tensor_with_reshape(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
```
KodiaqQ, line 275: No description.

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +205,83 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+        tensor_name = node.output[0]
+        shape = self.get_initializers_value(node.input[1])
+        tensor_value = self.get_initializers_value(node.input[0])
+        reshaped_tensor_value = tensor_value.reshape(shape)
+        return tensor_name, reshaped_tensor_value
+
+    def _get_tensor_from_zero_input(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
```
KodiaqQ, line 282: The same. Need to add a description.

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -103,18 +103,23 @@ def mean_minmax_statistic_collector(use_per_sample_stats: bool,
...
-        weight_tensor_name, _ = onnx_graph.get_weight_tensor(node)
-        for i, input_name in enumerate(node.input):
-            if input_name == weight_tensor_name:
-                return i
-        return None
+        onnx_node = onnx_graph.get_node_by_name(node.node_name)
+        return onnx_graph.get_weight_tensor_port_id(onnx_node)
+
+    @staticmethod
+    def get_weight_tensor_quantization_axis(model: onnx.ModelProto, node: NNCFNode,
```
KodiaqQ, line 116: There is no such method in the main backend entity.

tests/onnx/quantization/test_classification_models_graph.py
```diff
@@ -24,38 +24,35 @@
...
+                          (ModelToTest('squeezenet1_0', [1, 3, 224, 224]), models.squeezenet1_0(pretrained=True)),
+                          (ModelToTest('densenet121', [1, 3, 224, 224]), models.densenet121(pretrained=True)),
+                          (ModelToTest('mnasnet0_5', [1, 3, 224, 224]), models.mnasnet0_5(pretrained=True)),
                           ]
                          )
 def test_min_max_quantization_graph(tmp_path, model_to_test, model):
-    if model_to_test.model_name in ['inception_v3', 'googlenet']:
-        # The model skipping will be remove when correct NNCFGraph building will be merged
-        pytest.skip('Ticket 96177')
+    if model_to_test.model_name == 'mobilenet_v3_small':
```
KodiaqQ, line 46: Why should we skip this model?

tests/onnx/quantization/test_graphs.py
```diff
@@ -12,22 +12,18 @@
...
 from tests.onnx.models import MultiInputOutputModel
-from tests.onnx.models import ReshapeWeightModel
 
 
 @pytest.mark.parametrize('model_cls_to_test', ALL_SYNTHETIC_MODELS.values())
 def test_syntetic_models_graph(model_cls_to_test):
     if model_cls_to_test == MultiInputOutputModel:
         pytest.skip('min_max_quantize_model does not support many inputs for now.')
-    if model_cls_to_test == ReshapeWeightModel:
-        pytest.skip('There is incorrect quantization of the weight tensor. Should be fixed in #1362 PR.')
```
KodiaqQ, line 30: Why was this model removed? In the previous PR, there was a message to fix this.

nncf/common/insertion_point_graph.py
vshampor: Remove the comment, add `nncf_logger.debug()` message about it. Tests or not, we will have to handle such a case anyway unless we guarantee that the input graph object always has input nodes. Right now, there is no such guarantee.

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -29,25 +28,51 @@ def get_all_aliases(cls) -> List[str]:
         return cls.op_names
 
 
+class OpWeightDef:
```
vshampor, line 31: Use `@dataclass`

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -29,25 +28,51 @@ def get_all_aliases(cls) -> List[str]:
...
+    """
+    Contains information about the weight of operation.
+    """
+
+    def __init__(self, weight_port_id: Optional[int], channel_axis: int, bias_port_id: Optional[int]):
```
vshampor, line 36: Some arguments have `Optional`, but don't have a default parameter value.  Also the docstring does not say what happens when each optional parameter is not given.

nncf/experimental/onnx/graph/onnx_graph.py
vshampor: I tend to agree that the code is convoluted and that we can miss real RuntimeErrors in here. It was OK when the try-except block was limited to a single function, but no longer. Refactor the `get_` function to not throw exceptions as part of their normal functionality, and set up `def has_*(name) -> bool:` functions instead for each function that used to throw exceptions. Call the `has_*` function in an if-block before calling the `get_` functions and do early-exit if `has_` check failed.

Also, this much logic has to be unit-tested.

nncf/experimental/onnx/graph/onnx_graph.py
vshampor: Although I need to remind you that the docstring is not strictly necessary for private functions, in this case I think it might help with understanding this and the other functions' purpose.

nncf/quantization/algorithms/min_max/utils.py
```diff
@@ -44,41 +44,50 @@ def calculate_scale_level(max_val: Union[float, np.ndarray], min_val: Union[floa
...
+    return QuantizerLayerParameters(scales.tolist(), zero_points.tolist(), mode, axis)
 
 
 def calculate_activation_quantizer_parameters(statistics: MinMaxTensorStatistic,
-                                              quantizer_config: QuantizerConfig) -> QuantizerLayerParameters:
+                                              quantizer_config: QuantizerConfig,
+                                              axis: Optional[int]) -> QuantizerLayerParameters:
     """
     Calculates Quantizer/Dequantizer layer attributes for activation quantizer such as scale, zero_points and
     quantization mode: symmetric, asymmetric.
```
vshampor, line 81: Needs a description for each parameter in the docstring

tests/common/quantization/test_ignore_post_processing.py
```diff
@@ -271,7 +271,7 @@ def test_node_locator_finds_postprocessing_nodes(model_to_test):
     quantizable_layer_nodes = [QuantizableWeightedLayerNode(weight_node, [QuantizerConfig()]) for weight_node in
                                weight_nodes]
 
-    quant_prop_graph = QuantizerPropagationStateGraph(ip_graph)
+    quant_prop_graph = QuantizerPropagationStateGraph(ip_graph, quantizable_layer_nodes)
```
vshampor, line 274: Looks like an artifact from the previous attempt

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +205,83 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+                    if weight_port_id is None:
+                        # The node does not have weight
+                        continue
+                    weight_input = self.get_node_edge_names(current_node.name)['input'][weight_port_id]
+                    return self.get_initializer(weight_input).name, self.get_initializers_value(weight_input)
+                except RuntimeError:
+                    continue
+            raise RuntimeError('Could not find the weight value of the node') from exc
+
+    def _get_param_from_weight_definitions(self, node: onnx.NodeProto, parameter: str) -> int:
```
vshampor, line 236: Not good - you are basically wrapping around the `getattr` usage and using `getattr` is generally wrong save for special cases. If you really need to get something from the `OpWeightDef`, and the `OpWeightDef` may not be there, why not `_get_weight_definitions` and then just access the `OpWeightDef` statically defined members via `.` access?

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +405,39 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
     @staticmethod
     def get_post_hook_node_key(node_key: str) -> str:
         return InsertionPointGraph.POST_HOOK_ID_PREFIX + node_key
+
+
+class ConstantNodesFilter:
```
vshampor, line 410: Add tests for this functionality.

vshampor (changes requested):

nncf/common/insertion_point_graph.py
kshpv: Done

nncf/common/insertion_point_graph.py
kshpv: Guys, I can remove the comment and add debug message. 
But what should I do with the models without NNCF_INPUT_NODE? 
I start graph traversing from them and mark all visited node if any node wasn't marked I remove it from IP graph.
How can I adjust this logic for case when there is no NNCF_INPUT_NODE?

tests/onnx/quantization/test_graphs.py
kshpv: This model is not skipped. Skipping is removed for this model

nncf/common/insertion_point_graph.py
vshampor: Keep the logic, test for it.

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
KodiaqQ: It is not clear from the name of the variable and not clear even from the parameter description.

kshpv: Done

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: Done

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Done

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Done

nncf/quantization/algorithms/min_max/backend.py
kshpv: Done. Changed `get_weight_tensor_port_id `in `ONNXGraph` to `get_weight_port_id `

nncf/common/insertion_point_graph.py
kshpv: Done

nncf/common/insertion_point_graph.py
kshpv: Done

nncf/common/insertion_point_graph.py
kshpv: Done

tests/common/quantization/test_ignore_post_processing.py
kshpv: Done

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: This func was added to remove copy/paste code for the getting attributes of `OpWeightDef`. There are three of them:
```
get_weight_port_id
get_weight_channel_axis
get_bias_tensor_port_id
```

nncf/quantization/algorithms/min_max/utils.py
kshpv: Done

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
kshpv: Done

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
kshpv: Done

nncf/common/insertion_point_graph.py
```diff
@@ -243,6 +245,95 @@ def _get_default_post_hook_ip_list(nncf_graph: NNCFGraph) -> List[PostHookInsert
             allowed_post_hook_insertion_points.append(PostHookInsertionPoint(nncf_node.node_name))
         return allowed_post_hook_insertion_points
 
+    def get_input_nodes(self) -> List[str]:
+        """
+        Returns all Input Nodes, nodes having any of INPUT_NOOP_METATYPES.
+
+        :return: All Input Nodes.
```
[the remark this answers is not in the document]
kshpv: Done

nncf/common/insertion_point_graph.py
```diff
@@ -243,6 +245,95 @@ def _get_default_post_hook_ip_list(nncf_graph: NNCFGraph) -> List[PostHookInsert
...
+        return output
+
+    def get_merged_node_from_single_node_key(self, node_key: str) -> str:
+        """
+        Returns the merged node key if the corresponding node with 'node_key' was merged.
+        Otherwise retuns the 'node_key'.
+
+        :param node_key: The key of the node which is checking on the merged.
+        :return: The merged node key. 'node_key' if it was not merged.
+        """
```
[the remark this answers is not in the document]
kshpv: Updated. Please, take a look

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -96,8 +96,8 @@ def to_json(self) -> Dict[str, Union[str, float, int]]:
         Serialize all MinMaxQuantization parameters to JSON.
         """
 
-    def _determine_weight_activation_modes(self,
-                                           preset: QuantizationPreset) -> Tuple[QuantizationMode, QuantizationMode]:
+    def _determine_weight_activation_modes(self, preset: QuantizationPreset) -> Tuple[
+        QuantizationMode, QuantizationMode]:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/min_max/backend.py
```diff
@@ -148,13 +148,23 @@ def get_weight_tensor(model: TModel, node: NNCFNode) -> Tuple[str, np.ndarray]:
...
         Returns node's weight tensor input port ID.
 
         :param model: Backend-specific model for the initializer finding.
         :param node: NNCFNode to find its weight input port ID.
-        :return: The input port ID of the weight.  None if the weight tensor was not found.
+        :return: The input port ID of the weight.
+        """
+
+    @staticmethod
+    def get_channel_axis(model: TModel, node: NNCFNode) -> int:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/common/insertion_point_graph.py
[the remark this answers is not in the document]
kshpv: Updated. Please, have a look

kshpv:
run torch pre-commit tests

kshpv:
run pytorch pre-commit tests

nncf/experimental/onnx/graph/onnx_graph.py
vshampor: How does using the approach I described lead to copy/paste code?

nncf/common/graph/traversal.py
```diff
@@ -0,0 +1,51 @@
...
+ limitations under the License.
+"""
+
+from typing import TypeVar, Callable, List, Tuple, Optional, Any
+
+Node = TypeVar('Node')
+TraversingGraph = TypeVar('TraversingGraph')
+
+
+def traverse_graph(graph: TraversingGraph,
```
vshampor, line 20: You are writing a template function depending on an interface, but none of the actual objects that are to be used with this function implement the interface explicitly and the interface itself is not even specified. If you are going to extract the function, do your design work right.

nncf/common/graph/traversal.py
```diff
@@ -0,0 +1,51 @@
...
+                   traverse_function: Callable[[Node, List[Any]], Tuple[bool, List[Any]]],
+                   start_nodes: Optional[List[Node]] = None,
```
vshampor, line 22: Not clear in which order the traversal will start.

vshampor (changes requested):

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Updated. Please take a look

nncf/common/graph/traversal.py
kshpv: Added the interface - `TraversableGraph`, which is utilized by `NNCFGraph` and `InsertionPointGraph`

nncf/common/graph/traversal.py
kshpv: Added the description

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Done

tests/onnx/quantization/test_classification_models_graph.py
kshpv: Added the ticket number

vshampor (approved):

KodiaqQ (approved):

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -226,13 +226,22 @@ def _add_weight_quantization_target_point(self, quantization_point: SingleConfig
...
+    def _get_activation_quantization_axis(self, quantizer_config: QuantizerConfig) -> Optional[int]:
+        """
+        Returns quantization axis for activation quantizer.
+
+        :param quantizer_config: Quantization configuration
+        :return: Quantization axis in per-channel case. None in per-tensor case.
+        """
+        if quantizer_config.per_channel:
+            return 1
+        return None
```
l-bat, line 243: I think it should be backend specific function because OpenVINO and TF support NHWC layout.
@KodiaqQ, @andrey-churkin please correct me if I am wrong.

KodiaqQ: Agree, we should move it into the backend entity.

kshpv: Done

l-bat (approved):

andrey-churkin (changes requested):

nncf/common/insertion_point_graph.py
```diff
@@ -243,6 +247,89 @@ def _get_default_post_hook_ip_list(nncf_graph: NNCFGraph) -> List[PostHookInsert
...
+                        return node
+        return node_key
+
+    def get_all_node_keys(self) -> List[str]:
+        """
+        Returns all keys of all nodes.
+
+        :return: All keys of all nodes.
+        """
+        return list(self.nodes)
```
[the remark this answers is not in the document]
kshpv: The same as below. I feel it is better to have the API function than search for correct attributes.

nncf/common/graph/traversal.py
```diff
@@ -0,0 +1,83 @@
...
+                   traverse_forward: bool = True) -> List[Node]:
+    """
+    Traverses graph forward or backward starting from 'start_nodes'.
+    The traversing starts iteratively from the 'start_nodes' list.
+    If 'start_nodes' is None, then traversing starts from the input nodes of the graph.
+    The order of input nodes is based on the returned list of get_input_nodes function of TraversableGraph.
+    The traverse logic should be implemented through 'traverse_function'.
+
+    :param graph: Any graph that implements get_next_nodes, get_previous_nodes and get_input_nodes functions.
+    :param traverse_function: Function describing condition of traversal continuation/termination.
```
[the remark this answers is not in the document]
kshpv: Ok, I will add

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +401,41 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
     @staticmethod
     def get_post_hook_node_key(node_key: str) -> str:
         return InsertionPointGraph.POST_HOOK_ID_PREFIX + node_key
+
+
+class ConstantNodesFilter:
```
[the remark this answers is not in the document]
kshpv: I assume this should be a better design - when we divide the responsibility. This class aims to filter the graph. 
It allows to better test the code and not overwhelm the base class.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -96,8 +96,8 @@ def to_json(self) -> Dict[str, Union[str, float, int]]:
         Serialize all MinMaxQuantization parameters to JSON.
         """
 
-    def _determine_weight_activation_modes(self,
-                                           preset: QuantizationPreset) -> Tuple[QuantizationMode, QuantizationMode]:
+    def _determine_weight_activation_modes(self, preset: QuantizationPreset) \
+            -> Tuple[QuantizationMode, QuantizationMode]:
```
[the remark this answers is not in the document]
kshpv: Done

aiarkinx:
run build

nncf/common/graph/traversal.py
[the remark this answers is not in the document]
kshpv: Rebased traverse logic on networkx functions

nncf/common/insertion_point_graph.py
```diff
@@ -243,6 +247,89 @@ def _get_default_post_hook_ip_list(nncf_graph: NNCFGraph) -> List[PostHookInsert
...
+        """
+        return list(self.nodes)
+
+    def get_all_edges(self) -> List[str]:
+        """
+        Returns graph edges.
+
+        :return: Graph edges.
+        """
+        return self.edges
```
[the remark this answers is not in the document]
kshpv: Removed

nncf/common/insertion_point_graph.py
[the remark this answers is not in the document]
kshpv: Removed

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -47,15 +48,15 @@ def _is_valid_onnx_metatype(node: NodeProto) -> bool:
         """
         node_type = node.op_type
         metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node_type)
-        if metatype == ONNXConstantMetatype:  # We don't need to quantize Constants
-            nncf_logger.debug('The metatype is ONNXConstantMetatype, which means that the node is Constant.'
-                              'All constant nodes are not added to NNCFGraph. The node is skipped.')
+        if metatype == ONNXConstantMetatype:
+            # We don't need to quantize Constants
             return False
```
alexsu52, line 53: Does it mean you build graph without constant but with constant sub-graphs?

alexsu52 (approved):

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: Constant is a node of ONNX opset, whose determines a constant value which is not weight. So, them typically go as attributes for some nodes. They make graph looks messy and dont impact on quantization process. And it seems they are not useful to be added in NNCFGraph for any algorithms al all.
But if some algorithms want to get the value from this constant it is till possible through ONNXGraph interface.

nncf/experimental/onnx/graph/nncf_graph_builder.py
```diff
@@ -47,15 +48,15 @@ def _is_valid_onnx_metatype(node: NodeProto) -> bool:
...
         if metatype == UnknownMetatype:
             node_name = node.name
             nncf_logger.warning(
                 'The node with name {} with type {} was mapped to UnknownMetatype,'
                 ' which means that there was not registered such NNCF metatype. '
+                'It could lead to not the best optimized model.'
```
andrey-churkin, line 59:
```suggestion
                'It could lead to not the best optimized model. '
```

nncf/experimental/onnx/graph/nncf_graph_builder.py
andrey-churkin: Our agreement was to add all nodes as is and filter the insertion point graph to be able to apply QPS algorithm.

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +206,102 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+        weight_input = self.get_node_edge_names(node.name)['input'][weight_port_id]
+        if self.has_initializer(weight_input):
             return self.get_initializer(weight_input).name, self.get_initializers_value(weight_input)
-        except RuntimeError as e:
-            node = self.get_nodes_by_output(weight_input)[0]
-            metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node.op_type)
-            if metatype == ONNXIdentityMetatype:
-                tensor_name = self.get_initializer(node.input[0]).name
-                return tensor_name, self.get_initializers_value(tensor_name)
+        weight_port_id = self.get_weight_port_id(node)
```
andrey-churkin, line 213:
```suggestion
```
You already have this variable.

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +206,102 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+        nodes = deque([parent_node_on_weight_port])
+        while nodes:
+            current_node = nodes.popleft()
+            node_parents = self.get_parents(current_node)
+            nodes.extendleft(node_parents)
+            metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(current_node.op_type)
+            if metatype in [ONNXIdentityMetatype, ONNXQuantizeLinearMetatype]:
+                if self.has_initializer(current_node.input[0]):
+                    return self._get_tensor_from_zero_input(current_node)
+                continue
```
andrey-churkin, line 224: This implementation is not general. It works only for a limited scope of models. It looks like a hack.

nncf/experimental/onnx/graph/onnx_graph.py
```diff
@@ -202,24 +206,102 @@ def get_weight_tensor(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
...
+        :param node: Reshape node, whose input is weight tensor.
+        :return: The weight tensor name and its value with applied the reshape operation.
+        """
+        tensor_name = node.output[0]
+        shape = self.get_initializers_value(node.input[1])
+        tensor_value = self.get_initializers_value(node.input[0])
+        reshaped_tensor_value = tensor_value.reshape(shape)
+        return tensor_name, reshaped_tensor_value
+
+    def _get_tensor_from_zero_input(self, node: onnx.NodeProto) -> Tuple[str, np.ndarray]:
```
andrey-churkin, line 296: Do you really need this method? You use it once.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -307,17 +305,13 @@ def _apply(self,
...
-                    weight_tensor_name, weight_tensor = self._backend_entity.get_weight_tensor(model, node)
-                    # If the nodes share one weight tensor, we should have only one quantizer on that
-                    if weight_tensor_name in weight_initializer_names:
-                        continue
-                    weight_initializer_names.add(weight_tensor_name)
-                except RuntimeError as er:
-                    nncf_logger.exception(er)
+                weight_tensor_name, weight_tensor = self._backend_entity.get_weight_tensor(model, node)
+                # If the nodes share one weight tensor, we should have only one quantizer on that
+                if weight_tensor_name in weight_initializer_names:
```
andrey-churkin, line 310:
```suggestion
                if weight_tensor_name in weight_tensor_names:
```

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: Ok, I will remove the filtration

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +354,35 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
...
+            nncf_logger.debug('The Filtration of Constant nodes of InsertionPointGraph is skipped,'
+                              ' because the graph does not have Input nodes,'
+                              ' from which the traversing should be started.')
+            return ip_graph
+        weight_nodes = []
+        if quantizable_layer_node_keys is not None:
+            weight_nodes = [ip_graph.get_merged_node_from_single_node_key(weight_node) for weight_node in
+                            quantizable_layer_node_keys]
+        visited_nodes = set()
+        start_nodes = input_nodes + weight_nodes
```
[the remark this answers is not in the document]
kshpv: Not necessary. For some cases such as RNN block, there are weight nodes which is not reachable from the model input. 
There is a graph from `tests/torch/data/reference_graphs/quantized/symmetric/lstm_cell.dot`
tempsnip

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +354,35 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
...
+
+        :param ip_graph: The original InsertionPointGraph.
+        :param quantizable_layer_node_keys: Keys of the nodes which have weights.
+        :return: InsertionPointGraph without Constant nodes.
+        """
+        input_nodes = ip_graph.get_input_nodes()
+        if not input_nodes:
+            nncf_logger.debug('The Filtration of Constant nodes of InsertionPointGraph is skipped,'
+                              ' because the graph does not have Input nodes,'
+                              ' from which the traversing should be started.')
```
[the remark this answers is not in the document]
kshpv: In our tests we have such graphs, which was the main reason of adding this skipping. After the discussion with @vshampor it was decided to not spend time on fixing these tests as it could take big efforts.

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -29,25 +29,48 @@ def get_all_aliases(cls) -> List[str]:
...
+    """
+    Contains the information about the weight and bias of the operation.
+    :param weight_channel_axis: Axis for weight per-channel quantization, meaning the number of output filters.
+    :param weight_port_id: Input port of the node's weight.
+    If the value is None the weight_port_id should be determined dynamically.
+    :param bias_port_id: Input port of the node's bias.
+    If the value is None it means that the Metatype does not have bias.
+    """
+    weight_channel_axis: int
+    weight_port_id: Optional[int] = None
```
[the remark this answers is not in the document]
kshpv: In the layers where the weight could be placed on one of the input ports, e.g. MatMul

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Yes, it is. This functional will be extended and improved in the following PRs. I placed TODO

nncf/common/insertion_point_graph.py
```diff
@@ -314,3 +354,35 @@ def get_pre_hook_node_key(node_key: str, input_port_id: int = 0) -> str:
...
+            return ip_graph
+        weight_nodes = []
+        if quantizable_layer_node_keys is not None:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -29,25 +29,48 @@ def get_all_aliases(cls) -> List[str]:
...
 
+@dataclass
+class OpWeightDef:
+    """
+    Contains the information about the weight and bias of the operation.
+    :param weight_channel_axis: Axis for weight per-channel quantization, meaning the number of output filters.
+    :param weight_port_id: Input port of the node's weight.
+    If the value is None the weight_port_id should be determined dynamically.
+    :param bias_port_id: Input port of the node's bias.
+    If the value is None it means that the Metatype does not have bias.
```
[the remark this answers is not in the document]
kshpv: Done

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: Done

nncf/experimental/onnx/graph/onnx_graph.py
kshpv: To not overwhelm the main method I divide some logic into separate function. I don't see the problem here. Do you?

nncf/common/insertion_point_graph.py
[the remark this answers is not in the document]
andrey-churkin: Thanks for the explanation!

nncf/common/insertion_point_graph.py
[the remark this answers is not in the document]
andrey-churkin: Thanks! I got it.

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
andrey-churkin: Yes, but weight_port_id will not contain a None value in this case anyway. You will determine this port_id dynamically and put it into this variable.

nncf/common/graph/traversal.py
```diff
@@ -0,0 +1,83 @@
...
+    :param start_nodes: Nodes from which traversal is started.
+    :param traverse_forward: Flag specifying direction of traversal.
+    If it is set to True the traversing will be done in forward direction, if - False in the backward direction.
```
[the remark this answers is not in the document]
andrey-churkin: It was not fixed.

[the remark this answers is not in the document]
andrey-churkin: Should be
```
    :param weight_port_id: Input port of the node's weight.
        If the value is None the weight_port_id should be determined dynamically.
```

nncf/experimental/onnx/graph/nncf_graph_builder.py
andrey-churkin: It was not fixed.

nncf/experimental/onnx/graph/onnx_graph.py
andrey-churkin: Alright. But this code will not be moved from the experimental until this will not have been fixed.

nncf/experimental/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
kshpv: I prefer to have None in that case and whether it is None we apply the logic to find the weight_port_id

nncf/common/graph/traversal.py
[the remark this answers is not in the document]
kshpv: This is the old code. I removed it

nncf/experimental/onnx/graph/nncf_graph_builder.py
kshpv: I will let you know when I update it. There is a need to update some harware fussing patterns to support elementwise ops with constant

aiarkinx:
run build

kshpv:
run onnx pre-commit tests

kshpv:
run openvino pre-commit tests

kshpv:
run pytorch pre-commit tests

l-bat:
Do we still need this method?
https://github.com/openvinotoolkit/nncf/blob/2e1e3c9376b417fb93b48a64488fa139fa2085d0/nncf/quantization/algorithms/min_max/backend.py#L182-L191

nncf/experimental/onnx/graph/nncf_graph_builder.py

nncf/common/graph/operator_metatypes.py
```diff
@@ -116,6 +117,11 @@ def get_operator_metatype_by_op_name(self, op_name: str) -> Type[OperatorMetatyp
         :return: The operator metatype.
         """
         if op_name not in self._op_name_to_op_meta_dict:
+            nncf_logger.warning(
+                f'The node with name {op_name} was mapped to UnknownMetatype,'
+                ' which means that there was not registered such NNCF metatype. '
+                'It could lead to not the best optimized model. '
+                'Please, Inform the NNCF developers about this message.')
```
andrey-churkin, line 124: Do we really need this warning? In this case, we should add a meta type for each operation in the framework.
@alexsu52 @vshampor

andrey-churkin:
LGTM, I have only one comment above. If it is OK for everyone I think we can merge this PR.

nncf/common/graph/operator_metatypes.py
vshampor: Will move the warning to debug shortly

PR #1366, opened 2022-11-14
[Torch] Weekly test with mixed precision

daniil-lyakhov:
### Changes

Support of mixed precision training in NNCF PT classification sample by torch.cuda.amp autocasting. Mixed precision mode is being enabled by `--mixed-precision` key. Default behavior - FP32 training.
Weekly test with mixed precision is introduced.

### Reason for changes

To test FP16 kernels by the NNCF weekly test

### Related tickets

79587

### DIFF
--- a/tests/torch/conftest.py
+++ b/tests/torch/conftest.py
@@ -49,6 +49,10 @@ def pytest_addoption(parser):
         "--weekly-models", type=str, default=None, help="Path to models' weights for weekly tests"
     )
     parser.addoption(
+        "--weekly-with-mixed-precision", action="store_true", default=False, help="Enable mixed precision for the"
+                                                                                  " nncf weekly test"
+    )
+    parser.addoption(
         "--sota-checkpoints-dir", type=str, default=None, help="Path to checkpoints directory for sota accuracy test"
     )
     parser.addoption(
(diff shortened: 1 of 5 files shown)


alexsu52 (approved):
LGTM

tests/torch/conftest.py
```diff
@@ -48,6 +48,10 @@ def pytest_addoption(parser):
     parser.addoption(
         "--weekly-models", type=str, default=None, help="Path to models' weights for weekly tests"
     )
+    parser.addoption(
+        "--weekly-with-mixed-precision", action="store_true", default=False, help="Enable mixed precision for the"
```
ljaljushkin, line 52: Not sure, it's a good way to extend weekly/nightly tests. It could lead to handling many options that means almost the same.
Ideally, there should be a single pytest marker: `weekly_tests` and `nightly_tests` similar to that: 
`
@pytest.mark.webtest
`
https://docs.pytest.org/en/7.1.x/example/markers.html#mark-examples

ljaljushkin (changes requested):

tests/torch/conftest.py
daniil-lyakhov: I changed it, now you can run weekly test by
```python -m pytest -s tests/torch -m weekly  --weekly-models ...```
as well as nighlty test by
```python -m pytest -s tests/torch -m nightly ...```

Mixed precision for test sota checkpoint is not enabled by this PR. In case we need mixed precision for nightly test let's create a task for that

daniil-lyakhov: Please see 96744

ljaljushkin (approved):

PR #1343, opened 2022-10-27
Adaptive compression training loop improvements

nikita-savelyevv:
This PR finishes up the changes started in #1134.

### Changes
- Making use of EarlyExit loop as initial phase of AdaptiveCompression loop. This reduces the time required for initial training as it may now exit early.
- Introduction early exit -like logic to main AdaptiveCompression loop in the form of existing a training phase if positive accuracy budget is reached. This further reduces the training time.
- Minor changes:
  - Improved logging
  - Code inspection warnings fixes
  - Reduced the number of parameters which user needs to set manually
- Bugs fixed:
  - Failing updating initial scheduler learning rate ['base_lrs', 'init_lr’]
  - Compression scheduler disabling behavior was incorrect
  - Incorrect behavior of best model saving for sparsity backend

### Reason for changes

Improvement and bug fixing of AdaptiveCompression algorithm

### Related tickets

94288

### Tests

Existing tests were tweaked according to the new algorithm logic. New tests addition is left for the future ticket. 

Performed experiments on ResNet18 CIFAR10 for pruning backend. Experiments demonstrate improvement of training time without significant loss of accuracy or compression.
| Value | Before | This PR |
| ----------- | ----------- | ----- |
| Epochs trained | 350 |   137  |
| Absolute accuracy drop | 0.74 % |  0.84 %  |
| Compression rate | 47.5 %   |  47.5 %  |

Slighter lower accuracy drop for old algorithm may be due to overfitting to the validation dataset.

Further TODOs / self-review is in the [separate branch](https://github.com/nikita-savelyevv/nncf/tree/actl-todo).

PR #1374, opened 2022-11-16
Make OrderedEnum int based

nikita-savelyevv:
### Changes

Removed `OrderedEnum` and replaced its usages by `IntEnum`.

### Reason for changes

In general `OrderedEnum` should base on `IntEnum` because a method \_\_lt\_\_ is implemented for it where enum values are compared with "<" sign. This is not critical, but currently it complicates development of #1343. There, when an object of type `CompressionStage(OrderedEnum)` is being serialized by json, it throws an error, because it is ambiguous how to serialize a general enum object. Making `OrderedEnum` be based from `IntEnum` resolves this issue.

**Edit:** decided to remove `OrderedEnum` altogether and use `IntEnum` instead.


vshampor:
We might still need ordered string enums. You can try to use multiple inheritance with `IntEnum`/`StrEnum` and an `OrderedEnumBase` (defining comparison magic methods), and then have `class OrderedEnum(IntEnum, OrderedEnumBase)` and `class OrderedStrEnum(StrEnum, OrderedEnumBase)`.

nikita-savelyevv:
> 
There is no usage for OrderedStrEnum as of now. Maybe its better to create it when it will be required? 

Also, I don't think one would ever need an ordered string enum.

nikita-savelyevv:
> Whoever introduced the original ordered enum needed it. For reasons of compatibility should use `OrderedStrEnum` instead of an `OrderedEnum` wherever string enums were inherited from the original `OrderedEnum`.

According to the information I received. The only two use cases with string ordered enums were made by mistake. I fixed them by basing from regular enum.

alexsu52 (commented):
It looks like you can replace `OrderedEnum` to `IntEnum` and remove `OrderedEnum` class from repository.

nikita-savelyevv:
Replaced `OrderedEnum` with `IntEnum`

alexsu52 (approved):

vshampor (approved):

PR #1381, opened 2022-11-19
Extract shared test code into own folder

vshampor:
### Changes
Creates a separate folder for the reusable test code and moves such code there.

### Reason for changes
Currently such code is stored in the `common` subfolder, and it seems necessary to separate the responsibilities of these folders - the `common` test suite is an entity of the same level as the framework-specific suites, and it doesn't seem right, for example, for `tests.torch` to import from `tests.common` - better for them all to import from the shared location. This also allows for effective fixture and pytest parameter reuse.

### Related tickets

N/A

### Tests
Existing tests


tests/shared/README.md
```diff
@@ -0,0 +1,4 @@
+This folder contains test code that is meant to be reused in more than one backend-specific (or agnostic) test subfolder, such as reusable fixtures, parsers, comparators etc.
```
alexsu52, line 1: What do you think about renaming `tests/shared` -> `tests/utils` or `tests/test_utils`?

alexsu52 (approved):
cool staff!

tests/shared/README.md
vshampor: There's always an option to introduce `utils` into a name of the file or a foler at a lower level. At this level I need to make evident that this is the code that is designed to be "shared".

ljaljushkin (approved):

PR #1384, opened 2022-11-20
Fix export of SOTA checkpoints for e2e validation

ljaljushkin:
### Changes

- export to ONNX in EVAL mode by default
- export to ONNX in TRAINING mode for NAS in order to avoid hang. Most probably it happens because of incompatibility of skipped layer in Elastic Depth with optimizations/fusings on export in EVAL mode.
- remove error for per-channel Q-DQ export

### Reason for changes

The behavior of exporting in training mode had been changed, and it started producing invalid graph for evaluation, as result e2e validation started to fail on accuracy validation

### Related tickets

96710

### Tests

e2e torch

### DIFF
--- a/nncf/experimental/torch/nas/bootstrapNAS/elasticity/onnx_export.py
+++ b/nncf/experimental/torch/nas/bootstrapNAS/elasticity/onnx_export.py
@@ -1,0 +1,94 @@
+"""
+ Copyright (c) 2022 Intel Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


nncf/experimental/torch/nas/bootstrapNAS/elasticity/onnx_export.py
```diff
@@ -0,0 +1,94 @@
...
+from typing import List
+
+import torch
+from torch import nn
+from torch.utils.hooks import RemovableHandle
+
+from nncf.torch.exporter import PTExporter
+
+
+class BNTrainingStateSwitcher:
```
vshampor, line 22: Can we keep this in the main torch code? We might need this later, I don't think that this is only specific to NAS by definition (although I do see that currently will only be used for NAS).

ljaljushkin: Don't see when it's needed now. It's specific implementation with hooks that is needed for export in training.
BNAdaptation uses different implementation.

vshampor (approved):

PR #1390, opened 2022-11-26
Remove direct setup.py installation

vshampor:
### Changes

READMEs and tests are adjusted to use the `pip install` approach instead of the `python setup.py` approach.

### Reason for changes
`python setup.py` direct calls are deprecated in the community.

### Related tickets
N/A

### Tests
cross_fw/install/test_install

PR #1396, opened 2022-11-30
Fix torch install

vshampor:
### Changes
Fixed leftovers from #1390 to enable correct testing for the PT GPU installation

### Reason for changes
Install tests fail otherwise.

### Related tickets
N/A

### Tests
tests/cross_fw_install

### DIFF
--- a/tests/cross_fw/install/test_install.py
+++ b/tests/cross_fw/install/test_install.py
@@ -45,10 +45,15 @@ def run_install_checks(venv_path: Path, tmp_path: Path, package_type: str, backe
             package_path = find_file_by_extension(PROJECT_ROOT / 'dist', '.tar.gz')
         elif package_type == "build_w":
             package_path = find_file_by_extension(PROJECT_ROOT / 'dist', '.whl')
-
-        subprocess.run(
-            f'{pip_with_venv} install {package_path}[{backend}] ',
-                check=True, shell=True)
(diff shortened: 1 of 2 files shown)


tests/cross_fw/install/test_install.py
```diff
@@ -45,10 +45,15 @@ def find_file_by_extension(directory: Path, extension: str) -> str:
...
-        subprocess.run(
-            f'{pip_with_venv} install {package_path}[{backend}] ',
-                check=True, shell=True)
+        # Currently CI runs on RTX3090s, which require CUDA 11 to work.
+        # Current torch, however (v1.12), is installed via pip using .whl packages
+        # compiled for CUDA 10.2. Thus need to direct pip installation specifically for
+        # torch, otherwise the NNCF will only work in CPU mode.
+        torch_extra_index = " --extra-index-url https://download.pytorch.org/whl/cu116"
+        run_cmd_line = f'{pip_with_venv} install {package_path}[{backend}]'
+        if backend == "torch":
```
ljaljushkin, line 54: Can we reuse `--torch-with-cuda11` option for that? Move it to common conftest.py, for instance
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/conftest.py#L84

vshampor: I think it doesn't make sense to have this configurable - we don't have CUDA 10 anywhere on our machines and are targeting CUDA11 anyway.

PR #1398, opened 2022-11-30
Add CPUExecutionProvider to the ONNX E2E

kshpv:
### Changes

Add CPUExecutionProvider as a launcher to measure models metrics. 
Updated AC configs.
Add two options for e2e test to disabel ov-ep and enable cpu-ep

### Reason for changes

To vanish OV inference problem and control the quality of NNCF.

### Related tickets

No

### Tests

No


kshpv:
The last report with only OV-EP is 147 number
The last report with OV-EP and CPU-EP is 142 number

vshampor (approved):

alexsu52 (approved):
LGTM

tests/onnx/benchmarking/object_detection_segmentation/onnx_models_configs/fcn-resnet50-12.yml
```diff
@@ -14,6 +19,7 @@ models:
           has_background: True
           sort_annotations: True
           semantic_only: True
+          convert_mask: False
```
KodiaqQ, line 22: Will this change affect the final accuracy?

l-bat (approved):

tests/onnx/benchmarking/object_detection_segmentation/onnx_models_configs/fcn-resnet50-12.yml
kshpv: No, this is needed to not overwrite the existing masks

KodiaqQ (approved):

kshpv:
build is 175

aiarkinx (approved):

kshpv:
The build is 192

Makefile
```diff
@@ -34,6 +38,18 @@ test-install-onnx:
...
 		--junitxml ${JUNITXML_PATH}
 
+ONNX_E2E_PTQ_SIZE ?= 300
+ONNX_E2E_PTQ_EVAL_SIZE ?=
+ONNX_E2E_OPTIONS=--model-dir ${MODELS_DIR} --data-dir ${DATA_DIR}  \
+				 --output-dir ${OUTPUT_DIR} --anno-dir ${ANNO_DIR} \
+                 --junitxml ${JUNITXML_PATH} --ptq-size ${ONNX_E2E_PTQ_SIZE} \
+                 --eval-size ${ONNX_E2E_PTQ_EVAL_SIZE}
+
+test-e2e-ptq:
```
alexsu52, line 48: My general comment that I would like to avoid "super" makefile which includes all possible targets with many input arguments. Could you propose the solution of this problem?

vshampor: I think running e2e does not belong as a makefile target at all, it has too many parameters for that which must be forwarded via clunky make CLI. I think that if we want to decouple CI-defined, stable parameter set such as dataset folders from the set of parameters that may be changed frequently by developers, so that devs can make faster changes without involving a change in CI repo, we could make a wrapper script such as run_e2e_onnx.py, which will accept only the CI-dependent args (i.e. dataset folder) and then forward them to a pytest call via subprocess.

vshampor (changes requested):

alexsu52 (approved):

vshampor (approved):

andrey-churkin (approved):

PR #1399, opened 2022-12-01
Remove fixed protobuf versions from requirements

vshampor:
### Changes
As stated in the title.

### Reason for changes
Originally the versions were fixed as a workaround for direct setup.py installation issues. Now that we deprecate this installation approach, the hacks are not required.

### Related tickets
N/A

### Tests
cross_fw/install


vshampor:
TF install test 126 passed.

PR #1402, opened 2022-12-02
Update onnx models to have no weights

kshpv:
### Changes

Replace onnx models by the same models topology but having no weights. Add a function for loading such models and filling the weights by zero.

### Reason for changes

To reduce the repo size

### Related tickets

97804

### Tests

Tests were updated accordingly


kshpv:
There is a problem with Mask RCNN model which has zero-detection with random weights.
It leads to an error during statistics collection.
Because of that this model can not be quantized

tests/onnx/quantization/helper.py
```diff
@@ -0,0 +1,117 @@
...
+    for tensor in onnx_graph.onnx_model.graph.initializer:
+        if uses_external_data(tensor):
+            np_dtype = TENSOR_TYPE_MAP[tensor.data_type].np_dtype
+            np_tensor = np.random.random(list(tensor.dims)).astype(np_dtype)
+            tensor_with_zeros = onnx.numpy_helper.from_array(np_tensor, name=tensor.name)
+            tensor.CopyFrom(tensor_with_zeros)
+    return model
+
+
+def save_model_without_tensors(model: onnx.ModelProto, model_filename: Path, model_dir: Path) -> None:
```
vshampor, line 103: Use https://docs.python.org/3/library/tempfile.html to work in a directory guaranteed to be temporary.

vshampor (changes requested):
A great improvement! Still has a couple of things to be changed. Might want to add a README.md in `tests/onnx/data/models` describing that these modules are actually without weights and what to use in code to load and save such models.

tests/onnx/quantization/helper.py
```diff
@@ -0,0 +1,117 @@
...
+
+    :param model_path: Path to the onnx model to load.
+    :return: Onnx model with filled the all external tensors by random values.
+    """
+    model = onnx.load_model(model_path, load_external_data=False)
+    onnx_graph = ONNXGraph(model)
+    for tensor in onnx_graph.onnx_model.graph.initializer:
+        if uses_external_data(tensor):
+            np_dtype = TENSOR_TYPE_MAP[tensor.data_type].np_dtype
+            np_tensor = np.random.random(list(tensor.dims)).astype(np_dtype)
```
vshampor, line 97: Set seed before using random

tests/onnx/quantization/helper.py
kshpv: Done

tests/onnx/quantization/helper.py
kshpv: Replace to loading zeros weights, not random. It is better for degerming whether the model was loaded by provided method

tests/onnx/quantization/helper.py
```diff
@@ -0,0 +1,117 @@
+"""
```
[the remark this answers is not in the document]
kshpv: Done

tests/onnx/quantization/test_graphs.py
```diff
@@ -13,15 +13,26 @@
...
 from tests.onnx.quantization.common import infer_model, min_max_quantize_model
 from tests.onnx.quantization.common import compare_nncf_graph
+from tests.onnx.quantization.common import mock_calculate_activation_quantizer_parameters
+from tests.onnx.quantization.common import mock_get_statistics
+from tests.onnx.quantization.common import mock_collect_statistics
 from tests.onnx.models import ALL_SYNTHETIC_MODELS
 from tests.onnx.models import MultiInputOutputModel
 
-
+@patch('nncf.quantization.algorithms.min_max.algorithm.calculate_activation_quantizer_parameters',
```
vshampor, line 28: Use patching and mocking in the same manner as it is done in the majority of the rest of the tests. Search the code base for `mocker.patch` or `mocker.spy` for examples.

vshampor (changes requested):

tests/onnx/quantization/test_graphs.py
kshpv: Done

KodiaqQ (approved):

l-bat (approved):

ljaljushkin:
Will it actually reduce repository size? Shouldn't we carefully delete all these files from all previous commits to achieve that?
How did you measure it?

alexsu52 (approved):
LGTM

kshpv:
@alexsu52 @ljaljushkin 
`nncf/tests/onnx/` size before: **1.4 GB** 
`nncf/tests/onnx/` size after: **7.3 MB**

kshpv:
@alexsu52 can we merge this without @vshampor approve?

ljaljushkin (approved):

PR #1403, opened 2022-12-02
Test improvements

vshampor:
### Changes
Switched ONNX tests to dumping intermediate ONNX files in a temporary directory.
Improved sorting of the reference .dot files to actually make sense (used to be that the edges for which the *string representation of the concatenation of the start and end node IDs* was the key in sorting, made it so that the sorting is hierarchical based on the start node ID first and end node ID second)

### Reason for changes
General improvements

### Related tickets
N/A

### Tests
precommit graph tests


tests/common/graph/nx_graph.py
```diff
@@ -28,14 +30,56 @@ def sort_dot(path):
...
+            return False
+
+    def graph_key(line: str) -> LineOrder:
         key = line.split(' ')[0].replace('"', '')
         if '->' in line:
-            key += line.split(' ')[3].replace('"', '')
-            return int(key) + offset
-        return int(key)
+            start_id = int(key)
+            end_id_str = line.split(' ')[3].replace('"', '')
```
ljaljushkin, line 77: What if there are more than 2 attributes on the edge? it would more safe to split by '->' first  and take the first element on the right after split by space.

ljaljushkin (changes requested):

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -18,34 +18,27 @@
...
-from nncf.common.utils.dot_file_rw import write_dot_graph
 from nncf.experimental.onnx.graph.nncf_graph_builder import GraphConverter
 from nncf.experimental.onnx.model_normalizer import ONNXModelNormalizer
+from tests.common.graph.nx_graph import compare_nx_graph_with_reference
 from tests.onnx.conftest import ONNX_TEST_ROOT
 
 from tests.onnx.models import ALL_SYNTHETIC_MODELS
 from tests.shared.paths import TEST_ROOT
 from tests.onnx.quantization.common import ModelToTest
-from tests.onnx.quantization.common import check_nx_graph
```
ljaljushkin, line 30: 👍

tests/onnx/test_nncf_graph_builder.py
```diff
@@ -18,34 +18,27 @@
...
 
     nncf_graph = GraphConverter.create_nncf_graph(model_to_test.onnx_model)
     nx_graph = nncf_graph.get_graph_for_structure_analysis(extended=True)
 
-    if generate_ref_graphs:
-        write_dot_graph(nx_graph, path_to_dot)
-
-    expected_graph = read_dot_graph(path_to_dot)
-    check_nx_graph(nx_graph, expected_graph)
+    compare_nx_graph_with_reference(nx_graph, path_to_dot)
```
ljaljushkin, line 41: 👍

tests/common/graph/nx_graph.py
vshampor: Implementing splitting by ' -> ' and regex matching afterwards

ljaljushkin (approved):

PR #1405, opened 2022-12-05
No weights override on sparsity model's export

ljaljushkin:
### Changes

Export sparsity models with non-in-place (`mul` instead of `mul_`) multiply operation.

### Reason for changes

User doesn't expect that export to onnx would do something with parameters. 
It could break convergence if user wants export on each epoch (e.g. saving checkpoint in Optimum requires export to ONNX and conversion to IR).

### Related tickets

n/a

### Tests

- [x] test checking that after export model has the same parameters
- [x] add test for filter pruning
- [x] e2e for resnet50-rb-sparsity-int8 (build 128)
- [x] e2e for unet_mapillary_pruning_geometric_median (build 135)

### DIFF
--- a/nncf/torch/pruning/filter_pruning/layers.py
+++ b/nncf/torch/pruning/filter_pruning/layers.py
@@ -16,7 +16,6 @@ from torch import nn
 
 from nncf.torch.layer_utils import COMPRESSION_MODULES
 from nncf.common.graph import NNCFNodeName
-from nncf.torch.utils import is_tracing_state, no_jit_trace
 
 
 @COMPRESSION_MODULES.register()
@@ -51,13 +50,10 @@ class FilterPruningMask(nn.Module):
 
             # For weights self.mask_applying_dim should be used, for bias dim=0
             dim = 0 if param_name == 'bias' else self.mask_applying_dim
-            if is_tracing_state():
-                with no_jit_trace():
-                    new_params.append(inplace_apply_filter_binary_mask(self.binary_filter_pruning_mask, param_value,
-                                                                       node_name_for_logging=self.node_name, dim=dim))
-            else:
-                new_params.append(apply_filter_binary_mask(self.binary_filter_pruning_mask, param_value,
-                                                           node_name_for_logging=self.node_name, dim=dim))
+            new_params.append(
+                apply_filter_binary_mask(self.binary_filter_pruning_mask, param_value,
+                                         node_name_for_logging=self.node_name, dim=dim)
+            )
         return new_params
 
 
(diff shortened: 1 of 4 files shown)


ljaljushkin:
The main concern of using non-in-place operation `mul` instead of `mul_` is different representation of the model weights in ONNX:

image

E2E validation in OpenVINO for the sparse-int8 resnet model with matmul passed.
build 128(with mul) and build 126 (with mul_):
image
image

vshampor (approved):

kshpv (approved):

nncf/torch/pruning/filter_pruning/layers.py
```diff
@@ -51,13 +50,10 @@ def forward(self, **params):
 
             # For weights self.mask_applying_dim should be used, for bias dim=0
             dim = 0 if param_name == 'bias' else self.mask_applying_dim
-            if is_tracing_state():
-                with no_jit_trace():
-                    new_params.append(inplace_apply_filter_binary_mask(self.binary_filter_pruning_mask, param_value,
```
ljaljushkin, line 56: @daniil-lyakhov @vshampor I've completely removed `inplace_apply_filter_binary_mask`, because it was needed for export only.
Also, I've corrected documentation for `apply_filter_binary_mask`. Please review one more time.

ljaljushkin:
Also, restarted e2e validation for new changes: build 137

vshampor (approved):

nncf/torch/pruning/filter_pruning/layers.py
daniil-lyakhov: Why `no_jit_trace` was used before than? Is it working without it? Other thing: should we remove `no_jit_trace` from the sparsity algo as well?

ljaljushkin: removed `no_jit_trace` from sparsity as well

ljaljushkin: > Originally, `no_jit_trace` excludes operations from ONNX model. But in case of binary mask, I didn't find a difference in ONNX model with and without it with ONNX opset starting from 8 up to 13. Maybe it was actual for old versions of pytorch.

@alexsu52 @vshampor @AlexKoff88 do you remember why it was used for binary mask before?
This code was since migration to Github.
https://github.com/openvinotoolkit/nncf/blame/372d740fcc234f77157d8f8f6d3e8b59463a423f/nncf/torch/pruning/filter_pruning/layers.py#L39

ljaljushkin: e2e validation for filter pruning model passed (build 137) without `no_jit_trace`. Can't see any degradation if not using it.

alexsu52: Honestly, I don't remember. How did you check that ONNX model is the same with and without `no_jit_trace` with ONNX opset starting from 8 up to 13. What torch versions did you check?

ljaljushkin: There's `save_format` option in `export_model` function. For instance, one can specify `onnx_10` for opset 10. I used torch==1.12.1, we don't test and support lower versions of pytorch now.

daniil-lyakhov (approved):

ljaljushkin:
@alexsu52 
I've performed e2e validation on torch 1.8.2 with and without this PR. I used `unet_mapillary_pruning_geometric_median` model.
The IR's have the same filters removed:
image
The validation accuracy is the same:

The same structure in IR's generated by torch 1.9.1:
image

alexsu52 (approved):

PR #1406, opened 2022-12-05
Export Q-DQ for correct axis in torch

vshampor:
### Changes
Added detection of the per-channel axis in Q-DQ export in PT and proper setting of the `axis_i` attribute for the ONNX ops

### Reason for changes
mobilenet_v2_imagenet_int8 fails to be converted by MO in Q-DQ export mode due to presence of per-channel activation quantized eltwise convs which had `axis` erroneously set to 0 in Q-DQ ops

### Related tickets
N/A

### Tests
e2e_pt, unit tests TBA

### DIFF
--- a/nncf/torch/quantization/layers.py
+++ b/nncf/torch/quantization/layers.py
@@ -445,8 +445,8 @@ class BaseQuantizer(nn.Module):
                 # Impossible to determine proper axis for per-channel quantization because we have
                 # scale shape ~ [1, 1, 1, 1], therefore falling back to per-tensor style export
                 axis = 1  # default value by opset, ignored in per-tensor quantization anyway
-                y_scale = y_scale.flatten()
-                y_zero_point = y_zero_point.flatten()
+                y_scale = y_scale.flatten()[0]
+                y_zero_point = y_zero_point.flatten()[0]
             else:
                 axis = possible_axes[0]
 
--- a/nncf/torch/quantization/layers.py
+++ b/nncf/torch/quantization/layers.py
@@ -33,8 +33,9 @@ from nncf.common.quantization.quantizer_setup import QuantizerSetupBase
 from nncf.common.quantization.quantizer_setup import QuantizationPointId
 from nncf.torch.graph.transformations.commands import TargetType
 from nncf.torch.graph.transformations.commands import PTTargetPoint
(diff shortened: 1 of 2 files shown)


ljaljushkin (approved):

ljaljushkin:
LGTM, please share results of e2e for Q-DQ prior to the merge

nncf/torch/quantization/layers.py
```diff
@@ -428,14 +429,31 @@ def _prepare_fq_export_quantization(self, x: torch.Tensor):
...
-    def _prepare_qdq_export_quantization(self, x: torch.Tensor):
+    def _prepare_qdq_export_quantization(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, int]:
         x, level_high, level_low, input_low, input_high = self._prepare_export_quantization(x)
         with no_jit_trace():
             y_scale, y_zero_point = get_scale_zp_from_input_low_input_high(level_low,
                                                                            level_high,
                                                                            input_low,
                                                                            input_high)
-        return x, y_scale, y_zero_point
+            possible_axes = self._possible_per_channel_dimensions()
```
AlexKoff88, line 439: Suggest calling the method like `get_channel_axis()` and encapsulate all the checks there.

vshampor:
E2E PT build 132 works for mobilenet_v2_int8 for both FQ and Q/DQ

nncf/torch/quantization/layers.py
```diff
@@ -428,14 +429,31 @@ def _prepare_fq_export_quantization(self, x: torch.Tensor):
...
+            if len(possible_axes) > 1:
+                raise RuntimeError(
+                    f"Impossible to determine the per-channel axis for a scale shape {self.scale_shape} - "
+                    f"more than one dimension is >1")
+            if not possible_axes:
+                # Impossible to determine proper axis for per-channel quantization because we have
```
AlexKoff88, line 445: The exception should be raised in this case as well because it should indicate that there is a logic in FQ insertion logic.

AlexKoff88:
Seriously, I don't understand why we merged this PR. I suggest reverting it and introducing proper logic and clean code without one-line functions and blurred logic.

AlexKoff88:
cc'ed @andrey-churkin and @KodiaqQ as NNCF maintainers.

PR #1407, opened 2022-12-06
Switch to torch 1.13

vshampor:
### Changes
As stated in the title

### Reason for changes
Keeping up with the latest torch version

### Related tickets
98528

### Tests
PT and ONNX precommit scope


vshampor:
Test E2E build 3 shows no degradations to the regular E2E.

alexsu52 (approved):

examples/torch/common/models/segmentation/icnet.py
```diff
@@ -168,14 +168,23 @@ def __init__(self, input_size_hw, bin_dimensions=None, mode='sum'):
...
             pad_h = (dim - (input_size_hw[0] % dim)) % dim
             pad_w = (dim - (input_size_hw[1] % dim)) % dim
-            self.paddings[dim] = [0, pad_w, 0, pad_h]
+            self.paddings[dim] = (0, pad_w, 0, pad_h)
 
     def forward(self, inputs):
         x = inputs.clone()
 
         for dim in self.bin_dimensions:
-            padded_input = F.pad(inputs, self.paddings[dim], mode='constant', value=0)
```
ljaljushkin, line 177: why did you remove pad op?

nncf/torch/binarization/binarize_functions.py
```diff
@@ -19,8 +19,24 @@
...
+    # Unsqueeze handling for different opsets inspired by torch.onnx.symbolic_helper._unsqueeze_helper
+    # The original unsqueeze_helper cannot be used in 1.13 since it references
+    # an `.opset` attribute on the `g` argument which is not there. The original intent
+    # of this in pytorch was to allow accessing opset info in the symbolic fn's
+    # code (see PR #84728 in the pytorch repo), but somehow the `symbolic` functions
+    # that we define are called from the C++ code, and they pass the old torch._C.Graph
+    # object as `g` instead of the torch.onnx._internal.jit_utils.GraphContext as it
+    # should be.
+    # The effect is that we can only do opset>=13-style unsqueeze here and hope that
+    # the user did not request an older opset.
```
ljaljushkin, line 35: Some tests are still using opset lower than 13. Does it make sense to keep logic for lower opset versions?

vshampor: We really only need this logic for our own unsqueezes in binarization. I don't think that there are that many uses for binarization anyway either in test or among users.

examples/torch/common/models/segmentation/icnet.py
vshampor: `F.pad` is one operation that is incorrectly implemented in torch itself in terms of exporting with shape inference - it produces invalid shape data for the output tensor which then cannot be treated correctly in avg_pool_2d. I originally removed this because this allowed to export FP32 ICNet without the `ATEN_FALLBACK`, but then I hit the proble with INT8 ICNet because FQs don't specify shape inference too.
Anyway, it seems that the `avg_pool_2d` op does the same padding implicitly - I can claim this due to the fact that with and without padding the result accuracy metrics are equivalent.

ljaljushkin (approved):

PR #1408, opened 2022-12-06
Disable graph dump

alexsu52:
### Changes

Disable graph dump in the quantize function for PyTorch backend.

### Reason for changes

Do not litter:)

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/nncf/torch/quantization/quantize.py
+++ b/nncf/torch/quantization/quantize.py
@@ -205,7 +205,8 @@ def quantize_impl(model: torch.nn.Module,
         config=nncf_config,
         dummy_forward_fn=dummy_forward_fn,
         wrap_inputs_fn=wrap_inputs,
-        wrap_outputs_fn=wrap_outputs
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev:
I think we should add some option to enable to get graph, ex. env variable `NNCF_DUMP_DOT_GRAPH` or add parameter to config. 

And need update documentation in [ote_sdk/ote_sdk/test_suite/training_tests_actions.py](https://github.com/openvinotoolkit/nncf/blob/62ef4ceca588a2daab7618851b40eef5c10fe449/docs/Usage.md#exploring-the-compressed-model) and [ote_sdk/ote_sdk/test_suite/training_tests_actions.py](https://github.com/openvinotoolkit/nncf/blob/119e3120d14f38683e06793eb89ea31a04009fe8/docs/FAQ.md#how-can-i-exclude-certain-layers-from-compression).

PR #1415, opened 2022-12-08
Do not send telemetry if in pytest

vshampor:
### Changes
When launched in a pytest context, NNCF won't be sending telemetry. This applies even to usages of NNCF in pytest runs other than our own (i.e. in dependent tools' testing pipelines).

### Reason for changes
Keeping test runs out of telemetry.

### Related tickets
78624

### Tests
test_telemetry

### DIFF
--- a/nncf/telemetry/wrapper.py
+++ b/nncf/telemetry/wrapper.py
@@ -11,6 +11,7 @@
  limitations under the License.
 """
 import os
+import sys
 from unittest.mock import MagicMock
 
 from nncf import __version__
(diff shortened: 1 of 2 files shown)


nncf/telemetry/wrapper.py
```diff
@@ -32,5 +33,9 @@
     nncf_logger.debug("openvino_telemetry package not found. No telemetry will be sent.")
     NNCFTelemetry = None
 
-if os.getenv(NNCF_CI_ENV_VAR_NAME) or os.getenv(NNCF_DEV_ENV_VAR_NAME) or NNCFTelemetry is None:
+# Currently the easiest way to disable telemetry in tests. Will break telemetry if pytest is used in NNCF package code.
```
ljaljushkin, line 36: How will it work in a developer environment when pytest is included to it?

vshampor: `sys.modules` contains everything that has been imported. User code doesn't usually do `import pytest` in production - even in NNCF examples there isn't a single `import pytest` line.

ljaljushkin (approved):

PR #1418, opened 2022-12-08
Improve NNCF logging

vshampor:
### Changes
Excessive logs were either removed or brought down to the DEBUG level. Added some vanity logs at critical locations (i.e. during original `nncf` module import and torch extension loading). Converted log strings to f-strings. Expanded the style guide with the guidelines for logging.

### Reason for changes
NNCF outputs too much as logs right now, and little out of these logs makes sense to an average user.

### Related tickets
93380

### Tests
Manual


nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -511,7 +511,7 @@ def build(self, target_model: NNCFNetwork) -> ElasticKernelHandler:
                 # Padding
                 ap = ElasticKernelPaddingAdjustment(elastic_kernel_op)
                 pad_op = UpdatePadding(ap).to(device)
-                nncf_logger.warning('Padding will be adjusted for {}'.format(node_name))
+                nncf_logger.warning(f'Padding will be adjusted for {node_name}')
```
ljaljushkin, line 514: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -496,7 +496,7 @@ def build(self, target_model: NNCFNetwork) -> ElasticKernelHandler:
                 if max_kernel_size >= 3:
                     crop_op = ElasticKernelInputForExternalPadding(elastic_kernel_op, max_kernel_size)
                     op = UpdateInputs(crop_op).to(device)
-                    nncf_logger.warning('Padded input will be cropped for {}'.format(node_name))
+                    nncf_logger.warning(f'Padded input will be cropped for {node_name}')
```
ljaljushkin, line 499: Propose moving to debug
@jpablomch please confirm that it won't hurt user experience

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -464,7 +464,7 @@ def build(self, target_model: NNCFNetwork) -> ElasticKernelHandler:
             self._node_names_to_make_elastic = [node.node_name for node in all_elastic_kernel_nodes]
 
         for node_name in self._node_names_to_make_elastic:
-            nncf_logger.info("Adding Elastic Kernel op for Conv2D in scope: {}".format(node_name))
+            nncf_logger.info(f"Adding Elastic Kernel op for Conv2D in scope: {node_name}")
```
ljaljushkin, line 467: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/search/evaluator.py
```diff
@@ -101,8 +101,7 @@ def add_to_cache(self, subnet_config_repr: Tuple[float, ...], measurement: float
         :param measurement: value for the evaluator's metric.
         :return:
         """
-        nncf_logger.info("Add to evaluator {name}: {subnet_config_repr}, {measurement}".format(
-                         name=self.name, subnet_config_repr=subnet_config_repr, measurement=measurement))
+        nncf_logger.info(f"Add to evaluator {self.name}: {subnet_config_repr}, {measurement}")
```
ljaljushkin, line 104: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -820,8 +819,8 @@ def find_pairs_of_nodes_with_different_width(self, pairs_of_nodes: List[Tuple[st
                          f'Width: {start_width} vs {end_width}. Shapes: {start_output_shape} vs {end_output_shape}'
             else:
                 continue
-            nncf_logger.warning('The block [\n\t{},\n\t{}\n]\n can`t be skipped, because {}'.format(
-                start_node_name, end_node_name, reason))
+            nncf_logger.warning(
```
ljaljushkin, line 822: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -335,7 +335,7 @@ def _generate_width_list(max_width: int, params: ElasticWidthParams) -> List[int
                 if p.max_num_widths == len(width_list):
                     break
                 if 0 >= multiplier > 1:
-                    nncf_logger.warning("Wrong value for multiplier: {}. Skipping ".format(multiplier))
+                    nncf_logger.warning(f"Wrong value for multiplier: {multiplier}. Skipping ")
```
ljaljushkin, line 338: I'd throw an exception, as it happens on init. 
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/search/search.py
```diff
@@ -459,9 +459,8 @@ def _evaluate(self, x: List[float], out: Dict[str, Any], *args, **kargs) -> NoRe
             self._elasticity_handler.activate_subnet_for_config(sample)
             if sample != self._elasticity_handler.get_active_config():
                 nncf_logger.warning("Requested configuration was invalid")
-                nncf_logger.warning("Requested: {sample}".format(sample=sample))
-                nncf_logger.warning("Provided: {config}".format(
-                                    config=self._elasticity_handler.get_active_config()))
+                nncf_logger.warning(f"Requested: {sample}")
+                nncf_logger.warning(f"Provided: {self._elasticity_handler.get_active_config()}")
```
ljaljushkin, line 463: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_depth.py
```diff
@@ -405,9 +405,7 @@ def build(self, target_model: NNCFNetwork) -> ElasticDepthHandler:
         nncf_logger.info("Blocks for skipping (changing the depth of model):")
         new_line = '\n'
         str_bs = [str(block) for block in blocks_for_info]
-        nncf_logger.info(f"""[{new_line.join(str_bs)}
-                        ]
-                        """)
+        nncf_logger.info(f"[{new_line.join(str_bs)}]\n\n")
```
ljaljushkin, line 408:
```suggestion
        nncf_logger.info(f"\n[{new_line.join(str_bs)}]\n\n")
```

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
ljaljushkin: as well as in 261 line:
```python
 nncf_logger.warning(f"Width list for {node_name} "
                                    f"contains invalid values: {fixed_width_list}, {max_width}")
```

and line 265:
```python
 nncf_logger.warning(f"Max width for {node_name} is not aligned with pre-trained model")
```

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -1006,7 +1005,7 @@ def build(self, target_model: NNCFNetwork) -> ElasticWidthHandler:
             nodes = graph.get_nodes_by_metatypes([metatype])
             for node in nodes:
                 node_name = node.node_name
-                nncf_logger.info("Adding Dynamic Input Op for {} in scope: {}".format(metatype.name, node_name))
+                nncf_logger.info(f"Adding Dynamic Input Op for {metatype.name} in scope: {node_name}")
```
ljaljushkin, line 1008: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -1066,7 +1065,7 @@ def _create_elastic_conv_width_op(conv_layer_attrs: BaseLayerAttributes,
                                       params: ElasticWidthParams,
                                       fixed_width_list: Optional[List[int]] = None) -> ElasticOutputWidthConv2DOp:
         assert isinstance(conv_layer_attrs, ConvolutionLayerAttributes)
-        nncf_logger.info("Adding Dynamic Conv2D Layer in scope: {}".format(str(node_name)))
+        nncf_logger.info(f"Adding Dynamic Conv2D Layer in scope: {str(node_name)}")
```
ljaljushkin, line 1068: Propose moving to debug
@jpablomch

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -1080,7 +1079,7 @@ def _create_elastic_linear_width_op(linear_layer_attrs: BaseLayerAttributes,
         assert isinstance(linear_layer_attrs, LinearLayerAttributes)
         if fixed_width_list is None:
             fixed_width_list = []
-        nncf_logger.info("Adding Dynamic Linear Layer in scope: {}".format(str(node_name)))
+        nncf_logger.info(f"Adding Dynamic Linear Layer in scope: {str(node_name)}")
```
ljaljushkin, line 1082: Propose moving to debug
@jpablomch

ljaljushkin (changes requested):

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
jpablomch: @ljaljushkin Moving to debug works. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
jpablomch: @ljaljushkin Confirmed. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
jpablomch: @ljaljushkin Moving to debug works. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/search/evaluator.py
jpablomch: @ljaljushkin Moving to debug works for us. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/search/search.py
jpablomch: @ljaljushkin Moving to debug works for us. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_depth.py
jpablomch: LGTM

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
jpablomch: @ljaljushkin Moving to debug works for us. Thanks!

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
+print("This is an info-level log message")
+```
 
 For logging functions that expect a pattern-string (with %-placeholders) as
-their first argument: Always call them with a string literal (not an f-string!)
-as their first argument with pattern-parameters as subsequent arguments. Some
-logging implementations collect the unexpanded pattern-string as a queryable
-field. It also prevents spending time rendering a message that no logger is
-configured to output.
+their first argument - consider calling them with a string literal (not an f-string!)
```
AlexKoff88, line 739: I don't think that it is a good point and where we should guide and limit ourselves. I also wonder where it comes from. Some comparisons do not show significant gaps, e.g. https://blog.finxter.com/string-formatting-vs-format-vs-formatted-string-literal
I personally love f-strings and use them only.

nncf/__init__.py
```diff
@@ -22,28 +22,39 @@
...
-    ov_runtime = None
-    ov_pot = None
+    _LOADED_FRAMEWORKS["openvino"] = False
+
 
-if torch is None and tf is None and onnx is None and ov_runtime is None and ov_pot is None:
-    import warnings
-    warnings.warn("None of PyTorch, TensorFlow, ONNX, OpenVINO have been found. Please, install one of the frameworks")
+if not any(_LOADED_FRAMEWORKS.values()):
+    nncf_logger.error("Neither PyTorch, TensorFlow, ONNX or OpenVINO Python packages have been found in your Python "
```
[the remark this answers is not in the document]
AlexKoff88: I support. This is a normal practice.

examples/torch/common/utils.py
```diff
@@ -161,7 +161,7 @@ def configure_logging(sample_logger, config):
 
     nncf_log_file_handler = logging.FileHandler(osp.join(config.log_dir, NNCF_LOG_FILE_NAME))
     nncf_log_file_handler.setFormatter(logging.Formatter("%(levelname)s:%(name)s:%(message)s"))
-    from nncf.common.utils.logger import logger as nncf_logger
+    from nncf.common.logging import nncf_logger
```
AlexKoff88, line 164: Can we have `nncf_logger` imported from `nncf` namespace directly?

AlexKoff88:
@vshampor, can you post an example of new output in the case of quantization? PTQ or QAT is ok.

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
+
+DEBUG - for NNCF internal information that can be utilized during debug sessions.
+
+INFO - for good-to-know information such as progress bar or activity indicators, short summaries or effects of non-default, user-defined configuration on execution (i.e. which parts of the model were ignored due to application of "ignored_scopes" arguments).
+
+WARNING - for unexpected events during execution that the user should know about, or for non-obvious effects of the current configuration
+
+ERROR - for reporting failures that impair the functionality without causing a fatal exception
+
+CRITICAL - for logging information relevant to NNCF total failures. Currently not used since this functionality is instead achieved with the text in the exceptions.
```
MaximProshin, line 768: Do you see any case where it could be needed in the future? I would remove it if it's not used to avoid any confusion.

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
-configured to output.
+their first argument - consider calling them with a string literal (not an f-string!)
+as their first argument with pattern-parameters as subsequent arguments, if constructing the log message takes a long
+time or is otherwise hurtful to performance.
 
 ```python
-# Correct:
 import nncf
-from nncf.common.utils.logger import logger as nncf_logger
+from nncf.common.logging import nncf_logger
```
MaximProshin, line 745: I think it's required in each NNCF file so can we avoid the import every time?

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
+# Probably not OK:
+for i in range(1000000):
+    nncf_logger.info(f'Test message: {sum(range(10000000))}')
 ```
 
+Use proper logging levels (https://docs.python.org/3/library/logging.html#logging-levels) when printing out a message to the logger.
+
+DEBUG - for NNCF internal information that can be utilized during debug sessions.
+
+INFO - for good-to-know information such as progress bar or activity indicators, short summaries or effects of non-default, user-defined configuration on execution (i.e. which parts of the model were ignored due to application of "ignored_scopes" arguments).
```
MaximProshin, line 762: Can we add here or below the sentence like "Use the INFO level for messages useful for NNCF users rather than NNCF developers."?

MaximProshin:
@helena-intel , I remember you had some concerns about logging in nncf so please also take a look

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
+
+WARNING - for unexpected events during execution that the user should know about, or for non-obvious effects of the current configuration
```
KodiaqQ, line 764:
```suggestion
WARNING - for unexpected events during execution that the user should know about, or for non-obvious effects of the current configuration.
```

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,62 @@ long_string = textwrap.dedent('''\
...
+
+ERROR - for reporting failures that impair the functionality without causing a fatal exception
```
KodiaqQ, line 766:
```suggestion
ERROR - for reporting failures that impair the functionality without causing a fatal exception.
```

nncf/__init__.py
```diff
@@ -22,28 +22,39 @@
...
-    onnx = None
+    _LOADED_FRAMEWORKS["onnx"] = False
 
 try:
     import openvino.runtime as ov_runtime
     import openvino.tools.pot as ov_pot
 except ImportError:
-    ov_runtime = None
-    ov_pot = None
+    _LOADED_FRAMEWORKS["openvino"] = False
```
KodiaqQ, line 51: We have both `openvino` and `openvino_native`.
cc @l-bat

docs/styleguide/PyGuide.md
vshampor: https://stackoverflow.com/questions/54367975/python-3-7-logging-f-strings

I used the "consider" part and the "if" clause in the end of the sentence specifically to make sure that this recommendation is to be applied only where it is really necessary.

examples/torch/common/utils.py
vshampor: Actually, since I needed to import the logger to do the error logging in the `__init__.py`, it should already be possible to do so. However, the logger is for logging from the NNCF package code, not from the user code, so I don't think that this will bring much benefit. Let the users utilize their own logging approaches and not touch others (save for the `set_log_level` and related functions)

nncf/__init__.py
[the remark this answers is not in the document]
vshampor: You don't raise exceptions on imports in general if this is something you can control.
It's pretty easy to trigger this too - create a clean venv, install NNCF via `pip install .` and run `python -c "import nncf"`.

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/search/evaluator.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/search/search.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_depth.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: Done

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: Done

nncf/__init__.py
vshampor: As confirmed on the sync today, we will only have "openvino" and the native/POT distinction will be abstracted away.

examples/torch/common/utils.py
```diff
@@ -189,8 +189,10 @@ def create_code_snapshot(root, dst_path, extensions=(".py", ".json", ".cpp", ".c
 
 
 def print_args(config, logger=default_logger):
+    logger.info("\nConfiguration parameters:")
```
KodiaqQ, line 192: Somewhere we use double quotes, somewhere single quotes. It doesn't affect functionality (usually) and generally looks applicable. But the eye clings to it, in the PR, where we change the logging messages. I'm not advocating changing this, just FYI.

vshampor: Actually we have a specific style defined by the guideline, but I find it hard to enforce if we don't have a clean state to start from, and going through the code just to fix all the quotes, even automatically, has always been deprioritized.

vshampor:
Here's what a launch of `python examples/torch/classification/main.py --mode train -c examples/torch/classification/configs/quantization/mobilenet_v2_imagenet_int8.json --data /home/vshampor/work/data/imagenet -b 3 --epochs 0 --gpu-id 0` outputs right now in terms of logs. NNCF only outputs the lines that have `*level*:nncf:*msg*`, the rest comes from the sample code.
```
INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, tensorflow, onnx
INFO:nncf:Compiling and loading extensions for quantization...
INFO:nncf:Compiling and loading extensions for binarization...

Configuration parameters:
batch_size                 : 3
batch_size_init            : None
batch_size_val             : None
checkpoint_save_dir        : runs/mobilenet_v2_imagenet_int8/2022-12-09__12-08-00
compression                : {'algorithm': 'quantization', 'initializer': {'range': {'num_init_samples': 2560}}}
config                     : examples/torch/classification/configs/quantization/mobilenet_v2_imagenet_int8.json
cpu_only                   : False
current_gpu                : 0
dataset                    : None
dataset_dir                : /home/vshampor/work/data/imagenet
device                     : cuda:0
disable_compression        : False
dist_backend               : nccl
dist_url                   : tcp://127.0.0.1:8899
distributed                : False
epochs                     : 0
execution_mode             : single_gpu
gpu_id                     : 0
imshow_batch               : False
input_info                 : {'sample_size': [2, 3, 224, 224]}
intermediate_checkpoints_path: runs/mobilenet_v2_imagenet_int8/2022-12-09__12-08-00/intermediate_checkpoints
log_dir                    : runs/mobilenet_v2_imagenet_int8/2022-12-09__12-08-00
metrics_dump               : None
mixed_precision            : False
mlflow                     : <examples.torch.common.utils.SafeMLFLow object at 0x7f3816ce2dc0>
mode                       : ['train']
model                      : mobilenet_v2
multiprocessing_distributed: False
name                       : mobilenet_v2_imagenet_int8
nncf_config                : {'model': 'mobilenet_v2', 'pretrained': True, 'input_info': {'sample_size': [2, 3, 224, 224]}, 'num_classes': 1000, 'batch_size': 256, 'epochs': 5, 'optimizer': {'type': 'Adam', 'base_lr': 1e-05, 'schedule_type': 'multistep', 'steps': [5]}, 'compression': {'algorithm': 'quantization', 'initializer': {'range': {'num_init_samples': 2560}}}, 'log_dir': 'runs/mobilenet_v2_imagenet_int8/2022-12-09__12-08-00'}
num_classes                : 1000
optimizer                  : {'type': 'Adam', 'base_lr': 1e-05, 'schedule_type': 'multistep', 'steps': [5]}
pretrained                 : True
print_freq                 : 10
print_step                 : False
rank                       : 0
resuming_checkpoint_path   : None
save_freq                  : 5
seed                       : None
start_epoch                : 0
target_device              : None
tb                         : <torch.utils.tensorboard.writer.SummaryWriter object at 0x7f380c466940>
test_every_n_epochs        : 1
to_onnx                    : None
weights                    : None
workers                    : 4
world_size                 : 1


Loading model: mobilenet_v2
/home/vshampor/work/venv/lib/python3.8/site-packages/torchvision/models/_utils.py:208: UserWarning: The parameter 'pretrained' is deprecated since 0.13 and will be removed in 0.15, please use 'weights' instead.
  warnings.warn(
/home/vshampor/work/venv/lib/python3.8/site-packages/torchvision/models/_utils.py:223: UserWarning: Arguments other than a weight enum or `None` for 'weights' are deprecated since 0.13 and will be removed in 0.15. The current behavior is equivalent to passing `weights=MobileNet_V2_Weights.IMAGENET1K_V1`. You can also use `weights=MobileNet_V2_Weights.DEFAULT` to get the most up-to-date weights.
  warnings.warn(msg)
INFO:nncf:Collecting tensor statistics |█               | 94 / 854
INFO:nncf:Collecting tensor statistics |███             | 188 / 854
INFO:nncf:Collecting tensor statistics |█████           | 282 / 854
INFO:nncf:Collecting tensor statistics |███████         | 376 / 854
INFO:nncf:Collecting tensor statistics |████████        | 470 / 854
INFO:nncf:Collecting tensor statistics |██████████      | 564 / 854
INFO:nncf:Collecting tensor statistics |████████████    | 658 / 854
INFO:nncf:Collecting tensor statistics |██████████████  | 752 / 854
INFO:nncf:Collecting tensor statistics |███████████████ | 846 / 854
INFO:nncf:Collecting tensor statistics |████████████████| 854 / 854
INFO:nncf:BatchNorm statistics adaptation |█               | 74 / 667
INFO:nncf:BatchNorm statistics adaptation |███             | 148 / 667
INFO:nncf:BatchNorm statistics adaptation |█████           | 222 / 667
INFO:nncf:BatchNorm statistics adaptation |███████         | 296 / 667
INFO:nncf:BatchNorm statistics adaptation |████████        | 370 / 667
INFO:nncf:BatchNorm statistics adaptation |██████████      | 444 / 667
INFO:nncf:BatchNorm statistics adaptation |████████████    | 518 / 667
INFO:nncf:BatchNorm statistics adaptation |██████████████  | 592 / 667
INFO:nncf:BatchNorm statistics adaptation |███████████████ | 666 / 667
INFO:nncf:BatchNorm statistics adaptation |████████████████| 667 / 667
Statistics of the quantization algorithm:
+--------------------------------+-------+
|        Statistic's name        | Value |
+================================+=======+
| Ratio of enabled quantizations | 100   |
+--------------------------------+-------+

Statistics of the quantization share:
+----------------------------------+--------------------+
|         Statistic's name         |       Value        |
+==================================+====================+
| Symmetric WQs / All placed WQs   | 100.00 % (53 / 53) |
+----------------------------------+--------------------+
| Asymmetric WQs / All placed WQs  | 0.00 % (0 / 53)    |
+----------------------------------+--------------------+
| Signed WQs / All placed WQs      | 100.00 % (53 / 53) |
+----------------------------------+--------------------+
| Unsigned WQs / All placed WQs    | 0.00 % (0 / 53)    |
+----------------------------------+--------------------+
| Per-tensor WQs / All placed WQs  | 0.00 % (0 / 53)    |
+----------------------------------+--------------------+
| Per-channel WQs / All placed WQs | 100.00 % (53 / 53) |
+----------------------------------+--------------------+
| Placed WQs / Potential WQs       | 50.48 % (53 / 105) |
+----------------------------------+--------------------+
| Symmetric AQs / All placed AQs   | 100.00 % (64 / 64) |
+----------------------------------+--------------------+
| Asymmetric AQs / All placed AQs  | 0.00 % (0 / 64)    |
+----------------------------------+--------------------+
| Signed AQs / All placed AQs      | 43.75 % (28 / 64)  |
+----------------------------------+--------------------+
| Unsigned AQs / All placed AQs    | 56.25 % (36 / 64)  |
+----------------------------------+--------------------+
| Per-tensor AQs / All placed AQs  | 73.44 % (47 / 64)  |
+----------------------------------+--------------------+
| Per-channel AQs / All placed AQs | 26.56 % (17 / 64)  |
+----------------------------------+--------------------+

Statistics of the bitwidth distribution:
+--------------+---------------------+--------------------+--------------------+
| Num bits (N) | N-bits WQs / Placed |    N-bits AQs /    | N-bits Qs / Placed |
|              |         WQs         |     Placed AQs     |         Qs         |
+==============+=====================+====================+====================+
| 8            | 100.00 % (53 / 53)  | 100.00 % (64 / 64) | 100.00 % (117 /    |
|              |                     |                    | 117)               |
+--------------+---------------------+--------------------+--------------------+

```

nncf/torch/__init__.py
```diff
@@ -10,21 +10,18 @@
...
-    import warnings
-    warnings.warn("NNCF provides best results with torch=={bkc}, "
-                   "while current torch version is {curr} - consider switching to torch=={bkc}".format(
-         bkc=BKC_TORCH_VERSION,
-         curr=torch.__version__
-    ))
+    nncf_logger.warn(
+        f"NNCF provides best results with torch=={BKC_TORCH_VERSION}, "
+        f"while current torch version is {torch.__version__} - "
+        f"consider switching to torch=={BKC_TORCH_VERSION}")
```
helena-intel, line 24: I always find this warning kind of scary, wondering what might go wrong, but usually other PyTorch versions work fine and my understanding is they're expected to work, that the only reason this message is there is because this is the PyTorch version that is validated.

Consider switching is also a pretty big ask, not always possible. Wondering if the message could be something like f"NNCF is validated with torch=={BKC_TORCH_VERSION}, while the current torch version is {torch.__version__}. If you encounter issues, consider switching to torch=={BKC_TORCH_VERSION}"

nncf/__init__.py
```diff
@@ -22,28 +23,39 @@
...
 
-if torch is None and tf is None and onnx is None and ov_runtime is None and ov_pot is None:
-    import warnings
-    warnings.warn("None of PyTorch, TensorFlow, ONNX, OpenVINO have been found. Please, install one of the frameworks")
+from nncf.common.logging import nncf_logger
+if not any(_LOADED_FRAMEWORKS.values()):
+    nncf_logger.error("Neither PyTorch, TensorFlow, ONNX or OpenVINO Python packages have been found in your Python "
+                      "environment.\n"
+                      "Please install one of the supported frameworks above in order to use NNCF on top of it.")
```
helena-intel, line 58: Since you get a warning when you use an unsupported version of PyTorch or Tensorflow, would it be good to link to the installation guide that mentions this?

helena-intel:
> @helena-intel , I remember you had some concerns about logging in nncf so please also take a look

This is awesome! This is a much better default user experience. Also great to see the improved language in the messages in addition to the more logical loglevels.

nncf/torch/__init__.py
vshampor: Done

nncf/__init__.py
vshampor: Done

docs/styleguide/PyGuide.md
```diff
@@ -719,21 +719,64 @@ long_string = textwrap.dedent('''\
...
-configured to output.
+their first argument - consider calling them with a string literal (not an f-string!)
+as their first argument with pattern-parameters as subsequent arguments, if constructing the log message takes a long
+time or is otherwise hurtful to performance.
 
 ```python
-# Correct:
 import nncf
-from nncf.common.utils.logger import logger as nncf_logger
+from nncf.common.logging import nncf_logger
```
alexsu52, line 745: What reason of renaming `from nncf.common.utils.logger import logger as nncf_logger` -> `from nncf.common.logging import nncf_logger`? Honestly I don't see any value in it.

alexsu52 (approved):
Thanks for the PR! The output of the NNCF is more readable!

AlexKoff88:
> 

Thanks. I presume that the first part of the output is related to the sample only. 
I found that some information looks duplicated, for example, "Ratio of enabled quantization" (I can only guess what it is about) and "Statistics of the bitwidth distribution->N-bits Qs / Placed Qs". I suggest moving the first one into the debug logs as well.

Regarding the "Statistics of the quantization share". I also tend to think that it should be moved to debug because from the regular user standpoint it does not carry any meaningful information. 
So keeping only the "Statistics of the bitwidth distribution" looks reasonable to me.

I am fine with progress bars because they can provide any clue about the overall time required for the quantization, at least. But haveing a single progress meter would be better.

examples/torch/common/utils.py
AlexKoff88: Ok, I see. I make sense but I could not find `set_log_level` in the documentation. But it looks important from the debugging standpoint.

nncf/__init__.py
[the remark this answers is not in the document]
AlexKoff88: But what's the point of such an environment if nncf cannot work without any of the supported frameworks (openvino should be the case as well)?

vshampor:
> > 
> 
> Thanks. I presume that the first part of the output is related to the sample only. I found that some information looks duplicated, for example, "Ratio of enabled quantization" (I can only guess what it is about) and "Statistics of the bitwidth distribution->N-bits Qs / Placed Qs". I suggest moving the first one into the debug logs as well.
> 
> Regarding the "Statistics of the quantization share". I also tend to think that it should be moved to debug because from the regular user standpoint it does not carry any meaningful information. So keeping only the "Statistics of the bitwidth distribution" looks reasonable to me.
> 
> I am fine with progress bars because they can provide any clue about the overall time required for the quantization, at least. But haveing a single progress meter would be better.

The statistics output to the screen is not NNCF-owned in this case (by design), rather, it is done in the example code via a separate (think user-defined) logger, so it's the user who will decide on the log level for the output; we only return the string to print or log:
https://github.com/openvinotoolkit/nncf/blob/c6c8522d8ea66bc0fb2e3c44bcf8b03dcc053b6d/examples/torch/classification/main.py#L254-L256

docs/styleguide/PyGuide.md
vshampor: It is easier in IDE (PyCharm or other rich IDEs) to type `nncf_logger`, and then Ctrl+Enter to trigger automatic import of the symbol from the correct location since `nncf_logger` is a unique name.

nncf/__init__.py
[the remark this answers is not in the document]
helena-intel: One way that throwing an exception on import can go wrong is if you have an integration (like in Optimum) that checks if NNCF is installed and then does some imports. That's currently harmless, you may have NNCF installed in your environment for historic reasons, or because it came with openvino-dev (at some point). That doesn't mean that you are intentionally using it there. We can change the optimum integration - but it is just one example that shows how throwing an exception on import may have side effects.

IMO this is exactly what a warning with error level is meant for. An exception is generally expected the moment things go wrong, not because something is expected to go wrong at some point in the future.

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
jpablomch: Moving to debug works for the user. Thanks @ljaljushkin!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
jpablomch: @ljaljushkin Thanks! It is ok to throw exceptions.

AlexKoff88:
> > > 
> > 
> > 
> > Thanks. I presume that the first part of the output is related to the sample only. I found that some information looks duplicated, for example, "Ratio of enabled quantization" (I can only guess what it is about) and "Statistics of the bitwidth distribution->N-bits Qs / Placed Qs". I suggest moving the first one into the debug logs as well.
> > Regarding the "Statistics of the quantization share". I also tend to think that it should be moved to debug because from the regular user standpoint it does not carry any meaningful information. So keeping only the "Statistics of the bitwidth distribution" looks reasonable to me.
> > I am fine with progress bars because they can provide any clue about the overall time required for the quantization, at least. But haveing a single progress meter would be better.
> 
> The statistics output to the screen is not NNCF-owned in this case (by design), rather, it is done in the example code via a separate (think user-defined) logger, so it's the user who will decide on the log level for the output; we only return the string to print or log:
> 
> https://github.com/openvinotoolkit/nncf/blob/c6c8522d8ea66bc0fb2e3c44bcf8b03dcc053b6d/examples/torch/classification/main.py#L254-L256

AlexKoff88:
> > > > 
> > > 
> > > 
> > > Thanks. I presume that the first part of the output is related to the sample only. I found that some information looks duplicated, for example, "Ratio of enabled quantization" (I can only guess what it is about) and "Statistics of the bitwidth distribution->N-bits Qs / Placed Qs". I suggest moving the first one into the debug logs as well.
> > > Regarding the "Statistics of the quantization share". I also tend to think that it should be moved to debug because from the regular user standpoint it does not carry any meaningful information. So keeping only the "Statistics of the bitwidth distribution" looks reasonable to me.
> > > I am fine with progress bars because they can provide any clue about the overall time required for the quantization, at least. But haveing a single progress meter would be better.
> > 
> > 
> > The statistics output to the screen is not NNCF-owned in this case (by design), rather, it is done in the example code via a separate (think user-defined) logger, so it's the user who will decide on the log level for the output; we only return the string to print or log:
> > https://github.com/openvinotoolkit/nncf/blob/c6c8522d8ea66bc0fb2e3c44bcf8b03dcc053b6d/examples/torch/classification/main.py#L254-L256

Then it means that only the progress bars and some generic messages are visible by default. Make sense to me.

nncf/__init__.py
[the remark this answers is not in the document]
AlexKoff88: Thanks @helena-intel, this makes some sense. Let’s skip it for now. But even having openvino with nncf enables post-training scenarios. Single nncf without frameworks is useless.

AlexKoff88 (approved):

docs/styleguide/PyGuide.md
MaximProshin: Ok, I see that it's not just internal level but rather inherited from logging. Resolved

docs/styleguide/PyGuide.md
MaximProshin: ok, I thought maybe there is a nice way to do it to avoid import in every file.

MaximProshin (approved):

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: Will fix in a follow-up PR

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: Will fix in a follow-up PR

ljaljushkin (approved):

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: #1431

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
vshampor: #1431

ljaljushkin:
Discussed with Vasily and decided to close 2 open comments in a separate PR: #1431 
https://github.com/openvinotoolkit/nncf/pull/1418#discussion_r1043782184
https://github.com/openvinotoolkit/nncf/pull/1418#discussion_r1043780814

PR #1419, opened 2022-12-08
Removed specialized export for NAS models.

ljaljushkin:
### Changes

Removed specialized export for NAS models and rewrote assertions in nas tests.

When assertion happens on calculation of number of nodes it tries to output the whole expression (onnx proto), and sometimes it causes pytest to freeze:
```
E       assert 65 == 66
E        +  where 65 = len([input: "onnx::Conv_229"\n  output:...74,-\264\272\315\363O\273\373\302h<,\.....
```
Introduced a temporary variable for number of nodes to avoid the hang.

### Reason for changes

The custom export for NAS models was introduced as a temporary solution for a sporadic hang in NAS tests. Hang was root caused, and the reason is not export in eval mode, but issue with onnx expression on assertion.

### Related tickets

n/a

### Tests

pre-commit scope


tests/torch/nas/test_elastic_depth.py
```diff
@@ -204,25 +198,28 @@ def test_can_export_model_with_one_skipped_block_resnet18(tmp_path):
...
     onnx_resnet18_orig = load_model_fn(orig_onnx_model_path)
 
-    # count of node in skipped block  == 5
-    assert len(onnx_resnet18_orig.graph.node) == 69
-    assert len(onnx_resnet18_without_one_block.graph.node) == 67
+    # Temporary variable for number of nodes is required to avoid hang. When assertion happens on calculation of number
+    # of nodes it tries to output the whole expression (onnx node), and sometimes it causes pytest to freeze.
+    num_all_nodes = len(onnx_resnet18_orig.graph.node)
+    num_not_skipped_nodes = len(onnx_resnet18_without_one_block.graph.node)
+    assert num_all_nodes == 65
```
ljaljushkin, line 221: Number of nodes has changed, because NASExporter exported in training mode, but default PTExporter uses evaluation one.

jpablomch: Thanks for fixing this.

jpablomch (approved):
LGTM. Thanks!

vshampor (approved):

PR #1426, opened 2022-12-12
Removed `from_str` methods from Enum classes

ljaljushkin:
### Changes

Removed `from_str` methods from `Enum` classes

### Reason for changes

Enum has almost the same functionality in ctor. The difference is only in the more informative message that is not really needed.
Instead, there will be the following message:
```
ValueError: 'some_wrong_enum_value' is not a valid `DerivedEnumClass`
```

### Related tickets

n/a

### Tests

pre-commit scope

### DIFF
--- a/nncf/experimental/torch/search_building_blocks/search_blocks.py
+++ b/nncf/experimental/torch/search_building_blocks/search_blocks.py
@@ -117,14 +117,6 @@ class BuildingBlockType(Enum):
     FF = 'FF'
     Unknown = 'unknown'
 
-    @staticmethod
-    def from_str(config_value: str) -> 'BuildingBlockType':
-        if config_value == BuildingBlockType.MSHA.value:
-            return BuildingBlockType.MSHA
-        if config_value == BuildingBlockType.FF.value:
-            return BuildingBlockType.FF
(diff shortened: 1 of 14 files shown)


daniil-lyakhov (approved):

nncf/experimental/torch/search_building_blocks/search_blocks.py
```diff
@@ -117,14 +117,6 @@ class BuildingBlockType(Enum):
     FF = 'FF'
     Unknown = 'unknown'
 
-    @staticmethod
-    def from_str(config_value: str) -> 'BuildingBlockType':
-        if config_value == BuildingBlockType.MSHA.value:
-            return BuildingBlockType.MSHA
-        if config_value == BuildingBlockType.FF.value:
-            return BuildingBlockType.FF
-        return BuildingBlockType.Unknown
```
daniil-lyakhov, line 126: BTW in this case `from_str` never raise an error

vshampor (approved):

KodiaqQ (approved):

andrey-churkin (approved):

alexsu52 (approved):

PR #1430, opened 2022-12-14
[Torch] `replace_modules` fix

daniil-lyakhov:
### Changes

`replace_modules` now could modifies nested target operations

### Reason for changes

In case one will make a registered user module which contains other registered module (`nn.Conv2d` for example), user registered module WILL BE REPLACED by NNCFUserModule but nested registered module WILL REMAIN UNCHANGED. In case someone will decide to quantize such module `nncf_network.get_weighted_original_graph_nodes()` will return this internal not wrapped module and then model transformer will try to add preops to it, which will lead to runtime error.

### Tests

`test_custom_module_registering` is updated

### DIFF
--- a/tests/torch/test_nncf_network.py
+++ b/tests/torch/test_nncf_network.py
@@ -19,6 +19,7 @@ from pathlib import Path
 from typing import List
 
 import networkx as nx
+import torch.nn.functional as F
 import pytest
 import torch
 from torch import nn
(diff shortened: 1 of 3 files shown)


tests/torch/test_nncf_network.py
```diff
@@ -147,24 +148,36 @@ def test_weight_normed_modules_are_replaced_correctly():
     assert len(wrapped_conv._forward_pre_hooks) == 1
 
 
-@register_module()
```
alexsu52, line 150: Why did you do the following changes?
```
@register_module()
class RegisteredModuleOfUser(ModuleOfUser):
    pass
```

daniil-lyakhov: This is needed to check that non registered user module which calls function (like `F.conv2d`) isn't wrapped by NNCF. I forgot to add the check, will do shortly

ljaljushkin: I'd add clarification to the name of different modules: like `RegisteredModuleWithOperation`, `RegisteredModuleWithNestedModule`, `UnregisteredModuleWithOperation`

daniil-lyakhov:
As it was discussed with @vshampor, nested modules shouldn't be wrapped by NNCF, so this fix isn't valid

PR #1434, opened 2022-12-15
Use a fixed implementation of mobilenet-v2/v3 in tests

ljaljushkin:
### Changes

No dependency from `torchvision` in tests using mobilenet-v2/v3 models

### Reason for changes

`Mobilenet` is subject to modifications in almost each new version of `torchvision`. It makes tests fragile when change supported torch version.

### Related tickets

99051

### Tests

precommit tests
manual precommit tests on torch 1.8.2 and torch 1.9.1


daniil-lyakhov:
@ljaljushkin, in sota checkpoints test we still will be using torchvision version, precommit tests will be affected only, is it right?

vshampor:
Why do we use 0.9.2 as the target version for mobilenets sourcing? Should freeze the mobilenet code from the current torchvision version instead so that it is obvious from the PR that the references remain unchanged.

ljaljushkin:
> Why do we use 0.9.2 as the target version for mobilenets sourcing? Should freeze the mobilenet code from the current torchvision version instead so that it is obvious from the PR that the references remain unchanged.

This is what I've tried first. Unfortunately, latest `torchvision` has quite sophisticated implementation of mobilenet with deprecated API handling and by using some torchvision dependencies that are not available in previous versions.

That's why I've chosen an implementation that simple and has minimal dependency from torchvision utilities.

ljaljushkin:
> @ljaljushkin, in sota checkpoints test we still will be using torchvision version, precommit tests will be affected only, is it right?

SOTA checkpoints are updated for the latest validated `torchvision` version. They are not affected.

vshampor (approved):

daniil-lyakhov (approved):

alexsu52 (approved):
LGTM, Special thanks for changes in the .gitignore!

PR #1437, opened 2022-12-16
Adaptive Compression Training: removed exception when can't reach required accuracy drop

nikita-savelyevv:
### Changes

- Removed exception raise if during ACT we are not able to reach the required minimal accuracy drop and pruning rate.
- The algorithm now finishes returning the best model it succeeded to train. 
- Added statistics property to ACTL for ease of checking of the training results.

### Reason for changes

The algorithm shouldn't finish with a runtime error.

### Related tickets

98977

### Tests

Added a test `test_adaptive_compression_training_loop_failing` for a scenario when ACT does not succeed in obtaining the model meeting the requirements, but still finishes correctly.


examples/common/utils.py
```diff
@@ -0,0 +1,16 @@
+def print_maximal_degradation_warning(config, acc_aware_statistics, logger):
```
AlexanderDokuchaev, line 1: What you think about move it to runner?

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -351,6 +376,7 @@ def _run_initial_training_phase(self, model):
         self.runner.maximal_total_epochs = self.runner.initial_training_phase_epochs
         model = self._run_early_exit_training_loop(model)
         self.runner.maximal_total_epochs = maximal_total_epochs
+        self._final_model_compression_rate, self._final_model_accuracy = 0.0, 0.0
```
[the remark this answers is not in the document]
nikita-savelyevv: Done

examples/common/utils.py
nikita-savelyevv: Good point! Moved.

AlexanderDokuchaev (approved):

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -127,9 +134,26 @@ def _run_early_exit_training_loop(self, model):
 
             self.runner.update_learning_rate()
 
-        self.runner.load_best_checkpoint(model)
+        self._finish_early_exit_training_loop(model)
         return model
 
+    def print_maximal_degradation_warning(self, logger):
```
alexsu52, line 140: It looks like a user code. I would suggest returning the final stats and let the user decide how they want to proceed.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -142,6 +166,19 @@ def log_accuracy_statistics(uncompressed_model_accuracy,
         nncf_logger.info('Relative accuracy drop: {:.2f}%'.format(rel_accuracy_drop))
         nncf_logger.info('Accuracy budget: {:.4f}'.format(accuracy_budget))
 
+    @property
+    def final_statistics(self):
+        statistics_dict = {
```
alexsu52, line 171: Would you consider to reuse class based approach for the statistics form compression algorithms instead of dict?

nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
```diff
@@ -120,3 +120,5 @@ def configure_optimizers_fn():
                                                tensorboard_writer=tensorboard_writer,
                                                log_dir=log_dir)
     callbacks.on_train_end()
+
+    return acc_aware_training_loop
```
alexsu52, line 124: I would recommend to return a 'History' object with accuracy aware statistics  or accuracy statistics alone to be consistent with the fit method.

alexsu52 (changes requested):

nncf/common/accuracy_aware_training/training_loop.py
nikita-savelyevv: Done. I replaced call of this method with printing of final statistics after the loop finish. The idea is to demonstrate the user that there is such property as `final_statistics`.

nncf/common/accuracy_aware_training/training_loop.py
nikita-savelyevv: Done

nncf/tensorflow/accuracy_aware_training/keras_model_utils.py
nikita-savelyevv: Made it return the final statistics object

nikita-savelyevv:
run torch pre-commit tests

nikita-savelyevv:
run pytorch pre-commit tests

nncf/common/accuracy_aware_training/statistics.py
```diff
@@ -0,0 +1,35 @@
...
+ limitations under the License.
+"""
+
+from dataclasses import dataclass
+
+from nncf.api.statistics import Statistics
+
+
+@dataclass
+class TrainingLoopStatistics(Statistics):
```
alexsu52, line 20: Please, add a docstring

nncf/tensorflow/sparsity/magnitude/algorithm.py
```diff
@@ -230,6 +230,7 @@ def compression_rate(self, compression_rate: float) -> None:
 
     def disable_scheduler(self):
         self._scheduler = StubCompressionScheduler()
+        self._scheduler.target_level = 0.0
```
alexsu52, line 233: Looks like a crutch for `compression_stage` function. Would you suggest more general solution?

nikita-savelyevv: Yes, it is. It bugs me too. I have created a ticket to fix it properly some time ago: 96828. In short, in my opinion the whole approach with `StubCompressionScheduler` is a hack in some way. I think every scheduler should have the disabling/enabling functionality. Then, we won't need to replace scheduler within controller, and will be able to disable one already present.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -62,6 +62,13 @@ def run(self, model: TModel, train_epoch_fn, validate_fn, configure_optimizers_f
         :return: The fine-tuned model
         """
 
+    @property
+    @abstractmethod
+    def final_statistics(self):
```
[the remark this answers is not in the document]
nikita-savelyevv: At some point I had one property for non-final and final statistics, but then decided to separate them due to implementation complications and because only the `final_statistics` is really needed by the user. 

I have changed it now to a single property. Seems better.

nncf/common/accuracy_aware_training/runner.py
```diff
@@ -164,11 +164,21 @@ def initialize_logging(self, log_dir: Optional[Union[str, pathlib.Path]] = None,
         """
 
     @abstractmethod
-    def load_best_checkpoint(self, model: TModel) -> None:
+    def load_best_checkpoint(self, model: TModel) -> float:
         """
         Load the most accurate model state from the fine-tuning history.
 
         param model: The model object in which the state will be loaded.
+        return: Compression rate of the returned model or the current compression rate if no model was loaded
```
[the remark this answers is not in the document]
nikita-savelyevv: Done

nncf/common/accuracy_aware_training/statistics.py
nikita-savelyevv: Done

nikita-savelyevv:
run pytorch pre-commit tests

alexsu52 (approved):
Thanks for the contribution!

ljaljushkin (approved):

PR #1438, opened 2022-12-19
Restore logging imports and mark legacy import as deprecated

vshampor:
### Changes
As stated in the title

### Reason for changes
Some code seems to rely on the old import paths.

### Related tickets
N/A

### Tests
Manual (`from nncf.common.utils.logger import logger`, `from nncf.common.utils.logger import set_log_level` etc.)


nncf/internal/deprecation.py
```diff
@@ -0,0 +1,42 @@
+""""
```
alexsu52, line 1: Hello, new `utils` package :) What do you think about moving deprecation.py to nncf.common.utils?

alexsu52 (changes requested):

nncf/internal/__init__.py
```diff
@@ -0,0 +1,12 @@
+""""
```
AlexKoff88, line 1: Does the name `internal` mean that we have `external`? This name is not self-explainable and it is not better than `utils` or `helpers`.

nncf/internal/deprecation.py
```diff
@@ -0,0 +1,42 @@
...
+import warnings
+from typing import Callable
+
+
+def warning_deprecated(msg):
+    # Note: must use FutureWarning in order not to get suppressed by default
+    warnings.warn(msg, FutureWarning, stacklevel=2)
+
+
+class deprecated_call:
```
AlexKoff88, line 22: why it is not just `deprecated`? It can be applicable to classes and methods.

nncf/internal/deprecation.py
vshampor: This is internal code. Users are not supposed to touch this. I can make this even more obvious by calling the top-level module `_internal`, but then we will have problems with pylint, unless we ignore all `import-private-name` warnings.

nncf/internal/__init__.py
vshampor: > Does the name internal mean that we have external?
Not necessarily. Whatever is in `internal` right now is `internal`. Not everything that we have outside `internal` is "not internal", but now we have a way to make the situation better.

> This name is not self-explainable
You've guessed it yourself. `internal` is as oppossed to `external`, and as an explicit marker to not use the code in this module in external user scripts, because not only there are no guarantees of compatibility, there is also no real reason to use the code here anywhere outside NNCF.

nncf/internal/deprecation.py
vshampor: Will try to come up with an implementation that works for class instantiation as well.

nncf/internal/__init__.py
AlexKoff88: I want to split this PR into two where the first will address the backward compatibility problem for `logging` and we can continue this discussion.
@vshampor, can you create a separate PR where logger import is placed to its previous location?

vshampor: It is already there, in #1442

nncf/internal/deprecation.py
vshampor: Done, but the resulting decorator will only trigger a warning on instantiation of the class objects. Enums and static attributes of classes cannot be deprecated in this fashion.

nncf/internal/deprecation.py
alexsu52: First of all, we should think about is the friendly and intuitive modules structure in NNCF (https://github.com/openvinotoolkit/nncf/tree/develop/nncf), and if there is an opportunity not to add a new module, then It's better to do so.

vshampor:
@alexsu52 @AlexKoff88 please take another look at this PR - I retargeted and refactored it to not add the `internal` package for the time being and to only incrementally add the `@deprecated` decorator capabilities.

PR #1404, opened 2022-12-05
Add function check_config_matches_graph

AlexanderDokuchaev:
### Changes

Add `BaseCompressionAlgorithmController._check_scopes_in_graph(self, graph: NNCFGraph)` to check NNCF config matches model. 

In case of config contains contain unused patter the function raise RuntimeError:

```
E           RuntimeError: No match has been found among the model operations for the following ignored/target scope definitions:
E            - ignored_scope: ['unknown']
E           Refer to the original_graph.dot to discover the operations in the model currently visible to NNCF and specify the ignored/target scopes in terms of the names there.
```

### Related tickets

98455

### Tests

Added tests `test_raise_runtimeerror_for_not_matched_scope_names`

PR #1439, opened 2022-12-19
Add a paragraph on bad file names to the style guide

vshampor:


docs/styleguide/PyGuide.md
```diff
@@ -965,6 +965,11 @@ Always use a `.py` filename extension. Never use dashes.
 Python filenames must have a `.py` extension and must not contain dashes (`-`).
 This allows them to be imported and unit tested.
 
+Avoid file and module names of `utils`, `helpers` etc. and do not extend existing files and packages with these names.
```
AlexKoff88, line 968: I don't see this is an industry standard. Here are some examples:
- https://pytorch.org/docs/stable/data.html
- https://www.tensorflow.org/api_docs/python/tf/keras/utils/to_categorical
- https://github.com/huggingface/transformers/tree/main/utils

vshampor: In all of these examples the `utils` part can either be removed without loss of semantic meaning, or replaced to a more fitting name (e.g. the transformers example seems to contain mostly test and CI scripts and checkers - there are your two-three names (`tests`, `ci_checks`, `checks`) that could be used instead of `utils`.

AlexKoff88: PyTorch contains quite a lot of stuff in [utils](https://github.com/pytorch/pytorch/tree/master/torch/utils) moving to `torch` namespace will make it messy. The content of Keras also disapproves of your statement. 
How does `logging` fit the `common` namespace in your case? The name `common` should also fall into the rule you proposed in this PR.

vshampor: No, not really messier than it already is. I don't see how `torch.utils.jit.log_extract` from their code is better kept in `torch.utils.jit` rather than in the already existing `torch.jit`, and why `torch.jit._decomposition_utils` is not `torch.utils.jit._decomposition`. As for Keras - I'm sure that bad decisions can be found in any code base, Google's is no exception.

As mentioned, `common` is a descriptive name in a sense that it separates the backend-agnostic code and definitions from the backend-specific code. I can add a README.md file to that module like I did in `tests/common` and `tests/shared` to give a more documental definition for this.

AlexanderDokuchaev: Some point from https://maestros.io/whats-in-a-utils-file
> I came across to a utils.js -or something like that- it was a real headache. I had to guess what's in there.

I agree that avoid utils and helpers names is good practices. Put code to utils it's the easiest way, better find correct names that will display the functionality. But don't need to strictly follow this, otherwise there will be many small files.

AlexanderDokuchaev: For example my PR  https://github.com/openvinotoolkit/nncf/pull/1404/files, for me several function can be added. But it's can be already discussed.

docs/styleguide/PyGuide.md
```diff
@@ -965,6 +965,11 @@ Always use a `.py` filename extension. Never use dashes.
...
+Otherwise there is always a temptation to add an additional piece of code to this file/module, which ends up in having large files with unrelated functions and classes.
+Instead group your new code in dedicated files/modules that are named explicitly according to the purpose of code - either create new files/modules for this or extend the already existing, properly named ones.
+If you see an opportunity to break down a `utils.py` or `helpers.py` file into a better named and grouped set of files - take it.
```
alexsu52, line 971:
```suggestion
Avoid creating Python files like a "swiss knife" naming their with generic names such as `utils.py`, `helpers.py`, `common.py` and etc.
```

alexsu52 (changes requested):

docs/styleguide/PyGuide.md
AlexKoff88: > No, not really messier than it already is. I don't see how `torch.utils.jit.log_extract` from their code is better kept in `torch.utils.jit` rather than in the already existing `torch.jit`, and why `torch.jit._decomposition_utils` is not `torch.utils.jit._decomposition`. As for Keras - I'm sure that bad decisions can be found in any code base, Google's is no exception.
> 
> As mentioned, `common` is a descriptive name in a sense that it separates the backend-agnostic code and definitions from the backend-specific code. I can add a README.md file to that module like I did in `tests/common` and `tests/shared` to give a more documental definition for this.

`Common` can be descriptive for backend-agnostic code but logging is not related to backends at all. I recommend not mixing the repository folder layout and import layout because these are two different things. The former is our internal reasonable preferences while the latter is what and how we expose this to the user.
The initially proposed guidance does not make sense to me. @alexsu52, proposed a different rule of thumb in the style guide, and I support it.

vshampor: I have update the paragraph to integrate @alexsu52's wording, shortened it somewhat and added examples of good and bad layouts. I also kept a short sentence on what to do instead, so that the recommendation is not just a plain prohibition but also describes a course of action.

docs/styleguide/PyGuide.md
vshampor: Rewrote the paragraph, please re-review.

alexsu52: Let's handle `*.py` at the first.

alexsu52 (approved):
LGTM

PR #1441, opened 2022-12-19
[Torch][SOTA eval] Threshold for mobilenet_int4_int8 is updated

daniil-lyakhov:
### Changes

Threshold for mobilenet_v2_int4_int8 is increased

### Reason for changes

Quantized model has validation metric close to the threshold (-0.97%), any small deviation makes this model fall.
Model should be retrain in future to gain better validation metric

### Related tickets

99626


ljaljushkin (approved):

daniil-lyakhov:
PTSota eval build 153

alexsu52 (approved):

PR #1442, opened 2022-12-20
Restore logging imports and mark legacy import as deprecated

vshampor:
### Changes
As stated in the title

### Reason for changes
Some code seems to rely on the old import paths.

### Related tickets
N/A

### Tests
Manual (`from nncf.common.utils.logger import logger`, `from nncf.common.utils.logger import set_log_level` etc.)

### DIFF
--- a/nncf/common/utils/logger/__init__.py
+++ b/nncf/common/utils/logger/__init__.py
@@ -1,0 +1,26 @@
+"""
+ Copyright (c) 2022 Intel Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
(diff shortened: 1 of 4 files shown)


alexsu52 (approved):
LGTM

nncf/common/utils/logger/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+#pylint:disable=wrong-import-position
+from nncf.common.logging.logger import warning_deprecated
+
+warning_deprecated(
```
AlexKoff88, line 17: @MaximProshin, can we do such a deprecation considering that this is the LTS release of OpenVINO and, in fact, this rule should be spread to all its tools?

MaximProshin: There are no deprecation rules concerning LTS, only the recommendation that HW support should not be deprecated in LTS. Moreover, recent example with Python 3.6 shows that even removal is allowed. Anyway, I would say we should rely on whether a deprecation is really needed rather than on some rules. I personally don't have a strong opinion here. @vshampor , how is it important to state that this is deprecated? Can we just say it's not recommended?

AlexKoff88: If there is no problem with deprecation I would keep "deprecated" in the message because this is a common approach. No comments from my side.

AlexKoff88 (approved):

PR #1447, opened 2022-12-21
Adjust torch BKC version

vshampor:


tests/onnx/requirements.txt
```diff
@@ -1,7 +1,7 @@
 pytest-mock>=3.3.1
 pytest-dependency>=0.5.1
 pytest-ordering>=0.6
-torch==1.13.0
-torchvision==0.14.0
+torch~=1.13.0
```
ljaljushkin, line 4: @kshpv please take a look

ljaljushkin (approved):

tests/onnx/requirements.txt
kshpv, line 4: @vshampor why did you update torch version in tests? What if any bugs will be added in a new release version of torch? Moreover if they update the models inside torchvision?
I would prefer to keep it strict (==)

kshpv (changes requested):

tests/onnx/requirements.txt
vshampor: Patch version updates tend to fix bugs, not add them.

kshpv: Ok, since the 1.13 has the security issues let's update it.

kshpv (approved):

PR #1449, opened 2022-12-21
Extending makefile for OpenVINO, TensorFlow and PyTorch backends

alexsu52:
### Changes

Extending makefile for OpenVINO, TensorFlow and PyTorch backends

### Reason for changes

Simplify validation process

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 PYLINT_VERSION := 2.13.9
 JUNITXML_PATH ?= nncf-tests.xml
-DATA_ARG := ''
 
 ifdef DATA
 	DATA_ARG := --data $(DATA)
@@ -42,6 +41,8 @@ install-openvino-test:
 	pip install -e .[openvino]
 	pip install -r tests/openvino/requirements.txt
 	pip install -r tests/cross_fw/install/requirements.txt
+	pip install -r examples/experimental/openvino/bert/requirements.txt
+	pip install -r examples/experimental/openvino/yolo_v5/requirements.txt
 
 install-openvino-dev: install-openvino-test
 	pip install pylint==$(PYLINT_VERSION)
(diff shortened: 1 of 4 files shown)


vshampor:
`-dev` versions should probably install examples too.

Makefile
```diff
@@ -1,23 +1,117 @@
-install-onnx-dev:
+PYLINT_VERSION := 2.13.9
```
AlexKoff88, line 1: a widely used practice in Python repos is `make style` that fixes code structure and imports automatically. It is based on black and isort. Can we have the same?
Here is an example:
```
style_check:
	black --check .
	isort --check .

style:
	black .
	isort .
```

vshampor: @AlexKoff88 that it is used in transformers doesn't automatically mean that this is a widely used practice. 

It would, however, be nice to have a similar functionality. I considered adding `black` before, but the resulting style is pretty much unconfigurable and goes against our current code style at least in terms of used string quoting symbols. Furthermore, `black` and `isort` frequently [conflict with each other](https://github.com/PyCQA/isort/issues/1518), and also running black/isort does not guarantee that pylint will then pass.

I suggest to keep style out of this PR and review the options for style improvements at a later date.

alexsu52:
> `-dev` versions should probably install examples too.

You are absolutely right. Thanks for the comment.

Makefile
alexsu52: @AlexKoff88, @vshampor  I will try  to cover this aspect in another PR.

Makefile
```diff
@@ -1,23 +1,128 @@
...
+
+install-torch-dev: install-torch-test
+	pip install pylint==$(PYLINT_VERSION)
+
+test-torch:
+	pytest tests/common tests/torch --junitxml ${JUNITXML_PATH} $(DATA_ARG)
+
+pylint-torch:
+	pylint --rcfile .pylintrc               \
+		nncf/common                         \
```
ljaljushkin, line 109: Have you checked the output of the pylint in that way?
It's a bit different from what we have in CI now, there might be some additional pylint warnings.

ljaljushkin (approved):

Makefile
alexsu52: Yes, I checked locally and everything is fine.

alexsu52:
run pytorch pre-commit tests

PR #1450, opened 2022-12-21
Added upper bound for maximal compression rate of ACTL

nikita-savelyevv:
- Added `maximal_compression_rate` property for compression controllers. It returns the maximal compression rate that controller can compress the model.
- Added an upper bound for maximal compression rate Adaptive Compression Training Loop may try to reach
- Added some related tests

### Reason for changes

During ACTL, the model compression rate may get arbitrary high. But not all of the compression controllers support that, e.g. pruning controller with FLOPS target. This lead to a runtime error.

### Related tickets

98894

### Tests

Added the following tests:
- a test for correctness of computation of `compression_controller.maximal_compression_rate` (PT + TF)
- a test for ACTL where maximal compression rate is reached and algorithm still finished correctly


nikita-savelyevv:
run openvino pre-commit tests

nncf/api/compression.py
```diff
@@ -315,6 +315,13 @@ def disable_scheduler(self) -> None:
         it to a dummy one that does not change the compression rate.
         """
 
+    @property
+    @abstractmethod
+    def maximal_compression_rate(self) -> float:
+        """
+        Returns the maximal model compression rate supported by the compression controller
```
daniil-lyakhov, line 322:
```suggestion
        Returns the maximal model compression rate supported by the compression controller.
```

nncf/tensorflow/pruning/filter_pruning/algorithm.py
```diff
@@ -240,6 +240,12 @@ def set_pruning_level(self, pruning_level: float,
         if run_batchnorm_adaptation:
             self._run_batchnorm_adaptation()
 
+    @property
+    def maximal_compression_rate(self) -> float:
+        max_flops, _ = self._calculate_flops_and_weights_in_uniformly_pruned_model(1.0)
+        max_compression_rate = 1 - max_flops / max(self.full_flops, 1)
+        return max_compression_rate
```
daniil-lyakhov, line 247: 1) `max_flops` is a bit confusing name because it is minimum possible flops for the model after the pruning. `min_flops` is clearer in my opinion
2) `max(self.full_flops, 1)` is needed for a case when a model has 0 flops, right?

daniil-lyakhov:
By the way: according to `get_rounded_pruned_element_number` in case sparsity rate is 1. it always returns `total` elements even if it's not a multiply of 8. Is it bug or feature? @alexsu52

nncf/tensorflow/pruning/filter_pruning/algorithm.py
nikita-savelyevv: 1. I see. If I change, it will be inconsistent with this part [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/tensorflow/pruning/filter_pruning/algorithm.py#L143) . Should it be changed it to `min_` in both places?
2. Yes

daniil-lyakhov: 1) Yes
2) 👍

nikita-savelyevv: Following that trail, decided to put maximal compression rate computation inside existing `PrunedModelTheoreticalBorderline` where it seems to belong

nncf/tensorflow/pruning/filter_pruning/algorithm.py
```diff
@@ -240,6 +240,13 @@ def set_pruning_level(self, pruning_level: float,
         if run_batchnorm_adaptation:
             self._run_batchnorm_adaptation()
 
+    @property
+    def maximal_compression_rate(self) -> float:
+        stats = PrunedModelTheoreticalBorderline(
+            self._pruned_layers_num, self._prunable_layers_num, self._min_prunable_flops,
+            self._min_prunable_params, self.full_flops, self.full_params_num)
+        return stats.maximal_pruning_rate
```
daniil-lyakhov, line 248: I suggest to calculate `maximal_pruning_rage` explicitly. Copy paste will be unified in future when common pruning algorithm builder/controller will be introduced.

nikita-savelyevv: I agree, this is more clean

nncf/torch/pruning/filter_pruning/algo.py
```diff
@@ -613,6 +613,11 @@ def disable_scheduler(self):
         self._scheduler = StubCompressionScheduler()
         self._scheduler.current_pruning_level = 0.0
 
+    @property
+    def maximal_compression_rate(self) -> float:
+        max_compression_rate = 1 - self._min_prunable_flops / max(self.full_flops, 1) if self.prune_flops else 1.0
+        return max_compression_rate
```
daniil-lyakhov, line 619: IMHO
```suggestion
        if self.prune_flops:
            return 1.0
        return 1 - self._min_prunable_flops / max(self.full_flops, 1)
```

daniil-lyakhov (approved):

nncf/torch/pruning/filter_pruning/algo.py
nikita-savelyevv: 👍

daniil-lyakhov (approved):

ljaljushkin (changes requested):
Why not make it more consistent and implement `maximal_compression_rate` for all other algorithms that have overridden `compression_rate` method?
I think this method is needed for `magnitude`, `rb` and `movement` sparsity algorithms.

ljaljushkin:
> Why not make it more consistent and implement `maximal_compression_rate` for all other algorithms that have overridden `compression_rate` method? I think this method is needed for `magnitude`, `rb` and `movement` sparsity algorithms.

maximal_sparsity_rate = 1 should be ok for sparsity algorithms.

ljaljushkin (approved):

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -209,6 +209,11 @@ def __init__(self,
             raise RuntimeError('No compression algorithm supported by the accuracy-aware training '
                                'runner was specified in the config')
 
+        if maximal_compression_rate > self.adaptive_controller.maximal_compression_rate:
+            maximal_compression_rate = self.adaptive_controller.maximal_compression_rate
+            nncf_logger.warning(f'Updated training loop maximal compression rate by the maximal compression rate '
```
alexsu52, line 214: Probably, the maximum compression rate in 95% can not be archived for most models. I mean that the user will see warning if maximal_compression_rate will not defined and use default value 0.95. It can be confuse an user. I would recommend to handle it during the accuracy aware loop.

nncf/common/pruning/statistics.py
```diff
@@ -161,26 +161,26 @@ class PrunedModelTheoreticalBorderline(Statistics):
     def __init__(self,
                  num_pruned_layers: int,
                  num_prunable_layers: int,
-                 max_prunable_flops: float,
-                 max_prunable_params: float,
+                 min_prunable_flops: float,
+                 min_prunable_params: float,
```
alexsu52, line 165: Seems like `max_prunable_flops` makes more sense. 

But you can rename them as following, for example:
```suggestion
                 min_flops: float,
                 min_params: float,
```

nncf/tensorflow/pruning/filter_pruning/algorithm.py
```diff
@@ -240,6 +240,12 @@ def set_pruning_level(self, pruning_level: float,
         if run_batchnorm_adaptation:
             self._run_batchnorm_adaptation()
 
+    @property
+    def maximal_compression_rate(self) -> float:
+        if self.prune_flops:
+            return 1 - self._min_prunable_flops / max(self.full_flops, 1)
+        return 1.0
```
alexsu52, line 247: This is not correct statement. The pruning algorithm can not prune all model parameters.

alexsu52 (changes requested):

nncf/api/compression.py
```diff
@@ -315,6 +315,13 @@ def disable_scheduler(self) -> None:
         it to a dummy one that does not change the compression rate.
         """
 
+    @property
+    @abstractmethod
+    def maximal_compression_rate(self) -> float:
```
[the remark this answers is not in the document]
ljaljushkin: Since accuracy aware algorithm uses common compression interface, we can't introduce it for specific algorithms and not introduce for others.
`maximal_compression_rate`=1 is valid for sparsity, it means there's no technical limitation to compress everything.

The only concern is that nobody needs maximal compression rate equal to 1. This model will not make any sense.
Probably we should call it `upper_bound_for_compression_rate` or something like that.

@nikita-savelyevv @alexsu52 What do you think about another approach when there's no `maximal_compression_rate`, but everything is handled by set_compression_rate?
We can return a specific error or exception that means we set more than the algorithm can really compress.

nncf/tensorflow/pruning/filter_pruning/algorithm.py
nikita-savelyevv: Yes. But at the same time currently it is possible to set any (channel) pruning rate and it will not throw an exception to it. I think this logic needs to be implemented first in order to support the behavior of the property value you suggest.

nncf/api/compression.py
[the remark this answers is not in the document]
ljaljushkin: @daniil-lyakhov @alexsu52, should `maximal_compression_rate` be different when some layers are in ignored scope for filter pruning by channels and for sparsity?

nncf/common/accuracy_aware_training/training_loop.py
nikita-savelyevv: Agree. Let's keep only the logging during the loop itself.

nncf/common/pruning/statistics.py
nikita-savelyevv: `max_prunable_flops` is not quite correct, since this value represents the minimal FLOPS we are able to reach with the maximum compression rate. I don't have problems with renaming it to `min_flops` though.

nncf/api/compression.py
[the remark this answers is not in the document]
nikita-savelyevv: In an offline discussion, it was decided to keep the introduction of `maximal_compression_rate` property as it may be useful in the future. The property should also be implemented for the magnitude sparsity compression algorithm, because for it, the property is not equal to 1.0 by default.

nncf/tensorflow/sparsity/magnitude/algorithm.py
```diff
@@ -190,7 +190,7 @@ def _select_threshold(self, sparsity_level):
         if not all_weights:
             return 0.0
         all_weights_tensor = tf.sort(tf.concat(all_weights, 0))
-        index = int(tf.cast(tf.size(all_weights_tensor), all_weights_tensor.dtype) * sparsity_level)
+        index = int(tf.cast(tf.size(all_weights_tensor) - 1, all_weights_tensor.dtype) * sparsity_level)
```
nikita-savelyevv, line 193: Fixed a minor bug. Not consistent with PT [implementation ](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/sparsity/magnitude/algo.py#L145)

nncf/common/pruning/statistics.py
alexsu52: Possible, it depends on an interpretation. My opinion is the max prunable flops is number of flops when prunable flops were pruned to their maximum.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -209,6 +209,9 @@ def __init__(self,
             raise RuntimeError('No compression algorithm supported by the accuracy-aware training '
                                'runner was specified in the config')
 
+        if maximal_compression_rate > self.adaptive_controller.maximal_compression_rate:
+            maximal_compression_rate = self.adaptive_controller.maximal_compression_rate
+
```
alexsu52, line 214:
```suggestion
        maximal_compression_rate = min(
            maximal_compression_rate, 
            self.adaptive_controller.maximal_compression_rate)
```

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -209,6 +209,9 @@ def __init__(self,
             raise RuntimeError('No compression algorithm supported by the accuracy-aware training '
                                'runner was specified in the config')
 
+        if maximal_compression_rate > self.adaptive_controller.maximal_compression_rate:
+            maximal_compression_rate = self.adaptive_controller.maximal_compression_rate
```
alexsu52, line 213: Just to clarify, AdaptiveCompressiontrainingLoop is ablle to reach the `maximal_compression_rate `?

nncf/tensorflow/sparsity/magnitude/algorithm.py
```diff
@@ -256,6 +256,31 @@ def compression_stage(self) -> CompressionStage:
             return CompressionStage.FULLY_COMPRESSED
         return CompressionStage.PARTIALLY_COMPRESSED
 
+    @property
+    def maximal_compression_rate(self) -> float:
```
alexsu52, line 260: Could you explain why this implementation is consistent with "compression_rate"?

nncf/common/accuracy_aware_training/training_loop.py
nikita-savelyevv: If model compresses successfully, it will reach `(maximal_compression_rate - compression_rate_step)` compression rate, where `compression_rate_step` is the last compression rate change step of the algo. For example, if the compression rate is only increasing then the algorithm will exit at `(maximal_compression_rate - 0.1)` compression rate by default.

nncf/common/pruning/statistics.py
nikita-savelyevv: Renamed to `min_possible_*`. It is a good candidate and the variables are already used with that name inside `PrunedModelTheoreticalBorderline`.

nncf/common/accuracy_aware_training/training_loop.py
nikita-savelyevv: Done

nncf/tensorflow/sparsity/magnitude/algorithm.py
nikita-savelyevv: Decided to return `maximal_compression_rate` of 1.0 for sparsity compression algorithm

PR #1458, opened 2022-12-22
Renamed pruning_rate to pruning_level for TF pruning controller

nikita-savelyevv:
Renamed `pruning_rate` to `pruning_level` for TF pruning controller

### Reason for changes

Everywhere in the codebase `pruning_level` is used for compression rate of pruning controller except multiple places in TF implementation of pruning controller

### DIFF
--- a/docs/compression_algorithms/Pruning.md
+++ b/docs/compression_algorithms/Pruning.md
@@ -40,7 +40,7 @@ The zeroed filters are frozen afterwards and the remaining model parameters are 
 **Exponential scheduler**
 
 Similar to the Baseline scheduler, during `num_init_steps` epochs model is pretrained without pruning.
-During the next `pruning steps` epochs `Exponential scheduler` gradually increasing pruning level from `pruning_init` to `pruning_target`. After each pruning training epoch pruning algorithm calculates filter importances for all convolutional filters and prune (setting to zero) `current_pruning_rate` part of filters with the smallest importance in each Convolution.  After `num_init_steps` + `pruning_steps` epochs algorithm with zeroed filters is frozen and remaining model parameters only fine-tunes.
+During the next `pruning steps` epochs `Exponential scheduler` gradually increasing pruning level from `pruning_init` to `pruning_target`. After each pruning training epoch pruning algorithm calculates filter importances for all convolutional filters and prune (setting to zero) `current_pruning_level` part of filters with the smallest importance in each Convolution.  After `num_init_steps` + `pruning_steps` epochs algorithm with zeroed filters is frozen and remaining model parameters only fine-tunes.
 
 Current pruning level $P_{i}$ (on i-th epoch) during training calculates by equation:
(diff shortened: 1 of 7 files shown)


alexsu52 (commented):
Could you take a look at other places in the code where `pruning_rate` is used: For example: https://github.com/openvinotoolkit/nncf/blob/cc3b34aab39061791afae5c2552fcd339b139c4e/nncf/torch/pruning/filter_pruning/algo.py#L606

nikita-savelyevv:
> Could you take a look at other places in the code where `pruning_rate` is used: For example:
> 
> https://github.com/openvinotoolkit/nncf/blob/cc3b34aab39061791afae5c2552fcd339b139c4e/nncf/torch/pruning/filter_pruning/algo.py#L606

Do you mean we should replace it with `pruning_level` in all other places?

alexsu52:
> > Could you take a look at other places in the code where `pruning_rate` is used: For example:
> > https://github.com/openvinotoolkit/nncf/blob/cc3b34aab39061791afae5c2552fcd339b139c4e/nncf/torch/pruning/filter_pruning/algo.py#L606
> 
> Do you mean we should replace it with `pruning_level` in all other places?

What I mean is to take a look at another part of the code and check that it is legal to use `pruning_rate` in another part of the code.

nikita-savelyevv:
> What I mean is to take a look at another part of the code and check that it is legal to use `pruning_rate` in another part of the code.

In my opinion it would be better to rename `pruning_rate` to `pruning_level` everywhere for consistency sake, because in the most cases it is referred to as `pruning_level` at the moment. 

At the same time however, I think `pruning_rate` is a better name for it because it fits well together with `compression_rate`: `compression_rate` is a general compression backend -agnostic name and `pruning_rate` is how `compression_rate` is referred to when pruning backend is used. But I don't know the reasons why it was decided to make it `pruning_level` despite this so I am not confident enough to rename it all back to `pruning_rate`.

alexsu52:
> > What I mean is to take a look at another part of the code and check that it is legal to use `pruning_rate` in another part of the code.
> 
> In my opinion it would be better to rename `pruning_rate` to `pruning_level` everywhere for consistency sake, because in the most cases it is referred to as `pruning_level` at the moment.
> 
> At the same time however, I think `pruning_rate` is a better name for it because it fits well together with `compression_rate`: `compression_rate` is a general compression backend -agnostic name and `pruning_rate` is how `compression_rate` is referred to when pruning backend is used. But I don't know the reasons why it was decided to make it `pruning_level` despite this so I am not confident enough to rename it all back to `pruning_rate`.

Thank you for your comment! First, the naming must be consistent. Let's do it and then we'll discuss renaming.

alexsu52 (approved):
LGTM

ljaljushkin (approved):

daniil-lyakhov (approved):

PR #1460, opened 2022-12-22
[Torch][Microfix] PTTarget point translation fix

daniil-lyakhov:
### Changes

Type checking fixed in insertion type translation inside `PTInsertionPoint`

### Reason for changes

In case `PTInsertionType` is passed to `PTInsertionPoint` and its id will match `InsertionType` id, it will be translated in a wrong way

### DIFF
--- a/nncf/torch/nncf_network.py
+++ b/nncf/torch/nncf_network.py
@@ -126,7 +126,8 @@ class PTInsertionPoint:
     }
 
     def _get_pt_insertion_type(self, target_type: TargetType) -> PTInsertionType:
-        if target_type not in PTInsertionPoint.TARGET_TYPE_VS_PT_INSERTION_TYPE_DICT:
+        if not isinstance(target_type, TargetType) or\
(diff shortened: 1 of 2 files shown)


nncf/torch/nncf_network.py
```diff
@@ -126,7 +126,8 @@ class PTInsertionPoint:
     }
 
     def _get_pt_insertion_type(self, target_type: TargetType) -> PTInsertionType:
-        if target_type not in PTInsertionPoint.TARGET_TYPE_VS_PT_INSERTION_TYPE_DICT:
+        if not isinstance(target_type, TargetType) or\
```
ljaljushkin, line 129: Looks as some logical error: typehint says it is TargetType, but later in the function we check that it isn't instance of it.

daniil-lyakhov: It's just suggest to pass `TargetType`, but I accidentally put `PTInsertionType` and it didn't brake down but occasionally  made new  `PTInsertionType` from feeded  `PTInsertionType`

ljaljushkin (approved):

PR #1462, opened 2022-12-23
Fixed again the install test for TensorFlow

alexsu52:
### Changes

Fixed path to TensorFlow test requirements.txt

### Reason for changes

The test requirements install command failed with the following error:

```
ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/home/jenkins/agent/workspace/NNCF/nightly/ubuntu20_install_tf/src/tests/tf/requirements.txt'
```

### Related tickets

N/A

### Tests

install_tf 144

### DIFF
--- a/tests/cross_fw/install/test_install.py
+++ b/tests/cross_fw/install/test_install.py
@@ -121,7 +121,8 @@ class TestInstall:
             pytest.xfail('Disabled until NNCF with torch version supporting CUDA 11.6 backend is exposed in a release')
         venv_path = create_venv_with_nncf(tmp_path, package_type, venv_type, extra_reqs={backend})
         pip_with_venv = f'. {venv_path}/bin/activate && {venv_path}/bin/pip'
-        subprocess.call(
-            f'{pip_with_venv} install -r {PROJECT_ROOT}/tests/{backend}/requirements.txt',
+        backend_name = 'tensorflow' if backend == 'tf' else backend            
+        subprocess.check_call(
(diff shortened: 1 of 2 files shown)


tests/cross_fw/install/test_install.py
```diff
@@ -121,7 +121,8 @@ def test_install_with_tests_requirements(tmp_path: Path,
             pytest.xfail('Disabled until NNCF with torch version supporting CUDA 11.6 backend is exposed in a release')
         venv_path = create_venv_with_nncf(tmp_path, package_type, venv_type, extra_reqs={backend})
         pip_with_venv = f'. {venv_path}/bin/activate && {venv_path}/bin/pip'
-        subprocess.call(
-            f'{pip_with_venv} install -r {PROJECT_ROOT}/tests/{backend}/requirements.txt',
+        backend_name = 'tensorflow' if backend == 'tf' else backend
```
ljaljushkin, line 124: Why do we have such a duality in the backend name? 
In which cases is it `tensorflow` and in which is `tf`?

alexsu52: Historically, 'tf' is extra parameter for installation via pip but all folders are named as `tensorflow`.

ljaljushkin: Backend should be always `tf`. Why do you need `else`?
https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/install/conftest.py#L16

ljaljushkin (approved):

PR #1440, opened 2022-12-19
FastBC for OpenVINO backend

KodiaqQ:
### Changes

- Added OpenVINO backend support for the FastBC algorithm

### Reason for changes

- Extension of the set of algorithms for the backend

### Related tickets

- 98621

### Tests

- Added transformation unit-test

PR #1468, opened 2022-12-29
Full BiasCorrection for OpenVINO backend

KodiaqQ:
### Changes

- Added OpenVINO backend support for the BiasCorrection algorithm

### Reason for changes

- Extension of the set of algorithms for the backend

### Related tickets

- 98621

### Tests

- Added in [1440](https://github.com/openvinotoolkit/nncf/pull/1440)


alexsu52 (commented):
Could you provide comparison of yours and POT FullBiasCorrection implementation? I think it can be done on the models from the post training quantization conformance test.

PR #1473, opened 2023-01-02
FLOPS calculation for matmul operation with 4D inputs in NAS

ljaljushkin:
### Changes

This PR proposes how to support FLOPS calculation for (`matmul`, `bmm`) represented as torch functions on two tensors, not by module with weights. It's not a final solution, subject to discussion and modifications.

Note, there are many limitations: 
- All other operations associated with `matmul`  are not supported: `torch.addmm`, `torch.nn.functional.Linear`, `torch.mm`. 
- covered inputs of the same shape, but matmul also supports: 1D x 2D, 2D x 1D, 1D x ND, ND x 1D https://pytorch.org/docs/stable/generated/torch.matmul.html
- Filter pruning is not covered. Changes were made and tested for NAS part.
- FLOPS calculation was verified for FP32 model only. FLOPS of the pruned model is not verified.
- matmul, which takes learnable parameter on second input, is not supported (GNMT case) 
- `t.matmul(t)` is not supported

### Reason for changes

Currently, NNCF for Torch supports FLOPS calculation for modules only (e.g. `nn.Linear` or `nn.Conv2d`), because all required layer attributes are collected for modules now.
However, modern architectures (e.g., Transformers) are constructed not from modules only. Usually there is a multiplication of two activations represented as `torch.matmul` operation. For such operations, there was no extraction of layer attributes that is needed for FLOPS calculation.

### Related tickets

n/a

### Tests

`test_flops_and_weights_calculation`

### DIFF
--- a/nncf/common/graph/layer_attributes.py
+++ b/nncf/common/graph/layer_attributes.py
@@ -125,6 +125,15 @@ class LinearLayerAttributes(WeightedLayerAttributes):
         return 0
 
 
+class MatmulAttributes(BaseLayerAttributes):
+    def __init__(self,
+                 in_features: int,
+                 out_features: int):
+        super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
+
+
 class ConvolutionLayerAttributes(WeightedLayerAttributes):
     """
     This class stores attributes of convolution modules/layers
--- a/nncf/torch/dynamic_graph/wrappers.py
+++ b/nncf/torch/dynamic_graph/wrappers.py
@@ -173,7 +173,7 @@ def wrap_module_call(module_call):
 def _add_operation_attributes(node: 'DynamicGraphNode', ctx: TracingContext):
     op_address = node.op_exec_context.op_address
     # TODO(nlyalyus): support other operations: torch.addmm, torch.nn.functional.Linear, torch.bmm, torch.mm
-    if op_address.operator_name == 'matmul':
+    if op_address.operator_name in ('matmul', 'bmm'):
         graph = ctx.graph
         input_edges = graph.get_input_edges(node)
         # no input nodes when input_info tensor was bound to a *args or **kwargs on graph tracing
@@ -182,7 +182,7 @@ def _add_operation_attributes(node: 'DynamicGraphNode', ctx: TracingContext):
             activation_edge, weight_edge = input_edges
             activation_shape = activation_edge.activation_shape
             weight_shape = weight_edge.activation_shape
-            # TODO(nlyalyus): support other shapes: 1D x 1D, 1D x 2D, 2D x 1D, 1D x ND, ND x 1D
+            # TODO(nlyalyus): support other shapes: 1D x 2D, 2D x 1D, 1D x ND, ND x 1D
             # https://pytorch.org/docs/stable/generated/torch.matmul.html
             activation_dims = len(activation_shape)
             weight_dims = len(weight_shape)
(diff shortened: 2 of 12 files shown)


nncf/common/graph/layer_attributes.py
```diff
@@ -125,6 +125,15 @@ def get_target_dim_for_compression(self) -> int:
         return 0
 
 
+class MatmulAttributes(BaseLayerAttributes):
```
ljaljushkin, line 128: Introduced new class for `matmul` specifically, since `LinearLayerAttrbiutes` was designed for modules with weights and biases.

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -193,6 +214,8 @@ def _execute_op(op_address: 'OperationAddress',
             is_called_inside_nncf_module = isinstance(ctx.get_current_module(), _NNCFModuleMixin)
             node = ctx.maybe_add_node(processed_input, tensor_metas, op_address,
                                       layer_attrs, ignored_algos, is_called_inside_nncf_module)
+            if node is not None:
+                _add_operation_attributes(node, ctx)
```
ljaljushkin, line 218: Have to override existing attributes for operations that require graph traversal in order to find layer attributes.
input and output features for `matmul` is extracted from the shape of second input.

nncf/torch/pruning/utils.py
```diff
@@ -105,16 +112,21 @@ def collect_output_shapes(graph: NNCFGraph) -> Dict[NNCFNodeName, List[int]]:
...
     modules_out_shapes = {}
+
     output_shape_collecting_info = [
-       (NNCF_GENERAL_CONV_MODULES_DICT, slice(2, None)),
-       (NNCF_LINEAR_MODULES_DICT, slice(None)),
+        OutputShapeCollectionInfo(node_types=[v.op_func_name for v in NNCF_GENERAL_CONV_MODULES_DICT],
+                                  shape_slice=slice(2, None)),
+        OutputShapeCollectionInfo(node_types=[v.op_func_name for v in NNCF_LINEAR_MODULES_DICT],
+                                  shape_slice=slice(None)),
+        OutputShapeCollectionInfo(node_types=['matmul', 'bmm'],
```
ljaljushkin, line 121: @daniil-lyakhov Does it affect Filter Pruning algorithm?
Does Filter Pruning algorithm calculate FLOPS after setting some meaningful pruning masks, or it only calculates for FP32 model?

daniil-lyakhov: FLOPS calculation besides statistics prompt is used to calculate filter pruning level by given FLOPS pruning level. This change possibly affects masks configuration for models that have matmul operations with flops pruning targets in config

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -169,6 +170,30 @@ def wrapped(self, *args, **kwargs):
...
+    op_address = node.op_exec_context.op_address
+    # TODO(nlyalyus): support other operations: torch.addmm, torch.nn.functional.Linear, torch.bmm, torch.mm
+    if op_address.operator_name in ('matmul', 'bmm'):
+        graph = ctx.graph
+        input_edges = graph.get_input_edges(node)
+        # no input nodes when input_info tensor was bound to a *args or **kwargs on graph tracing
+        if len(input_edges):
+            assert len(input_edges) == 2
+            activation_edge, weight_edge = input_edges
+            activation_shape = activation_edge.activation_shape
```
daniil-lyakhov, line 183: @ljaljushkin, can we use `args` and `kwargs` for operation attributes? For some operations (like `transpose`) layer attributes collection by input edges is impossible (`input_shape=(5,5,5), transpose=(2, 1, 0)`). You can find example with torch implementation for transpose layer attributes [here](https://github.com/daniil-lyakhov/nncf/commit/079578ccab88a18640ada7088b33f548e63937f4#diff-6e72f489b20539eb2b9d90608c21d9288a65638482687360312d8aebbc546871)

PR #1481, opened 2023-01-05
Removing matplotlib from the mandatory requirements

alexsu52:
### Changes

matplotlib dependency has been moved under dev extra.

### Reason for changes

OpenVINO notebooks have compatibility issues with matplotlib dependency.

### Related tickets

Custom request

### Tests

N/A

### DIFF
--- a/nncf/common/utils/decorators.py
+++ b/nncf/common/utils/decorators.py
@@ -19,14 +19,14 @@ from nncf.common.logging import nncf_logger
 IMPORTED_DEPENDENCIES = {}
 
 
-def bypass_noreturn_function(dependencies: List[str]) -> Callable:
+def skip_if_dependency_unavailable(dependencies: List[str]) -> Callable:
     """
-    Decorator to bypass a noreturn function if dependencies are not met.
+    Decorator factory to skip a noreturn function if dependencies are not met.
 
-    :param func: A noreturn function to decorate
-    :return: Decorated function
+    :param dependencies: A list of dependencies
+    :return: A decorator
     """
-    def wrap(func: Callable) -> Callable:
+    def wrap(func: Callable[..., None]) -> Callable[..., None]:
         def wrapped_f(*args, **kwargs):
             for libname in dependencies:
                 if libname in IMPORTED_DEPENDENCIES:
--- a/nncf/common/utils/decorators.py
+++ b/nncf/common/utils/decorators.py
@@ -1,0 +1,51 @@
+"""
+ Copyright (c) 2022 Intel Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


nncf/experimental/torch/nas/bootstrapNAS/search/search.py
```diff
@@ -50,6 +48,16 @@
 from nncf.experimental.torch.nas.bootstrapNAS.elasticity.multi_elasticity_handler import SubnetConfig
 from nncf.torch.nncf_network import NNCFNetwork
 
+try:
+    import matplotlib.pyplot as plt
```
ljaljushkin, line 52: To be more consistent, it makes sense to do the same for hawq_debug.py file
could you do that in the PR?

alexsu52: Honestly, I thought that the debug code should fail if the dependencies are not meet But nevertheless, I did it.

nncf/common/utils/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+
+from importlib import import_module
+from typing import Callable, List
+
+from nncf.common.logging import nncf_logger
+
+IMPORTED_DEPENDENCIES = {}
+
+
+def bypass_noreturn_function(dependencies: List[str]) -> Callable:
```
vshampor, line 22: `skip_if_dependency_unavailable`

`typing.NoReturn` that you are referring to here is actually a wrong type hint - it marks functions that *do not return* (e.g. functions that always raise an exception), not functions that return `None` explicitly or implicitly.

nncf/common/utils/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Decorator to bypass a noreturn function if dependencies are not met.
+
+    :param func: A noreturn function to decorate
+    :return: Decorated function
+    """
+    def wrap(func: Callable) -> Callable:
```
vshampor, line 29: `Callable[..., None]` to mark that the callable to be wrapped should only be a function that does not return a specific return value other than `None`

nncf/common/utils/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+from nncf.common.logging import nncf_logger
+
+IMPORTED_DEPENDENCIES = {}
+
+
+def bypass_noreturn_function(dependencies: List[str]) -> Callable:
+    """
+    Decorator to bypass a noreturn function if dependencies are not met.
+
+    :param func: A noreturn function to decorate
```
vshampor, line 26: Should probably describe the `dependencies` parameter instead.

vshampor (changes requested):

nncf/common/utils/decorators.py
alexsu52: done

nncf/common/utils/decorators.py
alexsu52: done

nncf/common/utils/decorators.py
alexsu52: done

vshampor (approved):

PR #1484, opened 2023-01-09
Export of the NNCF graph for the Netron app was added

andrey-churkin:
### Changes

The `save_for_netron()` method was added. It takes the `NNCFGraph` object and `save_path` (path to the `.xml` file where the graph should be saved).

There is an ability to show the arbitrary number of the node attributes. For example, I added the metatype attribute
pic

### Reason for changes

- You can see the NNCF graph object in the Netron app now.

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/nncf/experimental/netron.py
+++ b/nncf/experimental/netron.py
@@ -1,0 +1,197 @@
+"""
+ Copyright (c) 2022 Intel Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import Optional
+import xml.etree.ElementTree as ET
+
+from nncf.common.graph import NNCFGraph
+
+
+class Tags:
+    NET = 'net'
+    NODES = 'layers'
+    EDGES = 'edges'
+    NODE = 'layer'
+    EDGE = 'edge'
+    DATA = 'data'
+    INPUT = 'input'
+    OUTPUT = 'output'
+    PORT = 'port'
+    DIM = 'dim'
+
+
+class PortDesc:
+
+    def __init__(self,
+                 port_id: str,
+                 shape: Optional[List[int]] = None,
+                 precision: str = None):
+        self.port_id = port_id
+        if shape is None:
+            shape = []
+        self.shape = shape
+        self.precision = precision
+
+    def as_xml_element(self) -> ET.Element:
+        port = ET.Element(Tags.PORT, id=self.port_id)
+        if self.precision:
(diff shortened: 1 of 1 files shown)


andrey-churkin:
Guys, could you please look at this stuff? I think we can make this method part of our public API. It may be very helpful for the creation of the ignored scope. If you have any comments/suggestions feel to free put them here.

andrey-churkin:
I tested this method only for the OpenVINO backend. But I think it should work for other backends as well.

nncf/experimental/netron.py
```diff
@@ -0,0 +1,197 @@
...
+                },
+                inputs=inputs,
+                outputs=outputs
+            )
+        )
+
+    return nodes, edges
+
+
+def save_for_netron(graph: NNCFGraph,
```
alexsu52, line 178: Could this be the NNCFGraph method? The user must have NNCFGraph object to call `save_for_netron`, so it looks like easily to call `graph.save(...) ` instead of `nncf.*.save_for_netron(graph, ...)`

alexsu52 (commented):
Cool staff!

nncf/experimental/netron.py
andrey-churkin: Do we really need to show the `NNCFGraph` object to the user? May be it is better do not show this object and use code like this
```python
quantized_model = nncf.quantize(model, calibration_dataset)
nncf.save_graph_for_netron(quantized_model, 'quantized_model.xml')
```

andrey-churkin: If I understand correctly, the user only needs the NNCFGraph to save it.

andrey-churkin: @AlexKoff88 What do you think about how we should expose this method to users?

AlexKoff88: I would agree that it is better to have it as a method of the wrapped object:
```python
quantized_model.save_for_netron("quantized_model.xml")
```

nncf/experimental/netron.py
```diff
@@ -0,0 +1,197 @@
...
+                    port_id=str(edge.output_port_id),
+                    shape=edge.tensor_shape,
+                )
+            )
+
+        nodes.append(
+            NodeDesc(
+                node_id=str(node.node_id),
+                name=node.node_name,
+                type=node.node_type,
```
ljaljushkin, line 166:
```suggestion
                type=node.node_type.title(),
```
image
There is more chances to have pretty graph with type in a title case.
Probably, some mapping from NNCF types to Netron ones would be helpful in the future.

ljaljushkin (changes requested):
Like it! Can we make it a bit more general?

aiarkinx:
run onnx pre-commit tests

aiarkinx:
run onnx pre-commit tests

aiarkinx:
run onnx pre-commit tests

ljaljushkin:
@alexsu52 @AlexKoff88  @vshampor @andrey-churkin 
Can we merge it as is as an experimental feature for internal usage and finalize exposure for external users in a separate PR?
I found it useful for debugging mask propagation.

vshampor (approved):

ljaljushkin:
> @alexsu52 @AlexKoff88 @vshampor @andrey-churkin Can we merge it as is as an experimental feature for internal usage and finalize exposure for external users in a separate PR? I found it useful for debugging mask propagation.

Created 106717 ticket for further improvements.

ljaljushkin:
run pylint pre-commit tests

PR #1485, opened 2023-01-09
Introduce CPU_SPR target device

alexsu52:
### Changes

Added CPU_SPR target device

### Reason for changes

OpenVINO backend supports CPU_SPR target device

### Related tickets

N/A

### Tests

N/A


nncf/onnx/quantization/quantize.py
```diff
@@ -46,6 +46,8 @@ def quantize_impl(model: onnx.ModelProto,
         raise RuntimeError('Quantization algorithm from the ONNX backend '
                            'does not support operation types in the ignored '
                            'scopes yet')
+    if target_device == TargetDevice.CPU_SPR:
```
kshpv, line 49: What happen if user uses directly `PostTrainingQuanitzation` for quantization pipeline?
Could you add the new `target_device` in `/tests/onnx/quantization/test_ptq_params.py` in `test_target_device()`?

kshpv (changes requested):

alexsu52:
> LGTM. @alexsu52, have you tested this functionality with the OpenVINO backend? Is the new device flowing through POT API correctly?

I added the test, please take a look.

nncf/onnx/quantization/quantize.py
alexsu52: done

kshpv (approved):

KodiaqQ (approved):

nncf/parameters.py
```diff
@@ -31,6 +31,7 @@ class TargetDevice(Enum):
     CPU = 'CPU'
     GPU = 'GPU'
     VPU = 'VPU'
+    CPU_SPR = 'CPU_SPR'
```
vshampor, line 34: Note that without additional changes the `"target_device": "CPU_SPR"` won't be parseable via NNCF .json config file. Furthermore, consider not failing with "CPU_SPR" on other backends, and instead mapping this value to regular "CPU" with a warning-level log message spawned.

alexsu52: Thanks for the comment. Please, take a look.

vshampor (approved):

KodiaqQ (approved):

kshpv (approved):

ljaljushkin (approved):

PR #1489, opened 2023-01-10
Add aliases for backend extra installation

vshampor:
### Changes
Add synonyms for the `tf` and `torch` in context of extra dependency installation, i.e. now can run `pip install .[tensorflow]` to do the same thing as `pip install .[tf]`.

### Reason for changes
Consistency with `openvino-dev` extras syntax makes for better UX.

### Related tickets
N/A

### Tests
N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,32 +123,48 @@ INSTALL_REQUIRES = ["ninja>=1.10.0.post2, <1.11",
                     "openvino-telemetry"]
 
 
+
+TF_EXTRAS = [
+        "tensorflow~=2.8.4",
+    ]
+
+TORCH_EXTRAS = [
+        "torch>=1.8.2,<1.14",
+    ]
+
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -124,32 +124,45 @@ def find_version(*file_paths):
...
+
+ONNX_EXTRAS = [
         "onnx==1.12.0",
         "onnxruntime-openvino==1.13.1"
-    ],
-    "openvino": [
+    ]
+
+OPENVINO_EXTRAS = [
         "openvino-dev~=2022.2.0"
```
helena-intel, line 141: Unrelated to this PR, but could this changed to 2022.3.0?

vshampor: I rebased this PR on top of develop, and the requirement has been relaxed there already.

ljaljushkin (approved):

PR #1494, opened 2023-01-11
Calibrate tool for OpenVINO backend

alexsu52:
### Changes

Added the calibrate tool for OpenVINO backend

### Reason for changes

To wide test scope

### Related tickets

ref:98129

### Tests

N/A

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -153,7 +153,16 @@ def map_quantization_parameters(pot_parameters):
     ignored_parameters = [
         'dump_intermediate_model',
         'inplace_statistics',
-        'num_samples_for_tuning'
+        'num_samples_for_tuning',
+        'batch_size',
+        'optimizer',
+        'loss',
+        'tuning_iterations',
+        'random_seed',
+        'use_ranking_subset',
+        'calibration_indices_pool',
+        'calculate_grads_on_loss_increase_only',
+        'weight_decay'
     ]
 
     return _map_parameters(pot_parameters, parameters_mapping,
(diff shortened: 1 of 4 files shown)


tests/openvino/tools/calibrate.py
```diff
@@ -0,0 +1,267 @@
...
+            if op.get('attributes') is not None:
+                raise ValueError(f'Attributes in the ignored operations '
+                                 f'are not supported')
+    return {'ignored_scope': nncf.IgnoredScope(names=ignored.get('scope'), 
+                                               types=ignored.get('types'))}
+
+
+def map_preset(preset):
+    preset = preset.lower()
+    if preset not in [p.value for p in nncf.QuantizationPreset]:
```
KodiaqQ, line 117: There is existing an conversion from the `accuracy` preset into `mixed` in the POT.
https://github.com/openvinotoolkit/openvino/blob/master/tools/pot/openvino/tools/pot/configs/config.py#L349
Do we need this functionality here also?

tests/openvino/tools/calibrate.py
```diff
@@ -0,0 +1,267 @@
...
+    }
+    
+    default_parameters = {
+      'use_layerwise_tuning': False
+    }
+
+    ignored_parameters = [
+        'dump_intermediate_model',
+        'inplace_statistics',
+        'num_samples_for_tuning'
```
KodiaqQ, line 158: There is also a `tuning_iterations` parameter for `LayerwiseTuning`. Also about the `sparsity_level` parameter (that might be useless now).

tests/openvino/tools/calibrate.py
alexsu52: I don't think so, because it maps the preset from POT config to the `nncf.QuantizationPreset`.

tests/openvino/tools/calibrate.py
alexsu52: done

KodiaqQ (approved):

l-bat (approved):

andrey-churkin (approved):

PR #1497, opened 2023-01-11
Fix the rest of the log string creation

vshampor:
### Changes

Forgot to set a proper log level at nncf/torch/pruning/base_algo.py:117 so that a message was printed for every weight pruned in INFO level, while it should have been printed to DEBUG. Also made the rest of the non-f-string-constructed parametrized log strings actually use f-strings.

### Reason for changes
Follow-up for the things that slipped away in the previous log-related PR.

### Related tickets
N/A

### Tests
N/A

### DIFF
--- a/nncf/torch/extensions/__init__.py
+++ b/nncf/torch/extensions/__init__.py
@@ -20,7 +20,7 @@ class ExtensionsType(enum.Enum):
 def get_build_directory_for_extension(name: str) -> Path:
     build_dir = Path(_get_build_directory('nncf/' + name, verbose=False)) / torch.__version__
     if not build_dir.exists():
-        nncf_logger.debug("Creating build directory: {}".format(str(build_dir)))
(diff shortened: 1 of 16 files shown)


nncf/torch/extensions/__init__.py
```diff
@@ -20,7 +20,7 @@ class ExtensionsType(enum.Enum):
 def get_build_directory_for_extension(name: str) -> Path:
     build_dir = Path(_get_build_directory('nncf/' + name, verbose=False)) / torch.__version__
     if not build_dir.exists():
-        nncf_logger.debug("Creating build directory: {}".format(str(build_dir)))
+        nncf_logger.debug(f"Creating build directory: {str(build_dir)}")
```
KodiaqQ, line 23: That's minor in general. Wouldn't it be better to use `.as_posix()` than `str()`?

KodiaqQ (approved):

PR #1501, opened 2023-01-16
Fix calibrate tool

andrey-churkin:
### Changes

The `transform_fn()` was fixed.

### Reason for changes

I got the following error:

```
  File "openvino/tools/pot/engines/ie_engine.py", line 242, in process_input
      is_loader_batchfied = input_dim(input_blob) == input_data[0].ndim
      AttributeError: 'list' object has no attribute 'ndim'
```

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -220,7 +220,7 @@ def quantize_model(xml_path, bin_path, accuracy_checcker_config,
         _, batch_annotation, batch_input, _ = data_item
         filled_inputs, _, _ = model_evaluator._get_batch_input(
             batch_input, batch_annotation)
-        return filled_inputs
+        return filled_inputs[0]
 
     calibration_dataset = nncf.Dataset(model_evaluator.dataset, transform_fn)
     if quantization_impl == 'pot':


vshampor (approved):

tests/openvino/tools/calibrate.py
```diff
@@ -220,7 +220,7 @@ def transform_fn(data_item):
         _, batch_annotation, batch_input, _ = data_item
         filled_inputs, _, _ = model_evaluator._get_batch_input(
             batch_input, batch_annotation)
-        return filled_inputs
+        return filled_inputs[0]
```
KodiaqQ, line 223: Can we be sure that `model_evaluator._get_batch_input` always returns a list as the first output? Can't we take the first batch this way, for example?

andrey-churkin: That's a good point. I will check it.

alexsu52 (approved):

tests/openvino/tools/calibrate.py
andrey-churkin: > Can we be sure that model_evaluator._get_batch_input always returns a list as the first output?

Yes, we can.

> Can't we take the first batch this way, for example?

The `filled_inputs` is a `List[Dict[str, np.ndarray]]`. Probably `len(filled_inputs)` always equals to 1.

KodiaqQ (approved):

PR #1502, opened 2023-01-16
Time for the dataset's length estimation was improved

andrey-churkin:
### Changes

`DatasetWrapper` was improved. It is a temporary class. It will be removed when we will switch on native implementation. 

### Reason for changes

The `nncf.quantize()` method is slower than POT.

**Before Fix:**
Elapsed Time: 00:04:01

**After Fix:**
Elapsed Time: 00:00:31

### Related tickets

Ref: 100984

### Tests

N/A

### DIFF
--- a/nncf/openvino/engine.py
+++ b/nncf/openvino/engine.py
@@ -72,11 +72,18 @@ class DatasetWrapper:
 
     def __len__(self) -> int:
         if self._length is None:
-            self._length = DatasetWrapper._get_length(self._iterable)
+            indices = getattr(self._iterable, '_indices', None)
+            data_source = getattr(self._iterable, '_data_source', None)
+            if indices:
+                self._length = len(indices)
+            iterable = data_source if data_source else self._iterable
+            self._length = DatasetWrapper._get_length(iterable)
         return self._length
 
     @staticmethod
     def _get_length(iterable) -> int:
+        if hasattr(iterable, '__len__'):
+            return len(iterable)
         length = 0
         for _ in iterable:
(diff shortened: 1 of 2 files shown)


nncf/openvino/engine.py
```diff
@@ -72,11 +72,18 @@ def __iter__(self):
 
     def __len__(self) -> int:
         if self._length is None:
-            self._length = DatasetWrapper._get_length(self._iterable)
+            indices = getattr(self._iterable, '_indices', None)
+            data_source = getattr(self._iterable, '_data_source', None)
```
vshampor, line 76: Is there a way for this be rewritten in terms of `isinstance` calls? `isinstance` is bad on its own, though, but plain `getattr`-based conditionals are even worse.

vshampor (approved):

tests/openvino/tools/calibrate.py
```diff
@@ -212,9 +212,6 @@ def quantize_model(xml_path, bin_path, accuracy_checcker_config,
     model_evaluator = create_model_evaluator(accuracy_checcker_config)
     model_evaluator.load_network([{'model': ov_model}])
     model_evaluator.select_dataset('')
-    # workaround to minimize time of estimation len of dataset
-    subset_size = quantization_parameters.get('subset_size', 300)
```
KodiaqQ, line 216: In case of dropping this, the quantization time would be increased, because the statistics will be collected from the whole dataset. Am I right?

KodiaqQ (approved):

PR #1503, opened 2023-01-16
Use new version of requesting deterministic algorithms in PyTorch

vshampor:
### Changes
A single-line change to reflect the new API in PyTorch.

### Reason for changes
The old way has been deprecated since 1.8. The newly re-enabled compression training tests caught this.

### Related tickets
N/A

### Tests
tests/torch/test_compression_training.py

### DIFF
--- a/examples/torch/common/execution.py
+++ b/examples/torch/common/execution.py
@@ -121,4 +121,4 @@ def set_seed(config):
         manual_seed(config.seed)
         cudnn.deterministic = True
         cudnn.benchmark = False
-        torch.set_deterministic(True)
+        torch.use_deterministic_algorithms(True)


examples/torch/common/execution.py
```diff
@@ -121,4 +121,4 @@ def set_seed(config):
         manual_seed(config.seed)
         cudnn.deterministic = True
         cudnn.benchmark = False
-        torch.set_deterministic(True)
+        torch.use_deterministic_algorithms(True)
```
ljaljushkin, line 124: We still claim a support of torch 1.8.2. Shouldn't be `if` condition for different torch version here?

vshampor: They wouldn't deprecate the old way without a new in place. The 1.8.1 [docs](https://pytorch.org/docs/1.8.1/generated/torch.use_deterministic_algorithms.html#torch.use_deterministic_algorithms) describe the new function, so the change keeps our desired compatibility scope.

ljaljushkin (approved):

PR #1507, opened 2023-01-17
Update eval results in tables across READMEs and align TF model names with PT

vshampor:
### Changes
Updated metric values in tables and in reference files. Made TF models conform to the way of naming in PT. Enabled in-place editing for the `tools/update_eval_results.py` script to do this automatically using only the `results.json` file from CI.

### Reason for changes
Catching up to the metric drift.

### Related tickets
N/A

### Tests
E2E TF for the model renaming part.


alexsu52 (changes requested):
What reason of these changes?

examples/tensorflow/classification/README.md
```diff
@@ -125,32 +125,40 @@ python main.py \
 To export a model to the OpenVINO IR and run it using the Intel® Deep Learning Deployment Toolkit, refer to this [tutorial](https://software.intel.com/en-us/openvino-toolkit).
 
 ### Results
+<a name="results"></a>
 
-|**Model**|**Compression algorithm**|**Dataset**|**Accuracy (Drop) %**|**NNCF config file**|**TensorFlow checkpoint**|
+|Model|Compression algorithm|Dataset|Accuracy (_drop_) %|NNCF config file|PyTorch checkpoint|
```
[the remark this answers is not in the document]
vshampor: Fixed

vshampor:
@alexsu52 the metrics measured by us in E2E change over time (probably due to framework version changes). As a result currently measured metrics may be consistently different from the expected results, while still remaining within 1% of FP32. The metrics in READMEs and reference metric values should be aligned with what we measure right now, and this is what this PR does. 

For PT, the script I'm modifying here was already able to update both reference values in a .json file and the values across READMEs automatically based on an artifact in CI. I have now extended this functionality to TF, cleaning up the TF's sota_checkpoints_eval.json file in the process (aligning the names of the models with the format in PT). I will rename the checkpoints themselves for the new format in parallel.

vshampor:
> > 
> 
> As for the TF backend, all metrics have been updated during the migration to TF 2.8 and do not require updating. Since we do not know the reason for changing the metric, I suggest you open bugs for models which accuracy metric from test and readme is not equal.

I only noticed just now that, due to a logic error introduced in the script somewhere between now and the last time I used it, the generated tables contained the reference FP32 metric everywhere instead of the currently measured compressed metric. @alexsu52 you can now see that the metrics have actually changed very little.

vshampor:
See E2E TF build 229 for results, can also review build 227 for preliminary results (it has only one failure for mobilenet_v2_hub in OV due to an error in AC config naming that I fixed in the latest commit and which should be green in build 229).

examples/tensorflow/classification/README.md
```diff
@@ -125,32 +125,40 @@ python main.py \
...
-|ResNet50|Sparsity 80% (RB)|ImageNet|74.36 (0.68)|[resnet50_imagenet_rb_sparsity.json](configs/sparsity/resnet50_imagenet_rb_sparsity.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/resnet50_rb_sparsity_80.tar.gz)|
-|ResNet50|INT8 (per-tensor, symmetric for weights; per-tensor, symmetric for activations) + Sparsity 65% (RB)|ImageNet|74.3 (0.74)|[resnet50_imagenet_rb_sparsity_int8.json](configs/sparsity_quantization/resnet50_imagenet_rb_sparsity_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/resnet50_int8_w_sym_t_half_a_sym_t_rb_sparsity_65.tar.gz)|
-|TensorFlow Hub MobileNet V2|Sparsity 35% (Magnitude)|ImageNet|71.90 (-0.06)|[mobilenet_v2_hub_imagenet_magnitude_sparsity.json](configs/sparsity/mobilenet_v2_hub_imagenet_magnitude_sparsity.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/tf1_mobilenet_v2_1.0_224_sparsity_35.tar.gz)|
+|Inception V3|None|ImageNet|77.91|[inception_v3_imagenet.json](configs/classification/inception_v3_imagenet.json)|-|
```
alexsu52, line 132: TF does not have configs/classification/inception_v3_imagenet.json

examples/tensorflow/classification/README.md
```diff
@@ -125,32 +125,40 @@ python main.py \
...
+|Inception V3|INT8 (per-tensor, symm. W, asymm. half-range A)|ImageNet|78.38 (-0.47)|[inception_v3_imagenet_int8.json](configs/quantization/inception_v3_imagenet_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/inception_v3_imagenet_int8)|
```
alexsu52, line 133: Could you explain why you change compression algorithm notation?
INT8 (per-tensor, symmetric for weights; per-tensor, symmetric for activations) -> INT8 (per-tensor, symm. W, asymm. half-range A)

alexsu52, line 133: The link does not work https://storage.openvinotoolkit.org/repositories/nncf/models/develop/inception_v3_imagenet_int8.

alexsu52, line 133: It looks like you want to update name of checkpoint. Could you statement in the contribution guide how the developer should naming checkpoints in this PR or in the separate PR.

tests/tensorflow/sota_checkpoints_eval.json
```diff
@@ -6,253 +6,254 @@
...
                     "diff_fp32_min": -0.5,
                     "diff_fp32_max": 1
                 },
-                "mobilenet_v2_int8_w_sym_t_half_a_sym_t_rb_sparsity_52": {
+                "mobilenet_v2_imagenet_rb_sparsity_int8": {
                     "config": "examples/tensorflow/classification/configs/sparsity_quantization/mobilenet_v2_imagenet_rb_sparsity_int8.json",
-                    "reference": "mobilenet_v2",
-                    "target": 71.0,
+                    "reference": "mobilenet_v2_imagenet",
+                    "target": 70.96,
```
alexsu52, line 90: If you are going to update the target metric every release, then we may end up in a situation where the metric will change significantly.

Сan we not change target metrics in the automatic flow?

alexsu52 (changes requested):

examples/tensorflow/classification/README.md
vshampor: Restored the notation.

vshampor: Well, the links with new identifiers won't work until I have uploaded the renamed models, but after you drew my attention to this I noticed that I forgot to add `.tar.gz` for TF cases. Fixed

tests/tensorflow/sota_checkpoints_eval.json
vshampor: The FP32 metric can only change if the upstream FP32 model is retuned or improved. If this leads to our existing checkpoints having less than 99% FP32, it will show. The entire script in `tools/update_eval_results.py` is only meant to be run manually, based off E2E results that we find acceptable in their entirety.

examples/tensorflow/classification/README.md
vshampor: Done in this PR, please review.

examples/tensorflow/classification/README.md
alexsu52: You have added the following config inception_v3_imagenet.json to the example folder. It appears that with the following config it is possible to train the inception_v3 on imagenet. Have you checked it? I highly recommend removing this row from the table.

alexsu52 (changes requested):
Please, resolve my comment.

examples/tensorflow/classification/configs/inception_v3_imagenet.json
```diff
@@ -0,0 +1,11 @@
+{
+    "model": "InceptionV3",
+    "pretrained": true,
+    "eval_only": true,
+    "input_info": {
+        "sample_size": [1, 299, 299, 3]
+    },
+    "batch_size": 256,
+    "dataset": "imagenet2012",
+    "dataset_type": "tfrecords"
```
alexsu52, line 10:
```suggestion
    "dataset_type": "tfds"
```

examples/tensorflow/object_detection/README.md
```diff
@@ -157,22 +157,23 @@ To export a model to the OpenVINO IR and run it using the Intel® Deep Learning
...
 
-<a name='results'></a>
-
 ## Results
+<a name="results"></a>
 
-|**Model**|**Compression algorithm**|**Dataset**|**mAP (drop) %**|**NNCF config file**|**TensorFlow checkpoint**|
+|Model|Compression algorithm|Dataset|mAP (_drop_) %|NNCF config file|Checkpoint|
 | :---: | :---: | :---: | :---: | :---: | :---: |
-|RetinaNet|None|COCO2017|33.44|[retinanet_coco.json](configs/retinanet_coco.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/retinanet.tar.gz)|
```
alexsu52, line 166: Could you explain why you removed it?

examples/tensorflow/classification/README.md
```diff
@@ -125,32 +125,40 @@ python main.py \
...
-| :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
-|ResNet50|None|ImageNet|75.04|7.75 (100%)|25.5 (100%)|-|-|
-|ResNet50|Filter Pruning 40%, geometric_median criterion|ImageNet|74.98 (0.06)|4.29 (55.35%)|15.8 (61.96%)|[Link](configs/pruning/resnet50_imagenet_pruning_geometric_median.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/resnet50_pruning_40.tar.gz)|
-|ResNet50|Filter Pruning 40%, geometric_median criterion + INT8 (per-tensor, symmetric for weights; per-tensor, symmetric for activations)|ImageNet|75.08 (-0.04)|4.27 (55.10%)|15.8 (61.96%)|[Link](configs/pruning_quantization/resnet50_imagenet_pruning_geometric_median_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/resnet50_int8_w_sym_t_half_a_sym_t_pruning_40.tar.gz)|
+<a name="filter_pruning"></a>
+
+|Model|Compression algorithm|Dataset|Accuracy (_drop_) %|NNCF config file|Checkpoint|
+| :---: | :---: | :---: | :---: | :---: | :---: |
+|ResNet-50|None|ImageNet|75.05|[resnet50_imagenet.json](configs/resnet50_imagenet.json)|-|
```
alexsu52, line 156: Could you explain why resnet-50 in this table but not in the table above? Maybe it's worth creating a separate table for FP32 models?

examples/common/sample_config.py
```diff
@@ -116,7 +117,26 @@ def create_sample_config(args, parser) -> SampleConfig:
...
+
+def create_sample_config(args, parser) -> SampleConfig:
+    sample_config = _parse_sample_config(args, parser)
+    _fail_if_training_with_eval_only_config(sample_config)
+    sample_config = _embed_nncf_config(args, sample_config)
+
+    return sample_config
+
+
+def create_sample_config_for_separate_train_eval_scripts(args, parser) -> SampleConfig:
```
alexsu52, line 139: What different between `create_sample_config` and `create_sample_config_for_separate_train_eval_scripts`? Why you cannot check  `_fail_if_training_with_eval_only_config` everywhere?

examples/tensorflow/common/experimental_patcher.py
```diff
@@ -0,0 +1,8 @@
+from nncf import NNCFConfig
```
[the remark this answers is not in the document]
vshampor: Done

examples/tensorflow/classification/configs/inception_v3_imagenet.json
vshampor: Done

examples/tensorflow/object_detection/README.md
vshampor: It was the result of a bug in the table-generating script - fixed the bug and the line reappeared.

examples/tensorflow/classification/README.md
vshampor: Same - it was a bug in the script.

examples/common/sample_config.py
vshampor: When train/eval scripts are separate, the train/eval mode is not determined by the `--mode` argument from the CLI (which is then also embedded in the config), but rather it is the choice of the script itself (either `train.py` or `eval.py`) that determines whether we are training or eval-ing.

examples/tensorflow/segmentation/README.md
```diff
@@ -140,12 +140,13 @@ To export a model to the OpenVINO IR and run it using the Intel® Deep Learning
...
+|Mask-R-CNN|None|COCO 2017|37.33|[mask_rcnn_coco.json](configs/mask_rcnn_coco.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco.tar.gz)|
+|Mask-R-CNN|INT8 (per-tensor symmetric for weights, per-tensor asymmetric half-range for activations)|COCO 2017|37.19 (0.14)|[mask_rcnn_coco_int8.json](configs/quantization/mask_rcnn_coco_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco_int8.tar.gz)|
+|Mask-R-CNN|Magnitude sparsity, 50%|COCO 2017|36.94 (0.39)|[mask_rcnn_coco_magnitude_sparsity.json](configs/sparsity/mask_rcnn_coco_magnitude_sparsity.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco_magnitude_sparsity.tar.gz)|
 |MaskRCNN|Sparsity 50% (Magnitude)|COCO2017|bbox: 36.93 (0.40)<br/>segm: 33.23 (0.33)|[Link](configs/sparsity/mask_rcnn_coco_magnitude_sparsity.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_sparsity_50.tar.gz)|
```
alexsu52, line 152: It looks like leftover.

examples/tensorflow/segmentation/README.md
```diff
@@ -140,12 +140,13 @@ To export a model to the OpenVINO IR and run it using the Intel® Deep Learning
...
-|MaskRCNN|INT8 (per-tensor, symmetric for weights; per-tensor, symmetric for activations)|COCO2017|bbox: 37.27 (0.06)<br/>segm: 33.54 (0.02)|[Link](configs/quantization/mask_rcnn_coco_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_int8_w_sym_t_half_a_sym_t.tar.gz)|
+|Mask-R-CNN|None|COCO 2017|37.33|[mask_rcnn_coco.json](configs/mask_rcnn_coco.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco.tar.gz)|
+|Mask-R-CNN|INT8 (per-tensor symmetric for weights, per-tensor asymmetric half-range for activations)|COCO 2017|37.19 (0.14)|[mask_rcnn_coco_int8.json](configs/quantization/mask_rcnn_coco_int8.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco_int8.tar.gz)|
+|Mask-R-CNN|Magnitude sparsity, 50%|COCO 2017|36.94 (0.39)|[mask_rcnn_coco_magnitude_sparsity.json](configs/sparsity/mask_rcnn_coco_magnitude_sparsity.json)|[Link](https://storage.openvinotoolkit.org/repositories/nncf/models/develop/tensorflow/mask_rcnn_coco_magnitude_sparsity.tar.gz)|
```
alexsu52, line 151: Please report both metrics as in the develop.

examples/tensorflow/segmentation/evaluation.py
```diff
@@ -84,7 +85,7 @@ def get_config_from_argv(argv, parser):
         {'dataset_type': 'tfrecords'}
     )
 
-    config_from_json = create_sample_config(args, parser)
+    config_from_json = create_sample_config_for_separate_train_eval_scripts(args, parser)
```
alexsu52, line 88:
```suggestion
    config_from_json = create_sample_config(args, parser, mode='test')
```

examples/tensorflow/segmentation/train.py
```diff
@@ -69,14 +71,18 @@ def get_argument_parser():
...
+
 def get_config_from_argv(argv, parser):
     args = parser.parse_args(args=argv)
 
     sample_config = SampleConfig(
         {'dataset_type': 'tfrecords'}
     )
 
-    config_from_json = create_sample_config(args, parser)
+    config_from_json = create_sample_config_for_separate_train_eval_scripts(args, parser)
```
alexsu52, line 82:
```suggestion
    config_from_json = create_sample_config(args, parser, mode='train')
```

alexsu52 (changes requested):

examples/tensorflow/segmentation/README.md
vshampor: Removed

examples/tensorflow/segmentation/README.md
vshampor: Ok, but we actually don't measure the second metric anywhere right now, have no fixed reference for it and don't check for it. I couldn't find a metric output for the second metric in the E2E tests quickly, so I'll find out the current metric and integrate it into the E2E and table-generating script in a follow-up PR.

examples/tensorflow/segmentation/evaluation.py
vshampor: Done

alexsu52 (approved):

PR #1511, opened 2023-01-19
Remove unnecessary dependencies

vshampor:
### Changes

Removed `addict` and `wheel` from NNCF package dependencies and moved `pillow` dependency to the `[dev]` extra.

### Reason for changes
These dependencies are not strictly necessary for NNCF to function across the supported frameworks.

### Related tickets
N/A

### Tests
Precommit scope

### DIFF
--- a/nncf/openvino/quantization/accuracy_aware.py
+++ b/nncf/openvino/quantization/accuracy_aware.py
@@ -10,12 +10,10 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 """
-
 from typing import List
 from functools import partial
 
 import numpy as np
-from addict import Dict
 
 from openvino.tools import pot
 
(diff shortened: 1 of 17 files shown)


nncf/openvino/quantization/accuracy_aware.py
```diff
@@ -35,21 +33,28 @@ class NMSEBasedAccuracyAware(pot.AccuracyAwareCommon):
     def __init__(self, config, engine):
         super().__init__(config, engine)
         if not engine.use_original_metric:
-            self._metrics_config['original_metric'].persample = Dict(
```
andrey-churkin, line 38: There is a problem here and below. POT API uses `persample.attr_name`  to get the `attr_name` value. I suggest using
```python
MetricAttrs = namedtuple('MetricAttrs', ['name', 'type', 'is_special', 'comparator', 'sort_fn'])
```
as workaround. This code will be removed soon.

vshampor: I figured, but here I am not actually replacing the underlying `ad.Dict` in `self._metric_config['original_metric']`. I specifically used `.clear` and `.update`, which are methods of `dict` (but also of `ad.Dict`, since it inherits from dict), in order to preserve the downstream code functionality. At the same time I can remove the `from addict import Dict` because I don't need to instantiate a new object. There is thus no necessity to introduce the tuple right now - basically everything should work as-is. Not sure if tests cover this exact code, though.

andrey-churkin: I got it. Thanks for the explanation!

andrey-churkin (approved):

PR #1525, opened 2023-01-25
[OV] Supported different configurations of input data for OVNativeEngine

l-bat:
### Changes

* Added inputs support for `OVNativeEngine.infer` such as `np.ndarray`, `List[np.ndarray]`, `Tuple[np.ndarray]`, `Dict[str, np.ndarray]`

### Reason for changes

* To support different transform_fn for openvino backend

### Related tickets

* 101929

### Tests

* Added test_engine


nncf/experimental/openvino_native/engine.py
```diff
@@ -36,19 +36,21 @@ def __init__(self, model: ov.Model, target_device: TargetDevice = TargetDevice.C
...
         :return output_data: models outputs.
         """
-        model_inputs = {}
-        for name in self.input_names:
-            if name not in input_data:
-                raise RuntimeError(f'Missing a required input: {name} to run the model.')
-            model_inputs[name] = input_data[name]
-
-        model_outputs = self.compiled_model(model_inputs)
+        if len(self.input_names) != len(input_data):
```
andrey-churkin, line 48: Does it correctly works when `input_data` is an `np.ndarray`?

nncf/experimental/openvino_native/engine.py
```diff
@@ -36,19 +36,21 @@ def __init__(self, model: ov.Model, target_device: TargetDevice = TargetDevice.C
...
+            raise RuntimeError(f'Model expects {len(self.input_names)} inputs, but {len(input_data)} are provided.')
+        if isinstance(input_data, dict):
+            for name in self.input_names:
+                if name not in input_data:
+                    raise RuntimeError(f'Missing a required input: {name} to run the model.')
```
andrey-churkin, line 53: I don't think that we need to check all of this. I mean that if something is wrong with `input_data` we will get an error from `compiled_model`, won't we? @alexsu52 What do you think?

tests/openvino/native/test_engine.py
```diff
@@ -0,0 +1,74 @@
...
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+"""
+
+import pytest
+
+import numpy as np
```
andrey-churkin, line 16:
```suggestion
import pytest
import numpy as np
```

tests/openvino/native/models.py
```diff
@@ -93,7 +93,9 @@ def _create_ov_model(self):
         input_2 = opset.parameter([1, 3, 28, 14], name="Input_2")
         multiply = opset.multiply(input_2, 1 / scale, name="Mul")
         add_1 = opset.add(multiply, (-1) * mean, name="Add_1")
-        transpose_fq_input = opset.fake_quantize(add_1, -1, 1, -1, 1, 256, name="Transpose/fq_input_0")
+        # OV bug with element types after fusing preprocessing
+        transpose_fq_input = opset.fake_quantize(add_1,
+             np.float32(-1), np.float32(1), np.float32(-1), np.float32(1), 256, name="Transpose/fq_input_0")
```
KodiaqQ, line 98: IDK, but maybe it would be more convenient to have a method that creates dummy fake quantize nodes with the `float32` values to avoid potential bugs & fuses from the backend's side.

nncf/experimental/openvino_native/engine.py
l-bat: In this case `compiled_model` raise error:
```
RuntimeError: Check '::getPort(port, name, {_impl->GetInputs(), _impl->GetOutputs()})' failed at src/inference/src/cpp/ie_infer_request.cpp:382:
Port for tensor name a was not found.
````

Comment (andrey-churkin) [reply nncf/experimental/openvino_native/engine.py:53]: I got it. It makes sense to have these checks. What do you think about moving the validation code to a separate static method? For example,
```python
@staticmethod
def _check_input_data_format(input_data, input_names):
```
My motivation is to separate logic from checks.

nncf/experimental/openvino_native/engine.py
l-bat: You are right. For `np.ndarray` len returns `np.ndarray.shape[0]`. Thanks for comment!

nncf/experimental/openvino_native/engine.py
alexsu52: I agree.

nncf/experimental/openvino_native/engine.py
```diff
@@ -36,19 +36,34 @@ def __init__(self, model: ov.Model, target_device: TargetDevice = TargetDevice.C
...
+        """
+        Checks correspondence of the model input names and the passed data.
+        If there is a mismatch, the method throws a more specific and readable error than
+        original error raised by the compiled model.
+
+        :param input_data: Provided inputs to infer the model.
+        :param input_names: Model input names.
+        """
+        if isinstance(input_data, np.ndarray) and len(input_names) != 1 or len(input_names) != len(input_data):
+            raise RuntimeError(f'Model expects {len(input_names)} inputs, but {len(input_data)} are provided.')
```
andrey-churkin, line 52:
```suggestion
            actual_num_inputs = 1 if isinstance(input_data, np.ndarray) else len(input_data)
            raise RuntimeError(f'Model expects {len(input_names)} inputs, but {actual_num_inputs} are provided.')
```

nncf/experimental/openvino_native/engine.py
```diff
@@ -36,19 +36,35 @@ def __init__(self, model: ov.Model, target_device: TargetDevice = TargetDevice.C
...
+        Checks correspondence of the model input names and the passed data.
+        If there is a mismatch, the method throws a more specific and readable error than
+        original error raised by the compiled model.
+
+        :param input_data: Provided inputs to infer the model.
+        :param input_names: Model input names.
+        """
+        actual_num_inputs = 1 if isinstance(input_data, np.ndarray) else len(input_data)
+        if actual_num_inputs != len(input_names):
+            raise RuntimeError(f'Model expects {len(input_names)} inputs, but {len(input_data)} are provided.')
```
andrey-churkin, line 53:
```suggestion
            raise RuntimeError(f'Model expects {len(input_names)} inputs, but {actual_num_inputs} are provided.')
```

andrey-churkin (approved):

KodiaqQ (approved):

alexsu52 (approved):
LGTM

nncf/experimental/openvino_native/engine.py
```diff
@@ -36,19 +36,35 @@ def __init__(self, model: ov.Model, target_device: TargetDevice = TargetDevice.C
...
+    def infer(self, input_data: Union[np.ndarray, List[np.ndarray], Tuple[np.ndarray], Dict[str, np.ndarray]]) \
+              -> Dict[str, np.ndarray]:
         """
         Runs model on the provided input via OpenVINO Runtime.
         Returns the dictionary of model outputs by node names.
 
-        :param input_data: inputs for the model.
-        :return output_data: models outputs.
+        :param input_data: Inputs for the model.
+        :return output_data: Models outputs.
```
andrey-churkin, line 66:
```suggestion
        :return output_data: Model's output.
```

PR #1526, opened 2023-01-25
Add prepare_for_inference function for PyTorch models

AlexanderDokuchaev:
### Changes

1. Add `prepare_for_inference` that convert compressed model to inference in PyTorch format without NNCF specific operations.
    - Convert `AsymmetricQuantizer` and `SymmetricQuantizer` to `FakeQuantize`.
    - Apply filter pruning masks to weights. by filing zeroes.
    - Apply sparsity binary masks to weights.
2. The class has also been fixed `ModelPruner`.
3. Fixed `get_scale_zp_from_input_low_input_high`, the zero_point type was incorrectly shifted by -1. 

### Reason for changes

Converting model to OpenVINO format directory.

### Related tickets

CVS-92247

### Tests

- test_converting_symmetric_quantizer
- test_converting_asymmetric_quantizer
- test_prepare_for_inference_quantization
- test_prepare_for_inference_pruning
- test_prepare_for_inference_quantization_and_pruning
- test_save_original_model

Added `data_generators.py` to create input tensors with values in the middle of the quant points. To detect errors in this points that happens by difference in calculation in nncf and torch quantization formulas.


AlexanderDokuchaev:
About the export, there are some problems with the Exporter not passing compression_ctrl (needed to check if the algorithm is supported) and config (to take the "export type" parameter). Any ideas how to pass these parameters without changing all controllers? 
https://github.com/openvinotoolkit/nncf/pull/1526/commits/7f2beeda4c3e300ba8a154261bf5fa07c971a29d

nncf/torch/prepare_for_inference.py
```diff
@@ -0,0 +1,213 @@
...
+        if controller.name == "filter_pruning":
+            graph = compressed_model.get_original_graph()
+            ModelPruner(compressed_model, graph, PT_PRUNING_OPERATOR_METATYPES, PrunType.FILL_ZEROS).prune_model()
+
+    clean_operators(compressed_model)
+
+    return compressed_model
+
+
+def clean_operators(model: NNCFNetwork) -> None:
```
AlexKoff88, line 83: we need to separate what is intend to be used as external API from internal helpers.

nncf/torch/prepare_for_inference.py
```diff
@@ -0,0 +1,213 @@
...
+from nncf.torch.pruning.operations import ModelPruner
+from nncf.torch.pruning.operations import PrunType
+from nncf.torch.quantization.layers import AsymmetricQuantizer
+from nncf.torch.quantization.layers import BaseQuantizer
+from nncf.torch.quantization.layers import SymmetricQuantizer
+
+SUPPORTED_ALGORITHMS = ["quantization", "filter_pruning"]
+
+
+def prepare_for_inference(
```
AlexKoff88, line 33: I would expect to see it as a method of Controller. Potentially, it can replace its .export() API. For other backends, it could be a stub.

tests/common/quantization/data_generators.py
```diff
@@ -0,0 +1,372 @@
...
+                    input_low[idx], input_range[idx], input_size[0:1] + input_size[2:], bits, rtol_for_mid_point
+                )
+                inputs[:, idx] = ch_input
+                is_near_mid_point[:, idx] = ch_is_near_mid_point
+                quant_lens[:, idx] = ch_quant_lens
+
+    return inputs, is_near_mid_point, quant_lens
+
+
+def check_outputs(arr_a: np.array, arr_b: np.array, is_near_mid_point: np.array, quant_lens: np.array, atol=0.000001):
```
AlexKoff88, line 332: This file looks like a mixture of various helper functions. Some of them is used in tests but some in core functionality. It should be split and called appropriately.

AlexanderDokuchaev: There is all functions used in tests. `check_outputs` to compare output, generators to generate test input. 
And `check_outputs` require outputs of generators from this file, because required additional arguments (is_near_mid_point. quant_lens). In my mind, it's looks like parts of one logic and should be in one file.

nncf/torch/composite_compression.py
```diff
@@ -132,3 +133,27 @@ def compression_rate(self) -> float:
                 sum_compression_rate += sum_compression_rate
                 not_none_compression_rate_cnt += 1
         return sum_compression_rate / max(not_none_compression_rate_cnt, 1)
+
+    def prepare_for_inference(self, save_original_model: bool = True) -> NNCFNetwork:
```
AlexanderDokuchaev, line 137: @AlexKoff88 @vshampor What you think about this implementation of api?
`prepare_of_inference.py` removed. Logic moved to algo controller. 
 - https://github.com/openvinotoolkit/nncf/pull/1526/files#diff-98e7d8799a3972a5bc5e857caad2fa52f69f4282d80ac3925d67faa25c11de39R661
 - https://github.com/openvinotoolkit/nncf/pull/1526/files#diff-a5e2ce6b7b41ac95fd0e90302455ad07cef0076e3b2eeba8c4d11319a6b80c8fR1453



But need to use protected-access to "model" property to save instance of the model. 
https://github.com/openvinotoolkit/nncf/pull/1526/files#diff-835a066a208019a262cb822315aa79cb087e6a2d4974a7e8c37b0b4f3e3a043dR153-R155

nncf/torch/composite_compression.py
```diff
@@ -132,3 +133,27 @@ def compression_rate(self) -> float:
...
+        """
+        Prepare NNCFNetwork for inference by converting NNCF modules to torch native format.
+
+        :param save_original_model: `True` means that a copy of the model will be modified.
```
AlexKoff88, line 141: `make_model_copy` instead of `save_original_model` sounds more relevant to me.

nncf/torch/composite_compression.py
AlexKoff88: Does it mean that each algorithm makes a copy of the model that it has a reference to and applies modifications fo this copy? I would expect that copy of the model is made once and each algo does corresponding modifications. @AlexanderDokuchaev, @vshampor, what do you think? is it technically possible within the current architecture?

AlexKoff88: I am ok with this API.

AlexanderDokuchaev: > Does it mean that each algorithm makes a copy of the model that it has a reference to and applies modifications fo this copy?

No, makes copy only once. But to avoid multiple copies of the model, I have to add to a modified protected property in `PTCompositeCompressionAlgorithmController`.


https://github.com/openvinotoolkit/nncf/blob/38a123a3d1fc2867918e4f1ce141e7f15d5092c7/nncf/torch/composite_compression.py#L153-L155

nncf/torch/quantization/prepare_for_inference.py
```diff
@@ -0,0 +1,146 @@
...
+import torch
+from torch.quantization.fake_quantize import FakeQuantize
+
+from nncf.torch.nncf_network import NNCFNetwork
+from nncf.torch.quantization.layers import AsymmetricQuantizer
+from nncf.torch.quantization.layers import BaseQuantizer
+from nncf.torch.quantization.layers import SymmetricQuantizer
+
+
+def replace_quantizer_to_torch_native_module(model: NNCFNetwork) -> None:
```
vshampor, line 25: Make this return the `model`. I know that the function works in-place, but it is more evident that this function modifies the object by making it return the `NNCFNetwork` as if it were a new object.

nncf/torch/quantization/prepare_for_inference.py
```diff
@@ -0,0 +1,146 @@
...
+                    op.op = convert_to_fakequantizer(op.op)
+
+        if hasattr(nncf_module, "post_ops"):
+            for key in list(nncf_module.post_ops.keys()):
+                op = nncf_module.get_post_ops(key)
+                if isinstance(op.op, BaseQuantizer) and op.op.is_enabled_quantization():
+                    op.op = convert_to_fakequantizer(op.op)
+
+
+def convert_to_fakequantizer(nncf_quantizer: BaseQuantizer) -> FakeQuantize:
```
vshampor, line 64: `convert_to_torch_fakequantize`

tests/common/quantization/data_generators.py
```diff
@@ -0,0 +1,372 @@
...
+    :param bits: Number of bits of quantization.
+
+    :return np.array: Length of quant.
+    """
+    input_range = value_high.astype(np.float64) - value_low
+    return get_quant_len_by_range(input_range, bits)
+
+
+def get_quant_points(value_low: np.array, quant_len: np.array, bits: int) -> np.array:
+    """Returns array of quant point for quantization.
```
vshampor, line 43: Need better explanation as to what a `quant point` actually means. `"quant points" are floating-point levels to which the corresponding input floating point values will be brought to after application of fake quantization.`

tests/common/quantization/data_generators.py
```diff
@@ -0,0 +1,372 @@
...
+    :param quant_len: Length of quant.
+    :param bits: Number of bits of quantization.
+
+    :return np.array: Array of quant points.
+    """
+    return np.array([value_low.astype(np.float64) + i * quant_len.astype(np.float64) for i in range(2**bits)])
+
+
+def get_mid_quant_points(value_low: np.array, quant_len: np.array, bits: int) -> np.array:
+    """Returns an array of dots in the middle between quants.
```
vshampor, line 55: `between quant points`

tests/common/quantization/data_generators.py
```diff
@@ -0,0 +1,372 @@
...
+            input_range = np.empty(scales_shape)
+            for idx in range(0, channel_count):
+                input_low[0, idx], input_range[0, idx] = generate_random_low_and_range()
+
+    return input_low, input_range
+
+
+def get_points_near_of_mid_points(input_data: np.array, mid_points: np.array, atol: float = 0.00005) -> np.array:
+    """
+    Get array of where 'True' means that point is in the middle between quants.
```
vshampor, line 194: Same

vshampor (changes requested):

tests/torch/test_prepare_for_inference.py
```diff
@@ -0,0 +1,345 @@
...
+    (16, 8, 12, 12),
+)
+
+
+@pytest.mark.parametrize("use_cuda", (True, False), ids=("cuda", "cpu"))
+@pytest.mark.parametrize("half_range", [False, True], ids=["full_range", "half_range"])
+@pytest.mark.parametrize("is_weights", (True, False), ids=("weights", "activation"))
+@pytest.mark.parametrize("is_signed", (True, False), ids=("signed", "unsigned"))
+@pytest.mark.parametrize("input_size", INPUT_TEST_SCALES, ids=_idfn)
+@pytest.mark.parametrize("is_per_channel", (True, False), ids=("per_channel", "per_tensor"))
```
vshampor, line 115: Do you think that you could reuse this parametrization for the regular NNCF FQs as well so as not to duplicate code and be sure that torch FQs are validated at the same parameter space as NNCF FQs? You can set up a `pytest.fixture` for that and reuse it both in `test_prepare_for_inference.py` and `test_functions.py`

vshampor (changes requested):

tests/torch/test_prepare_for_inference.py
AlexanderDokuchaev: Good idea. But i think better do it in new PR, as this requires a change in non-PR related tests, and can be reused for another tests not only in `test_prepare_for_inference.py` and `test_functions.py`.
But it can changes name of current tests (order of parametrized arguments and ids). It will not be a problem, is it? 

To be clear, do you suggest?

Add to contest.py fixtures for `use_cuda`, `half_range`, `is_weights`, `is_per_channel`, `is_signed`:

```
@pytest.fixture(params=[True, False], ids=["per_channel", "per_tensor"])
def is_per_channel(request):
    return request.param
```

Remove parametrize from tests:

`@pytest.mark.parametrize("is_weights", (True, False), ids=("weights", "activation"))
`

AlexanderDokuchaev:
@AlexKoff88  @vshampor  Fixed comments and update examples by adding `--prepare-for-inference` argument to test and export model with call `prepapare_for_inference`.

AlexanderDokuchaev:
Added support sparsity algorithms.

AlexKoff88:
I am ok with the changes but I wonder how it aligns with the current behavior of TF backend.

vshampor (approved):

alexsu52 (approved):

PR #1533, opened 2023-01-28
Remove unmatched ignored scopes from HF configs

vshampor:
### Changes
As stated in the title.

### Reason for changes
The scopes didn't actually match to anything, and now that we have the runtime checking of such situations the `third_party_sanity` tests have been failing for a while. Some of the ignored scopes referred to GELU activations that were blown up in previous transformers versions, but starting from some point in time they were replaced with proper `gelu` operator calls, so there was no need to work around that with ignored_scopes anymore. Other ignored scopes were written in an old style when  they could be matched partially without regex, which means that they have been invalid for quite a while and it is unknown whether the config with or without them actually produces the claimed quantizer setup and accuracy.

### Related tickets
N/A

### Tests
third_party_sanity


ljaljushkin (approved):

vshampor:
third_party build 32 still shows failures, but no longer due to ignored scopes.

PR #1549, opened 2023-02-02
Added YOLOv8 example

AlexKoff88:
### Changes
- Added YOLOv8n PTQ example

### Reason for changes
- Demonstrate new PTQ API

### DIFF
--- a/examples/post_training_quantization/openvino/yolov8/main.py
+++ b/examples/post_training_quantization/openvino/yolov8/main.py
@@ -139,7 +139,7 @@ def main():
     # Prepare validation dataset and helper
     validator, data_loader = prepare_validation(model, args)
     
-    # COnvert to OpenVINO model
+    # Convert to OpenVINO model
     ov_model, ov_model_path = prepare_openvino_model(model, MODEL_NAME)
     
     # Quantize mode in OpenVINO representation
--- a/examples/post_training_quantization/openvino/yolov8/main.py
+++ b/examples/post_training_quantization/openvino/yolov8/main.py
@@ -13,7 +13,7 @@
 import subprocess
 from tqdm import tqdm
 from pathlib import Path
-from typing import Dict
+from typing import Dict, Tuple, Any
 import re
 
 import numpy as np
@@ -27,9 +27,11 @@ from ultralytics.yolo.utils import DEFAULT_CONFIG
 from ultralytics.yolo.configs import get_config
 from ultralytics.yolo.utils import ops
 from ultralytics.yolo.data.utils import check_dataset_yaml
+from ultralytics.yolo.engine.validator import BaseValidator as Validator
 
 
-def validate(model:ov.Model, data_loader:torch.utils.data.DataLoader, validator, num_samples:int = None, ):
+def validate(model:ov.Model, data_loader:torch.utils.data.DataLoader, 
+             validator: Validator, num_samples:int = None) -> Tuple[Dict, int, int]:
     validator.seen = 0
     validator.jdict = []
     validator.stats = []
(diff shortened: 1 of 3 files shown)


examples/post_training_quantization/openvino/yolov8/main.py
```diff
@@ -0,0 +1,174 @@
...
+
+from ultralytics import YOLO
+from ultralytics.yolo.utils.metrics import ConfusionMatrix
+from ultralytics.yolo.utils import DEFAULT_CONFIG
+from ultralytics.yolo.configs import get_config
+from ultralytics.yolo.utils import ops
+from ultralytics.yolo.data.utils import check_dataset_yaml
+
+
+def validate(model:ov.Model, data_loader:torch.utils.data.DataLoader, validator, num_samples:int = None, ):
```
alexsu52, line 32:
```suggestion
def validate(model:ov.Model, data_loader:torch.utils.data.DataLoader, validator, num_samples:int = None):
```
Please, also update type hints for input arguments and returns here and below.

examples/post_training_quantization/openvino/yolov8/main.py
```diff
@@ -0,0 +1,174 @@
...
+    print(f"Floating-point model performance: {fp_model_perf} FPS")
+    
+    #Benchmark performance of quantized model
+    quantized_model_perf = benchmark_performance(quantized_model_path, args)
+    print(f"Quantized model performance: {quantized_model_perf} FPS")
+    
+    return fp_stats["metrics/mAP50-95(B)"], q_stats["metrics/mAP50-95(B)"], \
+        fp_model_perf, quantized_model_perf
+
+if __name__ == "__main__":
```
alexsu52, line 171: Do we use __main__  in the example?

examples/post_training_quantization/openvino/yolov8/requirements.txt
```diff
@@ -0,0 +1,4 @@
+ultralytics==8.0.5
+onnx
+openvino-dev
+nncf
```
alexsu52, line 4: NNCF is prerequisite for examples, is not it?

https://github.com/openvinotoolkit/nncf/blob/1223ee0c4075cb01d03f77f02f021bb08fd0c3e5/examples/post_training_quantization/openvino/yolov8/README.md?plain=1#L11

AlexKoff88: I would remove this text rather than nncf from the requirements.txt. From the user standpoint, it should be much more friendly to run only "pip install -r requirements.txt" rather then navigate to another document and read how to install NNCF.

examples/post_training_quantization/openvino/yolov8/main.py
AlexKoff88: The plan is to import the main from the example and use it in the test to validate this on regular basis.

examples/post_training_quantization/openvino/yolov8/main.py
AlexKoff88: Of course, I can do that but I am not sure about the added value of that for the example and the user.

examples/post_training_quantization/openvino/yolov8/requirements.txt
alexsu52: The example is part of the NNCF repository and is valid for current commit. nncf in the requirements.txt can be lead to use nncf version from the pypi instead of repository version. I think the dual behavior of setup of example's envariment It's not good practices.

AlexKoff88: Changes according to agreements

alexsu52 (approved):

PR #1559, opened 2023-02-08
[PTQ] Update logic of transformation usage in ModelTransformer

kshpv:
### Changes

In new logic ModelTransformer should be created once for particular model and be used to apply transformations directly on the provided model. ModelTransformer returns a new model with applied transformations.
ModelTransformer makes a deepcopy of the model only in the case of inline transformations. In case of creating a new model object - the original model will not be copied, improving memory efficiency.
[ONNX]: NNCFGraph is no longer built for every transformation onto the same model.
[ONNX]: Improve code style
Applied logic to FBC algorithm results in 20-30% speedup for ONNX backend.

### Reason for changes

Improving memory efficiency and reducing the execution time of the FBC.

### Related tickets

102768


kshpv:
run openvino pre-commit tests

kshpv:
Closed due to the commit history being broken. Create a new PR - https://github.com/openvinotoolkit/nncf/pull/1569

PR #1563, opened 2023-02-09
Support for the OpenVINO native quantization implementation in calibrate.py

alexsu52:
### Changes

Added support for the OpenVINO native quantization implementation in calibrate.py

### Reason for changes

Testing

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -26,7 +26,11 @@ from nncf.common.quantization.structs import QuantizationPreset
 from nncf.data.dataset import Dataset
 from nncf.experimental.openvino_native.quantization.quantize import \
     quantize_impl
-from nncf.parameters import IgnoredScope, ModelType, TargetDevice
+from nncf.parameters import (
+    IgnoredScope,
+    ModelType,
+    TargetDevice
+)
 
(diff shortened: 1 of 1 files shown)


tests/openvino/tools/calibrate.py
```diff
@@ -14,13 +14,21 @@
...
 from openvino.tools.accuracy_checker.evaluators.quantization_model_evaluator import \
     create_model_evaluator
 from openvino.tools.pot.configs.config import Config
 
 import nncf
+from nncf.common.quantization.structs import QuantizationPreset
+from nncf.data.dataset import Dataset
+from nncf.experimental.openvino_native.quantization.quantize import \
+    quantize_impl
+from nncf.parameters import IgnoredScope, ModelType, TargetDevice
```
KodiaqQ, line 29: Shouldn't this be formatted as in https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#22-imports?

KodiaqQ (approved):

tests/openvino/tools/calibrate.py
alexsu52: done

PR #1564, opened 2023-02-09
Fix ignored scope for weights & OV

KodiaqQ:
### Changes

- Fixed ignored scope for weights
- Fixed OpenVINO outputs creation in case of ignored scope

### Reason for changes

- Features alignment

### Related tickets

- 102187

### DIFF
--- a/nncf/experimental/openvino_native/graph/model_transformer.py
+++ b/nncf/experimental/openvino_native/graph/model_transformer.py
@@ -11,7 +11,7 @@
  limitations under the License.
 """
 
-from typing import List
+from typing import List, Tuple
 from enum import Enum
 
 import openvino.runtime as ov
@@ -111,12 +111,12 @@ class OVModelTransformer(ModelTransformer):
         self._model = self._insert_outputs(self._model, outputs=extra_model_outputs)
 
     def _get_extra_model_outputs(self,
-                                 transformations: List[OVOutputInsertionCommand]) -> List[ov.Output]:
+                                 transformations: List[OVOutputInsertionCommand]) -> List[Tuple[ov.Output, int]]:
         """
         Collects extra model outputs based on transformations.
 
         :param transformations: lisf of the OVOutputInsertionCommand.
-        :return: list of the output names.
+        :return: list of tuples with ov.Output & port_ids.
         """
         extra_model_outputs = []
         for transformation in transformations:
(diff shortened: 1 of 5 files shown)


nncf/experimental/openvino_native/graph/model_transformer.py
```diff
@@ -111,12 +111,12 @@ def _apply_output_insertion_transformations(self, transformations: List[OVOutput
...
 
     def _get_extra_model_outputs(self,
-                                 transformations: List[OVOutputInsertionCommand]) -> List[ov.Output]:
+                                 transformations: List[OVOutputInsertionCommand]) -> List[Tuple[ov.Output, int]]:
         """
         Collects extra model outputs based on transformations.
 
         :param transformations: lisf of the OVOutputInsertionCommand.
-        :return: list of the output names.
+        :return: list of tuples with ov.Output & port_ids.
```
l-bat, line 119:
```suggestion
        :return: list of tuples with ov.Output & port_id.
```

nncf/experimental/openvino_native/graph/model_transformer.py
```diff
@@ -125,30 +125,29 @@ def _get_extra_model_outputs(self,
...
         :return: modified model.
         """
         model_outputs = model.get_results()
         params = model.get_parameters()
         extra_model_outputs = []
-        for output in outputs:
+        for (output, port_id) in outputs:
             output_name = output.get_node().get_friendly_name()
-            port_id = output.get_index()
             result = opset.result(output, name=f'Result_{output_name}.{port_id}')
```
l-bat, line 151: I think we should support nodes with several outputs or add TODO here and open a ticket.

alexsu52:
@l-bat, please, review.

nncf/experimental/openvino_native/graph/model_transformer.py
KodiaqQ: Ticket - 103787

l-bat (approved):

kshpv:
This is a workaround for each backend, while the main fix should be done in Solver. Could you create a ticket on that?

KodiaqQ:
> This is a workaround for each backend, while the main fix should be done in Solver. Could you create a ticket on that?

@kshpv, what do you mean while saying `workaround`? As I understand, the MinMaxQuantization pipeline was incorrect at the beginning. Please, look at the pipeline in the PyTorch backend, where is ignored scope for weights filters outside the solver - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/quantization/algo.py#L265
Fixing the Solver is not a part of this PR.

kshpv:
> > This is a workaround for each backend, while the main fix should be done in Solver. Could you create a ticket on that?
> 
> @kshpv, what do you mean while saying `workaround`? As I understand, the MinMaxQuantization pipeline was incorrect at the beginning. Please, look at the pipeline in the PyTorch backend, where is ignored scope for weights filters outside the solver - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/quantization/algo.py#L265 Fixing the Solver is not a part of this PR.

We have duplicate code with ignored_scopes. First, we do filtration at each backends, then we do the same filtration by Solver. For me it doesn't make sene. While the filtration step is the same just by the name matching of the nodes.

KodiaqQ:
> We have duplicate code with ignored_scopes. First, we do filtration at each backend, then we do the same filtration by Solver. While the filtration step is the same just by the name matching of the nodes. For me, it doesn't make sense and I called this a workaround.

This is not a workaround, but a missing in Solver's design. Solver doesn't work with the quantizers for weights in terms of ignored scope. The filtration for the activation quantizers is a bit more complicated than for weights and pipelines are not the same.

KodiaqQ:
> There is no backend specific in the filtration of weight nodes anyway

Please, address your thoughts in the ticket and prioritize them related to our plan. I think this PR is not the place and time to make Solver's refactoring.

kshpv:
> > There is no backend specific in the filtration of weight nodes anyway
> 
> Please, address your thoughts in the ticket and prioritize them related to our plan. I think this PR is not the place and time to make Solver's refactoring.

103796

kshpv (approved):

alexsu52 (approved):

PR #1565, opened 2023-02-09
Disallow hooking in-place ops

vshampor:
### Changes
In-place ops in PyTorch will not allow setting up pre- and post-hooks for themselves. The detection of the in-place op is done via naming convention.

### Reason for changes
Quantizers cannot end up on pre- and post-hooks of in-place ops, because quantizers themselves are currently not in-place and therefore a) the result of quantization is never used, and b) the compressed graph breaks down.

### Related tickets
76386

### Tests
tests/torch/test_compressed_graph.py::test_synthetic_model_quantization[InplaceThenQuantizableOpModel]

### DIFF
--- a/nncf/common/insertion_point_graph.py
+++ b/nncf/common/insertion_point_graph.py
@@ -40,7 +40,7 @@ class PreHookInsertionPoint:
         self.input_port_id = input_port_id
 
     def __str__(self):
-        return str(self.input_port_id) + ' ' + self.target_node_name
+        return f"PreHookInsertionPoint {self.target_node_name} input_port={str(self.input_port_id)}"
 
 
 class PostHookInsertionPoint:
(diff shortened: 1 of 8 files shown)


ljaljushkin (approved):

vshampor:
Sadly, this approach leads to torch ResNets being quantized sub-optimally - the residual connections now have two FQs after the in-place `relu_` instead of one:
image
At the same time the in-place `relu_` in torch is actually used like a non-inplace op in terms of reference binding:
image

and setting up post-hooks at this exact relu_ usage actually works. Quantizing ResNet18 in the way that is shown on the picture is probably unacceptable for us, so we have the right way to do things that leads to bad results, and a wrong way to do things that functionally works.

vshampor:
I think we need to still keep this as a draft until we either come up with a way to intercept in-place ops (e.g. by implementing an in-place version of quantization hooks), or until we are sure that the extra FQs introduced by this solution are not that important from the performance standpoint (or that these extra FQs are successfully optimized out in the OV runtime). In the latter case we can take this draft, pull it to the latest state of the develop and merge it.

PR #1566, opened 2023-02-09
Migrate torch fused patterns to common representation

daniil-lyakhov:
### Changes

Torch fused patterns migrated to new API

### Reason for changes

Unified processing of fused patterns allows to use common code in PTQ algorithms (MinMax)

### Tests
`test_pattern_manager` for Torch


nncf/torch/hardware/fused_patterns.py
```diff
@@ -13,49 +30,124 @@
...
+    pattern = GraphPattern()
+    pattern.add_pattern_alternative(bn_act)
+    pattern.add_pattern_alternative(act_bn)
+    pattern.add_pattern_alternative(batch_norm)
+    pattern.add_pattern_alternative(activations)
+    return pattern
 
 
-PT_HW_FUSED_PATTERNS = _get_torch_hw_fused_patterns()
+def get_torch_hw_patterns():
```
KodiaqQ, line 152: Why do we need this method? Why not use `PatternsManager` as is?

nncf/torch/hardware/fused_patterns.py
```diff
@@ -13,49 +30,124 @@
...
-    batch_norm_activations_permutation = batch_norm + activations | activations + batch_norm | batch_norm | activations
 
-    retval.register(linear_ops + arithmetic_ops, 'LINEAR + ARITHMETIC', match=True)
-    retval.register(linear_ops + batch_norm_activations_permutation, 'LINEAR + BN_ACT_PERM',
-                    match=True)
-    retval.register(batch_norm + activations, 'BN + ACTIVATIONS', match=True)
-    retval.register(activations + batch_norm, 'ACTIVATIONS + BN', match=True)
-    retval.register(arithmetic_ops + batch_norm_activations_permutation,
-                    'ARITHMETIC + BN_ACT_PERM', match=True)
+def batch_norm_activations_permutations():
```
KodiaqQ, line 132: My suggestion is not to add so-called `permutations` of the patterns. This method contains the list of existing combinations such as `batch norm + activations`, `activations + batch norm` and etc., that should be described as BATCH_NORM_ACTIVATIONS and corresponding patterns.

nncf/torch/hardware/fused_patterns.py
daniil-lyakhov: I did it to simplify code inside torch algorithms: instead of importing BackendType, TargetDevice  and using them every time I made a default version. I believe this approach is easier to understand as Torch has no target specific patterns and logic yet

alexsu52:
@KodiaqQ, please, review.

nncf/torch/hardware/fused_patterns.py
daniil-lyakhov: Done, please check

nncf/experimental/torch/search_building_blocks/search_graph.py
```diff
@@ -261,7 +263,8 @@ def get_merged_original_graph_with_pattern(orig_graph: nx.DiGraph, hw_fused_ops:
     if not hw_fused_ops:
         return merged_graph
     # pylint: disable=protected-access
-    pattern_fusing_graph = PT_HW_FUSED_PATTERNS.get_full_pattern_graph()
+    pattern_fusing_graph =\
+        PatternsManager().get_full_pattern_graph(BackendType.TORCH, TargetDevice.ANY)
```
KodiaqQ, line 267: Can these patterns be applied only for ANY (CPU) device?

nncf/torch/hardware/fused_patterns.py
```diff
@@ -13,49 +27,148 @@
...
-    h_swish = create_h_swish_act()
-    activations = atomic_activations | swish | h_swish | h_sigmoid
-    retval.register(activations, 'ACTIVATIONS', match=False)
 
-    arithmetic_ops = GraphPattern()
-    arithmetic_ops.add_node(**ARITHMETIC_OPERATIONS)
-    retval.register(arithmetic_ops, ARITHMETIC_OPERATIONS['label'], match=False)
+@PT_HW_FUSED_PATTERNS.register(PatternNames.L2_NORM)
+def create_l2_norm_operations():
+    return create_l2_norm()
```
KodiaqQ, line 37: Can we reduce the number of files and replace methods from the `nncf.torch.graph.patterns` directly here? Since they are not used anymore, but only here.

nncf/torch/graph/pattern_operations.py
```diff
@@ -11,68 +11,69 @@
  limitations under the License.
 """
 from nncf.common.graph.patterns import merge_two_types_of_operations
+from nncf.common.graph.patterns import GraphPattern
 
-LINEAR_OPERATIONS = {'type': ['linear',
```
kshpv, line 15: I think we should update patterns on Metatypes in this PR how it did in ONNX and OV

kshpv:
I would suggest to add @vshampor to review torch patterns

nncf/common/graph/patterns/patterns.py
```diff
@@ -281,6 +281,9 @@ class PatternNames(Enum):
     and would not be quantized in compression pipeline.
     """
 
+    # ATOMIC OPERATIONS
+    L2_NORM = PatternDesc('l2_norm')
```
kshpv, line 285: Can we somehow divide Torch specific patterns from common patterns?

kshpv (changes requested):

nncf/torch/quantization/algo.py
```diff
@@ -347,7 +349,8 @@ def __init__(self, quant_config: Dict, target_model: NNCFNetwork,
                          precision_init_params, range_init_params,
                          hw_config)
 
-        self._pattern_fusing_graph = PT_HW_FUSED_PATTERNS.get_full_pattern_graph()
+        self._pattern_fusing_graph =\
+            PatternsManager().get_full_pattern_graph(BackendType.TORCH, TargetDevice.ANY)
```
KodiaqQ, line 353: Please, use the predefined `TargetDevice`, since the quantization is possible not only for `ANY` device.

nncf/common/graph/patterns/patterns.py
kshpv, line 285: We already have NORMALIZE_L2_MULTIPLY. What is the diff between them?

nncf/torch/graph/pattern_operations.py
daniil-lyakhov: The problem is that in contrast with ONNX backend, Torch backend uses operation names instead of metatypes for pattern matching. I suggest to cover this in a separate PR

nncf/experimental/torch/search_building_blocks/search_graph.py
daniil-lyakhov: Torch has no device specific logic in patterns

nncf/common/graph/patterns/patterns.py
daniil-lyakhov: I'll check it

tests/torch/test_nncf_network.py
```diff
@@ -622,7 +624,8 @@ def forward(self, input_):
     def test_get_ip_graph_with_merged_operations(self, mock_graph_factory, dot_file_name):
         mock_graph = mock_graph_factory()
         ip_graph = get_ip_graph_for_test(mock_graph)
-        pattern = PT_HW_FUSED_PATTERNS.get_full_pattern_graph()
+        pattern =\
+            PatternsManager().get_full_pattern_graph(BackendType.TORCH, TargetDevice.ANY)
```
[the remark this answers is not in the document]
daniil-lyakhov: I copied it from MinMax, I'll check

nncf/torch/quantization/metrics.py
```diff
@@ -329,7 +331,8 @@ def _marking_edges(self, graph, node_key, queue, mark=True):
                 self.stats.quantized_edges_in_cfg += 1
 
     def get_merged_original_graph_with_patterns(self, original_graph: PTNNCFGraph):
-        pattern = PT_HW_FUSED_PATTERNS.get_full_pattern_graph()
+        pattern =\
+            PatternsManager().get_full_pattern_graph(BackendType.TORCH, TargetDevice.ANY)
```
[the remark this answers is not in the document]
daniil-lyakhov: I afraid that device param wasn't indented to be here, I probably need to refactor whole quantization algo to pass this param which at the end will change nothing. My suggestion is: let's refactor this when device specific patterns will be introduced in Torch backend

nncf/torch/quantization/algo.py
daniil-lyakhov: Please see https://github.com/openvinotoolkit/nncf/pull/1566/files/01b9ab9f9143ec0361290f5bb322521e4bdf0f83#r1105622381

nncf/experimental/torch/search_building_blocks/search_graph.py
KodiaqQ: > Torch has no device specific logic in patterns

Do you mean quantization in general or some specific algorithms that use this method?

daniil-lyakhov:
> I would suggest to add @vshampor to review torch patterns

This code doesn't change any logic inside torch algorithms, it's just using different representation. I suggest to keep this PR as small as possible, and when some fussed pattern logic will have to be hanged we will refactor this code

nncf/torch/quantization/metrics.py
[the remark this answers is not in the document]
KodiaqQ: But the `MinMaxQuantization` algorithm has a `device` as an option. Or are these files not used in the `MinMaxQuantization` pipeline?

nncf/torch/graph/pattern_operations.py
kshpv: I know it, but the think that you need to do is only to change the name of the layer to Metatype.

kshpv:
> > I would suggest to add @vshampor to review torch patterns
> 
> This code doesn't change any logic inside torch algorithms, it's just using different representation. I suggest to keep this PR as small as possible, and when some fussed pattern logic will have to be hanged we will refactor this code

As @vshampor added most of the patterns, I suggest adding him to double check that all patterns are registered

nncf/experimental/torch/search_building_blocks/search_graph.py
daniil-lyakhov: There are no algorithms in NNCF Torch that's using information about device when collecting fussing pattern. Fussing patterns are always the same

KodiaqQ: @alexsu52, @AlexKoff88, is the statement above correct? Should PyTorch PTQ quantization use types of devices that differ from ANY?

nncf/torch/quantization/metrics.py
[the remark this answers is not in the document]
daniil-lyakhov: This PR is not related to MinMaxQuantization pipeline, this PR converts existing patterns to a new representation.
Besides no matter device Torch pattern manager will return the same set of patterns, so it's safe to use in `MinMaxQuantization`

nncf/common/graph/patterns/patterns.py
daniil-lyakhov: It's differs from Openvino `NORMALIZE_L2_MULTIPLY`

nncf/torch/graph/pattern_operations.py
daniil-lyakhov: What do you mean by the new API has patterns only on Metatypes? Could you please show me this new API?

tests/torch/test_nncf_network.py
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/torch/hardware/fused_patterns.py
daniil-lyakhov: Done, please check

nncf/torch/graph/pattern_operations.py
daniil-lyakhov: I want to use torch patterns inside `MinMaxQuantization`, and for that I don't need to change types here. I would do this in a separate PR with pattern matching algorithm refactoring

kshpv: Please, look at this PR - https://github.com/openvinotoolkit/nncf/commit/54540ba1e1f50ad961c8cb3e2b285f361b332ac7#diff-62c24dee1521c104745fab8ceee4a346d952825ea04e70152033a1ceb78265afL16
This is how the migration of patterns was done for ONNX.

All matching logic already is implemented.  About what refactoring did you write?

kshpv: > This method, for example, could be simplified when all backends will be migrated on metatypes https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/graph/graph_matching.py#L76-L95

This is true, but it should be done only when ALL backends are migrated. If we add only Torch I don't see any refactoring in matching logic

kshpv (approved):

nncf/experimental/torch/search_building_blocks/search_graph.py
```diff
@@ -261,7 +263,8 @@ def get_merged_original_graph_with_pattern(orig_graph: nx.DiGraph, hw_fused_ops:
     if not hw_fused_ops:
         return merged_graph
     # pylint: disable=protected-access
-    pattern_fusing_graph = PT_HW_FUSED_PATTERNS.get_full_pattern_graph()
+    pattern_fusing_graph =\
+        PatternsManager.get_full_pattern_graph(BackendType.TORCH, TargetDevice.ANY)
```
daniil-lyakhov, line 267: I suggest to set proper target device here in a separate PR

daniil-lyakhov: CC: @ljaljushkin

KodiaqQ (approved):

vshampor (approved):

PR #1573, opened 2023-02-13
Added patching of torch.jit.script_if_tracing

nikita-savelyevv:
The fact of importing of `nncf.torch` can lead to `torch.onnx.export` failing if the model being exported internally relies on [torch.jit.script_if_tracing](https://pytorch.org/docs/stable/generated/torch.jit.script_if_tracing.html). This is due to patching issues.

### Changes

- Added patching of torch.jit._script_if_tracing
- Added a couple of tests

### DIFF
--- a/nncf/torch/dynamic_graph/patch_pytorch.py
+++ b/nncf/torch/dynamic_graph/patch_pytorch.py
@@ -128,17 +128,17 @@ def torch_jit_script_wrapper(*args, **kwargs):
 
 
 def torch_jit_script_if_tracing(fn):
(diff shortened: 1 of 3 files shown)


vshampor (approved):

ljaljushkin:
Do we still need `torch_jit_script_wrapper`?

nikita-savelyevv:
> Do we still need `torch_jit_script_wrapper`?

Yes. Without it `torch.jit.script` wouldn't be patched.

ljaljushkin (approved):

PR #1556, opened 2023-02-07
Conformance test with Process

KodiaqQ:
### Changes

- Updated conformance test with Process while validation to avoid main process failure

### Reason for changes

- More robust conformance test

PR #1576, opened 2023-02-15
[PTQ][Tests] Remove progress bar during val step

daniil-lyakhov:
### Changes

Progress bar is removed from validation step of test_quantize_conformance

### Reason for changes

Progress dumped during CI validation makes output logs big and unreadable


KodiaqQ:
The same changes with https://github.com/openvinotoolkit/nncf/pull/1556

PR #1580, opened 2023-02-16
add nncf.quantize_with_accuracy_control and test with yolov8n-seg mod…

yangsu2022:
…el and coco128-seg dataset

### Changes
add nncf.quantize_with_accuracy_control 
run with yolov8n-seg model and coco128-seg dataset

### Reason for changes
extend the POT Yolov8 sample with accuracy control and segmentation

### DIFF
--- a/examples/post_training_quantization/openvino/mobilenet_v2/main.py
+++ b/examples/post_training_quantization/openvino/mobilenet_v2/main.py
@@ -35,13 +35,13 @@ DATASET_CLASSES = 10
 
 
 def download(url: str, path: str) -> Path:
-    downloader = FastDownload(base=path, 
-                              archive='downloaded', 
+    downloader = FastDownload(base=path,
+                              archive='downloaded',
                               data='extracted')
     return downloader.get(url)
 
 
-def validate(model: ov.Model, 
+def validate(model: ov.Model,
              val_loader: torch.utils.data.DataLoader) -> float:
     predictions = []
     references = []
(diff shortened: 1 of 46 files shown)


MaximProshin:
@AlexKoff88 , @alexsu52 , please take a look

alexsu52 (commented):
I would suggest to add command line parameter 'max_acc_drop' to the main.py sccript to feed the maximum accuracy drop value and identify that the quantize_with_accuracy_control should be run. Or do you prefer to make it a separate example? @yangsu2022 @AlexKoff88 what do you think?

yangsu2022:
> I would suggest to add command line parameter 'max_acc_drop' to the main.py sccript to feed the maximum accuracy drop value and identify that the quantize_with_accuracy_control should be run. Or do you prefer to make it a separate example? @yangsu2022 @AlexKoff88 what do you think?

How about a separate script in the same Yolov8 directory? @alexsu52 @AlexKoff88

AlexKoff88:
> > I would suggest to add command line parameter 'max_acc_drop' to the main.py sccript to feed the maximum accuracy drop value and identify that the quantize_with_accuracy_control should be run. Or do you prefer to make it a separate example? @yangsu2022 @AlexKoff88 what do you think?
> 
> How about a separate script in the same Yolov8 directory? @alexsu52 @AlexKoff88

Hi, @yangsu2022. 

First of all, thanks for the contribution!

To me, having two separate scripts make sense from the examples' perspective because two APIs require a different level of effort to be used. I more care about the value here.
Are you able to see the benefit from the accuracy-aware quantization when using 0.02 (2%) of the accuracy drop criterium? I believe that we can quantize this model by default and satisfy it without the need for accuracy-aware quantization.
Can you also please share an insight where this criterium comes from?

yangsu2022:
> 
Thanks, @AlexKoff88,
0.02 is just used to avoid the fallback in my test. Here, 0.005 would be more meaningful.

AlexKoff88:
> > 
> 
> Thanks, @AlexKoff88, 0.02 is just used to avoid the fallback in my test. Here, 0.005 would be more meaningful.

Thanks, @yangsu2022! Your PR is in a good shape, IMO. A couple of minor comments from my side:
- Can you please remove the commented code?
- I would suggest making accuracy drop a constant in the example. This change will allow us to validate this sample on a regular basis because it will be run as `python main.py` and we will make sure that other changes will not break it in the future.

PR #867, opened 2021-07-22
Allow to setup compression modules for the training-only modules

ljaljushkin:
It's needed for elastic width in case of Aux branches, e.g. in InceptionV3.
We still don't need to compress these branches, but we need Mixin modules with pre_ops to propagate changes in the width.

Suggested approach - simple replace eval graph with train one and mark all training-only nodes to filter them in all previous algorithms. 

**Why is it draft?** 
There're some unresolved issues with statistics and setting compression. Looks like some compression is set for training only layers. Also wrapping training-only layers leads to some issues with LSTM.

PR #1584, opened 2023-02-17
Rework NNCFNetwork for better integration

vshampor:


docs/Usage.md
```diff
@@ -268,7 +268,7 @@ NNCF has the capability to apply the model compression algorithms while satisfyi
 The following function is required to create the accuracy-aware training loop. One has to pass the `NNCFConfig` object and the compression controller (that is returned upon compressed model creation, see above).
 ```python
 from nncf.common.accuracy_aware_training import create_accuracy_aware_training_loop
-training_loop = create_accuracy_aware_training_loop(nncf_config, compression_ctrl)
+training_loop = create_accuracy_aware_training_loop(nncf_config, compression_ctrl, original_model_accuracy)
```
vshampor, line 271: Note that in the scope of this PR I had to rework the approach of passing the original model accuracy to the AA algos. It used to be (as a bad design decision) stored in the `NNCFNetwork` as an attribute, and also as an attribute to TF model objects - basically this means that AA functionality leaked into the NNCFNetwork object even if AA was not used. Now AA functions accept the original model accuracy as an argument where required, and in some places measure the original model accuracy based on config-defined dataloaders, as they should, instead of relying on an attribute set on the model object by someone else in some possibly unrelated part of code.

nncf/common/accuracy_aware_training/runner.py
```diff
@@ -453,7 +448,8 @@ def update_training_history(self, compression_rate, metric_value):
         self._compressed_training_history.append((compression_rate, accuracy_budget))
 
         if IMG_PACKAGES_AVAILABLE:
-            plt.figure()
+            plt.ioff()
```
vshampor, line 451: AA algo also spawned a lot of figures in an interactive mode, leading to bad UX. I fixed it.

nncf/experimental/torch/sparsity/movement/structured_mask_strategy.py
```diff
@@ -32,7 +32,7 @@ def detect_supported_model_family(model: NNCFNetwork) -> Optional[str]:
     :return: The string name of the model family that `model` belongs to. If `model` is not supported,
         then returns None.
     """
-    model_pymodules = inspect.getmodule(model.get_nncf_wrapped_model()).__name__.split('.')
```
vshampor, line 35: Now the compressed model object and the original object are one and the same object, so there is no need for additional calls here and below when we need to query the original model for something.

nncf/torch/checkpoint_loading.py
```diff
@@ -228,8 +228,8 @@ def _clip_keys_without_collisions(keys: List[str], keys_to_ignore: List[str]) ->
     @staticmethod
     def _key_clipper(key: str) -> str:
         new_key = key
-        from nncf.torch.nncf_network import MODEL_WRAPPED_BY_NNCF_ATTR_NAME #pylint: disable=cyclic-import
-        clip_patterns = [MODEL_WRAPPED_BY_NNCF_ATTR_NAME + '.', 'module.', '|OUTPUT', '|INPUT']
+        from nncf.torch.nncf_network import LEGACY_MODEL_WRAPPED_BY_NNCF_ATTR_NAME #pylint: disable=cyclic-import
```
vshampor, line 231: Internal structure of the compressed model object has changed with this PR, and the checkpoint format basically changes too. This code provides the functionality of loading old-style checkpoints with a deprecation warning, urging the user to re-save their checkpoint, which will then save the checkpoint in the new format.

nncf/torch/dynamic_graph/context.py
```diff
@@ -34,7 +34,12 @@
 from nncf.torch.dynamic_graph.trace_tensor import TensorMeta
 from nncf.common.utils.patcher import PATCHER
 
-_CURRENT_CONTEXT = None
+class ThreadLocalGlobalContext(threading.local):
+    def __init__(self):
+        super().__init__()
+        self.context = None
+
+_CURRENT_CONTEXT = ThreadLocalGlobalContext()
```
vshampor, line 42: While working on this PR, I fixed an obscure bug, making the NNCFNetwork actually thread-safe. Before the fix, `DataParallel` only worked due to sheer luck.

nncf/torch/exporter.py
```diff
@@ -128,40 +128,37 @@ def _export_to_onnx(self, save_path: str, opset_version: int) -> None:
...
         input_tensor_list = []
-        for info in self._model.input_infos:
+        for info in self._model.nncf.input_infos:
             single_batch_info = copy(info)
             input_shape = tuple([1] + list(info.shape)[1:])
             single_batch_info.shape = input_shape
             input_tensor_list.append(create_mock_tensor(single_batch_info, 'cpu'))
 
-        original_forward = model.forward
+        full_arg_forward = model.nncf.get_original_forward()
```
vshampor, line 137: The original forward function can now be retrieved using a well-defined function. The `.forward` on the NNCFNetwork function will always be an NNCF-wrapped `forward` version.

nncf/torch/knowledge_distillation/knowledge_distillation_loss.py
```diff
@@ -92,6 +96,19 @@ def match_fn(obj):
...
+                                f"and compressed models;\n"
+                                f"original has {len(orig_model_loss_outputs)} loss tensors,\n"
+                                f"compressed has {len(compressed_model_loss_outputs)}")
+        if not orig_model_loss_outputs:
+            nncf_logger.warning("KD: no loss outputs detected in original model, knowledge distillation not possible")
+            return None
+        if not compressed_model_loss_outputs:
+            nncf_logger.warning("KD: no loss outputs detected in compressed model, "
+                                "knowledge distillation not possible")
+            return None
```
vshampor, line 111: An improvement that I found useful not only when debugging this PR, but which could actually help users and future us to pinpoint what went wrong in the KD application.

nncf/torch/model_creation.py
```diff
@@ -171,34 +170,26 @@ def create_nncf_network(model,
...
-                               wrap_inputs_fn=wrap_inputs_fn,
-                               wrap_outputs_fn=wrap_outputs_fn,
-                               ignored_scopes=ignored_scopes,
-                               target_scopes=target_scopes,
-                               scopes_without_shape_matching=scopes_without_shape_matching,
-                               original_model_accuracy=original_model_accuracy)
-
-    nncf_network.get_tracing_context().disable_trace_dynamic_graph()
+    # Preserve `.training`/`.requires_grad` state since we will be building NNCFNetwork in `.eval` mode
+    with training_mode_switcher(model, is_training=False):
```
vshampor, line 174: @ljaljushkin please comment whether or not this impacts NAS approach. Maybe we need to extend NNCFNetwork with an option to build train-mode or eval-mode graphs based on a parameter, if NAS needs the training-mode graph to work.

tests/torch/data/reference_graphs/const_sparsity/inception_v3.dot
```diff
@@ -311,129 +311,118 @@ strict digraph  {
 "309 Inception3/InceptionC[Mixed_6e]/BasicConv2d[branch_pool]/NNCFBatchNorm2d[bn]/batch_norm_0" [id=309, type=batch_norm];
 "310 Inception3/InceptionC[Mixed_6e]/BasicConv2d[branch_pool]/relu__0" [id=310, type=relu_];
 "311 Inception3/InceptionC[Mixed_6e]/cat_0" [id=311, type=cat];
-"312 Inception3/InceptionAux[AuxLogits]/avg_pool2d_0" [id=312, type=avg_pool2d];
```
vshampor, line 314: In Inceptionv3, the aux branch is gone since we are now building graphs for the eval mode.

tests/torch/data/reference_graphs/const_sparsity/ssd_mobilenet.dot
```diff
@@ -420,16 +416,12 @@ strict digraph  {
...
 "201 MobileNetSSD/SSDDetectionOutput[detection_head]/view_8" -> "205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1";
 "202 MobileNetSSD/SSDDetectionOutput[detection_head]/view_9" -> "205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1";
 "203 MobileNetSSD/SSDDetectionOutput[detection_head]/view_10" -> "205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1";
 "204 MobileNetSSD/SSDDetectionOutput[detection_head]/view_11" -> "205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1";
 "205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1" -> "206 MobileNetSSD/SSDDetectionOutput[detection_head]/view_12";
-"205 MobileNetSSD/SSDDetectionOutput[detection_head]/cat_1" -> "209 MobileNetSSD/SSDDetectionOutput[detection_head]/view_14";
 "206 MobileNetSSD/SSDDetectionOutput[detection_head]/view_12" -> "207 MobileNetSSD/SSDDetectionOutput[detection_head]/softmax_0";
-"208 MobileNetSSD/SSDDetectionOutput[detection_head]/view_13" -> "210 /nncf_model_output_0";
-"209 MobileNetSSD/SSDDetectionOutput[detection_head]/view_14" -> "211 /nncf_model_output_1";
```
vshampor, line 434: Same in SSD, where the detection output in eval mode is actually under `with no_nncf_trace()` and therefore won't be present in the graph.

tests/torch/data/reference_graphs/quantized/asymmetric/densenet121.dot
```diff
@@ -726,9 +726,9 @@ strict digraph  {
 "724 DenseNet/Sequential[dense4]/Bottleneck[15]/cat_0" [id=724, type=cat];
 "725 DenseNet/NNCFBatchNorm2d[bn]/batch_norm_0" [id=725, type=batch_norm];
 "726 DenseNet/relu_0" [id=726, type=relu];
-"727 DenseNet/AsymmetricQuantizer/asymmetric_quantize_0" [id=727, type=asymmetric_quantize];
+"727 DenseNet/NNCFInterface[_nncf]/ModuleDict[external_quantizers]/AsymmetricQuantizer[DenseNet/relu_0|OUTPUT]/asymmetric_quantize_0" [id=727, type=asymmetric_quantize];
```
vshampor, line 729: Note that now the label of activation quantizers is more explicitly stated. The rest of the changes to the reference graphs are all because of this change - no structural changes in any except for the Inceptionv3 and SSD due to reasons outlined above.

tests/torch/nas/creators.py
```diff
@@ -86,6 +86,7 @@ def create_bnas_model_and_ctrl_by_test_desc(desc: MultiElasticityTestDesc):
 
     nncf_config = NNCFConfig.from_dict(config)
     model = desc.model_creator()
+    model.eval()
```
vshampor, line 89: @ljaljushkin now the graph building in NNCFNetwork preserves the train/eval state of the original module, while before it forced the model into eval. As a result, the NAS tests were written with implicit reliance on the model being in `.eval()` state. What is worse that the test actually fails in the `.train()` state. I added explicit `.eval()` calls so that the tests are now passing, but you may want to investigate this.

tests/torch/ptq/test_graphs.py
```diff
@@ -36,7 +36,7 @@
 @pytest.fixture(scope='function', params=ALGOS)
 def graph_dir(request):
     quantization_type = request.param
-    graph_dir = os.path.join('quantized', quantization_type)
+    graph_dir = Path('quantized') / 'ptq' / quantization_type
```
vshampor, line 39: Unfortunately, the current PTQ architecture is not compatible with reusing the PT `QuantizationBuilder` code. The PTQ algos for PT had to copy-and-paste code and do not produce the same result as the `create_compressed_model` approach. @daniil-lyakhov was offering to refactor `QuantizationBuilder` to make the quantizer creation more in-line with the `ModelTransformer` approach, and when that is done, we can make the test stronger by requiring the match to the quantized graphs of the QAT approach.

third_party_integration/huggingface_transformers/0001-Modifications-for-NNCF-usage.patch
```diff
@@ -1301,31 +1301,7 @@ index 214e7a978..530e80048 100755
...
--            signature = inspect.signature(self.model.forward)
-+            if isinstance(self.model, NNCFNetwork):
-+                signature = inspect.signature(self.model.get_nncf_wrapped_model().forward)
-+            else:
-+                signature = inspect.signature(self.model.forward)
-             self._signature_columns = list(signature.parameters.keys())
-             # Labels may be named label or label_ids, the default data collator handles that.
-             self._signature_columns += list(set(["label", "label_ids"] + self.label_names))
-@@ -1409,6 +1444,8 @@ class Trainer:
+@@ -1409,6 +1438,8 @@ class Trainer:
```
vshampor, line 1304: Now the integration into transformers won't need special signature and class handling for `NNCFNetwork` objects. 16 lines less in total in 2 places.

ljaljushkin: @vuiseng9 @yujiepan-work, potentially, it may affect NNCF integration to Optimum

nncf/torch/model_creation.py
ljaljushkin: It shouldn't affect the scope of working models. 
Potentially, the training-mode graph should help to resolve the issue with Aux branches in the Inception-v3 model. 
ticket 60976 (https://github.com/openvinotoolkit/nncf/pull/867)

tests/torch/nas/creators.py
ljaljushkin: I've investigated this issue. The root cause in changing of BN statistics in the train mode, not in the invalid weight reorg.
The test is not incorrect in train mode, since the output of the model is calculated with different statistics before and after weight reorg. 
Reordering indexes and order of updating weights is the same in eval and train modes. Propose to keep the fix with eval mode.

docs/NNCFArchitecture.md
```diff
@@ -60,24 +58,49 @@ class SimpleModule(torch.nn.Module):
...
+These string definitions are referred to as "scopes" in the NNCF configuration files (as in `"ignored_scopes"` or `"target_scopes"`), and help specify exact operations for inclusion into or exclusion from compression or for separate compression parameter specification.
 
- These string definitions are referred to as "scopes" in the NNCF configuration files (as in `"ignored_scopes"` or `"target_scopes"`), and help specify exact operations for inclusion into or exclusion from compression or for separate compression parameter specification.
 
 
-### Compression algorithm API and interaction with NNCFNetwork
+#### NNCFNetwork
+During NNCF compression, the incoming original model object is dynamically extended with NNCF-enabling functionality.
+This is done by replacing the model object's _class object_ with another class object that lists not only the original class object as its base, but also the `NNCFNetwork` object.
+The compressed model object can then be identified as passing the `isinstance(obj, NNCFNetwork)` checks, but also the `isinstance(obj, original_class)` checsk.
```
nikita-savelyevv, line 72:
```suggestion
The compressed model object can then be identified as passing the `isinstance(obj, NNCFNetwork)` checks, but also the `isinstance(obj, original_class)` checks.
```

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -435,8 +440,12 @@ def create_accuracy_aware_training_loop(nncf_config: NNCFConfig,
     """
     accuracy_aware_training_params = extract_accuracy_aware_training_params(nncf_config)
     accuracy_aware_training_mode = accuracy_aware_training_params.get('mode')
+    eval_fn = nncf_config.get_extra_struct(ModelEvaluationArgs).eval_fn
```
nikita-savelyevv, line 443: This will require some changes in OTX integration. For example, OTX provides `nncf_config` of instance `mmcv.ConfigDict`, not `NNCFConfig` as it should be.

cc @cih9088

nncf/torch/dynamic_graph/context.py
```diff
@@ -113,9 +118,12 @@ def __init__(self):
         self._ordinals_ids = None
 
     def __enter__(self):
-        global _CURRENT_CONTEXT
-        self._save_context = _CURRENT_CONTEXT
-        _CURRENT_CONTEXT = self
+        # For DataParallel, this relies on having one and the same compressed context for
```
nikita-savelyevv, line 121:
```suggestion
        # For DataParallel, this relies on having the same compressed context for
```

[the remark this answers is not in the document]
vshampor: Done

nikita-savelyevv (approved):

vshampor:
E2E build 252 shows no degradations.

nncf/torch/dynamic_graph/graph_tracer.py
```diff
@@ -108,8 +108,6 @@ def trace_graph(self, model: torch.nn.Module, context_to_use: Optional['TracingC
                     self.custom_forward_fn(model)
         model.load_state_dict(sd)
 
-        if isinstance(model, PostGraphBuildActing):
-            model.post_build_graph_actions()
```
AlexanderDokuchaev, line 112: `PostGraphBuildActing.post_build_graph_actions` no used anymore, looks like class `PostGraphBuildActing` can be removed

nncf/torch/quantization/quantize_functions.py
```diff
@@ -69,8 +69,7 @@ def backward(ctx: Any, *grad_outputs: Any) -> Any:
             )
         else:
             grad_input, _, grad_scale = QuantizedFunctionsCPU.get("Quantize_backward")(
-                grad_output, input_, input_low, input_range, levels, level_low, level_high, False
-            )
+                grad_output, input_, input_low, input_range, levels, level_low, level_high, False)
```
AlexanderDokuchaev, line 72: No changed this file, only bracket

tests/shared/command.py
```diff
@@ -46,6 +46,7 @@ def kill_process_tree(self, pid):
             print(err)
 
     def run(self, timeout=3600, assert_returncode_zero=True):
+        print(f"Running command: {self.cmd}")
```
AlexanderDokuchaev, line 49: Look like change for debug.

nncf/torch/checkpoint_loading.py
vshampor: Not really, because our scopes are specified in terms of structure of the original model, and this format has not been changed.

docs/NNCFArchitecture.md
```diff
@@ -60,24 +58,49 @@ class SimpleModule(torch.nn.Module):
...
+The resulting object serves internally as a receptacle for compression algorithm-related adjustments to the control flow graph of the model.
+In particular, `NNCFNetwork` redefines the `forward` calls so that the resulting object can function as usual in the outside training pipeline, but have compression-related functionality called at the right times.
```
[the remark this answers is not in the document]
vshampor: Removed `outside`. As for "right times", these are different from algo to algo and therefore cannot be summarized in one sentence. I added some more words to that account anyway.

docs/NNCFArchitecture.md
```diff
@@ -60,24 +58,49 @@ class SimpleModule(torch.nn.Module):
...
+
+More specifically, `NNCFNetwork` extends the regular model functionality with an ability to insert hooks into arbitrary places in the model (provided they can be properly addressed in the model control flow graph, see below).
```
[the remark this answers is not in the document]
vshampor: Incorporated some of your suggestions.

nncf/torch/nncf_network.py
```diff
@@ -714,10 +699,245 @@ def get_node_to_op_address_mapping(self) -> Dict[NNCFNodeName, OperationAddress]
...
+        """
+        return self.forward(*args, **kwargs)
+
+    @debuggable_forward
+    def forward(self, *args, **kwargs):
+        """
+        Wraps the original forward call, doing additional actions before and after the call to facilitate model
+        graph tracing and calling compression-related hooks.
+        """
+        #pylint:disable=protected-access
```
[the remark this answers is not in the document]
vshampor: Note that this is one class accessing protected members of another. Python has no concept of class friendship. At the same time these fields were protected for a reason.

nncf/torch/debug.py
```diff
@@ -62,12 +62,12 @@ def init_actual(self, owner_model):
 
 
 def debuggable_forward(forward_func):
-    def decorated(self, *args, **kwargs):
-        if self.debug_interface is not None:
-            self.debug_interface.pre_forward_actions(module=self)
+    def decorated(self: 'NNCFNetwork', *args, **kwargs):
+        if hasattr(self, 'nncf') and self.nncf.debug_interface is not None:
```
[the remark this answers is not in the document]
vshampor: I hope not, this is more of a precaution.

nncf/torch/dynamic_graph/graph_tracer.py
vshampor: Thanks for spotting this, removed

nncf/torch/quantization/quantize_functions.py
vshampor: Rolled back

tests/shared/command.py
vshampor: No, actually this is so that the tests that launch a python script in a subprocess would always dump the command itself into log - the developer can then copy-and-paste it and launch it directly when tracing a bug instead of having to launch the test in a debugger and `print`ing the `self.cmd` variable just to learn what command line was launched.

AlexanderDokuchaev (approved):

aiarkinx:
run pytorch pre-commit tests

aiarkinx:

nncf/torch/nncf_network.py
```diff
@@ -715,10 +700,255 @@ def get_node_to_op_address_mapping(self) -> Dict[NNCFNodeName, OperationAddress]
...
+        new_forward.__dict__.update(fn.__dict__)
+        new_forward.__signature__ = inspect.signature(original_class.forward)
+        new_class.forward = new_forward
+
+        # Make resulting class keep __module__ attributes of the original class,
+        # otherwise these will point to NNCF
+        new_class.__module__ = original_class.__module__
+        original_model.__class__ = new_class
+
+        if isinstance(original_model, torch.nn.Sequential):
```
ljaljushkin, line 779: Do we have such model in test scope? if not, makes sense to check it.

vshampor: `tests.torch.test_nncf_network.test_can_work_with_sequential_models` is added in this PR already

AlexKoff88:
@vuiseng9, @jpablomch, @yujiepan-work, FYI. Please take a look as well. Backward compatibility is preserved, as far as I understand.

ljaljushkin (approved):

docs/NNCFArchitecture.md
```diff
@@ -5,20 +5,18 @@ Neural Networks Compression Framework is a set of compression algorithms and too
...
+To abstract away the compression logic from specifics of the backend, NNCF builds an `NNCFGraph` object for each incoming model object to be compressed.
+`NNCFGraph` is a wrapper over a regular directed acyclic graph that represents a control flow/execution graph of a DNN. 
+Each node corresponds to a call of a backend-specific function ("operator").
+It is built both for the original, unmodified model, and for the model with compression algorithms applied (which, in general, may have additional operations when compared to the original model).
+
+### PyTorch-specific
+#### Model control flow graph tracing
 Unlike other frameworks such as TensorFlow, PyTorch does not have an easily accessible graph representation of a model, and thus no way to identify specific points in the control flow graph.
 For this reason NNCF performs tracing of the PyTorch operators, implemented via wrapping the corresponding function and module calls.
 Through this process of tracing, NNCF builds an internal representation of the model graph, which is then supplied as the point of reference for specification and insertion of hooks at proper places in the network.
```
AlexKoff88, line 19: While reading this design document I found that `NNCFNetwork` is mentioned prior to its definition. It looks that the description of the NNCFNetwork entity should go first.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -201,11 +203,12 @@ class AdaptiveCompressionTrainingLoop(BaseEarlyExitCompressionTrainingLoop):
     def __init__(self,
                  nncf_config: NNCFConfig,
                  compression_controller: CompressionAlgorithmController,
+                 original_model_accuracy: float,
                  lr_updates_needed=True, verbose=True,
                  minimal_compression_rate=0.0,
                  maximal_compression_rate=0.95,
                  dump_checkpoints=True):
-        super().__init__()
+        super().__init__(compression_controller)
```
alexsu52, line 211: I would like to know your motivation for these changes? if you are leaving `self.compression_controller = compression_controller` on the next line.

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -435,8 +440,12 @@ def create_accuracy_aware_training_loop(nncf_config: NNCFConfig,
     """
```
alexsu52, line 440: The create_accuracy_aware_training_loop function has the following signature in the documentation: 
```training_loop = create_accuracy_aware_training_loop(nncf_config, compression_ctrl, original_model_accuracy)```. Could you add `original_model_accuracy` or `uncompressed_model_accuracy` as the optional parameter of this function.

**Review by alexsu52 — DISMISSED** [review]

File: nncf/common/accuracy_aware_training/training_loop.py (hunk 27 above)
Comment (vshampor) [reply nncf/common/accuracy_aware_training/training_loop.py:211]: Forgot to remove the next line.

File: examples/tensorflow/classification/main.py [hunk 29]
```diff
@@ -271,7 +264,6 @@ def run(config):
                 steps_per_epoch=train_steps,
                 tensorboard_writer=SummaryWriter(config.log_dir, 'accuracy_aware_training'),
                 log_dir=config.log_dir,
-                uncompressed_model_accuracy=uncompressed_model_accuracy,
```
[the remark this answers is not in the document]
vshampor: Done

nncf/common/accuracy_aware_training/runner.py
```diff
@@ -188,8 +175,11 @@ class BaseAccuracyAwareTrainingRunner(TrainingRunner):
     initialized with the default parameters unless specified in the config.
     """
 
-    def __init__(self, accuracy_aware_training_params: Dict[str, object], verbose=True,
-                 dump_checkpoints=True, lr_updates_needed=True):
+    def __init__(self, accuracy_aware_training_params: Dict[str, object],
+                 original_model_accuracy: float,
```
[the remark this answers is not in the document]
vshampor: Let me know if you decide for another option, or if you find some names that I missed.

nncf/torch/sparsity/collector.py
```diff
@@ -71,7 +71,8 @@ def _collect_weights_descriptions(self) -> List[WeightDescription]:
             processed_modules.append(minfo.module)
 
         compression_types = tuple(COMPRESSION_MODULES.registry_dict.values())
-        for module_name, module in self._model.get_nncf_wrapped_model().named_modules():
+        modules_to_process = {k: v for k, v in self._model.named_modules() if not k.startswith('_nncf')}
```
daniil-lyakhov, line 74: Is this safe?

nncf/torch/nncf_network.py
```diff
@@ -715,10 +700,255 @@ def get_node_to_op_address_mapping(self) -> Dict[NNCFNodeName, OperationAddress]
...
+                                                    wrap_inputs_fn,
+                                                    scopes_without_shape_matching,
+                                                    ignored_scopes,
+                                                    target_scopes,
+                                                    wrap_outputs_fn)  # pylint:disable=protected-access
+        # The new class will also have an adjusted metaclass to avoid a "metaclass conflict" upon
+        # class creation
+        original_metaclass = type(original_model.__class__)
+        class_creation_kwds = {}
+        if original_metaclass is not type:
```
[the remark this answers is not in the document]
vshampor: It covers the case when the original model class has among its bases a class that has a metaclass. Class names are also objects and as such, can have types. When the class `Klass` does not have a metaclass, then `type(Klass) is type` (according to Python spec). If the class has a metaclass `Metaklass`, then `type(Klass) is Metaklass`. It becomes harder to explain when inheritance is involved, but basically the outcome is that we have to make our metaclass dynamic too if the original model class had a metaclass, and this is what this conditional is all about.

nncf/torch/sparsity/collector.py
vshampor: Safe in what sense specifically? This approach does implicitly mean that we should not store sparsity-processed modules in the `NNCFNetworkInterface` namespace, but we don't do that now and I don't see when we will need to. The possibility that the user's module would start with `_nncf` is quite low.

alexsu52 (commented):
LGTM

PR #1585, opened 2023-02-17
[OV] Align FQ params data type with target node  precision

l-bat:
### Changes

For models coming from a framework in FP16 precision, FakeQuantize parameters should be in FP16.
For models coming from a framework in FP32 precision and converted to FP16 via MO, FakeQuantize parameters should be in FP32.
FQ parameters precision aligned with model input precision.
Input FP32
image
Input FP16
image

### Reason for changes

* To align NNCF PTQ with POT

### Related tickets

* 103926, 101738

### DIFF
--- a/tests/openvino/native/models.py
+++ b/tests/openvino/native/models.py
@@ -196,7 +196,7 @@ class FPModel(OVReferenceModel):
 
     def _create_ov_model(self):
         input_shape = [1, 3, 4, 2]
-        input_1 = opset.parameter(input_shape, name="Input")
+        input_1 = opset.parameter(input_shape, name="Input", dtype=np.float32)
         data = self._rng.random((1, 3, 4, 5)).astype(self.precision)
         if self.precision == np.float16:
             data = opset.convert(data, np.float32)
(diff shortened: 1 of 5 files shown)


alexsu52 (commented):
General comment/question on this PR:
- As I understood from the PR, the precision of FQ's weights is set by input precision. In term of collecting statistics, statistics are collected using the input, so they have the same precision as the input, right?
- What precision are min max range values computed?

tests/openvino/native/models.py
```diff
@@ -210,6 +210,19 @@ def _create_ov_model(self):
         return model
 
 
+class FP16InputModel(OVReferenceModel):
```
KodiaqQ, line 213: Why not use/modify the existing `FPModel` above?

KodiaqQ:
@l-bat, could you please provide the quantized graph examples with the different precisions? (from Netron)

tests/openvino/native/models.py
l-bat: Added `ìnput_dtype` argument to `FPModel`

alexsu52 (approved):
LGTM

KodiaqQ (approved):

PR #1588, opened 2023-02-21
[PTQ] Constant statistics aggregation optimization

daniil-lyakhov:
### Changes

Statistic collector registers constant statistics only ones

### Reason for changes

Constants don't change during inference, so registering them one is enough

### DIFF
--- a/nncf/experimental/openvino_native/statistics/aggregator.py
+++ b/nncf/experimental/openvino_native/statistics/aggregator.py
@@ -34,6 +34,7 @@ class OVStatisticsAggregator(StatisticsAggregator):
         self._name_to_node_mapping = {
             op.get_friendly_name(): op for op in model.get_ops()
         }
+        self._registered_weights = set()
         super().collect_statistics(model)
 
     def _register_activation_statistic(self, statistic_point: StatisticPointsContainer,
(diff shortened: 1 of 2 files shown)


nncf/onnx/statistics/aggregator.py
```diff
@@ -34,6 +34,7 @@ class ONNXStatisticsAggregator(StatisticsAggregator):
     def collect_statistics(self, model: onnx.ModelProto) -> None:
         self._nncf_graph = NNCFGraphFactory.create(model)
         self._onnx_graph = ONNXGraph(model)
+        self._registered_weights = set()
```
kshpv, line 37: Could you avoid adding a new attribute?

kshpv (changes requested):

kshpv (approved):

nncf/onnx/statistics/aggregator.py
daniil-lyakhov: Statistic_point contains several statistic collectors, so we need to check all collectors have registered constant at least ones. I suggest to keep it as is

nncf/experimental/openvino_native/statistics/aggregator.py
```diff
@@ -69,7 +70,11 @@ def _register_statistics(self,
                     self._register_activation_statistic(statistic_point,
                                                         target_point, node_name, outputs)
                 elif target_point.type == TargetType.OPERATION_WITH_WEIGHTS:
-                    self._register_weight_statistic(statistic_point, target_point)
+                    # Register constant only once because it does not change
```
KodiaqQ, line 73: Are that changes for the shared weights? Is there any test for such a case added?

daniil-lyakhov: This optimization forces to register statistics for weights only ones, regardless whether it  is shared or not. `test_statistics_aggregator` checks weight statistic registered at least ones. 
Tests for shared weights are missing now in all backends, I suggest to add them in a different PR

daniil-lyakhov: Ticket: 104570

KodiaqQ (approved):

l-bat (approved):

daniil-lyakhov:
@KodiaqQ please merge it if it's ok

PR #1593, opened 2023-02-22
Fix E2E

vshampor:
### Changes
Properly renamed some of the PT .json configs and made TF classification sample correctly handle a situation where `base_lr` is not specified in the sample config.

### Reason for changes
Follow-up for #1507 - some E2E models failed after the merge.

### Related tickets
N/A

### Tests
SOTA eval runs for PT and TF


vshampor:
Will post passing SOTA build IDs once they are ready

examples/tensorflow/classification/main.py
```diff
@@ -205,12 +205,16 @@ def run(config):
...
-            scheduler = build_scheduler(
-                config=config,
-                steps_per_epoch=train_steps)
-            optimizer = build_optimizer(
-                config=config,
-                scheduler=scheduler)
+            train_kwargs = {}
+
+            if 'train' in config.mode:
+                scheduler = build_scheduler(
```
alexsu52, line 211: It looks like the following change is enough and does not need to change the main.py
```
base_lr = schedule_params.get('base_lr', optimizer_config.get('base_lr', 0.001))
```

alexsu52 (changes requested):

examples/tensorflow/classification/main.py
vshampor: Done

alexsu52 (approved):
LGTM

PR #1595, opened 2023-02-23
Added conv3d to first layer overflow fix check

nikita-savelyevv:
Overflow fix for a first layer check should consider `conv3d` besides `conv2d`.

In general, all weighted ops should be considered, but this requires some refactoring, see ticket [104645]


alexsu52 (approved):
LGTM

PR #1596, opened 2023-02-23
Added average pool 3d -like ops to pruning mask forwarding ops

nikita-savelyevv:
`PTAvgPool2dMetatype` and `PTMaxPool2dMetatype` were in `PTIdentityMaskForwardPruningOp` but `PTAvgPool3dMetatype` and `PTMaxPool3dMetatype` not. Added.

Similarly for TF.


daniil-lyakhov (approved):
Good catch, thanks!

daniil-lyakhov:
Could you please add a test on pruning groups with this new operations?

PR #1597, opened 2023-02-23
[Pruning] Fixed statistics computation for pruned layers

nikita-savelyevv:
### Summary
For PT filter pruning, before `compression_ctrl.export_model()` call the pruned layer statistics are printed. They however are currently empty and do not reflect the actual pruned state of the model.

### Reason
This is due the fact that `pruning_level_for_filters()` method used to retrieve statistics returns incorrect values. This in turn is due the fact that it relies on the number of zeros in the model weights. But weights are zeroed (pruned) only during model export and actual model weights are not zeroed during runtime.

### What's changed
Removed `pruning_level_for_weight()` and `pruning_level_for_filters()` altogether. Now `pruning_level_for_mask()` is the only method to compute the statistic.

#### Before
image
#### After
image

### DIFF
--- a/nncf/torch/pruning/base_algo.py
+++ b/nncf/torch/pruning/base_algo.py
@@ -241,23 +241,24 @@ class BasePruningAlgoController(PTCompressionAlgorithmController):
         """
         raise NotImplementedError
 
-    @staticmethod
-    def pruning_level_for_weight(minfo: PrunedModuleInfo):
+    def pruning_level_for_weight(self, minfo: PrunedModuleInfo):
         """
         Calculates sparsity level for all weight nodes.
         """
         weight = minfo.module.weight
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
Suggestion: remove `pruning_level_for_weight()` and `pruning_level_for_filters()` altogether since they return the same values as `pruning_level_for_mask()`.

daniil-lyakhov (approved):
This legacy methods were redundant, agree. LGTM!

vshampor (approved):
I'll merge this, but consider the fact that the users towards whom the controller interface is oriented probably don't have to know that there are masks involved in our pruning algo. They just want to know how strongly pruned the model is, or at most a given weight. So the previous methods were OK in that their names were something that is familiar to the user, in contrast with the "mask". @daniil-lyakhov says the methods were redundant, and maybe they are, but think about how the remaining methods are perceived by the users.

PR #1602, opened 2023-02-27
[Microfix] Fix ref before assignment

daniil-lyakhov:
### Changes
 ref before assignment error is fixed


KodiaqQ (approved):
Thanks for fix 👍

PR #1616, opened 2023-03-06
Metric is changed when compress_quantize_weights_transformation is applied

andrey-churkin:
### Changes

Infer model in custom plugin (TEMPLATE)

### Reason for changes

Metric is changed when compress_quantize_weights_transformation is applied.

### Related tickets

Ref: 103920

### Tests

N/A


andrey-churkin:
@alexsu52 Should we use this plugin everywhere in NNCF when we compile the model?

PR #1618, opened 2023-03-06
Calculated ignored scope for calibrate.py

daniil-lyakhov:
### Changes

Ignored scope for some models now is being calculated by model architecture and not depend on IR names

### Reason for changes

To make NNCF ignored scopes independent of MO names changes in OV convert test

### Related tickets

105384


daniil-lyakhov:
Validated by DLBenchmark runs 64 and 65

alexsu52 (approved):

PR #1624, opened 2023-03-09
[ONNX] Add port to nncf_input_node_next_onnx_nodes

kshpv:
### Changes

Fix a bug with using incorrect port while quantizing Input

### Reason for changes

104908

### Related tickets

104908

### DIFF
--- a/nncf/onnx/graph/model_transformer.py
+++ b/nncf/onnx/graph/model_transformer.py
@@ -26,6 +26,7 @@ from nncf.onnx.graph.transformations.commands import ONNXOutputInsertionCommand
 from nncf.onnx.graph.transformations.commands import ONNXQuantizerInsertionCommand
 from nncf.onnx.graph.transformations.commands import ONNXQDQNodeRemovingCommand
 from nncf.common.graph.model_transformer import ModelTransformer
+from nncf.onnx.graph.node_utils import get_input_edge
 
 
 class ONNXModelTransformer(ModelTransformer):
@@ -59,14 +60,8 @@ class ONNXModelTransformer(ModelTransformer):
         """
         if transform_type in [TargetType.PRE_LAYER_OPERATION, TargetType.OPERATION_WITH_WEIGHTS]:
             return onnx_graph.get_node_edge_names(node_name)['input'][port_id]
-        input_edges = set()
         if node_name in input_edges_mapping:  # ADD INPUT NODE CASE
-            next_nodes = input_edges_mapping[node_name]
-            for next_node in next_nodes:
-                name, port_id = next_node
-                input_edges.add(onnx_graph.get_node_edge_names(name)['input'][port_id])
-            assert len(input_edges) == 1
-            return input_edges.pop()
+            return get_input_edge(node_name, input_edges_mapping, onnx_graph)
         return onnx_graph.get_node_edge_names(node_name)['output'][port_id]
 
     def transform(self, transformation_layout: TransformationLayout) -> onnx.ModelProto:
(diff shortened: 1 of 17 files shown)


nncf/onnx/graph/node_utils.py
```diff
@@ -53,3 +56,21 @@ def get_bias_value(node_with_bias : NNCFNode, model: onnx.ModelProto) -> np.ndar
     if metatype == ONNXIdentityMetatype:
         return onnx_graph.get_initializers_value(node.input[0])
     raise RuntimeError('Could not find the bias value of the node')
+
+
+def get_nncf_input_node_next_onnx_nodes(nncf_graph: NNCFGraph) -> Dict[str, Tuple[str, int]]:
```
KodiaqQ, line 61: The method's name is weird. It would be better to call it `get_input_edges_mapping` or something like that.

nncf/onnx/graph/model_transformer.py
```diff
@@ -58,9 +58,12 @@ def _get_target_edge(self, port_id: int, node_name: str, transform_type: TargetT
         """
         if transform_type in [TargetType.PRE_LAYER_OPERATION, TargetType.OPERATION_WITH_WEIGHTS]:
             return onnx_graph.get_node_edge_names(node_name)['input'][port_id]
+        input_edges = set()
         if node_name in nncf_input_node_next_nodes:  # ADD INPUT NODE CASE
-            node_names = nncf_input_node_next_nodes[node_name]
-            input_edges = set(onnx_graph.get_node_edge_names(name)['input'][port_id] for name in node_names)
+            next_nodes = nncf_input_node_next_nodes[node_name]
+            for next_node in next_nodes:
+                name, port_id = next_node[0], next_node[1]
```
KodiaqQ, line 65: IMHO, it's better to use `dict` inside, instead of mapping by indexes of very different types of data.

nncf/onnx/graph/node_utils.py
```diff
@@ -53,3 +56,21 @@ def get_bias_value(node_with_bias : NNCFNode, model: onnx.ModelProto) -> np.ndar
...
+    :param nncf_graph: instance of NNCFGraph
+    :return: A mapping of NNNCF input node names and a tuple with the consumed node names and their input port ids.
+    """
+    nncf_input_node_next_onnx_nodes = {}
+    for input_node in nncf_graph.get_input_nodes():
+        nncf_input_node_next_onnx_nodes[input_node.node_name] = []
+        for next_node in nncf_graph.get_next_nodes(input_node):
+            for edge in nncf_graph.get_input_edges(next_node):
+                if edge.from_node == input_node:
+                    nncf_input_node_next_onnx_nodes[input_node.node_name].append((next_node.node_name,
```
KodiaqQ, line 74: Why just not store the correct edge name in the mapping? Instead of packaging the `port_id` and using it for further name calculation.

KodiaqQ:
What about the test? How could we catch this regression earlier?

nncf/onnx/graph/node_utils.py
kshpv: Yeah, this is probably better idea

nncf/onnx/graph/node_utils.py
kshpv: Done

nncf/onnx/graph/model_transformer.py
kshpv: Discussed

nncf/onnx/graph/node_utils.py
kshpv: Discussed

tests/onnx/data/reference_graphs/quantization/synthetic/multi_input_output_model.dot
```diff
@@ -0,0 +1,24 @@
+strict digraph  {
```
KodiaqQ, line 1: Where is the original model graph?

kshpv: The original is already added. The quantization of that model was skipped before. I removed skipping

KodiaqQ (approved):

PR #1626, opened 2023-03-09
Fix for torch.jit.script patching applied to classes

nikita-savelyevv:
Generally `torch.jit.script` is applied to methods, but in some cases it may be applied to classes, for example in `kornia` package. In such case NNCF patching of `torch.jit.script` is incorrect and leads to an error during import of the scripted class. 

The problem arises because patching inserts and additional item to call stack and `torch.jit.script` relies on it when retrieving `ResourceResolutionCallback` object representing the scopes for the trace methods of the class. All this results in the need for providing a custom `ResourceResolutionCallback` during patching.

Also, `ORIGINAL_OPERATORS` list was increasing in size after each unpatch-patch iteration resulting in duplication up to 10x. This is because it wasn't cleared before yet another patching operation. This issue was fixed in this PR too.

### Related tickets

81980

### Tests

Added a test for the case when a class is being scripted.

### DIFF
--- a/tests/torch/test_pytorch_patch.py
+++ b/tests/torch/test_pytorch_patch.py
@@ -1,6 +1,9 @@
+import inspect
+
 import torch
 
 from nncf.torch.dynamic_graph.patch_pytorch import MagicFunctionsToPatch
+from nncf.torch.dynamic_graph.patch_pytorch import _ORIG_JIT_SCRIPT
 from nncf.torch.graph.operator_metatypes import PT_OPERATOR_METATYPES
 from nncf.torch.dynamic_graph.context import TracingContext
 from nncf.torch.dynamic_graph.trace_tensor import TracedTensor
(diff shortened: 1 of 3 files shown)


tests/torch/test_pytorch_patch.py
```diff
@@ -67,3 +70,22 @@ def forward(self, x: torch.Tensor):
 def test_jit_if_tracing_script_source():
     # Run test case in a separate process to track patching of torch by NNCF
     run_pytest_case_function_in_separate_process(test_jit_if_tracing_script_source_equals)
+
+
+def test_jit_script_signature():
```
AlexKoff88, line 75: Can we also have a test for the model optimized with NNCF? Let's say we take a model from torchvision (MobilNetv2), apply quantization on a dummy dataset, call `.prepare_for_inference()`, and do the tracing/scripting then.

nikita-savelyevv: Supporting scripting/tracing after `.prepare_for_inference()` is planned for another PR (ticket 106013). This one is only to fix the issue with `kornia` package.

nikita-savelyevv:
run onnx pre-commit tests

vshampor (approved):

AlexKoff88 (approved):

PR #1630, opened 2023-03-10
Fix layer attribute collection for user modules

vshampor:
### Reason for changes
User modules with `addmm` operation could not be properly quantized because the operation did not have `layer_attributes` assigned to it at collection stage (only the builtin NNCF modules had this collection with old logic), but was of the matching metatype to ops that are supposed to have weights, which led to an assertion failure in quantization.

### Related tickets
N/A

### Tests
tests.torch.quantization.test_algo_quantization.test_can_quantize_user_module_with_addmm


daniil-lyakhov:
Could this logic with default `layer_attributes` broke quantization logic in some corner case?
Could we change logic in a way that user modules shouldn't leverage `layer_attributes` to enable quantization?

vshampor:
Third party sanity test run 34 passes (single failure due to some network issues).

vshampor:
> Could this logic with default `layer_attributes` broke quantization logic in some corner case? Could we change logic in a way that user modules shouldn't leverage `layer_attributes` to enable quantization?

This quick fix is a follow-up for your own change for `layer_attributes` and associated weighted node detection. I could not use the metatype-based `layer_attribute` collection because I ran into circular imports issues. Meanwhile I need a quick fix so that the third-party sanity tests can run (and GPT-2 quantization doesn't fail).

vshampor:
@daniil-lyakhov I uploaded a more consistent fix version as discussed, check it out

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -207,11 +208,12 @@ def _collect_module_attrs_and_ignored_algorithms(ctx: TracingContext,
                                                  op_name: str) -> Tuple[BaseLayerAttributes, List[str]]:
     layer_attrs = None
     ignored_algos = []
-    if op_name in OP_NAMES_REQUIRING_MODULE_ATTRS:
-        curr_module = ctx.get_current_module()
+    curr_module = ctx.get_current_module()
+    from nncf.torch.graph.operator_metatypes import OP_NAMES_WITH_WEIGHTS
+    if op_name in OP_NAMES_WITH_WEIGHTS:
```
daniil-lyakhov, line 213:
```suggestion
    from nncf.torch.graph.operator_metatypes import OP_NAMES_WITH_WEIGHTS #pylint: disable=cyclic-import
    if op_name in OP_NAMES_WITH_WEIGHTS:
            curr_module = ctx.get_current_module()
```

vshampor: Done

daniil-lyakhov (approved):
👍

vshampor:
run openvino pre-commit tests

PR #1636, opened 2023-03-14
[PTQ][OV] Remove IS calculation from `calibrate.py`, move it to separate tool

daniil-lyakhov:
### Changes

[PTQ][OV] Remove IS calculation from `calibrate.py`, move it to separate tool

### Reason for changes

To make calibration clearer

### DIFF
--- a/tools/get_ir_is_by_model_id.py
+++ b/tools/get_ir_is_by_model_id.py
@@ -1,0 +1,113 @@
+"""
+ Copyright (c) 2023 Intel Corporation
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


alexsu52 (approved):
LGTM

tools/get_ir_is_by_model_id.py
```diff
@@ -0,0 +1,113 @@
+"""
```
andrey-churkin, line 1: Could we give a more readable name for this tool than `get_ir_is_by_model_id.py?`

tools/get_ir_is_by_model_id.py
```diff
@@ -0,0 +1,113 @@
...
+                continue
+            next_nodes = get_next_nodes(node, 0)
+            assert len(next_nodes) == 1
+            next_node = next_nodes[0]
+            assert next_node.type_info.name == 'Convolution'
+            return [next_node.get_friendly_name()]
+    return None
+
+
+IS_TO_BUILDER_MAP = {
```
andrey-churkin, line 91: Could you please use a full name i.e. ignored scope instead of `is`?

daniil-lyakhov: Done, please check

andrey-churkin (approved):

PR #1640, opened 2023-03-15
[OV] Added scale_shift_activations fused pattern

l-bat:
### Changes

                                                          Before / After
image

### Reason for changes

* To align OV Native quantization with POT

### Related tickets

* 105079

### Tests

* added `ScaleShiftReluModel`


l-bat:
run pytorch pre-commit tests

alexsu52 (approved):
LGTM

nncf/experimental/openvino_native/hardware/fused_patterns.py
```diff
@@ -643,6 +643,14 @@ def create_clamp_mult_const():
     return pattern
 
 
+@OPENVINO_HW_FUSED_PATTERNS.register(PatternNames.SCALE_SHIFT_ACTIVATIONS)
+def create_scale_shift_activations():
```
KodiaqQ, line 647: Since the `SCALE_SHIFT_ACTIVATIONS` was added, let's drop the message about it from the `/tests/openvino/native/test_pattern_manager.py`.
Also, maybe it would be helpful to think about checking the patterns that exist, in the same test.

KodiaqQ (approved):

PR #1579, opened 2023-02-15
Add NNCF cache directory

kshpv:
### Changes

Add directory is used to store cache to run tests or samples.

### Reason for changes

Speed up ONNX pre-commit by caching dataset and models.

### Related tickets

No

### Tests

No

PR #1641, opened 2023-03-15
Added new AA config on CIFAR10 and ResNet18

nikita-savelyevv:
### Changes

- Added new Accuracy Aware config for quantization + pruning scenario on ResNet18 model and CIFAR10 dataset
- Added logic for downloading models

Results are given below. The AA metrics were averaged over 10 runs.
|   | Accuracy  | Compression rate  |  Duration |  
|---|---|---|---|
| Baseline FP32 | 88.61%  | 0%  | -  | 
| AA INT8 + Pruning|  87.78% | 58.13%  | 44.3 min  |

### Reason for changes

Existing AA configs run on ImageNet and take multiple days to finish which makes it hard for new users to get familiar with the approach and use it. The new config runs 45 mins on average with 8 workers on i9-10980XE CPU and single RTX 3090 GPU.


alexsu52 (approved):
Thanks for the contribution!

examples/torch/common/model_loader.py
```diff
@@ -26,6 +30,9 @@
 from nncf.torch.utils import safe_thread_call
 
 
+MODELS_PATH = '~/.cache/nncf/models'
```
vshampor, line 33: Bring this in line with the general approach in https://github.com/openvinotoolkit/nncf/pull/1579

vshampor (changes requested):

examples/torch/common/model_loader.py
nikita-savelyevv: Seems like I need #1579 to be merged first. Maybe better to merge this PR and change `MODELS_PATH` in #1579 ?

vshampor (approved):

PR #1642, opened 2023-03-16
[PTQ] Update ShapeOf pass with ReadValues

KodiaqQ:
### Changes

- Updated ShapeOf pass with ReadValues

### Reason for changes

- Bugfix

### Related tickets

- 105886


KodiaqQ:
run onnx pre-commit tests

KodiaqQ:
run openvino pre-commit tests

KodiaqQ:
run openvino pre-commit tests

KodiaqQ:
run onnx pre-commit tests

nncf/quantization/algorithms/min_max/backend.py
```diff
@@ -57,6 +57,13 @@ def shapeof_metatypes(self) -> List[OperatorMetatype]:
         Property for the backend-specific ShapeOf metatypes.
         """
 
+    @property
+    @abstractmethod
+    def similar_input_metatypes(self) -> List[OperatorMetatype]:
```
kshpv, line 62: Is it better than using input_metatypes which returns all possible meta types that are inputs?

nncf/quantization/passes.py
```diff
@@ -69,3 +80,36 @@ def remove_shapeof_subgraphs(nncf_graph: NNCFGraph, shapeof_metatypes: List[Oper
 
     nncf_graph.remove_nodes_from(nodes_to_drop)
     return nncf_graph
+
+
+def filter_constant_nodes(nncf_graph: NNCFGraph,
```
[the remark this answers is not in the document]
KodiaqQ: Since QAT NNCFGraphs doesn't contain `Constant` nodes, the behavior would be the same.

nncf/quantization/algorithms/min_max/backend.py
KodiaqQ: What do you mean by `input_metatypes`? `ReadValues` are not inputs themselves and are followed by `Constant` nodes every time.

kshpv: Ok, I see. Just the name `similar_input_metatypes` confused me. 
I see that this is applicable only for OpenVINO, while it could be extended to be applicable to all backends if this function will return all metatypes that should be treatable as inputs.

KodiaqQ: Yeah, it may be extended for any type that is needed. I've tried to present a more general solution under the conditions that metatypes are not common.

alexsu52 (approved):
LGTM

l-bat (approved):

kshpv (changes requested):
Looks good! Can you add a test for OpenVINO synthetic model with an added pass?

kshpv (approved):
I see that ReadValue metatype is already in LSTMModel, so this functional should be covered with tests

PR #1648, opened 2023-03-20
[PTQ] Fix rewriting results for models with the same name

kshpv:
### Changes

Fix conformance test

### Reason for changes

Fix bug with rewriting the results for the mode with different quantization configurations

### Related tickets

No

### Tests

No


alexsu52 (approved):

alexsu52 (commented):
Could you provide a build with these changes?

PR #1649, opened 2023-03-21
Align PT model names in AC configs

vshampor:
### Changes
Names in AC configs for PT E2E models were aligned with canonical names.

### Reason for changes
E2E test pipeline relies on AC config model names being the same as the model IDs in `sota_checkpoints_eval.json`, otherwise it can't find proper OV metrics to be displayed in the test report.

### Related tickets
N/A

### Tests
E2E PT


vshampor:
E2E PT build 255

vshampor:
Build green, problematic models no longer show issues.

PR #1658, opened 2023-03-24
Windows installation tests

ksilligan:
### Changes

Installation tests was designed to run only on unix platforms (for example: `{venv_path}/bin/activate`), and was never tested on Win32 platform

### Reason for changes

We want to run tests on Windows platform

### Related tickets

10585

### Tests

pytest tests/cross_fw/install/ -s --backend {backend.name} {backend.params}


ksilligan:
Linux installation tests passed for this PR

PR #1620, opened 2023-03-07
[PTQ] Add model_type parameter to MinMax

kshpv:
### Changes

Add support of model_type parameter.

### Reason for changes

To support correct quantization of Transfromer models.

### Related tickets

96707

PR #1664, opened 2023-03-27
FastBiasCorrection for PyTorch

AlexanderDokuchaev:
### Changes

Implementation of fast bias correction algorithm for PyTorch.

- Add `TemplateTestFBCAlgorithm`  to test FBC algorithm.
- Add function `replace_custom_modules_with_torch_native` with converting custom modules from timm which broke functions of torch backend to torch native modules.
- Set requirements `timm==0.6.13` for `tests/post_training`, 0.9.2 currently can not be exported to onnx with error  `torch.onnx.errors.UnsupportedOperatorError: Exporting the operator 'aten::scaled_dot_product_attention' to ONNX opset version 13 is not supported.`

### Related tickets

96770

### Tests
- TestTorchFBCAlgorithm
- TestONNXFBCAlgorithm
- TestOVFBCAlgorithm


nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @staticmethod
+    def create_bias_correction_command(node: NNCFNode,
+                                       bias_value: np.ndarray,
+                                       nncf_graph: NNCFGraph) -> PTBiasCorrectionCommand:
+        return create_bias_correction_command(node, bias_value)
+
+    @staticmethod
+    def model_extraction_command(inputs: List[str], outputs: List[str]) -> PTModelExtractionCommand:
+        return PTModelExtractionCommand(inputs[0])
```
KodiaqQ, line 80: Why only `[0]` input? Where are the `outputs`?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @staticmethod
+    def mean_statistic_collector(reduction_shape: ReductionShape,
+                                 num_samples: Optional[int] = None,
+                                 window_size: Optional[int] = None) -> PTMeanStatisticCollector:
+        return PTMeanStatisticCollector(reduction_shape, num_samples, window_size)
+
+    @staticmethod
+    def get_sub_input_output_names(subgraph: nn.Module) -> Tuple[str, str]:
+        # Pytorch does not have name for extracted node
```
KodiaqQ, line 90: Are the names of the nodes in the new sub-graph match with the same that was provided to the `model_extraction_command`?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,144 @@
...
+        blob = torch.zeros(shape, dtype=data[0].dtype)
+        for j, idx in enumerate(np.ndindex(blob.shape[channel_axis])):
+            index = tuple(slice(None) if i != channel_axis else idx for i in range(blob.ndim))
+            blob[index] = data[j]
+        return blob
+
+    @staticmethod
+    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: nn.Module) -> np.ndarray:
+        node_module = model.get_containing_module(node.node_name)
+        if node_module.bias is None:
```
KodiaqQ, line 104: Why do we need it? Why this behavior is not aligned between backends?

nncf/torch/composite_compression.py
```diff
@@ -24,8 +24,8 @@
 from nncf.torch.compression_method_api import PTCompressionAlgorithmController
 from nncf.torch.compression_method_api import PTCompressionLoss
 from nncf.torch.graph.transformations.layout import PTTransformationLayout
+from nncf.torch.model_transformer import PTModelTransformer
```
KodiaqQ, line 27: This PR is not for the styling fixes, by the way.

nncf/torch/graph/node_utils.py
```diff
@@ -0,0 +1,26 @@
...
+def is_node_with_bias(node: NNCFNode) -> bool:
+    """
+    Checks if the node has a bias or not.
+
+    :param node: The node to check.
+    :return: Return `True` if `node` corresponds to the operation
+        with bias (bias is added to the output tensor of that operation),
+        `False` otherwise.
+    """
+    return node.metatype in OPERATORS_WITH_BIAS_METATYPES
```
KodiaqQ, line 26: This method should checks not only the meta type consistency.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def create_blob(shape: Tuple[int], data: List[np.ndarray], channel_axis: int) -> np.ndarray:
+        blob = torch.zeros(shape, dtype=data[0].dtype)
+        for j, idx in enumerate(np.ndindex(blob.shape[channel_axis])):
+            index = tuple(slice(None) if i != channel_axis else idx for i in range(blob.ndim))
+            blob[index] = data[j]
+        return blob
+
+    @staticmethod
+    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: nn.Module) -> np.ndarray:
```
KodiaqQ, line 102: Why is the body of this method not in `nncf/<BACKEND>/graph/node_utils.py` as for the other backends?

nncf/torch/graph/transformations/command_creation.py
```diff
@@ -0,0 +1,31 @@
...
+from nncf.torch.graph.transformations.commands import PTTargetPoint
+
+
+def create_bias_correction_command(node: NNCFNode, bias_value: np.ndarray) -> PTBiasCorrectionCommand:
+    """
+     Creates bias correction command.
+
+    :param node: The node in the NNCF graph that corresponds to operation with bias.
+    :param bias_value: The new bias value that will be set.
+    :return: The `ONNXBiasCorrectionCommand` command to update bias.
```
KodiaqQ, line 28:
```suggestion
    :return: The `PTBiasCorrectionCommand` command to update bias.
```

nncf/torch/graph/transformations/command_creation.py
```diff
@@ -0,0 +1,31 @@
...
+
+import numpy as np
+
+from nncf.common.graph.graph import NNCFNode
+from nncf.common.graph.transformations.commands import TargetType
+from nncf.torch.graph.transformations.commands import PTBiasCorrectionCommand
+from nncf.torch.graph.transformations.commands import PTTargetPoint
+
+
+def create_bias_correction_command(node: NNCFNode, bias_value: np.ndarray) -> PTBiasCorrectionCommand:
```
KodiaqQ, line 22: The `PTBiasCorrectionCommand` instance supports only `bias_value` as `torch.Tensor`, Here is the `np.ndarray`.

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+from nncf.torch.graph.transformations.commands import PTInsertionCommand
+from nncf.torch.graph.transformations.commands import PTModelExtractionCommand
+from nncf.torch.graph.transformations.commands import PTTargetPoint
+from nncf.torch.graph.transformations.layout import PTTransformationLayout
+from nncf.torch.module_operations import UpdateWeight
+from nncf.torch.nncf_network import NNCFNetwork
+from nncf.torch.nncf_network import PTInsertionPoint
+
+
+class PTModelTransformer(ModelTransformer):
```
KodiaqQ, line 35: Missed description of the class and its methods.

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+            target_point = transformation_command.target_point  # type: PTTargetPoint
+            target_node_name = target_point.target_node_name
+            pt_ip = PTInsertionPoint(
+                target_type=target_point.target_type,
+                op_address=self._node_to_op_address_mapping[target_node_name],
+                input_port_id=target_point.input_port_id,
+            )
+            fn = transformation_command.fn
+            if target_point.type is TargetType.OPERATION_WITH_WEIGHTS:
+                fn = UpdateWeight(fn)
```
KodiaqQ, line 74: The method is called `_apply_output_insertion_transformations`. What is `UpdateWeight` doing here?

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+                fns_grouped_by_points[pt_ip] = [tup]
+            else:
+                fns_grouped_by_points[pt_ip].append(tup)
+
+        for pt_ip, fn_list_with_priority in fns_grouped_by_points.items():
+            fn_list_with_priority = sorted(fn_list_with_priority, key=lambda x: x[1])
+            self._model.insert_at_point(pt_ip, [x[0] for x in fn_list_with_priority])
+
+    def _apply_extraction_transformations(self, transformations: List[TransformationCommand]) -> nn.Module:
+        assert len(transformations) == 1, "Support only one node extraction"
```
KodiaqQ, line 86: Please, see the other backends to avoid such assert.

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+        extracted_module = self._model.get_containing_module(transformations[0].node_name)
```
KodiaqQ, line 87: Could you provide a graph example of the output model? I'm not sure that the output is correct.

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,97 @@
...
+        # Return copy of the module without pre and post operations.
+        extracted_module = copy.deepcopy(extracted_module)
+        extracted_module.reset()
+        return extracted_module
+
+    def _apply_bias_correction_transformations(self, transformations: List[TransformationCommand]) -> None:
+        for transformation in transformations:
+            node_name = transformation.target_point.target_node_name
+            node = self._model.get_containing_module(node_name)
+            node.bias.data = transformation.bias_value
```
KodiaqQ, line 97: We've agreed to avoid in-place transformations, didn't we?

tests/torch/ptq/test_fast_bias_correction.py
```diff
@@ -0,0 +1,49 @@
...
+    return desc.values[0].model_name
+
+
+TEST_MODELS_DESC = [
+    ModelDesc("resnet18", test_models.ResNet18, [1, 3, 32, 32]),
+    ModelDesc("BasicConvTestModel", BasicConvTestModel, [1, 1, 4, 4])
+]
+
+@pytest.mark.parametrize("desc", TEST_MODELS_DESC, ids=[get_model_name(m) for m in TEST_MODELS_DESC])
+def test_fast_bias_correction_algo(desc: ModelDesc):
```
KodiaqQ, line 37: What is being tested here? If we test the whole pipeline, why not test accuracy at the same time?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @staticmethod
+    def process_model_output(raw_data: Dict, output_name: str) -> PTNNCFTensor:
+        return PTNNCFTensor(raw_data)
+
+    @staticmethod
+    def is_quantized_weights(node: NNCFNode, nncf_graph: NNCFGraph, model: nn.Module) -> bool:
+        node_module = model.get_containing_module(node.node_name)
+        for pre_op in node_module.pre_ops.values():
+            if isinstance(pre_op.op, (BaseQuantizer, FakeQuantize)):
```
KodiaqQ, line 127: What happens in the case of only activation input quantization?

nncf/common/tensor_statistics/aggregator.py
```diff
@@ -13,12 +13,14 @@
 from abc import ABC
 from abc import abstractmethod
 from itertools import islice
-from typing import Dict, TypeVar, Any
+from typing import Any
```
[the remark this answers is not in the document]
AlexanderDokuchaev: It will be not actual after use isort

nncf/experimental/torch/quantization/quantize.py
```diff
@@ -10,58 +10,54 @@
...
+    params = PostTrainingQuantizationParameters(
+        number_samples=subset_size,
+        preset=preset,
+        target_device=target_device,
+        ignored_scopes=ignored_scope,
+        model_type=model_type,
+        fast_bias_correction=fast_bias_correction,
+    )
+
+    nncf_logger.warning('Bias correction algorithms are not supported by Torch backend by now.')
```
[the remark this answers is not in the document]
AlexanderDokuchaev: There is removed all algorithms expect MinMax (BiasCorrection and FastBiasCorrection). I changed to remove only BiasCorrection.
I think BiasCorrection in warning it's correct.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -268,6 +272,9 @@ def _create_input_data(self,
         :return: The dictionary of the blob by input name.
         """
         input_blob = self._backend_entity.create_blob(input_shape, input_fp, channel_axis)
+        if input_name is None:
+            # For unnamed inputs, as in pytorch
```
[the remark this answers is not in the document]
AlexanderDokuchaev: In pytorch you have no way to get sub graph by names like onnx and ov 
https://github.com/openvinotoolkit/nncf/blob/cd4171c2cc09ffbdd0ae0520e9f19d2840fe6c44/nncf/onnx/graph/model_transformer.py#L371
Only get node node by name, for FBC it works correctly , as i understand the algorithm.
https://github.com/openvinotoolkit/nncf/blob/cd4171c2cc09ffbdd0ae0520e9f19d2840fe6c44/nncf/quantization/algorithms/fast_bias_correction/algorithm.py#L228-L229

nncf/torch/model_transformer.py
AlexanderDokuchaev: UpdateWeight is operator that contains FQ module.

nncf/torch/model_transformer.py
AlexanderDokuchaev: ok, changed 
But now, if somebody will call several extraction_transformations will apply only last.

nncf/torch/model_transformer.py
AlexanderDokuchaev: Not sure. Why not, and what alternative? Create new Torch model and change bias in new model? 
@vshampor

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: Nothing, return false. It will mean that node (Conv, Matmul) will not quantize.
@vshampor may be you know another way to check quantized node?

tests/torch/ptq/test_fast_bias_correction.py
AlexanderDokuchaev: @KodiaqQ  Do you have any suggestion about how to better test it?

nncf/common/tensor_statistics/aggregator.py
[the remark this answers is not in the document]
KodiaqQ: It's actual for now.

nncf/experimental/torch/quantization/quantize.py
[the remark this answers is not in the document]
KodiaqQ: The warning is not correct. 
`Bias correction algorithms` = `FastBiasCorrection` + `BiasCorrection`. 
Since `FastBiasCorrection` is applicable for PyTorch, the warning is not correct.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
KodiaqQ: Please, provide the output graph. 
Also, the `ModelExctactionCommand` should be applicable for `BiasCorrection`. Thus, we need to have a general command in case of multiple inputs.

nncf/torch/model_transformer.py
KodiaqQ: We need to understand the part of the model's output after the transformation.
Without such understanding, we can't be sure that algorithm works correctly.

nncf/torch/model_transformer.py
daniil-lyakhov: `output_insertion` inside method name isn't reflecting meaning of this method. It inserts hooks rather that outputs

tests/torch/ptq/test_fast_bias_correction.py
KodiaqQ: I think, in such a scenario (with one real and one synthetic model) test is useless since we have a conformance test. It's better to provide a full sanity test for the PTQ for the PyTorch backend.

nncf/experimental/torch/quantization/quantize.py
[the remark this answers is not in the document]
AlexanderDokuchaev: Is it means that i should return
```
min_max_params = params.algorithms[MinMaxQuantization]
params.algorithms = {MinMaxQuantization: min_max_params}
```

nncf/torch/model_transformer.py
AlexanderDokuchaev: And it's code that just moved from nncf_network.py and moved to function.

nncf/common/tensor_statistics/aggregator.py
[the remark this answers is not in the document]
AlexanderDokuchaev: It's actual but it's only recommended, not wrong.
`Try to avoid, but this is also applicable:`

nncf/onnx/graph/node_utils.py
```diff
@@ -55,15 +56,15 @@ def get_bias_value(node_with_bias: NNCFNode, model: onnx.ModelProto) -> np.ndarr
     if onnx_graph.has_initializer(bias_input_name):
         return onnx_graph.get_initializers_value(bias_input_name)
     node = onnx_graph.get_nodes_by_output(bias_input_name)[0]
-    metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node_with_bias.op_type)
+    metatype = ONNX_OPERATION_METATYPES.get_operator_metatype_by_op_name(node.op_type)
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Now it's falls that node_with_bias have no `op_type` attribute.
image

nncf/torch/model_transformer.py
AlexanderDokuchaev: PTModelTransformation does it in-place. https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/nncf_network.py#L948
It can be change but i prefer do it in another PR. 
@vshampor what you think about it?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
AlexanderDokuchaev: If export pt node to onnx graph (input/output nodes from onnx, not available in pt), extracted graphs looks like:
image

General command it will be hard to implement. PT model is not a graph, node have no names you have no any function to extract subgraphs by names as in onnx.

nncf/common/tensor_statistics/aggregator.py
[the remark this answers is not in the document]
KodiaqQ: All of us doing the one-line approach, and you've decided to increase the number of lines. Go ahead.

nncf/onnx/graph/node_utils.py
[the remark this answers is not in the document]
KodiaqQ: Have you tested that the new behavior works correctly?
Also, such code may not be actual for now. Let's remove it.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
kshpv: Can we make this function backend specific? It will not lead to a if-clause. 
Moreover, now this function doesn't do much. It just wraps the return value to the correct type.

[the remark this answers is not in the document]
KodiaqQ: Thank you for the graph. This is what I've asked about.
> General command it will be hard to implement.

So, what should we do for the `BiasCorrection` algorithm in PyTorch? Any suggestions?

nncf/quantization/algorithms/fast_bias_correction/backend.py
```diff
@@ -175,3 +181,23 @@ def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
...
+        Calculates bias shift magnitude based on the current and updated values.
+
+        :param current_bias_value: The original bias value.
+        :param updated_bias_value: The updated bias value.
+        :return: Magnitude between original and updated bias values.
+        """
+
+    @staticmethod
+    def post_process_output_data(data: List[TTensor]) -> TTensor:
+        """
```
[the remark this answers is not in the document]
KodiaqQ: @AlexanderDokuchaev, do not resolve not fixed comments, please.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -118,50 +125,49 @@ def _apply(self,
...
-                output_name=sub_output_name)
+                output_name=sub_output_name
+            )
 
             if bias_value.ndim > 1:
-                axes = [i for i in range(bias_value.ndim) if i != channel_axis]
-                bias_shift = np.expand_dims(bias_shift, axes)
+                new_shape = [1] * bias_value.ndim
+                new_shape[channel_axis] = bias_shift.shape[0]
+                bias_shift = bias_shift.reshape(new_shape)
```
kshpv, line 167: I would make a specific function that could be tested if needed. One more good point is that it will have typehints with all possible types of `bias_shift`. It is not explicit now

nncf/quantization/algorithms/fast_bias_correction/backend.py
```diff
@@ -13,21 +13,25 @@
...
-from nncf.common.graph import NNCFGraph
-from nncf.common.graph import NNCFNode
-from nncf.common.tensor_statistics.collectors import TensorStatisticCollectorBase
 from nncf.common.tensor_statistics.collectors import ReductionShape
+from nncf.common.tensor_statistics.collectors import TensorStatisticCollectorBase
 from nncf.common.utils.registry import Registry
 
-
 TModel = TypeVar('TModel')
+TTensor = TypeVar('TTensor')
```
kshpv, line 34: Is it Union[np.ndarray, torch.tensor]?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -118,50 +125,49 @@ def _apply(self,
 
         nncf_graph = NNCFGraphFactory.create(model)
         node_and_bias_value = (
-            (node, self._backend_entity.get_bias_value(node, nncf_graph, model)) \
-                for node in nncf_graph.get_all_nodes() if self._backend_entity.is_node_with_bias(node, nncf_graph)
+            (node, self._backend_entity.get_bias_value(node, nncf_graph, model))
+            for node in nncf_graph.get_all_nodes()
```
KodiaqQ, line 129: Such formatting looks weird.

> for node in nncf_graph.get_all_nodes()

Looks like missed `:`, not as part of the unnamed expression.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -292,28 +301,14 @@ def _get_bias_shift(self,
         raw_output = engine.infer(input_blob)
         q_outputs = self._backend_entity.process_model_output(raw_output, output_name)
```
kshpv, line 302: Should we split this function into two? The first is common with obtaining `raw_output` and the second one is backend specific with returning `bias_shift`?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
KodiaqQ: Also, as I can see, there exists the name of the sub-graph. 
Is the condition `if input_name is None:` really work? What would be in case we feed the named `input_data`?

tests/torch/ptq/test_fast_bias_correction.py
kshpv: We could test that the algo does its work, e.g. update bias values of the exact layers, and test that these values are expected. I would expect that test for all backends btw

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
KodiaqQ: Based on the code, this method will return `True`, because there is no check for the `weights` value. Checking based only on the `isinstance` of one of the inputs is wrong since we need to check the quantized **weights** not **activations or weights**.

nncf/torch/graph/node_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Returns the bias tensor for the biased node.
+
+    :param node: The node that corresponds to the operation with bias.
+    :param model: The model that contains this operation.
+    :return: The bias value that is applied to the output tensor of the node's operation.
+    """
+    node_module = model.get_containing_module(node.node_name)
+    if node_module.bias is None:
+        return None
```
KodiaqQ, line 43: So, what would happen in case of `None`? How this condition might work in case if we check `is_node_with_bias` previously?

nncf/torch/graph/node_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    Checks if the node has a bias or not.
+
+    :param node: The node to check.
+    :param model: The model that contains this operation.
+    :return: Return `True` if `node` corresponds to the operation
+        with bias (bias is added to the output tensor of that operation),
+        `False` otherwise.
+    """
+    node_module = model.get_containing_module(node.node_name)
+    return node.metatype in OPERATORS_WITH_BIAS_METATYPES and node_module.bias is not None
```
KodiaqQ, line 30: As I know, some values are not biased because the wrong shape for the "bias" constant exists.
@kshpv, have found it previously. Should we check it here too?

kshpv: After a discussion with @alexsu52 we decided that we should update all biases, neither their shape

nncf/torch/model_transformer.py
KodiaqQ: What means the new name `_apply_insertion_transformations`? Is it inserts `outputs`? Or, maybe `quantizers`?
Also, how is the `UpdateWeight` connected with the `insertion_transformations`?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
AlexanderDokuchaev: > Also, as I can see, there exists the name of the sub-graph.
It's exported graph by `torch.onnx.export(extracted_model, ...)`, and names in graphs on screenshot is default names for exported model. 

> If export pt node to onnx graph **(input/output nodes from onnx, not available in pt)**

Extracted model is instance of NNCFConv2d with prehook for FQ weights.

nncf/torch/model_transformer.py
AlexanderDokuchaev: `_apply_insertion_transformations` means that it works for all `PTInsertionCommand`. Is can insert all that support PTInsertionCommand.  

> Also, how is the UpdateWeight connected with the insertion_transformations?

Just `UpdateWeight` used for insertion target point to pytorch model in some cases.

https://github.com/openvinotoolkit/nncf/blob/818ad9c29e6c31e0a1c8cf33df8529dd92454e0b/nncf/torch/nncf_network.py#L950-L975

nncf/quantization/algorithms/fast_bias_correction/backend.py
AlexanderDokuchaev: Do you suggest to add backend specific import in common code?

nncf/torch/graph/node_utils.py
AlexanderDokuchaev: > So, what would happen in case of None? 

For FBC, it checks in function `is_node_with_bias` before using `get_bias_value`, that guarantied that bias will be `not None`.


>  How this condition might work in case if we check is_node_with_bias previously?

What do you mean? 
How is a check to `None` in function `get_bias_value` connected with `is_node_with_bias`?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: But it will be 3 functions absolutely identity expect of typehint. 
But np.array typehint, better change to 'TTensor' like used `TModel`.

kshpv: I mean one common function

nncf/torch/model_transformer.py
KodiaqQ: Since all OV and ONNX backends have different types of commands for output- and quantizer insertions, I've expected to see the same behavior in PyTorch. 

> Just UpdateWeight used for insertion target point to pytorch model in some cases.

That means that the name of the method is not correct. Let's fix it. Also, target points are needed not only for `weight update`.

AlexanderDokuchaev: > Since all OV and ONNX backends have different types of commands for output- and quantizer insertions, I've expected to see the same behavior in PyTorch.

Where in FBC used output insertion? Quantizer? How it connected with this PR?

> That means that the name of the method is not correct. 
Why? Could you explain me what wrong in naming function as `_apply_insertion_transformations` for `PTInsertionCommand`? And how `UpdateWeight` is connected with it?

> Also, target points are needed not only for weight update.
So `target points` used not only for `weight update`.

I dont see no any reason to rework using `PTInsertionCommand` in this PR.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: How you suggest to add `typehints with all possible types of bias_shift` in one common function?

nncf/torch/graph/node_utils.py
AlexanderDokuchaev: > Ok. Then, this code is useless. Am I right?

No, it separated functions. `get_bias_value` can be reused without `is_node_with_bias`. 
In your logic `get_bias_value` should raise exception `NoneType is not data attribute` if node without bias?

Which is problem here?

KodiaqQ: The problem with the returned value is that can be `None`. Is that the `None` value catches somewhere further in the code? No. This is the problem.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: Is not sub-graph. No names of node and no node, no graph in PT model.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
kshpv: `typehints with all possible types of bias_shift` - it was my misunderstanding. Now I see that this is can't be done in the common code.

nncf/torch/model_transformer.py
AlexanderDokuchaev: > How are the _apply_insertion_transformations connected with the UpdateWeight?

Adding `UpdateWeight` operator to nn.Module is not insertion transformation? Yes, why not? 

May be the problem that you think that `UpdateWeight` is something like `BiasCorrection` that once change value of weight. 
But no, it's used for adding FQ or PruningMasks modules to the PT node, that calls on each forward.
It's way in PT to transform the model. It's used because PT models is not a graph. 
 
https://github.com/openvinotoolkit/nncf/blob/fbfd7ed7b3c1192b08011bbc5e8a96cc59897804/nncf/torch/module_operations.py#L61-L67
 
is it confusing your?

I really don't understand what problem here? 
Please, explain clearly what is wrong and what you propose.

KodiaqQ: For me, the `UpdateWeight` class/method/etc., naming is wrong, because it's not clear and straightforward in terms of functionality. I mean, it's not obvious that `UpdateWeight` does something related to the `Quantizer`. The naming of such variables is not transparent in terms of usage in `insertion transformation`.

> It's used because PT models is not a graph.

It's not a problem. Just name such methods as `AddQuantizer` or `AddPruningMask`, or whatever.

nncf/torch/graph/node_utils.py
AlexanderDokuchaev: > Is that the None value catches somewhere further in the code? 

1. As i said for FBC, it checks in function `is_node_with_bias` before using get_bias_value, that guarantied that bias will be not None.
2. If `get_bias_value` without `is_node_with_bias`, yes need add check to None.
3. Do you suggest add `if bias is None:` after `get_bias_value`? It will be double check. Are your really sure that it need?

KodiaqQ: > As i said for FBC, it checks in function is_node_with_bias before using get_bias_value, that guarantied that bias will be not None.

It means that `if node_module.bias is None:` is useless, isn't it? 

> If get_bias_value without is_node_with_bias, yes need add check to None.

So, add it, please.

> Do you suggest add if bias is None: after get_bias_value? It will be double check. Are your really sure that it need?

No. Just do not add unnecessary code. Why unnecessary? See point 1.

nncf/torch/model_transformer.py
AlexanderDokuchaev: Ok, It's only your option about name of core module that exists before NNCF was published on Github. 

How i connected to `UpdateWeight` and why i should change it now?

KodiaqQ: I prefer to stop this discussion now. If you don't care about the code understanding, it's up to you.

nncf/torch/graph/node_utils.py
AlexanderDokuchaev: > It means that if node_module.bias is None: is useless, isn't it?

No, `get_bias_value` function is not FBC specific private function, it can be reused separately from `is_node_with_bias`. And function should not raise Exception.


So, you says that check `if node_module.bias is None:` in `get_bias_value` is useless, because it checks before `is_node_with_bias` and it will be always false. In same time you want to add `if bias is None:` that will be always False for same reason.
Do I correctly understand you?

>So, add it, please.
Deal, I will rework this moment.

nncf/torch/model_transformer.py
AlexanderDokuchaev: https://github.com/openvinotoolkit/nncf/pull/1664#discussion_r1157904428
Is it ok?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: Not sure

tests/torch/ptq/test_fast_bias_correction.py
AlexanderDokuchaev: Changed test to check that bias was changed, and dont fall on `bias is None`
Plus add test for reshape_bias_shift on torch.Tensors.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: Add check that FQ for weight

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: Moved to function, added test

nncf/torch/graph/node_utils.py
AlexanderDokuchaev: Remove `node_and_bias_value` generator, looks so large. Moved all checks in loop `for node in nncf_graph.get_all_nodes():` with check None.

AlexanderDokuchaev:
run openvino pre-commit tests

AlexanderDokuchaev:
run onnx pre-commit tests

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -124,51 +130,49 @@ def _apply(self,
...
+
             node_name = node.node_name
+            bias_value = self._backend_entity.get_bias_value(node, nncf_graph, model)
 
-            if not self._backend_entity.is_quantized_weights(node, nncf_graph):
+            if not self._backend_entity.is_quantized_weights(node, nncf_graph, model):
                 nncf_logger.debug(f'Skipping node {node_name} because weights were not quantized')
                 continue
 
+            if bias_value is None:
```
KodiaqQ, line 149: Why do we should check `if bias_value is None:` if we have check `if not self._backend_entity.is_node_with_bias(node, nncf_graph, model):` few lines above?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -179,13 +183,30 @@ def _apply(self,
...
         transformed_model = model_transformer.transform(transformation_layout)
 
         return transformed_model
 
+    @staticmethod
+    def reshape_bias_shift(bias_shift: TTensor, bias_value: TTensor, channel_axis: int) -> TTensor:
+        """
+        Reshape bias_shift tensor in case of dimensions of bias_value is more then 1.
+
+        :param bias_shift: Bias shit tensor.
```
KodiaqQ, line 198:
```suggestion
        :param bias_shift: Bias shift tensor.
```
:|

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -179,13 +183,30 @@ def _apply(self,
...
+        :param bias_value: Bias value tensor.
+        :param channel_axis: Axis to update bias.
+
+        :return TTensor: Updated bias_shift.
+        """
+        if bias_value.ndim > 1:
+            new_shape = [1] * bias_value.ndim
```
KodiaqQ, line 205: Is it eligible for `PyTorch` & `Tensorflow` data format?

tests/torch/ptq/test_fast_bias_correction.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.parametrize("with_bias, ref_bias", ((False, None), (True, Tensor([-1.9974, -1.9974]))))
+def test_fast_bias_correction_algo(with_bias, ref_bias):
+    """
+    Check working on fast bias correction algorithm and compare bias in quantized model with reference
+    """
+    model = ConvTestModel(bias=with_bias)
+    input_shape = [1, 1, 4, 4]
+    nncf_network = get_nncf_network(model, input_shape)
+    nncf_network.register_compression_module_type(ExtraCompressionModuleType.EXTERNAL_QUANTIZER)
```
daniil-lyakhov, line 34: Why you call this method here?

tests/onnx/quantization/test_fast_bias_correction.py
daniil-lyakhov, line 1: I suggest to make this test common as [this test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/test_ptq_params.py). This should speed up test writing for other backends (OV, for example)

tests/torch/ptq/test_fast_bias_correction.py
```diff
@@ -0,0 +1,65 @@
...
+    quantization_algorithm = get_min_max_and_fbc_algo_for_test()
+
+    def transform_fn(data_item):
+        images, _ = data_item
+        return images
+
+    torch.manual_seed(42)
```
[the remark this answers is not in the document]
kshpv: See `_seed` at `tests/torch/conftest.py`

tests/onnx/quantization/test_fast_bias_correction.py
AlexanderDokuchaev: done

tests/torch/ptq/test_fast_bias_correction.py
AlexanderDokuchaev: removed

AlexanderDokuchaev:
run openvino pre-commit tests

AlexanderDokuchaev:
run openvino pre-commit tests

ksilligan:
run openvino pre-commit tests

daniil-lyakhov (approved):

nncf/experimental/openvino_native/quantization/algorithms/fast_bias_correction/openvino_backend.py
```diff
@@ -78,13 +78,14 @@ def get_sub_input_output_names(subgraph: ov.Model) -> Tuple[str, str]:
         return subgraph.inputs[0].node.friendly_name, subgraph.outputs[0].node.friendly_name
 
     @staticmethod
-    def create_blob(shape: Tuple[int], data: List[np.ndarray], channel_axis: int) -> np.ndarray:
+    def create_input_data(shape: Tuple[int], data: List[np.ndarray], input_name: str, channel_axis: int) -> np.ndarray:
```
kshpv, line 81: Looks like return typehint should be updated

tests/onnx/quantization/test_fast_bias_correction.py
```diff
@@ -0,0 +1,60 @@
...
+    def get_model(with_bias, tmp_dir):
+        model = ConvTestModel(bias=with_bias)
+        onnx_path = f"{tmp_dir}/model.onnx"
+        torch.onnx.export(model, torch.rand([1, 1, 4, 4]), onnx_path, opset_version=13)
+        onnx_model = onnx.load(onnx_path)
+        return onnx_model
+
+    @staticmethod
+    def get_dataset(model):
+        np.random.seed(42)
```
kshpv, line 51: you don't need to use seed here. It is set inside `get_random_dataset_for_test`

tests/onnx/quantization/test_fast_bias_correction.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import List
+
+import numpy as np
+import onnx
+import torch
+
+from nncf.quantization.algorithms.fast_bias_correction.onnx_backend import ONNXFastBiasCorrectionAlgoBackend
+from tests.onnx.quantization.common import get_random_dataset_for_test
+from tests.post_training.test_fast_bias_correction import TemplateTestFBCAlgorithm
+from tests.torch.ptq.helpers import ConvTestModel
```
kshpv, line 22: Is it okay that we import the code from the torch-specific tests?

nncf/experimental/openvino_native/quantization/algorithms/fast_bias_correction/openvino_backend.py
AlexanderDokuchaev: updated

tests/onnx/quantization/test_fast_bias_correction.py
AlexanderDokuchaev: We used timm in tests for all backends that return torch model, looks okay.

kshpv (approved):
Nice! 👍

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -106,13 +105,17 @@ def _set_backend_entity(self, model: TModel) -> None:
...
-                ONNXFastBiasCorrectionAlgoBackend
+            from nncf.quantization.algorithms.fast_bias_correction.onnx_backend import ONNXFastBiasCorrectionAlgoBackend
             self._backend_entity = ONNXFastBiasCorrectionAlgoBackend()
         elif model_backend == BackendType.OPENVINO:
-            from nncf.experimental.openvino_native.quantization.algorithms.fast_bias_correction.openvino_backend import\
-                OVFastBiasCorrectionAlgoBackend
+            # pylint: disable=line-too-long
+            from nncf.experimental.openvino_native.quantization.algorithms.fast_bias_correction.openvino_backend import (
+                OVFastBiasCorrectionAlgoBackend,
+            )
```
KodiaqQ, line 114: Correct the formatting, please.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -179,13 +176,29 @@ def _apply(self,
...
+        """
+        Reshape bias_shift tensor in case of dimensions of bias_value is more then 1.
+
+        :param bias_shift: Bias shift tensor.
+        :param bias_value: Bias value tensor.
+        :param channel_axis: Axis to update bias.
+
+        :return TTensor: Updated bias_shift.
+        """
+        if bias_value.ndim > 1:
```
KodiaqQ, line 196: Why do we need this condition? Are there any cases when the bias had more than different dimensions? What would happen with bias if the shape would be <= 1?

nncf/quantization/algorithms/fast_bias_correction/backend.py
```diff
@@ -170,11 +176,43 @@ def process_model_output(raw_data: OutputType, output_name: str) -> NNCFTensor:
...
+        """
+        Convert data to backend specific type.
+
+        :param data: List of data.
+        :return TTensor: Converted data.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def reshape_tensor(data: TTensor, new_shape: List[int]) -> TTensor:
```
KodiaqQ, line 212: I would rephrase the method's naming cause it's too general and seems to be in common part.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,137 @@
...
+    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: NNCFNetwork) -> np.ndarray:
+        return get_bias_value(node, model)
+
+    @staticmethod
+    def get_activation_port_ids_for_bias_node(node: NNCFNode) -> Tuple[int, int]:
+        return 0, 0
+
+    @staticmethod
+    def process_model_output(raw_data: Dict, output_name: str) -> PTNNCFTensor:
+        return PTNNCFTensor(raw_data)
```
KodiaqQ, line 110: Why you don't use `output_name` here? The model can contain multiple outputs and the sub-model too.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @staticmethod
+    def is_quantized_weights(node: NNCFNode, nncf_graph: NNCFGraph, model: NNCFNetwork) -> bool:
+        node_module = model.get_containing_module(node.node_name)
+        for pre_op in node_module.pre_ops.values():
+            if isinstance(pre_op, UpdateWeight) and isinstance(pre_op.op, (BaseQuantizer, FakeQuantize)):
```
KodiaqQ, line 116: Why there are no checks for quantized weights? Can quantizers be followed by the `Reshape` or other types of layers before weighted layers?

tests/onnx/quantization/test_fast_bias_correction.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def get_data_from_node(model: onnx.ModelProto, node_name: str):
+    data = [t for t in model.graph.initializer if t.name == node_name]
+    if data:
+        return onnx.numpy_helper.to_array(data[0])
+    return None
+
+
+class TestTorchFBCAlgorithm(TemplateTestFBCAlgorithm):
```
KodiaqQ, line 32: Nevertheless, it is `TestTorchFBCAlgorithm` or `TestONNXFBCAlgorithm`?

tests/onnx/quantization/test_fast_bias_correction.py
```diff
@@ -0,0 +1,59 @@
...
+    @staticmethod
+    def list_to_backend_type(data: List) -> np.array:
```
KodiaqQ, line 34:
```suggestion
    def list_to_backend_type(data: List) -> np.ndarray:
```

tests/onnx/quantization/test_fast_bias_correction.py
```diff
@@ -0,0 +1,59 @@
...
+        return np.array(data)
+
+    @staticmethod
+    def get_backend() -> ONNXFastBiasCorrectionAlgoBackend:
+        return ONNXFastBiasCorrectionAlgoBackend
+
+    @staticmethod
+    def get_model(with_bias, tmp_dir):
```
KodiaqQ, line 42: Missed type hints.

tests/openvino/native/quantization/test_fq_params_calculation.py
```diff
@@ -78,16 +78,16 @@ def fixture_inplace_statistics(request):
...
     quantized_model = quantize_model(model.ov_model, {'preset': preset, 'inplace_statistics': inplace_statistics})
     nodes = get_fq_nodes_stats_algo(quantized_model)
 
     ref_stats_name = model.ref_graph_name.split(".")[0] + f'_{preset.value}.json'
     ref_stats_path = REFERENCE_SCALES_DIR / ref_stats_name
 
-    # Unkomment lines below to generate reference for new models.
-    # from tests.shared.helpers import dump_to_json
+    # Uncomment lines below to generate reference for new models.
+    # from tests.openvino.native.common import dump_to_json
```
KodiaqQ, line 90: `tests.openvino.native.common import dump_to_json` or `from tests.shared.helpers import dump_to_json`?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: It's part from https://github.com/openvinotoolkit/nncf/pull/1620. 
As i understand, it for case combination layers `conv` + `add`  where `add` can have differed shape. Example model `vimformer` for OV backend.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: Because there's no names of output. Sub-model in case of PT is CONV module (weights, bias, FQ for weights). 

Do you know case when `conv` module return multiply or named output?

KodiaqQ: Have you checked the pipeline of the sub-graph creation? Are you sure that sub-graph CAN NOT contain multiple outputs? In the ONNX/OV all outputs are named.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: > Why there are no checks for quantized weights? 

`if isinstance(pre_op, UpdateWeight) and isinstance(pre_op.op, (BaseQuantizer, FakeQuantize)):` is check for quantized weights. If Conv module have `UpdateWeight` operator that run FQ for weights, it means that node is quantized.
Or what about this question?

> Can quantizers be followed by the Reshape or other types of layers before weighted layers?

No, in torch FQ add in preops for weight directly.
It's limitation of PT quantization, that supported only weights that collected in `weights` attribute of `conv` module (only default conv module from torch). If weights is output of another node NNCF will no quantize this module.

KodiaqQ: `UpdateWeight` may be used for `outputs insertion` as we found previously. So, there is no guarantee that `UpdateWeight` means `weights quantization`, especially when we check all outputs, instead of the `weighted` port.

tests/openvino/native/quantization/test_fq_params_calculation.py
AlexanderDokuchaev: Fixed, want to fix only `unkomment`

tests/post_training/test_fast_bias_correction.py
```diff
@@ -0,0 +1,86 @@
...
+
+        algo = FastBiasCorrection(FastBiasCorrectionParameters(number_samples=1,inplace_statistics=False))
+        # pylint: disable=protected-access
+        algo._backend_entity = self.get_backend()
+        new_bias_shift = algo.reshape_bias_shift(bias_shift, bias_value, channel_axis)
+        assert list(new_bias_shift.shape) == ref_shape
+
+    @staticmethod
+    @abstractmethod
+    def get_model(with_bias, tmp_dir):
```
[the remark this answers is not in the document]
AlexanderDokuchaev: added for all functions of TemplateTestFBCAlgorithm

tests/onnx/quantization/test_fast_bias_correction.py
AlexanderDokuchaev: added

tests/onnx/quantization/test_fast_bias_correction.py
AlexanderDokuchaev: TestONNXFBCAlgorithm

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: No. `UpdateWeight` can update only weight, and if FQ is `op` of `UpdateWeight` it's means that `weights quantization`
`UpdateWeight` used for add NNCF operation (FQ, prunning by mask) for work with weights.

`outputs insertion` name of function was incorrected and changed it to `_apply_insertion_transformations`.

And again in torch no output/input nodes in graph, and it's not a graph.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: > Have you checked the pipeline of the sub-graph creation? Are you sure that sub-graph CAN NOT contain multiple outputs?
Yes, sure that it's not sub-graph. It just one of instance of `Conv` module.


> In the ONNX/OV all outputs are named. 
I know it, but in PT no.

KodiaqQ: No matter. Are you sure that the `Conv` module can contain only one output?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: `UpdateWeight` used for add NNCF operation (FQ, pruning by mask) for work with weights.

What should be detailed in this answer to answer your question?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: >Are you sure that the Conv module can contain only one output?

Ofc. In which backend `conv` layer can return more that one output?
https://pytorch.org/docs/stable/nn.html#convolution-layers

nncf/quantization/algorithms/fast_bias_correction/backend.py
AlexanderDokuchaev: It's general name for general function. Why is it problem?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: It is correct for long import line.

AlexanderDokuchaev:
conformance_test vs develop:
[torch_ptq_fbc.xlsx](https://github.com/openvinotoolkit/nncf/files/11287709/torch_ptq_fbc.xlsx)

AlexanderDokuchaev:
| {num_all_conv} | {num_conv_with_bias} | {num_updated_bias} | {skipped_by_magnitude}
-- | -- | -- | -- | --
darknet53 | 52 | 0 | 0 | 0
dla34 | 38 | 1 | 0 | 1
dpn68 | 72 | 1 | 0 | 1
efficientnet_b0 | 81 | 32 | 24 | 8
efficientnet_lite0 | 49 | 0 | 0 | 0
hrnet_w18 | 325 | 4 | 0 | 4
inception_resnet_v2 | 244 | 40 | 18 | 22
levit_128 | 4 | 0 | 0 | 0
mobilenetv2_050 | 52 | 0 | 0 | 0
regnetx_002 | 44 | 0 | 0 | 0
resnest14d | 27 | 8 | 2 | 6
resnet18 | 20 | 0 | 0 | 0
swin_base_patch4_window7_224 | 1 | 1 | 0 | 1
tf_inception_v3 | 94 | 0 | 0 | 0
vgg11 | 10 | 10 | 0 | 10
wide_resnet50_2 | 53 | 0 | 0 | 0
xception | 74 | 0 | 0 | 0

KodiaqQ:
> Problem: In most models `conv` layer with out bias, but have batchnorm after. In OV `conv+bn` merging in `conv+add`.

Are you going to check & solve this issue in this PR?

AlexanderDokuchaev:
Build with replacing nodes: 
https://ci-adas-icv.iotg.sclab.intel.com/view/all/job/NNCF/job/manual/job/post_training_quantization/6/artifact/results.html (some models falls by 500)
https://ci-adas-icv.iotg.sclab.intel.com/view/all/job/NNCF/job/manual/job/post_training_quantization/8/artifact/results.html (crossvit_9_240)

nncf/torch/graph/node_utils.py
```diff
@@ -0,0 +1,140 @@
...
+from torch.quantization.fake_quantize import FakeQuantize
+
+from nncf.common.graph.graph import NNCFNode
+from nncf.torch.graph.operator_metatypes import OPERATORS_FUSED_METATYPES
+from nncf.torch.graph.operator_metatypes import OPERATORS_WITH_BIAS_METATYPES
+from nncf.torch.module_operations import UpdateWeight
+from nncf.torch.nncf_network import NNCFNetwork
+from nncf.torch.quantization.layers import BaseQuantizer
+
+
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Don`t think that moving function to class methods is need. But moved some function to `model_transformer.py` and new fie `model_analyzer.py`.

nncf/torch/graph/operator_metatypes.py
```diff
@@ -161,6 +162,7 @@ class PTDepthwiseConv1dSubtype(PTDepthwiseConvOperatorSubtype):
     name = "Conv1DOp"
     hw_config_name = [HWConfigOpName.DEPTHWISECONVOLUTION]
     module_to_function_names = {NamespaceTarget.TORCH_NN_FUNCTIONAL: ["conv1d"]}
+    output_channel_axis = 1
```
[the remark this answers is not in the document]
AlexanderDokuchaev: https://github.com/openvinotoolkit/nncf/pull/1664/files#diff-ee203b9c6ad42581af42f6b15e3a9d34a84cfaba77ca1812ebf0b450b90b335aR21-R30

Add to `OperatorMetatype` with update docstring, but add default value as `None`. Think about `1` and but not sure about another backends and some metatype like Relu. Added `none` that needs to be added for each type there it's used.

vshampor (changes requested):

vshampor (approved):

nncf/torch/model_transformer.py
```diff
@@ -0,0 +1,132 @@
...
+    def _apply_bias_correction_transformations(self, transformations: List[PTBiasCorrectionCommand]) -> None:
+        for transformation in transformations:
+            update_fused_bias(
+                target_node_name=transformation.target_point.target_node_name,
+                new_bias=transformation.bias_value,
+                model=self._model,
+            )
+
+
+def update_fused_bias(target_node_name: str, new_bias: Tensor, model: NNCFNetwork):
```
KodiaqQ, line 100:
```suggestion
def update_fused_bias(target_node_name: str, new_bias: Tensor, model: NNCFNetwork) -> None:
```

tests/torch/experimental/replace_cusstom_modules/test_replace_custom_modules.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2023 Intel Corporation
```
KodiaqQ, line 1: Wrong folder name.

nncf/experimental/torch/replace_custom_modules/timm_custom_modules.py
```diff
@@ -0,0 +1,137 @@
...
+        device=module.weight.device,
+        dtype=module.weight.dtype,
+    )
+    copy_parameters(module, new_norm)
+    new_drop = deepcopy(module.drop)
+    new_act = deepcopy(module.act)
+    return nn.Sequential(new_norm, new_drop, new_act)
+
+
+REPLACE_FN_MAP = {
```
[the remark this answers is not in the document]
AlexanderDokuchaev: May be yes, but how to define map with undefined variables?

nncf/torch/graph/operator_metatypes.py
```diff
@@ -843,3 +863,20 @@ def get_operator_metatypes() -> List[Type[OperatorMetatype]]:
...
+OPERATORS_WITH_BIAS_METATYPES = [
+    PTModuleConv1dMetatype,
+    PTModuleConv2dMetatype,
+    PTModuleConv3dMetatype,
+    PTDepthwiseConv1dSubtype,
+    PTDepthwiseConv2dSubtype,
+    PTDepthwiseConv3dSubtype,
+    PTModuleConvTranspose1dMetatype,
+    PTModuleConvTranspose2dMetatype,
+    PTModuleConvTranspose3dMetatype,
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Good point, but suggest to merge as it. I will add it, but in next PR, need to run post_training tests for it.
In onnx too (https://github.com/openvinotoolkit/nncf/blob/develop/nncf/onnx/graph/metatypes/onnx_metatypes.py#LL506C1-L506C31).

nncf/torch/graph/transformations/commands.py
```diff
@@ -37,13 +37,15 @@ class PTTargetPoint(TargetPoint):
     ]
     _HOOK_TYPES = [TargetType.OPERATOR_PRE_HOOK, TargetType.OPERATOR_POST_HOOK]
 
+    _LAYER_TYPE = [TargetType.LAYER]
```
vshampor, line 40: Looks OK.

nncf/torch/model_analyzer.py
```diff
@@ -0,0 +1,103 @@
...
+    Checks if the node has a fused bias.
+
+    :param node: The node to check.
+    :param model: The model that contains this operation.
+    :return: Return `True` if `node` corresponds to the operation
+        with bias (bias is added to the output tensor of that operation),
+        `False` otherwise.
+    """
+    fused_node = get_potential_fused_node(node.node_name, model)
+    node_module = model.get_containing_module(node.node_name)
```
alexsu52, line 57: This call generate the following warning:
```
/home/susloval/work/projects/nncf/nncf/torch/nncf_network.py:934: FutureWarning: Old style of accessing NNCF-specific attributes and methods on NNCFNetwork objects is deprecated. Access the NNCF-specific attrs through the NNCFInterface, which is set up as an `nncf` attribute on the compressed model object.
For instance, instead of `compressed_model.get_graph()` you should now write `compressed_model.nncf.get_graph()`.
The old style will be removed after NNCF v2.5.0
```
Please migrate to the new NNCFNetwork representation here and elsewhere. cc' @vshampor

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -262,7 +284,7 @@ def _add_statistic_point(self, container: StatisticPointsContainer, point: Targe
         )
 
     def _create_input_data(
```
alexsu52, line 286: I couldn't find where you use this function after your changes. Maybe you can delete it safely.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -281,9 +306,9 @@ def _get_bias_shift(
         model: TModel,
         input_blob: Dict[str, NNCFTensor],
```
alexsu52, line 307:
```suggestion
        input_blob: Union[TTensor, Dict[str, TTensor]],
```

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,142 @@
...
+    def get_activation_port_ids_for_bias_node(node: NNCFNode) -> Tuple[int, int]:
+        return 0, 0
+
+    @staticmethod
+    def process_model_output(raw_data: Dict, output_name: str) -> PTNNCFTensor:
+        return PTNNCFTensor(raw_data)
+
+    @staticmethod
+    def is_quantized_weights(node: NNCFNode, nncf_graph: NNCFGraph, model: NNCFNetwork) -> bool:
+        return is_quantized_weights(node, model)
```
alexsu52, line 112: Could you briefly explain why you can not determinate quantized weights by `NNCFGraph`?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -0,0 +1,142 @@
...
+
+    @staticmethod
+    def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph, model: NNCFNetwork) -> bool:
+        return is_node_with_fused_bias(node, model)
```
alexsu52, line 116: Could you briefly explain why you cannot determine fused bias by NNC Graph? I guess such functionality can be general and implemented during collecting layer attributes.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -299,40 +324,29 @@ def _get_bias_shift(
...
-    @staticmethod
-    def _get_bias_shift_magnitude(current_bias_value: np.ndarray, updated_bias_value: np.ndarray) -> float:
-        """
-        Calculates bias shift magnitude based on the current and updated values.
-
-        :param current_bias_value: The original bias value.
-        :param updated_bias_value: The updated bias value.
-        :return: Magnitude between original and updated bias values.
-        """
-        bias_shift_magnitude = np.inf
```
alexsu52, line 314: Could you implement this logic using `NNCFTensor`? cc' @daniil-lyakhov

nncf/experimental/torch/replace_custom_modules/replace_custom_modules.py
```diff
@@ -0,0 +1,78 @@
...
+    :param module: The module to try to convert.
+    :return nn.Module: The replaced module if the function succeeded in replacing the module, None otherwise.
+    """
+    replaced_module = None
+    if replace_timm_modules is not None:
+        replaced_module = replace_timm_modules(module)
+    return replaced_module
+
+
+def replace_modules(target_module: nn.Module):
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Without changes not, but i add `predicate_fn` arguments to set condition function to collect modules to replace and it's works. Thanks for suggestion.

nncf/torch/model_analyzer.py
AlexanderDokuchaev: Fixed

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: nncf_graph does not contain information about quantization and existing bias (like conv2d without bias) for torch model.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
AlexanderDokuchaev: I'm not sure that it's need for general implementation, it looks like torch specific case.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: What do you mean?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
alexsu52: I'm surprised that this information is not in NNCFGraph. Let's discuss it offline.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
alexsu52: We have abstractions `NNCFTensorProcessors` and `NNCFTensor` to implement shared logic of work with tensor. Is it possible to use it here?

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
alexsu52: We are going to minimize backend interfaces but this PR extending it. One way is used backed agnostic abstractions such as NNCFGraph, to do it. Thus I want to understand why we cannot extend NNCFGraph and implement algorithm that will find bias nodes based NNCFGraph, because It looks like the general functionality between frameworks. Let's discuss it offline, may be we will find solution.

KodiaqQ (approved):

alexsu52 (approved):

PR #1667, opened 2023-03-29
[PTQ] Add softmax_matmul pattern

kshpv:
### Changes

Add pattern Softmax -> MatMul <- Non_Pattern_Node

### Reason for changes

To align the quantization scheme with POT - don't insert FQ into these red areas.

image

### Related tickets

106873

### Tests

Tested manually


alexsu52 (commented):
For which model does this apply? Could you provide link to the POT pattern?

PR #1673, opened 2023-03-30
Register controller in NNCFNetwork at create_compressed_model

vshampor:
### Changes
NNCFNetwork objects leaving `create_compressed_model` now have a controller reference available in them.

### Reason for changes
Much ore convenient to access the controller through the model object in certain cases.

### Related tickets
N/A

### Tests
tests.torch.test_algo_common.test_compressed_model_has_controller_references

### DIFF
--- a/nncf/torch/nncf_network.py
+++ b/nncf/torch/nncf_network.py
@@ -288,8 +288,7 @@ class NNCFNetworkInterface(torch.nn.Module):
                                                           ShapeIgnoringTensorMetaComparator())
         self._load_listener = None
 
-
-
+        self.compression_controller = None  # type: PTCompressionAlgorithmController
 
     @property
     def _model_ref(self) -> 'NNCFNetwork':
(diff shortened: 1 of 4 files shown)


nncf/torch/nncf_network.py
```diff
@@ -699,6 +698,9 @@ def get_node_to_op_address_mapping(self) -> Dict[NNCFNodeName, OperationAddress]
             retval[nncf_node.node_name] = op_address
         return retval
 
+    def set_compression_controller(self, ctrl: 'PTCompressionAlgorithmController'):
+        self.compression_controller = ctrl
```
AlexanderDokuchaev, line 702: How about add short name for `compression_controller`, for example `model.nncf.ctrl` or  `model.nncf.controller`?`compression_controller` looks too long.

vshampor:
run openvino pre-commit tests

alexsu52 (approved):
LGTM

nncf/torch/nncf_network.py
vshampor: I thought that it would be better to be more explicit here in the naming.

PR #1674, opened 2023-03-30
Skip BN adaptation in some cases to enable shorter compression configs

vshampor:
### Changes
Now `create_compressed_model` won't fail no BN adaptation data loader is given, but there was nothing in the NNCFConfig about BN adaptation anyway. If we skip BN adaptation in this way, log a corresponding message at `INFO` level.

### Reason for changes
This allows using shorter configs such as `"compression": { "quantization" }` without calling `register_default_init_args`. Examples would be test cases where we don't need BN adaptation mostly, but need a model with quantizers at proper places and do not want to define a dummy dataloader or to explicitly specify 0 BN adaptation samples in config. Also might make it easier for the user to quickly evaluate NNCF even if they missed the `register_default_init_args` call. 

### Related tickets
N/A

### Tests
tests.torch.test_algo_common.test_can_apply_algo_with_single_line, also some tests that couldn't pass without setting `"num_bn_adaptation_samples": 0` were adjusted and are now passing without it.

### DIFF
--- a/nncf/common/compression.py
+++ b/nncf/common/compression.py
@@ -27,8 +27,10 @@ from nncf.common.schedulers import StubCompressionScheduler
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
 from nncf.common.utils.registry import Registry
+from nncf.config.extractors import BNAdaptDataLoaderNotFoundError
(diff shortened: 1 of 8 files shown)


vshampor:
run openvino pre-commit tests

alexsu52 (approved):
LGTM

PR #1679, opened 2023-04-03
Fix prune shape

jpablomch:
### Changes

Allow next_nodes_cluster to include nodes in clusters.

### Reason for changes

1. Code is simpler.
2. Current flops count function is incorrect when dealing with some specific designs of 'overwrite_groups' in elastic width.

For example:

```
"overwrite_groups": [[
        "ResNet/Sequential[layer2]/Bottleneck[0]/NNCFConv2d[conv1]/conv2d_0",
        "ResNet/Sequential[layer2]/Bottleneck[0]/NNCFConv2d[conv2]/conv2d_0"
 ]],
"overwrite_groups_widths":[[128, 96, 64]]
```

conv1 and conv2 are connected, conv2 is not a depthwise conv.

When we activate the minimum subnet:

output_channel[conv1] = 64, output_channel[conv2] = 64
input_channel[conv2] = 64

However, if we use the current flops count function, input_channel[conv2]=128. This is because conv2 is excluded from next_nodes_cluster and conv2 is not a depthwise conv.

### Related tickets

N/A

### Tests

Added a test in test_flops.py to check the flops count when using 'overwrite_groups' in elastic width.

### DIFF
--- a/nncf/common/pruning/shape_pruning_processor.py
+++ b/nncf/common/pruning/shape_pruning_processor.py
@@ -126,8 +126,6 @@ class ShapePruningProcessor:
 
         for node in cluster.elements:
             output_channels[node.node_name] -= pruned_elems
-            if node.is_depthwise:
-                input_channels[node.node_name] -= pruned_elems
 
         # Prune in channels in all next nodes
         next_nodes_info = pruning_groups_next_nodes[cluster.id]
@@ -201,7 +199,6 @@ class ShapePruningProcessor:
                 curr_next_nodes = get_next_nodes_of_types(graph, nncf_cluster_node, self._prunable_types)
 
                 next_nodes_cluster = next_nodes_cluster.union(curr_next_nodes)
-            next_nodes_cluster = next_nodes_cluster - cluster_nodes
             next_nodes[cluster.id] = []
             for next_node in next_nodes_cluster:
                 sparse_multiplier = self._get_next_node_sparse_multiplier(graph, next_node, cluster)
--- a/tests/torch/nas/test_flops.py
+++ b/tests/torch/nas/test_flops.py
@@ -51,6 +51,26 @@ LIST_OF_ME_DESCS = [
         blocks_to_skip=RESNET50_BLOCK_TO_SKIP,
     ),
     MultiElasticityTestDesc(
+        name='resnet50_width_overwrite_groups',
+        model_creator=resnet50_cifar10,
+        ref_model_stats=RefModelStats(
+            supernet=ModelStats(651_599_872, 23_467_712),
+            kernel_stage=ModelStats(651_599_872, 23_467_712),
+            depth_stage=ModelStats(615_948_288, 23_398_080),
+            width_stage=ModelStats(589_209_600, 23_238_336)
+        ),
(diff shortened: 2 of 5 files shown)


nncf/common/pruning/shape_pruning_processor.py
```diff
@@ -126,8 +126,6 @@ def prune_cluster_shapes(self,
 
         for node in cluster.elements:
             output_channels[node.node_name] -= pruned_elems
-            if node.is_depthwise:
-                input_channels[node.node_name] -= pruned_elems
```
jpablomch, line 130: The reason to remove this part is that depthwise nodes are in `next_nodes_info`. (Yi)

tests/torch/nas/test_flops.py
```diff
@@ -50,6 +50,26 @@
...
+        model_creator=resnet50_cifar10,
+        ref_model_stats=RefModelStats(
+            supernet=ModelStats(651_599_872, 23_467_712),
+            kernel_stage=ModelStats(651_599_872, 23_467_712),
+            depth_stage=ModelStats(615_948_288, 23_398_080),
+            width_stage=ModelStats(589_209_600, 23_238_336)
+        ),
+        blocks_to_skip=RESNET50_BLOCK_TO_SKIP,
+        algo_params={'width': {
+            "overwrite_groups": [
```
ljaljushkin, line 64: Why do you need to overwrite groups? I doubt that eventually user will use it.

ljaljushkin (changes requested):

tests/torch/nas/test_flops.py
jpablomch: @ljaljushkin  Overwrite groups allow the user to have a more granular control over the levels of elastic width that might be applied. It has proven to be a very useful knob. Thanks!

ljaljushkin: But it would mean no automation for elastic width, right?

jpablomch: @ljaljushkin No, automation is still important. For instance, we first try with elasticity recommended by NNCF. However, it is important to also allow the user to go beyond the automated recommendation. Thanks!

tests/torch/nas/test_flops.py
```diff
@@ -50,6 +50,26 @@
...
+                [
+                    "ResNet/Sequential[layer2]/Bottleneck[0]/NNCFConv2d[conv1]/conv2d_0",
+                    "ResNet/Sequential[layer2]/Bottleneck[0]/NNCFConv2d[conv2]/conv2d_0"
+                ],
+            ],
+            "overwrite_groups_widths":[[128, 96, 64]]
```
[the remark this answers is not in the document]
jpablomch: @ljaljushkin , it is only one group and the 3 values are the possible widths that the group can have. Thanks!

nncf/common/pruning/shape_pruning_processor.py
daniil-lyakhov: Yes, this part here is important and will brake shape pruning in the filter pruning algo

jpablomch:
Closing this PR because Daniil has solved the issue in a different commit. Thanks!

PR #1657, opened 2023-03-23
Draft of the automatic structured pruning for JPQD

ljaljushkin:

PR #1680, opened 2023-04-03
[2/N] Added automatic structured pruning for DistillBERT, CLIP.

ljaljushkin:
### Changes

Continue of the #1657 
Supported finding pruning groups for DistillBERT, CLIP models
Proper filtering groups - no opened/closed branches, invalidate when reach "bad" operation (stop_mask, ignored_scope, output).
Introduced block hierarchy class for finding groups on the leaves and for better visualization.
Refactored the functionality.

Block hierarchy with colored nodes: red - for invalid, green - for final leaves, grey - intermediate, not active.
image

added rules for DistillBERT case when pruning masks intersect with attention masks:
image

### Reason for changes

Expand scope of the supported models in JPDQ

### Related tickets

106556

### Tests

test_groups


nncf/experimental/common/pruning/operations.py
```diff
@@ -0,0 +1,700 @@
...
+        propagated_mask = None
+        if i1 != i2:
+            nncf_logger.warning(f"expand_as takes the shape from the node with propagation mask and pruning "
+                                "dimension in the mask matches the dimension in the expanded input. Currently, "
+                                "it's not supported and mask is invalidated. node_name={node.node_name}")
+            if i1 != None:
+                i1.invalidate_groups()
+        else:
+            propagated_mask = i1
+        node.data['output_mask'] = propagated_mask
```
ljaljushkin, line 700: image
limited masked fill for the case when on both branches the same mask for the sake of simple implementation, since it's not often presented in the models.

nncf/experimental/common/pruning/operations.py
```diff
@@ -0,0 +1,700 @@
...
+
+class ConvolutionPruningOp(BasePruningOp):
+    @classmethod
+    def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
+                         tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> None:
+        input_masks = get_input_masks(node, graph)
+        output_mask = node.data.get('output_mask', None)
+
+        if is_grouped_conv(node):
+            output_mask = None
```
daniil-lyakhov, line 113: Should we invalidate masks here?

nncf/experimental/common/pruning/nodes_grouping.py
```diff
@@ -422,38 +145,26 @@ def get_attributes_fn(node: NNCFNode) -> Dict[str, Any]:
...
+    """
+    finished_producers = set()
+    sorted_groups = sorted(pruning_groups, key=lambda group: len(group.dim_blocks), reverse=True)
+    result = []
+    for group in sorted_groups:
+        producers = {block.producer_id for block in group.dim_blocks}
+        if not producers.intersection(finished_producers):
+            finished_producers.update(producers)
+            result.append(group)
+    return result
```
daniil-lyakhov, line 170: New line missed

nncf/experimental/common/pruning/propagation_data.py
```diff
@@ -0,0 +1,221 @@
...
+class PropagationGroup:
+    """
+    Defines a group of propagation blocks and links it with the list of children groups.
+    The group is initialized on producers of pruning and is propagated until consumers within PropagationMask.
+    """
+
+    def __init__(self, blocks: List[PropagationBlock]) -> None:
+        self._blocks = blocks
+        self._children: List['PropagationGroup'] = []
+        self.is_invalid = False
```
daniil-lyakhov, line 129: Can we define `is_invalid` attr as property?

nncf/experimental/common/pruning/operations.py
```diff
@@ -93,16 +92,13 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
                          tensor_processor: Type[NNCFPruningBaseTensorProcessor]) -> None:
         node.data['output_mask'] = None
 
-
 class OutputPruningOp(BasePruningOp):
-    @classmethod
-    def accept_pruned_input(cls, node: NNCFNode) -> bool:
-        return True
-
```
daniil-lyakhov, line 101: I think this method is required for proper mask propagation

ljaljushkin: It's not used anywhere. propose to add when will actually needed

nncf/experimental/common/pruning/nodes_grouping.py
ljaljushkin: strange that linter is OK with it

nncf/experimental/common/pruning/operations.py
daniil-lyakhov: Ah, this method is for legacy node selector algo, you are right

nncf/experimental/common/pruning/propagation_data.py
daniil-lyakhov: Attribute `is_invalid` should be set only by `invalidate` call, so I suggest to make `is_valid` access rule read only

nncf/experimental/common/pruning/operations.py
daniil-lyakhov: Important: Ngraph pruning transformation most likely won't work in such cases, masks will be invalidated. Should be checked

nncf/experimental/common/pruning/operations.py
```diff
@@ -0,0 +1,700 @@
...
+            source_shape = input_edges[0].tensor_shape
+            for dim, groups in mask.dim_groups_map.items():
+                if target_shape[dim] == source_shape[dim]:
+                    nncf_logger.warning(f"expand_as takes the shape from the node with propagation mask and pruning "
+                                        "dimension in the mask matches the dimension in the expanded input. Currently, "
+                                        "it's not supported and mask is invalidated. node_name={node.node_name}")
+                    for group in groups:
+                        group.invalidate()
+            propagated_mask = mask
+        node.data['output_mask'] = propagated_mask
```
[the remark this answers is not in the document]
ljaljushkin: Discussed locally. NGraph implementation has the same limitation. It will work only if attention mask is always filled with zeros, which is the case for inference. 
cc: @AlexKoff88

nncf/experimental/common/pruning/operations.py
ljaljushkin: Discussed in person, currently assume that mask is always zero and doesn't affecrt on inference

daniil-lyakhov (approved):

nncf/experimental/common/pruning/operations.py
```diff
@@ -668,15 +689,59 @@ def mask_propagation(cls, node: NNCFNode, graph: NNCFGraph,
...
+            target_shape = input_edges[1].tensor_shape
+            source_shape = input_edges[0].tensor_shape
+            for dim, groups in mask.dim_groups_map.items():
+                if target_shape[dim] == source_shape[dim]:
+                    nncf_logger.warning(f"expand_as takes the shape from the node with propagation mask and pruning "
+                                        "dimension in the mask matches the dimension in the expanded input. Currently, "
+                                        "it's not supported and mask is invalidated. node_name={node.node_name}")
+                    for group in groups:
+                        group.invalidate()
+            # TODO: assume that expand_as is on constant path that does not affect pruning, otherwise pruning of
```
alexsu52, line 725: please, format TODO regarding the style guide.

alexsu52 (approved):
LGTM

nncf/experimental/common/pruning/operations.py
ljaljushkin: Discussed in person, will be resolved in https://github.com/openvinotoolkit/nncf/pull/1704

PR #1683, opened 2023-04-05
Fix incorrect work of `is_quantized_weights` in ONNX backend of PTQ

AlexanderDokuchaev:
### Changes

Use `get_input_edges` that return sorted edges by port_id instead of `get_previous_nodes`

### Reason for changes

`get_previous_nodes` returns unsorted nodes, in result `is_quantized_weights` can return false negative result.

### DIFF
--- a/nncf/quantization/algorithms/bias_correction/onnx_backend.py
+++ b/nncf/quantization/algorithms/bias_correction/onnx_backend.py
@@ -117,7 +117,7 @@ class ONNXBiasCorrectionAlgoBackend(BiasCorrectionAlgoBackend):
 
     @staticmethod
     def is_quantized_weights(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
-        input_nodes = nncf_graph.get_previous_nodes(node)
+        input_nodes = [edge.from_node for edge in nncf_graph.get_input_edges(node)]
         weight_port_id = node.metatype.weight_definitions.weight_port_id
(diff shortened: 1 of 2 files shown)


KodiaqQ (approved):

kshpv:
@AlexanderDokuchaev, thanks.
How did you detect this issue?

AlexanderDokuchaev:
> How did you detect this issue?

Same problem was in filter_pruning some time ago.

kshpv:
> > How did you detect this issue?
> 
> Same problem was in filter_pruning some time ago.

Can we fix it on the side of NNCFGraph then?

AlexanderDokuchaev:
> Can we fix it on the side of NNCFGraph then?
What fix? NNCFGraph works as described.

kshpv:
> > Can we fix it on the side of NNCFGraph then?
> 
> What fix? NNCFGraph works as described.

I mean if the same issue has already occurred and it happens again, probably, it is better to update the NNCFGraph function itself. It could return the nodes in the sorted order, but it should be checked whether it brakes other logic.

AlexanderDokuchaev:
@kshpv You are free to do it. 
I just fixed the bug.

kshpv:
> @kshpv You are free to do it. I just fixed the bug.

You fixed a consequence of the bug.
I suppose that the initial bug is inside of NNCFGraph, don't you think so?

kshpv (approved):

PR #1685, opened 2023-04-05
Move solver logs into debug

KodiaqQ:
### Changes

- Move solver logs into debug ("Not adding activation input quantizer...")
- Fixed minor PostprocessingNodeLocator bug

### Reason for changes

- Huge number of messages for users

### Related tickets

- 106545


alexsu52 (approved):
LGTM

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -322,7 +322,7 @@ def forward_traverse_function(node_key: str, output: List[str],
                                                               output=[], traverse_forward=False)
             if self._post_processing_marker_encountered:
                 ignored_node_keys = self._get_ignored_node_keys(node_keys)
-                output.extend(ignored_node_keys)
+                output.update(ignored_node_keys)
```
kshpv, line 325: Update typehint of the method

kshpv (changes requested):

nncf/common/quantization/quantizer_propagation/solver.py
KodiaqQ: Done

kshpv (approved):

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -908,7 +908,7 @@ def setup_initial_quantizers(self,
                 num_input_activations = quant_prop_graph.get_num_input_activations(node_key)
                 self._num_potential_quantized_activations += num_input_activations
                 if node_key in quant_prop_graph.ignored_node_keys:
-                    nncf_logger.info(f"Not adding activation input quantizer for operation: {node_key}")
+                    nncf_logger.debug(f"Not adding activation input quantizer for operation: {node_key}")
```
vshampor, line 911: The use case of notifying the user in the logs about the effect of the "ignored_scopes" is now lost. This is not the way to go.

Possible solutions:
1. Suppress this type of output in cases where the automatic ignored scope generation is used (e.g. in PTQ). This can be done using https://docs.python.org/3/library/logging.html#filter-objects

or


2. Clearly separate the ignored scopes that were specified by the user in the config file from the auto-generated ignored scopes, and only print logs for the former. This would probably require a signature change for the solver's `__init__` method.

or


3. Collect the operations for which the activation input quantizers were not collected, and if the number of such ops exceeds a predetermined number, only log to info the number of the ignored ops, and the complete list of op names - to debug.

vshampor (changes requested):

nncf/common/quantization/quantizer_propagation/solver.py
KodiaqQ: Suppressing such logs in the long term is not a solution, agree. But it is a quick fix for a current problem that is described in the ticket and discussed several times.
To be clear about the root cause of the problem: we will see such messages 1000+ times for networks that contain decomposed DetectionOutput.
I support the second way, which implies the separation of the user's and internal `ignored scopes`. But do we really need to do it right now?
cc @alexsu52, @AlexKoff88

alexsu52: I can agree with @vshampor statement. I would recommend to implement it in the follow up PR. @KodiaqQ Please open a ticket.

KodiaqQ: What should we do, @vshampor?

vshampor (approved):

KodiaqQ:

KodiaqQ:

PR #1692, opened 2023-04-06
[skip ci] Add pre-commit action

AlexanderDokuchaev:
### Changes

Add pre-comit config and github action for:
- black
- isort

### Related tickets

CVS-74733


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,16 @@
...
+      - opened
+      - reopened
+      - synchronize
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
```
alexsu52, line 15: Which version will be setup?

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,16 @@
...
+    - uses: pre-commit/action@v3.0.0
```
alexsu52, line 16: It looks like using `pre-commit/action` lead to double cloning NNCF repository.

alexsu52, line 16: As you know, we use make to run tests before commit, and I think it would be good if we use one entry point to test and apply style:
`make style`
`make check-style`
In this case, we can implement the pre-commit action on our side like this `github action` -> `make check-style` -> `pre-commit run`

examples/common/sample_config.py
```diff
@@ -28,9 +28,18 @@
 class ActionWrapper(argparse.Action):
     def __init__(self, action):
         self._action = action
-        super().__init__(action.option_strings, action.dest, nargs=action.nargs, const=action.const,
-                         default=action.default, type=action.type, choices=action.choices, required=action.required,
-                         help=action.help, metavar=action.metavar)
+        super().__init__(
```
alexsu52, line 31: I propose to move the formatting examples into a separate PR so as not to delay the merging of this PR.

pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[tool.black]
+line-length = 120
+
+[tool.isort]
+force_single_line = true
+line_length = 120
```
alexsu52, line 6:
```suggestion
force_single_line = true
single_line_exclusions: ('typing',)
line_length = 120
```

alexsu52 (changes requested):
During review of the PR I found the https://pypi.org/project/pyink/. It looks like it can avoid huge formatting of the repository.

cc' @vshampor

.github/workflows/pre-commit.yml
AlexanderDokuchaev: Added `make pre-commit`, it's more general name, if we will add another check like pylint, mypy.
`make check-style` i think it's no need, as code should be formatted. And i did not found direct flags to do it in pre-commit.

examples/common/sample_config.py
AlexanderDokuchaev: yeap, good idea.

.github/workflows/pre-commit.yml
AlexanderDokuchaev: Set python3.8.10 as our test environment

.github/workflows/pre-commit.yml
AlexanderDokuchaev: I remove pre-commit to use make instead. 
But no, fetch repository happens only on checkout step. https://github.com/openvinotoolkit/nncf/actions/runs/4630679783/jobs/8192572026

alexsu52 (approved):
LGTM

PR #1700, opened 2023-04-12
Support for ignored operation types in calibrate.py

alexsu52:
### Changes

- Support for peration types in calibrate.py
- Refactor IgnoredScope class to the datatclass.

### Reason for changes

This caused operation types from the ignored scope to be skipped.

### Related tickets

ref: 105079

### Tests

N/A

### DIFF
--- a/nncf/scopes.py
+++ b/nncf/scopes.py
@@ -11,13 +11,16 @@
  limitations under the License.
 """
 
+from dataclasses import dataclass
+from dataclasses import field
 import re
 from typing import List, Optional
 
-from nncf.common.logging import nncf_logger
 from nncf.common.graph.graph import NNCFGraph
+from nncf.common.logging import nncf_logger
 
 
+@dataclass
 class IgnoredScope:
     """
     Dataclass that contains description of the ignored scope.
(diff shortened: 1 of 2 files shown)


l-bat (approved):

nncf/scopes.py
```diff
@@ -53,21 +56,15 @@ class IgnoredScope:
...
-        """
-        self.names = names if names is not None else []
-        self.patterns = patterns if patterns is not None else []
-        self.types = types if types is not None else []
+    :param names: List of ignored node names.
+    :param patterns: List of regular expressions that define patterns for names of
+        ignored nodes.
+    :param types: List of ignored operation types.
+    """
+    names: List[str] = field(default_factory=list)
```
andrey-churkin, line 65: Only Python 3.10 supports this

https://docs.python.org/3/library/dataclasses.html#dataclasses.field

alexsu52: Maybe something new in 3.10. But it is also supported in Python 3.7.

https://docs.python.org/3.7/library/dataclasses.html#dataclasses.field

alexsu52:

PR #1701, opened 2023-04-12
API management and documentation mechanisms

vshampor:
### Changes
Added a decorator for marking entities as API and a set of Github actions to check whether a PR changes API (if so, then the PR will be automatically labeled as such) and to generate and push API documentation automatically to Github pages using Sphinx, AutoAPI (by readthedocs) and custom API name collection code.

### Reason for changes
Both devs and users will finally have an understanding of what is considered API and what is not, and have an opportunity to browse API documentation without having to look into source code.

### Related tickets
99696

### Tests
N/A


vshampor:
I had to experiment on my fork to properly debug the Github actions. The end result on Github pages would look like this:
https://vshampor.github.io/nncf/

(It used to be that this page contained only the schema, and now it has been extended to API docs. The schema is still there at https://vshampor.github.io/nncf/schema)

The page contains only the description of entities that were decorated with `@api()` and of the containing Python modules in the import path. The module descriptions are currently licenses, because we actually currently use the module docstrings to store licenses (will fix this by moving license content to comments and setting up proper docstrings for API-impacted modules). The class and function descriptions may have invalid formatting in HTML and missing descriptions - all this is fixable by adding or reformatting corresponding docstrings which will be done in a follow-up PR.

For demo of auto PR labeling, see:
https://github.com/vshampor/nncf/pull/6 - modification of an API function signature, PR was marked as API automatially
https://github.com/vshampor/nncf/pull/7 - addition of an API entity when compared to the develop state, PR was marked as API automatically
https://github.com/vshampor/nncf/pull/8 - a change inside an API function that does not explicitly impact API as seen by the external user, PR was not marked as API automatically - it is up to reviewers to determine whether such PR changes API behaviour or not.

NB: automatic PR labeling basically works by comparing API reference HTML content between develop and the state in a given PR.

andrey-churkin:
I think we should also include
- nncf.parameters.TargetDevice
- nncf.parameters.ModelType
- nncf.scopes.IgnoredScope

vshampor:
> I think we should also include
> 
> * nncf.parameters.TargetDevice
> * nncf.parameters.ModelType
> * nncf.scopes.IgnoredScope
> * nncf.data.dataset.Dataset

Done

docs/api/source/_autoapi_templates/python/attribute.rst
```diff
@@ -0,0 +1 @@
+{% extends "python/data.rst" %}
```
vshampor, line 1: These and other .rst files control the layout of the documentation and are copy-and-pasted from defaults, for future adjustments if necessary.

docs/api/source/conf.py
```diff
@@ -0,0 +1,114 @@
+import importlib
```
vshampor, line 1: This file controls the documentation build process that is executed during `sphinx-build`.

docs/api/source/conf.py
```diff
@@ -0,0 +1,114 @@
...
+    print(f"Skipped: {skipped_str}\n")
+
+    print("API entities:")
+    for api_fqn in api_fqns:
+        print(api_fqn)
+    return api_fqns
+
+
+with mock(['torch', 'torchvision', 'onnx', 'onnxruntime', 'openvino', 'tensorflow', 'tensorflow_addons']):
+    api_fqns = collect_api_entities()
```
vshampor, line 80: This code allows to collect API entities without having to install all backends at once.

AlexKoff88:
It looks great really, thanks @vshampor . I have three major comments:
- Can we remove the license from every page? It reduces the useful screen space and you need to scroll every time. 
- It would be great to have examples of configs along with descriptions of the features. For example, if we look at the KD description [here](https://vshampor.github.io/nncf/autoapi/nncf/torch/knowledge_distillation/algo/index.html), having a snippet of the config that enables this increases the usefulness of this page a lot, IMO.
- The pages do not reflect the fact that some of the APIs are exposed at the top level of the "nncf" namespace. For example, if to look at the `quantize` API [here](https://vshampor.github.io/nncf/autoapi/nncf/quantization/quantize/index.html), it is mentioned that it is a part of the "nncf.quantization.quantize" namespace but it can be imported directly from the "nncf". One way to work around it is to provide more documentation.

I think some follow up issues and PRs should be merged to address the last two issues.

vshampor:
@AlexKoff88 
> * Can we remove the license from every page? It reduces the useful screen space and you need to scroll every time.

Sure, will do this in a follow-up PR along with docstring formatting adjustments.

> * It would be great to have examples of configs along with descriptions of the features. For example, if we look at the KD description [here](https://vshampor.github.io/nncf/autoapi/nncf/torch/knowledge_distillation/algo/index.html), having a snippet of the config that enables this increases the usefulness of this page a lot, IMO.

Basically everything that can be written in a docstring will be rendered, so there should be no problem to add config snippets and even code snippets via [doctest](https://thomas-cokelaer.info/tutorials/sphinx/doctest.html) later.

> * The pages do not reflect the fact that some of the APIs are exposed at the top level of the "nncf" namespace. For example, if to look at the `quantize` API [here](https://vshampor.github.io/nncf/autoapi/nncf/quantization/quantize/index.html), it is mentioned that it is a part of the "nncf.quantization.quantize" namespace but it can be imported directly from the "nncf". One way to work around it is to provide more documentation.

I'll see if I can add another wrapper function to do this via code, maybe in this PR even.

nncf/quantization/quantize.py
```diff
@@ -26,6 +27,7 @@
 from nncf.parameters import TargetDevice
 
 
+@api(canonical_alias="nncf.quantize")
```
vshampor, line 30: @AlexKoff88 check this out - in this way it can be specified that the canonical import name should be "nncf.quantize", and the docs will be generated with the canonical import name in mind. Note that this also requires that the symbol itself is imported in corresponding `__init__.py` file (in this case, `nncf/__init__.py`), otherwise the doc build will fail and the doc build check will be red.

nncf/torch/layers.py
```diff
@@ -385,6 +386,7 @@ def from_module(module):
 NNCF_WRAPPED_USER_MODULES_DICT = {}
 
 
+@api(canonical_alias="nncf.torch.load_state")
 def register_module(*quantizable_field_names: str, ignored_algorithms: list = None):
```
KodiaqQ, line 390: Nevertheless is it `load_state` or `register_module`?

KodiaqQ (commented):
General: what about PTQ entities like `PostTrainingQuantization`, `MinMaxQuantization`, and other algorithms?
cc @alexsu52

docs/api/source/conf.py
```diff
@@ -0,0 +1,136 @@
...
+module_fqns = set()
+
+for fqn in api_fqns:
+    path_elements = fqn.split('.')
+    for i in range(1, len(path_elements)):
+        intermediate_module_path = '.'.join(path_elements[:i])
+        module_fqns.add(intermediate_module_path)
+
+
+def skip_non_api(app, what, name, obj, skip, options):
```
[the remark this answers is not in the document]
vshampor: The description is literally next to this line. The callback has the signature following the original function documented at https://sphinx-autoapi.readthedocs.io/en/latest/reference/config.html#event-autoapi-skip-member. It has no type hints, so I won't add any as well.

nncf/onnx/engine.py
```diff
@@ -18,7 +18,6 @@
 
 from nncf.common.engine import Engine
 
-
```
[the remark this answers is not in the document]
vshampor: Guess so.

nncf/torch/layers.py
vshampor: Fixed, and added a check in `conf.py` to guard against this.

KodiaqQ (approved):

nncf/quantization/quantize.py
AlexKoff88: I cannot find quantize() in the latest version of the API build, only quantize_with_accuracy_control().

vshampor: Fixed - make sure to reload the page with Ctrl-Shift-R so that you don't get a cached version.

One of the reasons was that had a function `quantize` exposed in `__init__.py` in the same subpackage that also contained the module called `quantize.py`. This meant that the `quantize.py` file was effectively impossible to import from, apart from the exposed `quantize` and `quantize_with_accuracy_control` functions, and this broke down the doc search mechanism, and is anyway considered to be bad practice (since the imported name shadows the subpackage name). Had to rename the `quantize.py` to `quantize_fns.py`, if you have suggestions for a better name, please post them.

AlexKoff88 (approved):

alexsu52:
> > General: what about PTQ entities like `PostTrainingQuantization`, `MinMaxQuantization`, and other algorithms? cc @alexsu52
> 
> If you are sure about API-ness of certain PTQ entities, tell me and I'll mark these as such. If not, feel free to open a PR marking the right entities as API after this is merged.

I think the coverage of `quantize` and `quantize_with_accuracy_control` is enough.

nncf/quantization/quantize.py
alexsu52: If only rename `quantize.py` -> `quantize_model.py`

alexsu52 (commented):
Great staff! It looks like some changes need to be done to the `quantize` and `quantize_with_accuracy_control` functions to make the documentation look good. I mean changing type hints and making them callable. Could you do it in this PR?

nncf/common/accuracy_aware_training/training_loop.py
```diff
@@ -168,6 +169,7 @@ def _accuracy_criterion_satisfied(self):
         return accuracy_budget >= 0 and self.runner.is_model_fully_compressed(self.compression_controller)
 
 
+@api(canonical_alias="nncf.tensorflow.EarlyExitCompressionTrainingLoop")
```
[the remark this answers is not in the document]
vshampor: Somehow it is only imported in `nncf/tensorflow/__init__.py`. Removed the canonical alias for the time being.

nncf/tensorflow/__init__.py
```diff
@@ -18,14 +18,20 @@
...
 from pkg_resources import parse_version
 
-tensorflow_version = parse_version(tensorflow.__version__).base_version
+try:
+    _tf_version = tensorflow.__version__
+    tensorflow_version = parse_version(_tf_version).base_version
+except:
+    nncf_logger.debug("Could not parse tensorflow version")
+    _tf_version = '0.0.0'
+    tensorflow_version = parse_version(_tf_version).base_version
```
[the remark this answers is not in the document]
vshampor: `WARNING:nncf:NNCF provides best results with tensorflow==2.11.*, while current tensorflow version is 0.0.0. If you encounter issues, consider switching to tensorflow==2.11.*`

We only really need this for when the tensorflow module is mocked externally by the documentation collector and therefore has no version set. User-side, this could only get triggered if tensorflow package is not installed, and in this case this file would produce an exception way above at line 18, so this code won't be triggered.

nncf/tensorflow/helpers/model_creation.py
```diff
@@ -58,7 +59,7 @@ def create_compression_algorithm_builder(config: NNCFConfig,
 
     return TFCompositeCompressionAlgorithmBuilder(config, should_init)
 
-
```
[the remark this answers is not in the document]
vshampor: Done

nncf/torch/__init__.py
```diff
@@ -17,7 +17,15 @@
...
 from pkg_resources import parse_version
-torch_version = parse_version(torch.__version__).base_version
+
+try:
+    _torch_version = torch.__version__
+    torch_version = parse_version(_torch_version).base_version
+except:
+    nncf_logger.debug("Could not parse torch version")
+    _torch_version = '0.0.0'
+    torch_version = parse_version(_torch_version).base_version
```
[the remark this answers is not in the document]
vshampor: The behaviour will be similar

nncf/torch/checkpoint_loading.py
```diff
@@ -16,10 +16,11 @@
 
 import torch
 
+from nncf.common.api_marker import api
 from nncf.common.logging import nncf_logger
 from nncf.common.deprecation import warning_deprecated
 
-
```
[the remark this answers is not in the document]
vshampor: Done

alexsu52 (approved):
please, rebase PR.

nncf/tensorflow/__init__.py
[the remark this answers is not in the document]
alexsu52: Honestly, version 0.0.0 in the warning message confuses me. I would prefer more relevant message like this "Could not parse tensorflow version, please, reinstall tensorflow"

PR #1704, opened 2023-04-13
[3/N] Added automatic structured pruning for MobileBERT

ljaljushkin:
### Changes

Continue of the #1657 and #1680 
Added better handling of producers and consumers. As result, MobileBERT is supported and Microsoft implementation of Swin finds all groups correctly.
Simplified group and block.

image

### Reason for changes
further generalization of structured pruning in JPQD

### Related tickets
96282

### Tests
test_groups


nncf/experimental/common/pruning/propagation_data.py
```diff
@@ -11,24 +11,54 @@
...
+from dataclasses import dataclass
 import json
-from typing import Dict
+from typing import Dict, Optional, Set
 from typing import List
 
+@dataclass
+class ProducerInfo:
+    """
+    Defines a node that produce pruning. Operation that corresponds to the node has built-in weights.
```
daniil-lyakhov, line 22:
```suggestion
    Defines a node that produce pruning masks. Operation that corresponds to the node has built-in weights.
```

nncf/experimental/common/pruning/propagation_data.py
```diff
@@ -11,24 +11,54 @@
...
+    def __hash__(self) -> int:
+        return hash((self.node_id, self.pruning_dimension))
+
+    def __str__(self) -> str:
+        return str(self.node_id)
+
+@dataclass
+class ConsumerInfo:
+    """
+    Defines the node that absorbs (consume) pruning. There are 2 types of consumer. First, have built-in weights,
```
daniil-lyakhov, line 41:
```suggestion
    Defines the node that absorbs (consume) pruning masks. There are 2 types of consumer. First, have built-in weights,
```

nncf/experimental/common/pruning/propagation_data.py
```diff
@@ -11,24 +11,54 @@
...
+    therefore can consume and produce pruning along the input channels. Consumers of the second type don't have built-in
+    weights, instead weights are coming as a second input. In that case, operation can only consume pruning and
+    pruning dimension is None.
+
+    :param node_id: id of the node that consume pruning.
+    :param pruning_dimension: axis number from 0 to N-1 in weights along which the pruning structure is defined.
+        N is total number of dimensions. Equals None for the second type of consumers.
```
daniil-lyakhov, line 48:
```suggestion
    Defines the node that absorbs (consume) pruning masks. There are 2 types of consumer. First, have built-in weights,
    therefore can consume and produce pruning along the input channels. Consumers of the second type don't have built-in
    weights, instead weights are coming as a second input. In that case
    pruning dimension is None.

    :param node_id: id of the node that consume pruning.
    :param pruning_dimension: axis number from 0 to N-1 in weights along which the pruning structure is defined.
        N is total number of dimensions. Equals None for the second type of consumers.
```
Second type of consumers could be possibly pruned, maybe in future we will support it! I suggest to drop the line `In that case, operation can only consume pruning`

nncf/experimental/common/pruning/propagation_data.py
```diff
@@ -150,51 +177,51 @@ def join_groups(*args: 'PropagationGroup') -> 'PropagationGroup':
...
-                if block not in blocks:
-                    blocks.append(block)
-        for block in blocks:
-            retval.add_block(block)
-        return retval
-
+            group.add_child(new_group)
+            new_group.add_producers(group.get_producers())
+            new_group.add_consumers(group.get_consumers())
+        return new_group
```
daniil-lyakhov, line 189: Looks far cleaner than before 👍

nncf/experimental/common/pruning/propagation_data.py
ljaljushkin: Originally, I wanted to say in general about removing elements, without implementation details with masks. my version looks better for me.

nncf/experimental/common/pruning/operations.py
```diff
@@ -173,29 +175,16 @@ def _both_dim_blocks_exist(left_idx, right_idx):
                 right = right_dim_groups[input_shape_len - 2]
                 assert len(left) == 1 and len(right) == 1, "multiple groups are not supported"
                 group = PropagationGroup.join_groups(left[0], right[0])
-                cls.add_consumer_block(group, node)
+                pruning_dimension = 1
+                if node.layer_attributes is None:
+                    pruning_dimension = None
+                consumer = ConsumerInfo(node_id=node.node_id, pruning_dimension=pruning_dimension)
```
daniil-lyakhov, line 181: I think current node is not a consumer in a sense of a node that have a weight that should be adjusted (for example, consumer conv input channels should be adjusted to make pruning possible), so pruning dimension always should be None?

nncf/experimental/common/pruning/operations.py
```diff
@@ -173,29 +175,16 @@ def _both_dim_blocks_exist(left_idx, right_idx):
...
+                group.add_consumers({consumer})
+        else:
+            output_mask = node.data.get('output_mask', None)
```
daniil-lyakhov, line 184: IMHO this if is too big, I suggest to move condition bodies (code inside if's) to a separate private method

nncf/experimental/common/pruning/operations.py
ljaljushkin: you are right, corrected

nncf/experimental/common/pruning/operations.py
ljaljushkin: refactored

nncf/experimental/common/pruning/propagation_data.py
ljaljushkin: Discussed in person, corrected

nncf/experimental/common/pruning/propagation_data.py
ljaljushkin: Done

daniil-lyakhov (approved):
👍

alexsu52 (approved):
LGTM

PR #1710, opened 2023-04-17
[PTQ][OV] Update model with optimization info

KodiaqQ:
### Changes

- Added optimization info into the model meta section (OpenVINO)

### Related tickets

- 94920

### Tests


nncf/experimental/openvino_native/quantization/quantize.py
```diff
@@ -34,6 +34,24 @@
...
+    :param path: Optional list of the paths.
+    """
+    for key, value in parameters.items():
+        # Special condition for composed fields like IgnoredScope
+        if isinstance(value, IgnoredScope):
+            dump_parameters(model, value.__dict__, [key])
+            continue
+        value = value if value else None
+        rt_path = ['optimization'] + path + [key]
+        model.set_rt_info(value, rt_path)
```
AlexKoff88, line 52: this `.set_rt_info` is quite limited in terms of the ability to represent and save objects of various types. There can be problems with ignored scope (exception). You should test it.

nncf/experimental/openvino_native/quantization/quantize.py
```diff
@@ -34,6 +34,24 @@
...
+    :param parameters: Incoming dictionary with parameters to save.
+    :param path: Optional list of the paths.
+    """
+    for key, value in parameters.items():
+        # Special condition for composed fields like IgnoredScope
+        if isinstance(value, IgnoredScope):
+            dump_parameters(model, value.__dict__, [key])
+            continue
+        value = value if value else None
+        rt_path = ['optimization'] + path + [key]
```
AlexKoff88, line 51: I think "optimization" is quite an ambiguous name here. My take is that we should call it "nncf" and add another level internally and call it like "ptq" or "quantization" or whatever because we can come up with other post-training APIs at some point, e.g. "weight_compression" and we should also mentioned this in the rt info.

ksilligan:
run openvino pre-commit tests

l-bat (approved):

nncf/experimental/openvino_native/quantization/quantize.py
KodiaqQ: Added to string conversion & test for this.

nncf/experimental/openvino_native/quantization/quantize.py
```diff
@@ -34,6 +34,24 @@
...
+    :param path: Optional list of the paths.
+    """
+    path = path if path else []
+    for key, value in parameters.items():
+        # Special condition for composed fields like IgnoredScope
+        if isinstance(value, IgnoredScope):
+            dump_parameters(model, value.__dict__, [key])
+            continue
+        rt_path = ['nncf', 'quantization'] + path + [key]
+        model.set_rt_info(str(value), rt_path)
```
alexsu52, line 52: Please add try/catch block for these code.

alexsu52 (changes requested):

nncf/experimental/openvino_native/quantization/quantize.py
KodiaqQ: Why do we need to add try/catch here? Since the value is a string, the only thing that can be wrong is adding to the incorrect field.

AlexKoff88 (approved):

nncf/experimental/openvino_native/quantization/quantize.py
KodiaqQ: Added.

KodiaqQ:
.xml output:
image

alexsu52 (approved):

PR #1717, opened 2023-04-19
[PTQ][OV] Fix rank zero output tensor for inplace statistics

daniil-lyakhov:
### Changes

Nodes with zero output ranks raised runtime error because empty `np.array` has `np.float` type and openvino reduction operations require `np.int64`. Now case `rank==0` works as expected.

### Reason for changes

To fix runtime error during statistic collection for models with nodes with target quantize nodes that have `rank==0` output.

### Related tickets

108784

### Tests

New test in openvino `test_model_transformer.py` that checks corner case with a target point node that has `rank==0` output.

### DIFF
--- a/nncf/experimental/openvino_native/graph/node_utils.py
+++ b/nncf/experimental/openvino_native/graph/node_utils.py
@@ -138,7 +138,7 @@ def get_reduce_node_name(output_name: str, node_type: str, port_id: int) -> str:
     return f'{output_name}_{node_type}.{port_id}'
 
 
-def get_inplace_reduce_op(op: Type[ov.Node], node_type: str, reduction_axes: Tuple[int, ...],
+def get_inplace_reduce_op(op: Type[ov.Node], node_type: str, reduction_axes: Optional[Tuple[int, ...]],
                           use_abs: bool) -> InplaceInsertionFnType:
     """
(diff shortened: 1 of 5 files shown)


nncf/experimental/openvino_native/graph/node_utils.py
```diff
@@ -155,7 +155,7 @@ def get_reduce_op(node: ov.Node, output_port_id: int) -> ov.Node:
         name_output_port_id = output_port_id
         if reduction_axes_ is None:
             partial_shape = get_partial_shape_safe(node, output_port_id)
-            reduction_axes_ = np.array(range(partial_shape.rank.get_length()))
+            reduction_axes_ = np.arange(partial_shape.rank.get_length()).astype(np.int64)
```
KodiaqQ, line 158: Are you sure that every `op` supports `np.int64` type?

daniil-lyakhov: Yes, it's covered by tests and described [here](https://docs.openvino.ai/latest/openvino_docs_ops_reduction_ReduceMean_1.html) 
```
T_IND : any supported integer type.
```

KodiaqQ (approved):

PR #1718, opened 2023-04-19
[PTQ][Common] Fix for preventing infinity loop in graph traversal

andreyanufr:
for preventing infinity loops.

### Changes

Return empty list in case of empty quantization_points

### Reason for changes

It is possible to receive infinity loop if quantization_points is empty.


daniil-lyakhov:
Could you please add a small test to `tests/post_training/test_ptq_params.py`?

andreyanufr:
> Could you please add a small test to `tests/post_training/test_ptq_params.py`?

I think I can close this PR by the reason of similar PR with better fix: https://github.com/openvinotoolkit/nncf/pull/1715/files#diff-439982cdce21bd7deb841ee26d4601401f3ca51bd4af3fbe13b18a4d7407da2bR488

PR #1721, opened 2023-04-20
[DOCS] Readme Usage and Samples sections improvement

nikita-savelyevv:
### Changes

- Usage section
  - Switched POT and QAT, placing PTQ first
  - Rephrased descriptions for PTQ and QAT making them more compact
  - Hide QAT code blocks by default; open a code block for OV PTQ by default
- Samples section
  - Put in the following order: (1) PTQ notebook, QAT notebooks (2) PTQ sample scripts (3) QAT sample scripts
  - Rephrased slightly for compactness

### Reason for changes

PTQ is easier to grasp, which makes it more user friendly and should be placed first.
General NNCF Readme improvement.


README.md
```diff
@@ -57,102 +57,19 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
...
-
-To start the algorithm, provide the following entities:
-* Original model.
-* Validation part of the dataset.
-* [Data transformation function](./docs/compression_algorithms/post_training/Quantization.md#data-transformation-function) transforming data items from the original dataset to the model input data. 
-
-
-The basic workflow steps:
-1) Create the [data transformation function](./docs/compression_algorithms/post_training/Quantization.md#data-transformation-function).
+To run the algorithm you only need your model and the validation part of the dataset without labels. The basic workflow steps:
```
nikita-savelyevv, line 63: I think it's important to explicitly state that POT may be used without labels. This broadens its usage space. Before, it was not so obvious because this knowledge is hidden in the code snippet.

README.md
```diff
@@ -57,102 +57,19 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
...
-NNCF provides [samples](#post-training-quantization-samples), which demonstrate Post-Training Quantization usage for PyTorch, TensorFlow, ONNX, OpenVINO.
-
-To start the algorithm, provide the following entities:
-* Original model.
-* Validation part of the dataset.
-* [Data transformation function](./docs/compression_algorithms/post_training/Quantization.md#data-transformation-function) transforming data items from the original dataset to the model input data. 
-
-
-The basic workflow steps:
-1) Create the [data transformation function](./docs/compression_algorithms/post_training/Quantization.md#data-transformation-function).
```
nikita-savelyevv, line 147: A separate bullet point for such a minor detail seems like a waste. Besides it's also mentioned one more time below.

README.md
```diff
@@ -57,102 +57,19 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
...
+To run the algorithm you only need your model and the validation part of the dataset without labels. The basic workflow steps:
+1) Load a model and a data source.
 2) Create an instance of `nncf.Dataset` class by passing two parameters:
-* `data_source` - Iterable python object that contains data items for model calibration.
-* `transform_fn` - Data transformation function from the Step 1.
-3) Run the quantization pipeline.
+   * `data_source` — Iterable object containing data items for model calibration.
+   * `transform_fn` — [Data transformation function](./docs/compression_algorithms/post_training/Quantization.md#data-transformation-function), if your data source returns both images and labels.
+3) Run the quantization pipeline. It will return the quantized model for faster inference.
```
nikita-savelyevv, line 68: **faster** :zap:

README.md
```diff
@@ -57,102 +57,19 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
...
 
 Below are the usage examples for every backend.
 
-<details><summary><b>PyTorch</b></summary>
+<details open><summary><b>PyTorch</b></summary>
```
nikita-savelyevv, line 72: Open by default only one snippet (the most important one IMO) as not to flood the page with excess code

README.md
nikita-savelyevv, line 147: A separate bullet point for such a minor detail seems like a waste. Besides it's also mentioned one more time below.

README.md
nikita-savelyevv, line 63: I think it's important to explicitly state that POT may be used without labels. This broadens its usage space. Before, it was not so obvious because this knowledge is hidden in the code snippet.

README.md
nikita-savelyevv, line 68: **faster inference** ⚡

README.md
nikita-savelyevv, line 72: Only one code block is open by default now (the most important one IMO) as not to flood the page with excess code

README.md
```diff
@@ -260,36 +177,123 @@ quantized_model = nncf.quantize(model, calibration_dataset)
 
 </details>
 
+
+[//]: # (NNCF provides full  [samples]&#40;#post-training-quantization-samples&#41;, which demonstrate Post-Training Quantization usage for PyTorch, TensorFlow, ONNX, OpenVINO.)
+
+### Training-time Compression
+
+The basic workflow steps:
```
nikita-savelyevv, line 185: Split large text block describing QAT into bullets, hence more readable

README.md
```diff
@@ -260,36 +177,123 @@ quantized_model = nncf.quantize(model, calibration_dataset)
...
+1) Create an NNCF config object from (a) JSON configuration file and (b) dataloader for the compression initialization.
+2) Pass your model together with the config to the `create_compressed_model` function which returns two entities:
+   1) Compression controller — an object allowing you to control the compression, for example during training.
+   2) Compressed model which may further be used as a regular PyTorch model for fine-tuning and inference.
+
+Expand the blocks below for PyTorch and Tensorflow usage examples: 
+
+<details><summary><b>PyTorch</b></summary>
```
nikita-savelyevv, line 193: Added expanding code blocks for QAT too. They are rather large, hidden by default

README.md
```diff
@@ -57,102 +57,19 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
...
-
-# ... the rest of the usual TensorFlow-powered training pipeline
-
-# Export to Frozen Graph, TensorFlow SavedModel or .h5  when done fine-tuning 
-compression_ctrl.export_model("compressed_model.pb", save_format='frozen_graph')
-```
-
-For a more detailed description of NNCF usage in your training code, see [this tutorial](docs/Usage.md). 
-For in-depth examples of NNCF integration, browse the [sample scripts](#compression-aware-training-samples) code, or the [example patches](#third-party-repository-integration) to third-party repositories.
-For FAQ, visit this [link](./docs/FAQ.md).
```
nikita-savelyevv, line 134: Removed these two For's. Sample scripts and third part integration sections are just below. FAQ link is already present in the end of the readme.

vshampor (approved):

nikita-savelyevv:
run onnx pre-commit tests

nikita-savelyevv:
run openvino pre-commit tests

README.md
alexsu52, line 72: I would recommend putting the OpenVINO backend in first place.

README.md
alexsu52, line 63: What do you think about reusing the introduction from the https://docs.openvino.ai/latest/basic_qauntization_flow.html#doxid-basic-qauntization-flow

README.md
```diff
@@ -260,36 +177,121 @@ quantized_model = nncf.quantize(model, calibration_dataset)
...
+[//]: # (NNCF provides full  [samples]&#40;#post-training-quantization-samples&#41;, which demonstrate Post-Training Quantization usage for PyTorch, TensorFlow, ONNX, OpenVINO.)
+
+### Training-time Compression
+
+The basic workflow steps:
+1) Create an NNCF config object from (a) JSON configuration file and (b) dataloader for the compression initialization.
+2) Pass your model together with the config to the `create_compressed_model` function which returns two entities:
+   1) Compression controller — an object allowing you to control the compression, for example during training.
+   2) Compressed model which may further be used as a regular PyTorch model for fine-tuning and inference.
+
```
alexsu52, line 190: It looks like you forgot the following steps:
- model fine-tuning
- model export

MaximProshin (approved):
lgtm

README.md
nikita-savelyevv: I suppose PyTorch models are the most frequent target for NNCF PTQ among NNCF users. From that point of view PyTorch should be first.

Besides, this is way we'll attract users who mainly use PyTorch to try out OpenVINO since quantized models must be converted to OV IR to experience faster inference.

README.md
nikita-savelyevv: `The basic quantization flow is the simplest way to apply 8-bit quantization to the model.` -- I agree, added in a slightly modified way

`- It is available for models in the following frameworks: PyTorch, TensorFlow 2.x, ONNX, and OpenVINO.` -- this is obvious from the code blocks lower in the README

`- Set up an environment and install dependencies.` -- too obvious IMO

`- Prepare the calibration dataset that is used to estimate quantization parameters of the activations within the model.` -- (1) too abstract compared to the one that is present (2) I don't think we need to explain to a user what it is used for at this stage

`- Call the quantization API to apply 8-bit quantization to the model.` -- not meaningfully different from the current version

README.md
alexsu52: IMHO: I don't agree with this statement, as far as I know we don't have statistics on PTQ API usage per backends yet. Anyway, OpenVINO PTQ is the base way to obtain a quantized model for OpenVINO because by using OpenVINO representation the user is guaranteed of getting an accurate model with best performance in OpenVINO than using the framework model representation. I think, we can firstly of all recommend what will give the user the best model for OpenVINO.

README.md
alexsu52: Could you reformulate this sentence `To run the algorithm you only need your model and the validation part of the dataset without labels.` because the user doesn't need `the validation part of the dataset without labels.` but needs calibration dataset. Usually, the calibration dataset is a small, no label, representative dataset (~100-500 samples) that is used to estimate the range, i.e. (min, max) of all floating point activation tensors in the model, to initialize the quantization
parameters. The easiest way to define a calibration dataset is to use a training or validation dataset and a transformation function to remove labels from the data item and prepare model input data.

README.md
nikita-savelyevv: Thanks, done.

README.md
nikita-savelyevv: I think it's safe to assume that the distribution of backends people would want to try NNCF with is strongly correlated with the popularity of the backends themselves. For OV and PT the number of downloads last month was 132K vs 10M. We should focus on a larger target audience.

I agree with your point, but as I see it, the purpose of the readme is to a greater degree to retain the visitors of the page and make them try out NNCF. For this, it is important to make the examples easy-to-follow and relatable. For most of the people PT is much more relatable then OV and they have their models in PT format and not OV format. I'm afraid a regular visitor wouldn't bother with the PT->OV conversion. 

I believe there's no information in the readme that the OV is the preferred backend. We should add such note so that the determined users would transition to OV.

README.md
```diff
@@ -260,36 +177,121 @@ quantized_model = nncf.quantize(model, calibration_dataset)
...
+3) Finetune the compressed model as a regular PyTorch/Tensorflow module.
+4) Export the finetuned model using the compression controller.
+
+Expand the blocks below for PyTorch and Tensorflow usage examples: 
+
+<details><summary><b>PyTorch</b></summary>
+
+```python
+import torch
+import nncf.torch  # Important - must be imported before any other external package that depends on torch
```
AlexanderDokuchaev, line 197: As found in PR with auto formatting. Isort can automatically move move nncf below other imports.  It can confused users who used isort or .
Suggest to add to readme `# isort: off
```python
# isort: off
# Important - must be imported before any other external package that depends on torch
import nncf.torch  
from nncf import NNCFConfig
from nncf.torch import create_compressed_model, register_default_init_args
# isort: on
```

README.md
AlexanderDokuchaev: What you think about move examples of using to QuickStart.md

In main readme leave only a short example for OV backend, as more interesting for users.

README.md
nikita-savelyevv: Thanks for the detailed information. I reformulated the dataset description, but omitting details. Please take a look.

Regarding the detailed description of calibration dataset, we could add it to a separate more detailed PTQ Usage description documentation file, when we decide to create such in the future. The current `Usage.md` is for NNCF QAT only.

README.md
nikita-savelyevv: I haven't found a `QuickStart.md` you are referring to. Did you imply to create it? Honestly, I think it's ok to keep all code examples as they're hidden anyway.

But it's worth considering moving QAT usage section from readme to an improved `Usage.md` because it's kinda heavy for readme in my opinion. But that's in the future, I'd like to move in small incremental steps.

AlexanderDokuchaev (approved):

README.md
```diff
@@ -260,36 +177,121 @@ quantized_model = nncf.quantize(model, calibration_dataset)
...
 
+
+[//]: # (NNCF provides full  [samples]&#40;#post-training-quantization-samples&#41;, which demonstrate Post-Training Quantization usage for PyTorch, TensorFlow, ONNX, OpenVINO.)
+
+### Training-Time Compression
+
+Training-Time Compression pipeline allows to fine-tune the quantized model in order to improve on the possible accuracy drop introduced by quantization. The basic workflow steps are:
+1) Create an NNCF config object from (a) JSON configuration file and (b) initialization dataloader.
+2) Pass your model together with the config to the `create_compressed_model()` function which returns (a) compression controller and (b) compressed model.
+3) Finetune the compressed model as a regular PyTorch/Tensorflow module.
```
alexsu52, line 188:
```suggestion
3) Finetune the compressed model as a regular PyTorch/Tensorflow model.
```

README.md
alexsu52: If using your logic TensorFlow should be first. Why is it not?

nikita-savelyevv: While PT and TF downloads numbers are of the same order, NNCF experience with PT is better than with TF I believe.

README.md
nikita-savelyevv: "It will return the quantized model for **faster inference**."

nikita-savelyevv: Ah, you probably suggested to add that the faster inference is when converted OV IR. Added :+1:

I've noticed that there are almost no references in the readme that compressed models must be converted to OV IR to experience runtime boost. I have some plans to make it obvious for users in the future.

README.md
nikita-savelyevv: Missed it. Done.

README.md
```diff
@@ -60,29 +60,25 @@ _Preview means that this is a work in progress and NNCF does not guarantee the f
 
 ### Post-Training Quantization
 
-The NNCF PTQ is the simplest way to apply 8-bit quantization. To run the algorithm you only need your model and a small (~300 samples) calibration no-label dataset. The basic workflow steps:
-1) Load a model and a data source.
```
nikita-savelyevv: Decided to remove text description of the steps, because the code is self-explanatory

README.md
nikita-savelyevv: Decided that it's not important

README.md
nikita-savelyevv: Placed OV first because PyTorch PTQ functionality is currently lagging behind the OV functionality

nikita-savelyevv:
@MaximProshin @vshampor @AlexanderDokuchaev please re-review, as some major changes are made: removed text description of the code samples steps

alexsu52 (approved):

PR #1722, opened 2023-04-21
deduplicate python/pip executable function

ksilligan:
### Changes

Shared function to return correct python/pip executable depending on platform

### Reason for changes

Fix and beautify code to run precommit tests on Windows and Linux platforms 

### Related tickets

107218

### DIFF
--- a/tests/torch/test_extensions_build.py
+++ b/tests/torch/test_extensions_build.py
@@ -9,6 +9,8 @@ import torch
 from tests.shared.paths import TEST_ROOT
 from tests.shared.command import Command
 from tests.shared.helpers import create_venv_with_nncf
+from tests.shared.helpers import get_python_executable_with_venv
+from tests.shared.helpers import get_pip_executable_with_venv
 
 EXTENSIONS_BUILD_FILENAME = 'extensions_build_checks.py'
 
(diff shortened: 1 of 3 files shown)


tests/torch/test_extensions_build.py
```diff
@@ -34,9 +35,12 @@ def test_force_cuda_build(tmp_path):
             pytest.skip('There is no CUDA on the machine. The test will be skipped')
 
     torch_build_dir = tmp_path / 'extensions'
-    export_env_variables = "export CUDA_VISIBLE_DEVICES='' export TORCH_EXTENSIONS_DIR={}".format(torch_build_dir)
 
-    python_executable_with_venv = ". {0}/bin/activate && {1} && {0}/bin/python".format(venv_path, export_env_variables)
+    export_env_variables = [
+        "CUDA_VISIBLE_DEVICES=''",
+        "TORCH_EXTENSIONS_DIR={}".format(torch_build_dir)
+    ]
```
vshampor, line 42: Now the `tests.shared.command.Command` has an `env` constructor argument that you could pass. It is then passed to `subprocess.Popen` and the env vars are being set automatically in a platform-specific way. I think it is better to use this than to bake env variables into the command line.

vshampor (changes requested):

vshampor (approved):

PR #1734, opened 2023-04-26
Update pandas

p-wysocki:
### Changes

Allow pandas patch releases in order to enable CI in https://github.com/openvinotoolkit/openvino/pull/17117 and https://github.com/intel-innersource/frameworks.ai.openvino.ci.product-configs/pull/2087


alexsu52:
install onnx: 95
install ov: 124
install pt cpu: 121
install pt gpu: 123
install tf: 238

PR #1738, opened 2023-04-26
Align package versions with OpenVINO

p-wysocki:
### Changes

Align package versions with OpenVINO


alexsu52:
install onnx: 97
install ov: 127
install pt cpu: 123
install pt gpu: 125
install tf: 240

alexsu52 (approved):

PR #1747, opened 2023-04-28
Disable quantization for post-processing ops

l-bat:
### Changes

The algorithm for automatically adding post-processing nodes to ignored scope has been redesigned.

Starting from the output nodes all the nodes are added to path, until the quantizable nodes with weights are faced.
If the path with the nodes has the post-processing marker node, all the nodes in this path (except outputs and nodes with weights) will be added into ignored scope.

### Reason for changes

* To fix loading cascade_rcnn_resnet101_fpn_1x_coco, pointrend-resnet50-fpn to OpenVINO runtime

### Related tickets

* 104185

### Tests

* Added test cases to tests/common/quantization/test_ignore_post_processing.py

### DIFF
--- a/nncf/common/quantization/quantizer_propagation/solver.py
+++ b/nncf/common/quantization/quantizer_propagation/solver.py
@@ -244,7 +244,8 @@ class PostprocessingNodeLocator:
     def get_post_processing_node_keys(self) -> Set[str]:
         """
         Finds out the nodes of the QuantizerPropagationStateGraph, which are in post-processing part of the model.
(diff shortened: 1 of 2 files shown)


kshpv (approved):

alexsu52 (commented):
Are ignored nodes printed from preprocessing to output?

l-bat:
> Are ignored nodes printed from preprocessing to output?

No. I didn't add a logger message, because `pointrend-resnet50-fpn` has more than 300 post-processing  ignored nodes.

alexsu52 (approved):

PR #1748, opened 2023-04-28
[Torch] Fix torch quantize import

daniil-lyakhov:
### Changes

Import fixed

### Reason for changes

Runtime error during quantize conformance


vshampor (approved):

vshampor:
pre-commit will be fixed in follow-up PRs

PR #1675, opened 2023-03-30
[OV] Disable FQ propagation through ReLU

l-bat:
### Reason for changes

To align NNCF quantization with POT

### Related tickets

105079

PR #1759, opened 2023-05-02
Added quantization agnostic ops list for all backends

l-bat:
### Changes

* Mark metatype as `NON_QUANTIZABLE` if `QuantizationTrait` is missed 

### Tests

* pytorch sota: 340
* tf sota: 335


l-bat:
run openvino pre-commit tests

nncf/torch/quantization/default_quantization.py
```diff
@@ -62,11 +62,27 @@
         operator_metatypes.PTDeformConv2dMetatype,
         operator_metatypes.PTModuleDeformConv2dMetatype,
         UnknownMetatype,
-        # Ticket: 108478
-        operator_metatypes.PTAbsMetatype,
-        operator_metatypes.PTExpMetatype,
-        operator_metatypes.PTLogMetatype,
-        operator_metatypes.PTSqrtMetatype,
+    ],
+    QuantizationTrait.QUANTIZATION_AGNOSTIC: [
```
vshampor, line 66: Looks like #1675 was merged without my approval. ReLU is a quantization-agnostic operation, in a sense that it doesn't matter if the quantization operation is placed immediately before or immediately after it, and should be marked as such everywhere. 

NNCF is right to propagate FQ through RELU if the RELU operation in question does not belong to any pattern. If you encounter models where this is an issue, analyze the model for a pattern and add the pattern.

vshampor (changes requested):

nncf/onnx/quantization/default_quantization.py
```diff
@@ -38,13 +38,31 @@
         onnx_metatypes.ONNXDequantizeLinearMetatype,
```
alexsu52, line 38: Do you still need QuantizationTrait.NON_QUANTIZABLE here and below?

alexsu52 (commented):
Thanks for contribution!

nncf/torch/quantization/default_quantization.py
vshampor: Please add a comment in all backends near this list stating just why RELU specifically is not considered to be QUANTIZATION_AGNOSTIC, namely:
1) that it's frequently better for the end accuracy to have quantizers set up after the RELU so that the input distribution to the quantizer is non-negative and we can therefore have better quantization resolution while preserving the original dynamic range, and:
2) that the runtime currently does not provide performance benefits from quantizing stand-alone RELU's, and maybe even on the contrary, and add a ticket reference.

nncf/onnx/quantization/default_quantization.py
```diff
@@ -9,10 +9,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from nncf.common.graph.operator_metatypes import UnknownMetatype
 from nncf.common.quantization.quantizer_propagation.structs import QuantizationTrait
 from nncf.onnx.graph.metatypes import onnx_metatypes
 
+# If there are no some metatypes it means that they are considered as QuantizationTrait.NON_QUANTIZABLE
```
vshampor, line 15: `If a metatype is not in this list, then it is considered to be QuantizationTrait.NON_QUANTIZABLE.`

nncf/onnx/quantization/default_quantization.py
```diff
@@ -62,6 +55,11 @@
         onnx_metatypes.ONNXCastLikeMetatype,
         onnx_metatypes.ONNXDropoutMetatype,
         onnx_metatypes.ONNXFlattenMetatype,
+        # ONNXReluMetatype is not considered to be QUANTIZATION_AGNOSTIC, because:
+        # 1. Runtime doesn't provide performance benefits by quantizing the stand-alone RELU's (ticket: 59548)
+        # 2. That it's frequently better for the end accuracy to have quantizers set up after the RELU
```
vshampor, line 60: Remove `That`

vshampor (approved):

alexsu52:
run pylint pre-commit tests

PR #1763, opened 2023-05-02
[PTQ][Tests] Add model reshape to calibrate.py

daniil-lyakhov:
### Changes

`allow_reshape_input` Accuracy checker param now emulated in `calibrate.py`

### Reason for changes

To enable models that require reshape or dynamic shapes before inference on a calibration  dataset

### Related tickets

108977

### Tests
Checked locally on bert-base-cased, fbcnn and human-pose-estimation
TODO

- [x] Check on all models with `allow_reshape_input`  param in ACConfig

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -22,6 +22,7 @@ import numpy as np
 import openvino.runtime as ov
 from openvino.tools.accuracy_checker.evaluators.quantization_model_evaluator import ModelEvaluator
 from openvino.tools.accuracy_checker.evaluators.quantization_model_evaluator import create_model_evaluator
+from openvino.tools.accuracy_checker.launcher.openvino_launcher import OpenVINOLauncher
 from openvino.tools.pot.configs.config import Config
 
 import nncf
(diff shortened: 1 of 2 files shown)


tests/openvino/tools/calibrate.py
```diff
@@ -577,10 +578,27 @@ def get_nncf_algorithms_config(compression_config):
...
+            if "allow_reshape_input" in launcher_config:
+                return launcher_config["allow_reshape_input"]
+    return False
+
+
+def maybe_reshape_model(model, accuracy_checker_config, transform_fn, dataset):
+    if not get_allow_reshape_input(accuracy_checker_config):
+        return model
+
+    inputs = transform_fn(next(iter(dataset)))
```
KodiaqQ, line 593: Can we be sure that reshaping would be needed only once and only based on 1st sample?
I mean, it might be that model need to be reshaped from sample to sample.

tests/openvino/tools/calibrate.py
```diff
@@ -577,10 +578,27 @@ def get_nncf_algorithms_config(compression_config):
...
+    shapes = {name: tensor.shape for name, tensor in inputs.items()}
+    return OpenVINOLauncher.reshape_network(model, shapes)
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

alexsu52 (commented):
Could you check this fix on the validation models?

KodiaqQ:
Why don't we provide this functionality in the internal `Engine`? How will users reshape the model using NNCF API while optimizing the model?

daniil-lyakhov:
> Why don't we provide this functionality in the internal `Engine`? How will users reshape the model using NNCF API while optimizing the model?

Users specifies dataset, and it's not NNCF responsibility to check if model inputs aligned with dataset, its defined by user

daniil-lyakhov:
run pytorch pre-commit tests

tests/openvino/tools/calibrate.py
```diff
@@ -612,16 +689,26 @@ def transform_fn(data_item):
...
     calibration_dataset = nncf.Dataset(model_evaluator.dataset, transform_fn)
+
+    if get_allow_reshape_input(accuracy_checker_config):
+        ov_model = maybe_reshape_model(
+            ov_model,
+            calibration_dataset,
+            quantization_parameters.get("subset_size", 300),
+            model_evaluator.launcher.input_to_tensor_name,
+        )
+        model_evaluator.load_network([{"model": ov_model}])
```
alexsu52, line 700: Is `model_evaluator.load_network([{"model": ov_model}])` mandatory?

daniil-lyakhov: Yes, it's needed to notify AC that model has new shape (static or dynamic). Skipping this leads to wrong dataset reshaping as AC prepares dataset based on loaded model inputs

daniil-lyakhov: Yes, it's needed to notify AC that model has new shape (static or dynamic). Skipping this leads to wrong dataset reshaping as AC prepares dataset based on loaded model inputs

PR #1765, opened 2023-05-02
nn.MHA quantization in case of not fast path

ljaljushkin:
### Changes

Not patch `multi_head_attention_forward` to add internal operations of the `nn.MultiheadAttention` to the NNCF graph.
**This is applicable for not “fast path” only.** 
If `torch._native_multi_head_attention` is called on “fast path”, there's no way split into internal operations. 
image

### Reason for changes

`nn.MultiheadAttention` was represented in NNCF graph with just a single `F.multi_head_attention_forward` op. 
Now it's represented as a graph, and it opens opportunities for compression.

### Related tickets

109727

### Tests


tests/torch/requirements.txt
```diff
@@ -14,4 +14,7 @@ transformers==4.23.1
 # Required for movement_sparsity tests
 datasets==2.6.2
 evaluate==0.3.0
-timm
\ No newline at end of file
+timm
+
+# Use case with quantization of nn.MultiheadAttention
+open_clip_torch==2.18.0
```
AlexKoff88, line 20: hope it will not increase the test time which already takes time as far as I heard. Potentially, you could create a model with one primitive to the text of use something from torchvision which should be more aligned with torch versions because now your test depend on the stability of the OpenCLIP library.

ljaljushkin: It takes the specific version of the library, shouldn't depend much. 
test takes 11 seconds

vshampor (approved):

tests/torch/requirements.txt
ljaljushkin: There are many conditions checked to make a decision about implementation of MHA
image

ljaljushkin: Removed dependency from open_clip

AlexKoff88 (approved):

MaximProshin (approved):
approved for merging to the release branch

PR #1769, opened 2023-05-03
Fix last win tests

vshampor:


vshampor:
TF Win precommit 7 and OV Win precommit 6 are green.

PR #1774, opened 2023-05-05
[TF] Range initialization fix

nikita-savelyevv:
### Changes

Included 0 in quantization ranges computation for TF.

### Reason for changes

MobileNets had accuracy degradation to 0 after quantization initialization. The issue was with the range initialization for outputs of HardSigmoid function which always returned 1 for samples from initializing dataset. This resulted in `input_low=0.95` and `input_range=0.1`, and then TF quantization turns these into `input_low=0`, `input_range=0.1` which doesn't include 1.

The fix forces initializer to include 0 in the range, as it should've done. This in the end results in `input_low=0` and `input_range=1` which allows for proper quantization of HardSigmoid outputs and resolves the accuracy issue.

### Related tickets

56945

### Tests

Added a test for `apply_range_initialization` method.

### DIFF
--- a/nncf/tensorflow/quantization/quantizers.py
+++ b/nncf/tensorflow/quantization/quantizers.py
@@ -492,6 +492,8 @@ class AsymmetricQuantizer(Quantizer):
         return _default_quantize()
 
     def apply_range_initialization(self, weights, min_values, max_values, min_range=0.1, eps=0.01):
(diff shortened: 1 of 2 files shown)


alexsu52 (approved):
Great staff!

l-bat (approved):

PR #1775, opened 2023-05-05
API documentation cleanup

vshampor:
### Changes
Missing API entities were marked as such to make the documentation consistent; docstrings were formatted for correct rendering to HTML.

### Reason for changes
Better user-facing documentation.

### Related tickets
N/A

### Tests
N/A


alexsu52 (commented):
Great staff! Could you please provide a link where I can view the generated documentation.

vshampor:
I pushed the state here to my fork's GH pages, so you can see `https://vshampor.github.io/nncf` if you don't want to download the build artifacts from the API changes check action.

alexsu52 (commented):
Some general comments from my side:
- Are you going to use some sphinix's theme or color scheme aligned with OpenVINO in the API documentation.
- Looks like the header "OpenVINO Neural Network Compression Framework" should be added.
- The titile doesn't look very pretty. May be rename to OpenVINO NNCF v2.4.0
image
- Surprisingly, classes and functions are in subpackages and we don't see them in the list.
image
 
https://vshampor.github.io/nncf/autoapi/nncf/index.html

vshampor:
@alexsu52 I switched to the Furo theme which seems to do what you expect from the contents bar without me spending effort on learning Jinja templating. I also removed the version altogether because we will probably use another mechanism for versioning docs. As for color alignment etc., I will let the OV doc team open PRs to that accord.

alexsu52 (approved):
It looks better. Thanks.

AlexKoff88:
I am ok with the changes. But the resource alone is hard to percept alone: https://vshampor.github.io/nncf/. However, it can be used already for a reference from other parts of NNCF documentation, e.g. OpenVINO product docs or NNCF developer docs (READMEs).

AlexKoff88 (approved):

vshampor:
The TF precommit failure is caused by CI issues.

vshampor:
The TF precommit failure is caused by CI issues.

PR #1777, opened 2023-05-05
[PTQ][OV] Insert null biases pass for BC

KodiaqQ:
### Changes

- Added `insert_null_biases` pass into the BiasCorrection algorithm.

### Reason for changes

- This pass allows us to correct the bias in the layers that could have it.
### Related tickets

104117

### Tests

conformance run - 123

ONNX ticket - 110696
PT ticket - 110697

### DIFF
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -448,7 +448,7 @@ class OVModelTransformer(ModelTransformer):
         for transformation in transformations:
             node_name = transformation.target_point.target_node_name
             node = name_to_node_mapping[node_name]
-            # Since layers that may have biases mostly are Convolution or MatMul variations, 
+            # Since layers that may have biases mostly are Convolution or MatMul variations,
             # we may use only 0 output port.
             node_shape = node.output(0).partial_shape.get_max_shape()
             node_output_port = node.output(transformation.target_point.port_id)
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -448,7 +448,8 @@ class OVModelTransformer(ModelTransformer):
         for transformation in transformations:
             node_name = transformation.target_point.target_node_name
             node = name_to_node_mapping[node_name]
-            # Since layers that may have biases mostly are Convolution or MatMul variations, we may use only 0 output port.
+            # Since layers that may have biases mostly are Convolution or MatMul variations, 
+            # we may use only 0 output port.
             node_shape = node.output(0).partial_shape.get_max_shape()
             node_output_port = node.output(transformation.target_point.port_id)
             node_output_source_ports = node_output_port.get_target_inputs()
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -49,7 +49,7 @@ class OVModelTransformer(ModelTransformer):
             (OVModelExtractionCommand, self._apply_model_extraction_transformation),
             (OVInplaceFnInsertionCommand, self._apply_insert_operation),
             (OVOutputInsertionCommand, self._apply_output_insertion_transformations),
-            (OVBiasInsertionCommand, self._apply_bias_insertion_transformations)
+            (OVBiasInsertionCommand, self._apply_bias_insertion_transformations),
         ]
 
     @staticmethod
@@ -435,7 +435,9 @@ class OVModelTransformer(ModelTransformer):
         raise RuntimeError(f"Transform type {transform_type} is not supported")
 
     @staticmethod
-    def _apply_bias_insertion_transformations(model: ov.Model, transformations: List[OVBiasInsertionCommand]) -> ov.Model:
+    def _apply_bias_insertion_transformations(
+        model: ov.Model, transformations: List[OVBiasInsertionCommand]
+    ) -> ov.Model:
         """
         Inserts bias operation after corresponding layer.
 
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -448,11 +448,12 @@ class OVModelTransformer(ModelTransformer):
         for transformation in transformations:
             node_name = transformation.target_point.target_node_name
             node = name_to_node_mapping[node_name]
+            node_shape = node.output(0).partial_shape.get_max_shape()
             node_output_port = node.output(transformation.target_point.port_id)
             node_output_source_ports = node_output_port.get_target_inputs()
 
-            bias_shape = [1] * len(node.shape)
-            bias_shape[1] = node.shape[1]
+            bias_shape = [1] * len(node_shape)
+            bias_shape[1] = node_shape[1]
             const_value = np.zeros(bias_shape, dtype=node.get_element_type().to_dtype())
             bias_const_node = opset.constant(const_value, dtype=node.get_element_type())
             bias_const_output_port = bias_const_node.output(0)
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -24,6 +24,7 @@ from nncf.common.graph.transformations.commands import TargetType
 from nncf.common.graph.transformations.layout import TransformationLayout
 from nncf.openvino.graph.node_utils import get_result_node_name
 from nncf.openvino.graph.transformations.commands import OVBiasCorrectionCommand
+from nncf.openvino.graph.transformations.commands import OVBiasInsertionCommand
 from nncf.openvino.graph.transformations.commands import OVFQNodeRemovingCommand
 from nncf.openvino.graph.transformations.commands import OVInplaceFnInsertionCommand
 from nncf.openvino.graph.transformations.commands import OVModelExtractionCommand
@@ -48,6 +49,7 @@ class OVModelTransformer(ModelTransformer):
             (OVModelExtractionCommand, self._apply_model_extraction_transformation),
             (OVInplaceFnInsertionCommand, self._apply_insert_operation),
             (OVOutputInsertionCommand, self._apply_output_insertion_transformations),
+            (OVBiasInsertionCommand, self._apply_bias_insertion_transformations)
         ]
 
     @staticmethod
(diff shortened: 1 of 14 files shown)


nncf/openvino/graph/model_transformer.py
```diff
@@ -431,3 +433,34 @@ def _insert_inplace_operation(
...
+        Inserts bias operation after corresponding layer.
+
+        :param transformations: List of the bias insertion transformations.
+        :returns: Transformed model with null biases.
+        """
+        name_to_node_mapping = OVModelTransformer._get_name_to_node_mapping(model)
+        for transformation in transformations:
+            node_name = transformation.target_point.target_node_name
+            node = name_to_node_mapping[node_name]
+            node_shape = node.output(0).partial_shape.get_max_shape()
```
alexsu52, line 451: if the node will have several different outputs, what will happen in this case?

alexsu52 (commented):
Thanks for the contribution, it looks like it requires additional validation:
- please provide passed conformance build for PTQ
- Please file a ticket to support for insertion bias in the ONNX backend.

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -446,7 +447,8 @@ def output_filter_func(point):
     def get_statistic_points(self, model: TModel) -> StatisticPointsContainer:
         self._set_backend_entity(model)
         model_copy = self._remove_fq_from_inputs(copy_model(model))
-        nncf_graph = NNCFGraphFactory.create(model_copy) if self.nncf_graph is None else self.nncf_graph
+        model_copy = self._insert_null_biases(model_copy)
```
alexsu52, line 450: Is it applicable for the fast bias correction algorithm?

nncf/openvino/graph/model_transformer.py
KodiaqQ: In fact, the `Convolution` layer has only one output tensor that can be fed into several inputs. 
Several types (like `Split`) had multiple outputs, but this is not about the `Convolution` and similar types. 
https://docs.openvino.ai/latest/openvino_docs_ops_convolution_Convolution_1.html

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Yes, applicable. We just need to provide the same method for the FastBC algorithm.

alexsu52: Are you going to do it in this PR or in a follow-up PR?

nncf/openvino/graph/model_transformer.py
```diff
@@ -431,3 +433,34 @@ def _insert_inplace_operation(
...
+            node_output_port = node.output(transformation.target_point.port_id)
+            node_output_source_ports = node_output_port.get_target_inputs()
+
+            bias_shape = [1] * len(node_shape)
+            bias_shape[1] = node_shape[1]
+            const_value = np.zeros(bias_shape, dtype=node.get_element_type().to_dtype())
+            bias_const_node = opset.constant(const_value, dtype=node.get_element_type())
+            bias_const_output_port = bias_const_node.output(0)
+
+            add_node = opset.add(node_output_port, bias_const_output_port, name=f"{node_name}/add_")
```
daniil-lyakhov, line 461: Can we specify in node name that this add is dummy and adds 0?

nncf/openvino/graph/model_transformer.py
```diff
@@ -431,3 +433,34 @@ def _insert_inplace_operation(
...
+
+    @staticmethod
+    def _apply_bias_insertion_transformations(
+        model: ov.Model, transformations: List[OVBiasInsertionCommand]
+    ) -> ov.Model:
+        """
+        Inserts bias operation after corresponding layer.
+
+        :param transformations: List of the bias insertion transformations.
+        :returns: Transformed model with null biases.
```
daniil-lyakhov, line 445: Could you please specify that inserted biases always adds zero i.e. do nothing? Or maybe it worth to feed a value of bias through `OVBiasInsertionCommand`?

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Since we need to have green conformance, I think that it would be better to have in this PR to avoid double validation.

nncf/openvino/graph/model_transformer.py
daniil-lyakhov: Did you check nodes with dynamic shapes / dynamic ranks in tests?

nncf/openvino/graph/model_transformer.py
KodiaqQ: We can, but why do we need it?

nncf/openvino/graph/transformations/commands.py
```diff
@@ -145,3 +145,19 @@ def __init__(self, inputs: List[str], outputs: List[str]):
     def union(self, other: "Command") -> "Command":
         # Have a look at nncf/torch/graph/transformations/commands/PTInsertionCommand
         raise NotImplementedError()
+
+
+class OVBiasInsertionCommand(TransformationCommand):
+    """
+    Inserts null bias for the corresponding node.
+    """
```
daniil-lyakhov, line 153: Should we name it `OVNullBiasInsertionCommand`?

nncf/openvino/graph/model_transformer.py
alexsu52: I absolutely agree with your statement if we take into account the bias correction algorithm. But this function in the OVModelTransformer and I don't know that it works only for convolution. Your doc string is "Inserts bias operation after corresponding layer."

nncf/openvino/graph/model_transformer.py
daniil-lyakhov: I would suggest to add this for the sake of explicitness
 
```suggestion
        """
        Inserts null bias operation after corresponding layer.

        :param transformations: List of the bias insertion transformations.
        :returns: Transformed model with null biases.
```

nncf/openvino/graph/model_transformer.py
daniil-lyakhov: Just for the sake of explicitness and to prevent naming conflicts

nncf/openvino/graph/model_transformer.py
KodiaqQ: Done.

nncf/openvino/graph/model_transformer.py
KodiaqQ: Done.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -49,6 +54,10 @@ def tensor_processor(self) -> ONNXNNCFCollectorTensorProcessor:
     def target_point(target_type: TargetType, target_node_name: str, port_id: int) -> ONNXTargetPoint:
         return ONNXTargetPoint(target_type, target_node_name, port_id)
 
+    @staticmethod
+    def create_bias_insertion_command(node: NNCFNode) -> ONNXNullBiasInsertionCommand:
+        return ONNXNullBiasInsertionCommand(node)
```
alexsu52, line 59: Does the ONNX model transformer support `ONNXNullBiasInsertionCommand` insertion command? What happens if such command will be feed to the ONNX model transformer?

KodiaqQ: It's not support now. Currently, nothing will happen, just skipping in the `ONNXModelTransformer`.

KodiaqQ:
There wouldn't be any major updates here anymore, I think. @alexsu52, @kshpv, review, please.

kshpv:
@KodiaqQ can you clear my understanding? If we insert null biases, they will not be corrected anyway. This is because of the condition here:
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/fast_bias_correction/algorithm.py#L169

kshpv:
> > @KodiaqQ can you clear my understanding? If we insert null biases, they will not be corrected anyway. This is because of the condition here: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/fast_bias_correction/algorithm.py#L169
> 
> That is correct. But let's not forget about the possibility to change the `threshold` value from parameters/config.

Thanks!

nncf/openvino/graph/model_transformer.py
```diff
@@ -431,3 +433,36 @@ def _insert_inplace_operation(
...
+        """
+        name_to_node_mapping = OVModelTransformer._get_name_to_node_mapping(model)
+        for transformation in transformations:
+            node_name = transformation.target_point.target_node_name
+            node = name_to_node_mapping[node_name]
+            # Since layers that may have biases mostly are Convolution or MatMul variations,
+            # we may use only 0 output port.
```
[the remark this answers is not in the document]
KodiaqQ: This check should be done before calling the transformer.

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -130,6 +130,7 @@ def _apply(
         dataset: Optional[Dataset] = None,
     ) -> TModel:
         self._set_backend_entity(model)
+        model = self._insert_null_biases(model)
```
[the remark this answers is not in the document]
KodiaqQ: I've replaced this method. If it would be accepted, I will do the same with the `_remove_fq_from_inputs` as well in a separate PR (#1812).

nncf/quantization/algorithms/bias_correction/openvino_backend.py
```diff
@@ -49,6 +55,16 @@ def tensor_processor(self) -> OVNNCFCollectorTensorProcessor:
...
         return FAKE_QUANTIZE_OPERATIONS
 
+    @property
+    def types_to_insert_bias(self):
+        return [
+            OVConvolutionMetatype,
+            OVGroupConvolutionMetatype,
+            OVDepthwiseConvolutionMetatype,
+            OVConvolutionBackpropDataMetatype,
+            OVGroupConvolutionBackpropDataMetatype,
```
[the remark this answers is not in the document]
KodiaqQ: For now, it's done to align with the POT - https://github.com/openvinotoolkit/openvino/blob/master/tools/pot/openvino/tools/pot/graph/passes.py#L737
In the future PRs, I support your proposal to do it for `MatMul` as well.

kshpv (approved):

alexsu52 (approved):

PR #1779, opened 2023-05-08
[Release v2.5] nn.MHA quantization in case of not fast path (#1765)

ljaljushkin:
### Changes

Not patch `multi_head_attention_forward` to add internal operations of the `nn.MultiheadAttention` to the NNCF graph.
**This is applicable for not “fast path” only.** 
If `torch._native_multi_head_attention` is called on “fast path”, there's no way split into internal operations. 
image

### Reason for changes

`nn.MultiheadAttention` was represented in NNCF graph with just a single `F.multi_head_attention_forward` op. 
Now it's represented as a graph, and it opens opportunities for compression.

### Related tickets

109727, #1765 

### Tests


vshampor (approved):

MaximProshin:
approved for merging to the release

MaximProshin (approved):

PR #1783, opened 2023-05-08
Use torch 2.0.0

vshampor:
### Changes
Allow torch up to ~2.0 to be used with NNCF and set BKC version accordingly.

### Reason for changes
Keeping up with the backend progress.

### Related tickets
N/A

### Tests
PT precommit and E2E


MaximProshin:
I aborted NNCF TF precommit (Ubuntu 20.04) job to reduce the queue as this PR doesn't impact TF and there is still red status for PT parts.

tests/torch/data/reference_graphs/quantized/synthetic_model/MHA.dot
```diff
@@ -11,71 +11,61 @@ strict digraph  {
 "9 TestModel/MultiheadAttention[_layer]/SymmetricQuantizer/symmetric_quantize_1" [id=9, type=symmetric_quantize];
```
vshampor, line 11: @ljaljushkin looks like the MHA in torch has changed. The new quantizer setup looks OK to me, but please check as well.

setup.py
```diff
@@ -132,7 +132,7 @@ def find_version(*file_paths):
 ]
 
 TORCH_EXTRAS = [
-    "torch>=1.9.1,<1.14;python_version < '3.11'",
+    "torch>=1.13.0,<2.1;python_version < '3.11'",
```
MaximProshin, line 135: If some NNCF user already has it installed with older PT version and then installs a newer NNCF version with your changes, will it fail or what will happen?

vshampor:
run pytorch pre-commit tests

vshampor:
run pytorch pre-commit tests

MaximProshin (approved):

vshampor:
run pytorch pre-commit tests

vshampor:
Experiencing intermittent errors in the torch precommit:

1) segfault in sanity sample tests (`classification-resnet18_cifar100_bin_xnor`)
2) some kind of long overflow in test_backward_compat.test_quantization_ckpt_without_wrapped_bn_loading

```
self = SymmetricQuantizer(bit=2660, ch=True)
val = 274636635817014472...3094503977137471487

    @level_high.setter
    def level_high(self, val: int):
>       self._level_high.fill_(val)
E       RuntimeError: Overflow when unpacking long
```

tests/torch/data/reference_graphs/quantized/synthetic_model/MHA.dot
ljaljushkin: Looks good, as there's no sum operation and quantizers can be merged.
image
image

ljaljushkin (approved):

PR #1795, opened 2023-05-10
Fixed const axis for Sequential models

l-bat:
### Changes

Before
image
After

### Reason for changes

* To fix bug with per-channel LSTM, GRU sequence quantization

### Tests

tests/openvino/native/quantization/test_fq_params_calculation.py::test_syntetic_models_fq_scales[inplace-LSTMSequenceModel-mixed]
tests/openvino/native/quantization/test_fq_params_calculation.py::test_syntetic_models_fq_scales[inplace-LSTMSequenceModel-performance]


alexsu52 (approved):
LGTM

PR #1797, opened 2023-05-10
Fix converting FQ to native torch given  `narrow_range` 

AlexanderDokuchaev:
### Changes

- When converting FQ to native torch levels, increase levels if `narrow_range` is `True` for `SymmetricQuantizaer`.
- Clamp weights on strip model, if `narrow_range` is `True`.
- Add tests converting to ONNX for 8bit.

### Reason for changes

Error on export to ONNX
```
SymbolicValueError: For (quant_min, quant_max), ONNX allows only (0, 127), (0, 255) and (-128, 127). Got (-127, 127)
```

### Related tickets

110498


nncf/torch/quantization/layers.py
```diff
@@ -743,9 +743,15 @@ def get_parameters_for_torch_fq(self) -> Tuple[int, int, torch.Tensor, torch.Ten
 
             scale, zero_point = get_scale_zp_from_input_low_input_high(level_low, level_high, input_low, input_high)
 
+            if self.narrow_range:
+                if level_low < 0:
+                    level_low -= 1
+                else:
+                    level_high += 1
```
l-bat, line 750: Can we align it with?
https://github.com/openvinotoolkit/nncf/blob/68c7d39815927eee7cd5334ffb83b70f3eeaff0d/nncf/common/quantization/quantizers.py#L39-L43

tests/torch/quantization/test_strip.py
```diff
@@ -255,10 +252,13 @@ def test_strip_quantization(mode, overflow_fix, num_bits):
 
     inference_model = compression_ctrl.strip()
     x_torch = inference_model(input_tensor)
-    # overflow_fix = False, because target_device = TRIAL
-    check_quantizer_operators(inference_model, 2**num_bits - 1, overflow_fix=False)
+    check_quantizer_operators(inference_model, 2**num_bits - 1)
+
+    assert torch.all(torch.isclose(x_nncf, x_torch, atol=1e-6)), f"{x_nncf.view(-1)} != {x_torch.view(-1)}"
 
-    assert torch.all(torch.isclose(x_nncf, x_torch)), f"{x_nncf.view(-1)} != {x_torch.view(-1)}"
```
vshampor, line 261: Where does the 1e-6, and absolute at that, comes from? We did not need to have any custom tolerances before, why do we need them now?

nncf/torch/quantization/layers.py
AlexanderDokuchaev: It's backward operation to get level_low and level_hight without narrow_range to get levels (255,256) for FQ that can be converted to ONNX

l-bat (approved):

AlexanderDokuchaev:
run pytorch pre-commit tests

vshampor (approved):

alexsu52 (approved):
LGTM

PR #1806, opened 2023-05-12
Restore compatibility with PT 1.9.1

vshampor:
### Changes
If running on Torch < 1.12, do not use `register_load_state_dict_post_hook` and do level range updates on each forward instead.

### Reason for changes
E2E tests for torch 1.9.1 failing, because `torch.nn.Module.register_load_state_dict_post_hook` has only been added in 1.12.

### Related tickets
N/A

### Tests
E2E PT 1.9.1 run TBA

### DIFF
--- a/nncf/torch/quantization/layers.py
+++ b/nncf/torch/quantization/layers.py
@@ -14,6 +14,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
 import torch
+from pkg_resources import parse_version
 from torch import distributed
 from torch import nn
 
@@ -322,6 +323,7 @@ class BaseQuantizer(nn.Module):
                 self.hook.remove()
 
         self.load_listener = LoadStateListener(self)
+        self._old_level_range_setting = False
 
     def enable_gradients(self):
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


nncf/torch/quantization/layers.py
```diff
@@ -347,6 +349,8 @@ def forward(self, x):
         # TODO: refactor to get rid of extra if's and calls on each forward
         if not self.is_enabled_quantization():
             return x
+        if self._old_level_range_setting:
+            self.set_level_ranges()
```
AlexanderDokuchaev, line 353: It's should be run after load pretrained model.
But in this case levels will be incorrected while not run forward, it's affect to strip model if run `strip` after `load_state`.

AlexanderDokuchaev: Nope, `strip` will not be affected `set_level_ranges` calls https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/quantization/strip.py#L75

AlexanderDokuchaev (approved):

vshampor:
Build 22 is OK, also running regular E2E PT build 311 to make sure nothing got worse there.

vshampor:
Build 311 OK.

PR #1808, opened 2023-05-12
Add option no_strip_on_export argument for torch examples

AlexanderDokuchaev:
### Changes

Add `no_strip_on_export ` argument for for examples.
Need to export with FQ custom op, instead of using torch native FQ.


vshampor:
E2E PT build shows that now the int4-int8 FQ models are loadable, which suggests that the fix is working.

vshampor (approved):

PR #1823, opened 2023-05-16
Fix compression training tests

vshampor:
### Changes
LeGR compression training threshold for comparing checkpoint's saved eval metric value with the currently measured one was increased. Levels in `BaseQuantizer`s have been made into torch buffers.

### Reason for changes
#1578 removed the level updating code on forward and sadly this broke down the DDP training because the levels, which were not torch buffers but instead simple Python attributes, were not synchronized across DDP processes and no calls to level-updating property setters (`num_bits` or `signed`) have been made, so while the model in the 0-th (main) DDP process started out the training OK, the other DDP replicas started from pretty much 0 accuracy.

### Related tickets
N/A

### Tests
PT compression training weekly tests

### DIFF
--- a/examples/torch/classification/main.py
+++ b/examples/torch/classification/main.py
@@ -345,6 +345,7 @@ def train(
     best_acc1=0,
 ):
     best_compression_stage = CompressionStage.UNCOMPRESSED
+
     for epoch in range(config.start_epoch, config.epochs):
         # update compression scheduler state at the begin of the epoch
         compression_ctrl.scheduler.epoch_step()
--- a/examples/torch/common/example_logger.py
+++ b/examples/torch/common/example_logger.py
@@ -2,6 +2,7 @@ import logging
 import sys
 
 logger = logging.getLogger("example")
+logger.propagate = False
 
 if not logger.hasHandlers():
     logger.setLevel(logging.INFO)
--- a/nncf/quantization/fake_quantize.py
+++ b/nncf/quantization/fake_quantize.py
@@ -16,6 +16,7 @@ import numpy as np
 
 from nncf.common.quantization.quantizers import calculate_asymmetric_level_ranges
 from nncf.common.quantization.quantizers import calculate_symmetric_level_ranges
+from nncf.common.quantization.quantizers import get_num_levels
 from nncf.common.quantization.structs import QuantizationMode
 from nncf.common.quantization.structs import QuantizerConfig
 from nncf.common.quantization.structs import QuantizerGroup
@@ -230,10 +231,12 @@ def calculate_quantizer_parameters(
     else:
         num_bits = quantizer_config.num_bits
         if quantizer_config.mode == QuantizationMode.SYMMETRIC:
-            _, _, levels = calculate_symmetric_level_ranges(num_bits, signed=True, narrow_range=narrow_range)
+            level_low, level_high = calculate_symmetric_level_ranges(num_bits, signed=True, narrow_range=narrow_range)
+            levels = get_num_levels(level_low, level_high)
             input_low, input_high = symmetric_range(min_values, max_values, levels, quantizer_config, quant_group)
         else:
-            _, _, levels = calculate_asymmetric_level_ranges(num_bits, narrow_range=narrow_range)
+            level_low, level_high = calculate_asymmetric_level_ranges(num_bits, narrow_range=narrow_range)
+            levels = get_num_levels(level_low, level_high)
             input_low, input_high = asymmetric_range(min_values, max_values, quantizer_config, quant_group)
 
     if not quantizer_config.per_channel:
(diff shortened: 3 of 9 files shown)


nncf/quantization/fake_quantize.py
```diff
@@ -230,10 +231,12 @@ def calculate_quantizer_parameters(
     else:
         num_bits = quantizer_config.num_bits
         if quantizer_config.mode == QuantizationMode.SYMMETRIC:
-            _, _, levels = calculate_symmetric_level_ranges(num_bits, signed=True, narrow_range=narrow_range)
+            level_low, level_high = calculate_symmetric_level_ranges(num_bits, signed=True, narrow_range=narrow_range)
+            levels = get_num_levels(level_low, level_high)
```
alexsu52, line 235: What do you think about such simplification?
```suggestion
            levels = calculate_levels(num_bits, narrow_range=narrow_range)
```

alexsu52 (commented):
Will these changes break checkpoint loading? Could you provide e2e test results for this PR?

nncf/quantization/fake_quantize.py
vshampor: The signatures are different for the asymmetric and symmetric cases, I could introduce such a function but I doubt that there would be much savings in the number of code lines.

vshampor:
> Will these changes break checkpoint loading? Could you provide e2e test results for this PR?

The new `_level_low` and `_level_high` buffers are marked as `persistent=False` specifically so that they don't end up in the model's state dict (since they are not first-order parameters, but are instead derived from `num_bits` and `signed_tensor`, which *are* state dict parameters and end up in the checkpoint), but are still tracked by Torch's DDP and are broadcasted properly. E2E is currently not very informative, but will show that the models are at least loadable, so I will schedule a run.

vshampor:
Compression training build 63 OK

nncf/quantization/fake_quantize.py
vshampor: On develop:
`def calculate_symmetric_level_ranges(num_bits: int, signed: bool, narrow_range: bool = False) -> Tuple[int, int, int]:`
`def calculate_asymmetric_level_ranges(num_bits: int, narrow_range: bool = False) -> Tuple[int, int, int]:`

examples/torch/classification/main.py
```diff
@@ -345,6 +345,7 @@ def train(
     best_acc1=0,
 ):
     best_compression_stage = CompressionStage.UNCOMPRESSED
+
```
AlexanderDokuchaev, line 348: Looks like unnecessary change

examples/torch/common/example_logger.py
```diff
@@ -2,6 +2,7 @@
 import sys
 
 logger = logging.getLogger("example")
+logger.propagate = False
```
AlexanderDokuchaev, line 5: What is it change?

nncf/quantization/fake_quantize.py
alexsu52: The calculation levels function does not depend on signed parameter, isn't it?

examples/torch/common/example_logger.py
vshampor: I find out regularly that some Python packages such as `rich` try to set up custom handlers to the global logger that suppress its output. If `.propagate` is not False in such a case, then it looks like the "example" logger will pass the logs to the global logger, whose handlers will then just ignore the logs, and we end up in a situation where no example logs are visible, which is off-putting.

vshampor:
E2E run 330

vshampor:
E2E run 330 passed with a single degradation in unet_mapillary_int8, which currently is unexplained, but seems to be sporadic. Re-launched the unet_mapillary models in PT E2E 335.

tests/torch/test_compression_training.py
```diff
@@ -391,7 +391,7 @@ def _get_weight_path(self, weekly_models_path):
     .expected_accuracy(68.11)
     .weights_filename("mobilenet_v2_32x32_cifar100_68.11.pth")
     .absolute_tolerance_train(1.5)
-    .absolute_tolerance_eval(2e-2),
+    .absolute_tolerance_eval(3e-2),
```
AlexanderDokuchaev, line 394: Do you check reason of changes result, what changed? Can it be error in PR?

nncf/quantization/fake_quantize.py
```diff
@@ -230,10 +231,12 @@ def calculate_quantizer_parameters(
...
             input_low, input_high = symmetric_range(min_values, max_values, levels, quantizer_config, quant_group)
         else:
-            _, _, levels = calculate_asymmetric_level_ranges(num_bits, narrow_range=narrow_range)
+            level_low, level_high = calculate_asymmetric_level_ranges(num_bits, narrow_range=narrow_range)
```
AlexanderDokuchaev, line 238: What do you think about add function like `calculate_asymmetric_num_level(num_bits, narrow_range=narrow_range)`. Looks like unnecessary code to get `num_levels`

vshampor:
E2E run 335 is green for the models launched.

PR #1827, opened 2023-05-17
[Release][PTQ][OV] Inplace statistic reducers names and hashes fix

daniil-lyakhov:
Copy of #1796 

### Changes
Names of reducers is used instead of types of reducers to name statistic subnet operations
Reducer names now differs one from another if they have different attributes and aren't merged
Hash of reducer now take into account reduction_shape

### Reason for changes
To fix problem when two different statistics that are instances of one class and share target point have equal name

### Related tickets
109989

### Tests

Test on reducers merging
Test with a model that have target point with two different reducers that are instances of one class

### DIFF
--- a/nncf/experimental/common/tensor_statistics/collectors.py
+++ b/nncf/experimental/common/tensor_statistics/collectors.py
@@ -50,9 +50,9 @@ class TensorReducerBase(ABC):
     def output_port_id(self) -> int:
         return 0
 
-    @classmethod
-    def name(cls):
-        return cls.__name__
+    @property
+    def name(self):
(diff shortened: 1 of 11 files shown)


daniil-lyakhov:
Full check showed on regression on this PR.

@MaximProshin please approve merge of this fix

MaximProshin (approved):

MaximProshin:
approved for merging

KodiaqQ (approved):

daniil-lyakhov:
run pytorch pre-commit tests

PR #1828, opened 2023-05-17
Force range initialization to be applied in eval mode

nikita-savelyevv:
### Changes

Force range initialization to be applied in eval mode.

### Reason for changes

If model in train mode is given to `create_compressed_model()`, the graph is built for the model in eval mode, but range initialization statistics are collected in train mode. This may result in wrong statistics collection if model behaves differently in train and eval modes.

### Tests

Added a test failing before the fix.

### DIFF
--- a/tests/torch/quantization/test_range_init.py
+++ b/tests/torch/quantization/test_range_init.py
@@ -810,9 +810,18 @@ def test_quantize_range_init_sets_correct_scale_shapes(quantizer_range_init_test
 
 
 def test_range_initialization_in_train_mode():
-    # Check that if a model in train mode is being compressed, range initialization still runs in eval mode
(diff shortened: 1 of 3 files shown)


tests/torch/quantization/test_range_init.py
```diff
@@ -809,16 +809,21 @@ def test_quantize_range_init_sets_correct_scale_shapes(quantizer_range_init_test
             assert False  # options above should be exhaustive
 
 
-class AbsTwosDataset:
```
nikita-savelyevv, line 812: This class was not used anywhere

vshampor (approved):

ljaljushkin (approved):

PR #1829, opened 2023-05-17
Fix setting forward/backward hooks on NNCFNetwork

vshampor:
### Changes

#1584 seems to have broken down the use case where the original model object has forward or backward hooks set up via `torch.nn.Module.register_forward_hook` etc. This PR restores the corresponding functionality.

### Reason for changes

Although the use case is not common, leaving things as-is means that a limitation is introduced by NNCF in regards to the fundamental torch behaviour, which is not something we want.

### Related tickets

N/A

### Tests

tests.torch.test_nncf_network.test_forward_hooks_are_preserved

### DIFF
--- a/tests/torch/ptq/test_calculation_quantizer_params.py
+++ b/tests/torch/ptq/test_calculation_quantizer_params.py
@@ -291,12 +291,12 @@ def calculate_fq_params(model, input_data):
     conv2_w = model.conv2.weight
     conv2_w_stats = calculate_statistics(conv2_w, QuantizationMode.SYMMETRIC, QuantizerGroup.WEIGHTS)
     return {
-        "FakeQuantize_6": conv1_stats,
-        "relu/FakeQuantize": bn1_stats,
-        "avg_pool/FakeQuantize": avg_pool_stats,
-        "bn1/FakeQuantize": conv2_stats,
-        "conv1/pre_ops.0/op/FakeQuantize": conv1_w_stats,
-        "conv2/pre_ops.0/op/FakeQuantize": conv2_w_stats,
+        "/FakeQuantize": conv1_stats,
+        "/relu/FakeQuantize": bn1_stats,
+        "/avg_pool/FakeQuantize": avg_pool_stats,
+        "/bn1/FakeQuantize": conv2_stats,
+        "/conv1/pre_ops.0/op/FakeQuantize": conv1_w_stats,
+        "/conv2/pre_ops.0/op/FakeQuantize": conv2_w_stats,
     }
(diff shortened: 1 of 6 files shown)


tests/torch/test_nncf_network.py
```diff
@@ -1090,3 +1090,29 @@ def wrapper(*args, **kwargs):
...
+    original_obj.register_forward_hook(hook)
+
+    hook.enabled = False
+    nncf_net = NNCFNetwork(original_obj, [ModelInputInfo(SimplestModel.INPUT_SIZE)])
+    hook.enabled = True
+
+    assert len(nncf_net._forward_hooks) == 1
+    assert next(iter(nncf_net._forward_hooks.values())) is hook
+    nncf_net(torch.ones(SimplestModel.INPUT_SIZE))
+    assert hook.call_count == 1
```
ljaljushkin, line 1118: What was before the fix? Was count equal 0?

vshampor: Indeed.

ljaljushkin (approved):

l-bat (approved):

tests/torch/ptq/test_calculation_quantizer_params.py
```diff
@@ -291,12 +291,12 @@ def calculate_fq_params(model, input_data):
...
-        "bn1/FakeQuantize": conv2_stats,
-        "conv1/pre_ops.0/op/FakeQuantize": conv1_w_stats,
-        "conv2/pre_ops.0/op/FakeQuantize": conv2_w_stats,
+        "/FakeQuantize": conv1_stats,
+        "/relu/FakeQuantize": bn1_stats,
+        "/avg_pool/FakeQuantize": avg_pool_stats,
+        "/bn1/FakeQuantize": conv2_stats,
+        "/conv1/pre_ops.0/op/FakeQuantize": conv1_w_stats,
+        "/conv2/pre_ops.0/op/FakeQuantize": conv2_w_stats,
     }
```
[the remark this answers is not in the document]
vshampor: Looks like the name building algorithm in ONNX tracing code uses the same `Scope`-like mechanism now like we do to disambiguate operation names. This is probably triggered by a `torch.nn.Module.__call__`, and the top-level version of this `__call__` was missing after the NNCFNetwork refactor. These are the reference names of the ONNX nodes after export, and the new format is correct for the non-NNCF case, so the fix in this PR also fixes the ONNX node name building. The wrong format ended up being in the tests because this test had been added after the NNCFNetwork refactor.

PR #1835, opened 2023-05-19
[PTQ][OV] Drop null biases for FBC

KodiaqQ:
### Changes

- Temporary remove inserting null biases from FBC

### Reason for changes

- Null biases do not affect accuracy for now due to the magnitude formula

### DIFF
--- a/nncf/quantization/algorithms/fast_bias_correction/algorithm.py
+++ b/nncf/quantization/algorithms/fast_bias_correction/algorithm.py
@@ -27,7 +27,6 @@ from nncf.common.tensor import NNCFTensor
 from nncf.common.tensor_statistics.statistic_point import StatisticPoint
 from nncf.common.tensor_statistics.statistic_point import StatisticPointsContainer
 from nncf.common.utils.backend import BackendType
-from nncf.common.utils.backend import copy_model
 from nncf.common.utils.backend import get_backend
 from nncf.quantization.algorithms.algorithm import Algorithm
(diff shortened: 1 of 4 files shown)


alexsu52 (approved):
Thanks!

KodiaqQ:
run pylint pre-commit tests

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -123,7 +122,6 @@ def _apply(
         dataset: Optional[Dataset] = None,
     ) -> TModel:
         self._set_backend_entity(model)
-        model = self._backend_entity.insert_null_biases(model)
```
AlexanderDokuchaev, line 126: `insert_null_biases` method still exists in backend_entity.
And why it's temporary if it's not affect for FBC?

KodiaqQ: We wonder to check out the algorithm's behavior with the null biases and unlimited threshold in the future.

alexsu52: I am for clean code, if the function is not used then there is no point in keeping it in the code base. It can always be brought back from history.

KodiaqQ: No problem, dropped it.

AlexanderDokuchaev (approved):

PR #1843, opened 2023-05-22
Remove convert_to_mixed_preset parameter

andrey-churkin:
### Changes

- The `convert_to_mixed_preset`  parameter was removed from the `AdvancedAccuracyRestorerParameters`.

### Reason for changes

- As we decided we will convert a model to a mixed preset as part of the tune hyperparameters algorithm. So we don't need this parameter anymore.

### Related tickets

N/A

### Tests

N/A


alexsu52 (approved):
Thanks!

PR #1852, opened 2023-05-24
Ignore 0-th port of baddbmm in quantization

vshampor:
### Changes
Add a field to `OperatorMetatype` that specifies input ports to be ignored and use it to disable quantization on the 0-th input port of the `torch.baddbmm` operation.

### Reason for changes
The 0-th input for baddbmm is a bias addition in essence, and if the input tensor is a constant, then adding a quantizer there breaks down a hardware-fused pattern and reduces performance. It is still possible that the 0-th input may be a runtime-dynamic activation tensor, in which case the quantization of this input would probably still be relevant, but right now there is no way in NNCF PT to discern activation and constant inputs.

### Related tickets
110848

### Tests

### DIFF
--- a/nncf/common/graph/operator_metatypes.py
+++ b/nncf/common/graph/operator_metatypes.py
@@ -22,11 +22,13 @@ class OperatorMetatype:
     :param name: The name of the operator.
     :param hw_config_names: The names of the hardware configurations.
     :param output_channel_axis: The axis along which the output channels of the operator are arranged.
+    :param ignored_input_ports: Input ports of the operations that should not be considered for purposes of compression.
     """
 
     name: str = ""
     hw_config_names: List[str] = []
     output_channel_axis: Optional[int] = None
+    ignored_input_ports: List[int] = []
 
     @classmethod
     def get_all_aliases(cls) -> List[str]:
--- a/nncf/common/quantization/quantizer_propagation/solver.py
+++ b/nncf/common/quantization/quantizer_propagation/solver.py
@@ -1155,6 +1155,11 @@ class QuantizerPropagationSolver:
                 pred_node_type
             ), "Invalid insertion point graph supplied for quantizer propagation!"
 
+            ip = pred_node[QuantizerPropagationStateGraph.QUANT_INSERTION_POINT_DATA_NODE_ATTR]
+            input_port_id = ip.input_port_id
+            if input_port_id in metatype.ignored_input_ports:
+                continue
+
             edge = quant_prop_graph.edges[pred_ip_key, operator_node_key]
             if not edge[QuantizerPropagationStateGraph.IS_INTEGER_PATH_EDGE_ATTR]:
                 pred_ip_key_vs_qconf_dict[pred_ip_key] = qconf_list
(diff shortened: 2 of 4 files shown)


AlexanderDokuchaev:
run pytorch pre-commit tests

AlexKoff88:
I confirm that it works as expected.

AlexKoff88 (approved):

nncf/common/graph/operator_metatypes.py
```diff
@@ -22,11 +22,13 @@ class OperatorMetatype:
     :param name: The name of the operator.
     :param hw_config_names: The names of the hardware configurations.
     :param output_channel_axis: The axis along which the output channels of the operator are arranged.
+    :param ignored_input_ports: Input ports of the operations that should not be considered for purposes of compression.
     """
 
     name: str = ""
     hw_config_names: List[str] = []
     output_channel_axis: Optional[int] = None
+    ignored_input_ports: List[int] = []
```
KodiaqQ, line 31: Shouldn't the required argument be placed before optional one?

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -1155,6 +1155,11 @@ def _setup_initial_quantizers_for_operator_node(
                 pred_node_type
             ), "Invalid insertion point graph supplied for quantizer propagation!"
 
+            ip = pred_node[QuantizerPropagationStateGraph.QUANT_INSERTION_POINT_DATA_NODE_ATTR]
```
KodiaqQ, line 1158: Shouldn't we check that `QUANT_INSERTION_POINT_DATA_NODE_ATTR` exists first?

nncf/torch/graph/operator_metatypes.py
```diff
@@ -498,11 +498,22 @@ class PTMatMulMetatype(PTOperatorMetatype):
...
 
+@PT_OPERATOR_METATYPES.register()
+class PTBaddBmmMetatype(PTOperatorMetatype):
+    name = "MatMulOp"
+    module_to_function_names = {NamespaceTarget.TORCH: ["baddbmm"]}
+    hw_config_names = [HWConfigOpName.MATMUL]
+    # 0-th arg to the baddbmm is basically a (b)ias to be (add)ed to the (bmm) operation,
+    # presuming that most runtime implementations will fuse the bias addition into the matrix multiplication
+    # and therefore won't quantize the bias input, as this would break the hardware-fused pattern.
+    ignored_input_ports: List[int] = [0]
```
KodiaqQ, line 514: Why do we need a type definition here?

vshampor: Doesn't seem to hurt

nncf/common/graph/operator_metatypes.py
vshampor: This is not a dataclass.

nncf/common/quantization/quantizer_propagation/solver.py
vshampor: The `assert` above serves the same purpose. It is not pretty, and I will have to refactor this when I get to refactoring the many main and auxiliary graph objects.

KodiaqQ (approved):

PR #1853, opened 2023-05-24
Make uncompressed_model_accuracy argument required for AATL creation

nikita-savelyevv:
### Changes

Made `uncompressed_model_accuracy` argument to be required for accuracy aware training loop creation

### Reason for changes

A usage scenario where this parameter is not provided results in wrong behavior. This is because we evaluate a model which is already compressed. Hence, the computed accuracy is not in fact the uncompressed model accuracy.


vshampor (approved):

alexsu52:
run pylint pre-commit tests

alexsu52 (approved):
LGTM

PR #1854, opened 2023-05-24
Fix PTQ example versions

vshampor:


AlexanderDokuchaev:
run pytorch pre-commit tests

vshampor:
Checked locally, affected examples still work with the new versions.

PR #1876, opened 2023-05-31
Known issues update

KodiaqQ:
### Changes

- Added known issues section for PTQ into ReleaseNotes.md

### Reason for changes

- Known issues

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -25,6 +25,11 @@ Post-training Quantization:
   - Added improvements for statistic collection process (collect weights statistics only once).
   - (PyTorch, OpenVINO, ONNX) Introduced unified quantizer parameters calculation.
 
+- Known issues:
+  - The accuracy of the optimized model on architectures such as DenseNets may a slight drops due to a malfunctioning scales unification functionality.
+  - On transformer architectures, the BiasCorrection algorithm (option `fast_bias_correction`: false) can cause unexpected hangs in the pipeline.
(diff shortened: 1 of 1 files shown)


alexsu52 (commented):
Not clear statements and what the user should do in these cases.

ReleaseNotes.md
```diff
@@ -25,6 +25,11 @@ Post-training Quantization:
   - Added improvements for statistic collection process (collect weights statistics only once).
   - (PyTorch, OpenVINO, ONNX) Introduced unified quantizer parameters calculation.
 
+- Known issues:
+  - The accuracy of the optimized model on architectures such as DenseNets may a slight drops due to a malfunctioning scales unification functionality.
+  - On transformer architectures, the BiasCorrection algorithm (option `fast_bias_correction`: false) can cause unexpected hangs in the pipeline.
+  - On architectures such as MobileNets, there may be a slight drops in accuracy due to the absence of Channel Alignment algorithm.
+
```
MaximProshin, line 32: I suggest to rephrase it to:
 - quantize() method (Post-Training Quantization) can generate inaccurate int8 results for models with the DenseNet architecture. Use quantize_with_accuracy_control() in such case.
 - quantize() method (Post-Training Quantization) can hang on models with Transformer architecture when 'fast_bias_correction' optional parameter is set to False. Don't set it to False or use quantize_with_accuracy_control() in such case.
 - quantize() method (Post-Training Quantization) can generate inaccurate int8 results for models with the MobileNet architecture on non-VNNI architecture.

MaximProshin (changes requested):

MaximProshin (approved):

PR #1878, opened 2023-06-01
Fix openvino version

alexsu52:
### Changes

Fixed OpenVINO version `openvino==2022.3.0`

### Reason for changes

Make pre-commits green 

### Related tickets

N/A

### Tests

pre-commits


KodiaqQ (approved):

vshampor:
Remember, we have a makefile now. For this kind of a temporary fix it might be better to add a `pip --force-reinstall` step there instead of in the setup.py.

alexsu52:
> Remember, we have a makefile now. For this kind of a temporary fix it might be better to add a `pip --force-reinstall` step there instead of in the setup.py.

Thanks for the comment. I think we can delete framework dependencies from the setup.py at all.

alexsu52:
run openvino pre-commit tests

PR #1885, opened 2023-06-01
[OV] Update tests references

KodiaqQ:
### Changes

- **Reverted**. Added `pytest.xfail` for `mobilenet-v3-small-1.0-224-tf` model due to OMZ failures (fixed in master - https://github.com/openvinotoolkit/open_model_zoo/pull/3777);
- Updated `googlenet-v3`, `mobilenet-v2`, `resnet-18` architecture examples (.dot) due to input **processing pattern changes** (see below);
- Updated `yolo-v4-tiny` and `mobilenet-v3-small` architecture examples (.dot) and scales (.json) due to layers names changes that provides new MO TF frontend (see below);
- Updated `mobilenet-v2`, `resnet-18` architecture examples scales (.json) due to input **processing pattern changes** (see below);
- Updated required OpenVINO package version to 2023.0.0;
- Updated OMZ package version;

### Reason for changes

- Red CI

### Details
- **OV 2023.0.0** output graph (resnet-18):
image
- **OV 2022.3.0** output graph (resnet-18):
image

As we can see, **only** the output processing subgraph was changed. 
For example, the FakeQuantize node for the second Convolution does not change its parameters.

In addition to the fact that the names have changed due to the new front end, the position of the layers has also changed. 
For example:
- **OV 2023.0.0** original graph (mobilenet-v3-small):
image
- **OV 2022.3.0** original graph (mobilenet-v3-small):
image

This is the reason for changing the number of FakeQuantize layers (+1 activation for separate Add) in the model (22.3 vs. 23.0).


tests/openvino/native/quantization/test_sanity.py
```diff
@@ -38,6 +38,8 @@
 def test_compression(data_dir, tmp_path, model, dataset, ref_metrics):
     if is_windows() and model == "mobilefacedet-v1-mxnet":
         pytest.xfail("OMZ for Windows has version 1.2.0 pinned that is incompatible with Python 3.8+")
+    if model == "mobilenet-v3-small-1.0-224-tf":
```
alexsu52, line 41: Open Model Zoo is used from the master https://github.com/openvinotoolkit/nncf/blob/847bbcede03989c703310e6bf056ce646568a76b/Makefile#L51 not form openvino-dev package. Is it possible to update open model zoo and remove pytest.xfail for "mobilenet-v3-small-1.0-224-tf"

alexsu52 (changes requested):

tests/openvino/native/quantization/test_sanity.py
KodiaqQ: Are you sure that it would work since OMZ is part of OpenVINO, not NNCF? I'm not.
Also, for CI it wouldn't work.

KodiaqQ: Done.

alexsu52 (approved):
LGTM

PR #1901, opened 2023-06-06
Added quantization config for ReduceSum operation

alexsu52:
### Changes

Added quantization config for ReduceSum operation

### Reason for changes

OpenVINO supports int8 precision for ReduceSum operation

### Related tickets

https://github.com/openvinotoolkit/openvino/issues/17389

### Tests

N/A

### DIFF
--- a/nncf/common/hardware/configs/cpu.json
+++ b/nncf/common/hardware/configs/cpu.json
@@ -252,6 +252,12 @@
                 "weights": "q8_w_sym"
             }
         },
+        {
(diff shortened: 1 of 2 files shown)


alexsu52:
> @alexsu52, which performance numbers can we achieve using this new quantization scheme? Do we have some evidence on real cases?

Adding ReduceSum does not bring huge speed up of whole model but ReduceSum is 1.34x faster in int8 precision.

KodiaqQ (approved):

PR #1903, opened 2023-06-06
BNAS shape pruning processor.

jpablomch:
### Changes

Handles input masks in NAS by inheriting shape pruning class from common folder. Current "get_input_mask" function doesn't consider "add_dynamic_inputs" in elastic width, which can be used to tweak the elasticity configuration. Currently, in some cases _get_next_node_sparse_multiplier raises error: "doesn't have closing mask". This PR fixes this issue.

### Reason for changes

Enable more flexibility in NAS to tweak elastic configuration. 

### Related tickets

N/A

### Tests

Included tests for dynamic inputs.


jpablomch:
@ljaljushkin, @yzheng124 Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -672,35 +675,41 @@ def activate_subnet_for_config(self, config: ElasticWidthConfig) -> None:
...
-                        for previous in previous_nodes:
-                            if "output_mask" in previous.data:
-                                if previous.data["output_mask"] is not None:
-                                    input_masks.append(previous.data["output_mask"])
-                                    input_masks = [i for i in input_masks if i]
-                                else:
-                                    nodes_to_check.append(previous)
+    def get_input_masks_with_add_dynamic_inputs(self, node: NNCFNode, graph: NNCFGraph) -> List[Optional[NNCFTensor]]:
+        """
+        Returns input masks for all inputs of given NNCFNode.
```
ljaljushkin, line 680: Could you elaborate in the method's description the difference between common `get_input_masks` method and this one?

nncf/experimental/torch/nas/bootstrapNAS/elasticity/nas_shape_pruning_processor.py
```diff
@@ -0,0 +1,48 @@
...
+        prunable_types: List[str],
+        pruning_operations_metatype: List[str],
+        get_input_masks_func: Callable[[NNCFNode, NNCFGraph], List[Optional[NNCFTensor]]] = get_input_masks,
+    ):
+        super().__init__(prunable_types, pruning_operations_metatype)
+        self.get_input_masks_func = get_input_masks_func
+
+    def _get_next_node_sparse_multiplier(
+        self, graph: NNCFGraph, next_node: NNCFNode, cluster: Clusterization[PrunedLayerInfoBase]
+    ) -> int:
```
ljaljushkin, line 40: Could you please elaborate which exactly scenario this fix?

yzheng124: This function mainly fixes the scenario where [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139) is used in elastic width. Some operators are only prunable in some specific cases. For security reasons, those operators are usually set as unprunable in the common function. For some specific models, in order to further improve performance, we need to add output nodes of those operators in [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139) to force enable their elasticity.

For example, here is part of the graph in OFA-MBV3:
image
If we do not use [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139), node104(marked in yellow) is unprunable because the node107 and node108(marked in green) are mean type, which sets output_mask = None. However, we can force enable node104 to be prunable by adding node109 to [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139) and adding node104 into [overwrite_groups](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L137).


Compared with [get_input_masks](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/common/pruning/utils.py#L361), `get_input_masks_with_add_dynamic_inputs` considers input masks of operators in [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139). Here the function `_get_next_node_sparse_multiplier` uses `get_input_masks_with_add_dynamic_inputs` instead of  [get_input_masks](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/common/pruning/utils.py#L361) to get input masks, so it can handle the scenario with [add_dynamic_inputs](https://github.com/openvinotoolkit/nncf/blob/dd37d445e3112edcd727fcef0196bc1caede4398/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py#L139).

Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
```diff
@@ -672,35 +675,41 @@ def activate_subnet_for_config(self, config: ElasticWidthConfig) -> None:
...
+        :param node: Given NNCFNode.
+        :param graph: Graph to work with.
+        :return: Input masks.
+        """
+        retval = []
+        input_masks = [input_node.data["output_mask"] for input_node in graph.get_previous_nodes(node)]
+        if self._add_dynamic_inputs:
```
yzheng124, line 687: Here is the place that `get_input_masks_with_add_dynamic_inputs` computes input masks of operators in `add_dynamic_inputs`.

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
yzheng124: Done. Thank you!

ljaljushkin: thanks!

daniil-lyakhov:
Hi @jpablomch, @yzheng124! Thank you for the contribution!
From the problem statement, it looks like average pooling inside the SE block of MNV3 implemented by two mean layers ruins pruning. In fact, this SE block always cause a lot of trouble for pruning (once one NNCF user implemented SE block with linear layers instead of 1x1 convs, so we worked hard to support linear layers in NNCF)

I believe we can process mean layer in a common way: if you check Average Pool operation in NNCF Pruning, you find that it makes identity mask propagation. This is not fair in common way, so last step of pruning analysis (symbolic mask propagation) checks that amount of output channels in prunable layers (producers) and amount of input channels in closing prunable layers (consumers) are equal. There are two possible ways average pooling affect pruning, let's elaborate on them:

1) AVG pooling pools some dimensions, and channel dimension is not in pooled dimensions. So, symbolic  mask propagation 
inits C channels for producer conv, passes C through pooling layer and compares with input channels of consumer layer. Consumer layer has C input channels as well, so in this case this subgraph will be marked as  prunable
image

2) AVG pooling pools some dimensions, and channel dimension is  in  pooled dimensions. So, symbolic  mask propagation 
inits C channels for producer conv, passes C through pooling layer and compares with input channels of consumer layer. Consumer layer has C* input channels, not equal to C output channels of producer conv (C* < C), so in this case this subgraph will be marked  as unprunable
image

As you can check, average pooling is actually equivalent to mean layer, so, you can just add mean metatype to the `Pass through pruning metatype`, and everything should work as intended. Please try this, check it in practice and add tests to filter pruning test cases to make solution more common therefore more safe :)

yzheng124:
Hi @daniil-lyakhov, thanks for the great explanation! As for "you can just add mean metatype to the `Pass through pruning metatype`", could you please elaborate it a bit more? I cannot find the `Pass through pruning metatype` in NNCF code. Thanks!

jpablomch:
> > Hi @daniil-lyakhov, thanks for the great explanation! As for "you can just add mean metatype to the `Pass through pruning metatype`", could you please elaborate it a bit more? I cannot find the `Pass through pruning metatype` in NNCF code. Thanks!
> 
> It's about metatypes registered in PTIdentityMaskForwardPruningOp

Thanks, @daniil-lyakhov and @ljaljushkin! We have made changes based on @daniil-lyakhov 's suggestions. We have also added the required tests. Please have a look. Thanks!

ljaljushkin (approved):

daniil-lyakhov (approved):
LGTM, thanks for the contribution!

PR #1799, opened 2023-05-11
[Style] Enable pylint for each python file in repo

daniil-lyakhov:
### Changes

* Added missed requirements to Makefile
* Files for pylint now are searching by rules: if file is in a directory with name==backend or file name started with backend - it's checked by backend pytlint
Common files are all `.py` files that weren't chosen by any known backend

PR #1904, opened 2023-06-07
Fixed pylint issue in the object detection example

alexsu52:
### Changes

Follow up https://github.com/openvinotoolkit/nncf/pull/1799


alexsu52:
run openvino pre-commit tests

PR #1914, opened 2023-06-14
Update pymoo to 0.6

yzheng124:
### Changes

Update pymoo to 0.6. This package is used in BootstrapNAS search. Currently BootstrapNAS uses pymoo 0.5, which was released two years ago. 

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ INSTALL_REQUIRES = [
     # Ticket: 69520
     "packaging>=20.0",
     "pyparsing<3.0",
-    "pymoo==0.5.0",
+    "pymoo @ git+https://github.com/anyoptimization/pymoo.git@695cb26923903f872c7256a9013609769f3cc2bd",
     "jsonschema>=3.2.0",
     "pydot>=1.4.1",
     "jstyleson>=0.0.2",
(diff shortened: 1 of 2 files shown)


yzheng124:
@ljaljushkin Please take a look at this PR. Thanks!
@jpablomch

ljaljushkin:
Does it affect tests for search?

yzheng124:
> Does it affect tests for search?

No. This change passes all tests for search. There is no performance degradation. Thanks!

ljaljushkin (approved):

setup.py
```diff
@@ -113,7 +113,7 @@ def find_version(*file_paths):
     # Ticket: 69520
     "packaging>=20.0",
     "pyparsing<3.0",
-    "pymoo==0.5.0",
+    "pymoo @ git+https://github.com/anyoptimization/pymoo.git@695cb26923903f872c7256a9013609769f3cc2bd",
```
yzheng124, line 116: The reason we use a specific commit instead of tag/0.6.0 is because BootstrapNAS needs this commit: https://github.com/anyoptimization/pymoo/issues/388. This commit is now only included in the main branch.

ljaljushkin: ok, thanks for clarification

PR #1916, opened 2023-06-14
Fix padding adjustment issue in elastic kernel

yzheng124:
### Changes

- Consider the original padding value when computing the kernel size list.

-  Consider the original padding value in `ElasticKernelPaddingAdjustment`. 

### Reason for changes

The old `ElasticKernelPaddingAdjustment` directly sets `padding value = active_kernel_size // 2`, which means it always sets output shape of the node equal to its input shape with different active kernel sizes.

This PR considers the original padding value, which always sets output shape of the node equal to its original output shape (sometimes the original output shape != input shape) with different active kernel sizes.

For example:
Suppose there is a node with kernel_size = 7 and padding = 1 in the original model. In this case, its original output shape != input shape because the original padding value != 2.
With elastic kernels enabled:

- The old code has kernel list = [7, 5, 3]. Padding adjustment method updates padding=2 for active kernel size =7, padding=1 for active kernel size =5, padding=0 for active kernel size=3.

- This PR has kernel list = [7, 5]. Padding adjustment method updates padding=1 for active kernel size=7 (keep same as the original model) and padding=0 for active kernel size =5.

### Related tickets

N/A

### Tests

test_elastic_kernel_output_shape


nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -206,7 +215,7 @@ def generate_kernel_size_list(self, max_kernel_size: KernelSizeType) -> List[Ker
             return [1]
         kernel = max_kernel_size
         ks_list = []
-        while kernel > 1:
+        while kernel >= max(max_kernel_size - 2 * original_padding_value, 3):
```
yzheng124, line 218: Using max(max_kernel_size - 2 * original_padding_value, 3) instead of max_kernel_size - 2 * original_padding_value is mainly for consistency with the old code. The old code doesn't allow 1 to appear in elastic kernel lists.

tests/torch/nas/creators.py
```diff
@@ -133,9 +133,11 @@ def create_supernet(
     return ctrl.multi_elasticity_handler, model
 
 
-def create_single_conv_kernel_supernet(kernel_size=5, out_channels=1) -> Tuple[ElasticKernelHandler, NNCFNetwork]:
+def create_single_conv_kernel_supernet(
+    kernel_size=5, out_channels=1, padding=2
```
yzheng124, line 137: Set `padding = 2` so that after enabling elastic kernel, this function can create the same supernet as the old one. Other modification of the padding value are for the same reason.

tests/torch/nas/test_search_space.py
```diff
@@ -93,7 +93,7 @@ def __str__(self):
         ElasticityDim.DEPTH,
         model_cls=DepthBasicConvTestModel,
         params={"min_block_size": 1, "hw_fused_ops": False},
-        ref_search_space=[[]],
+        ref_search_space=[[0], []],
```
yzheng124, line 96: After modifying the original padding value of `DepthBasicConvTestModel`, some output shape are different from the one in the original old model. So elastic depth has a larger search space.

yzheng124:
@ljaljushkin Please take a look at it. Thanks!
@jpablomch

ljaljushkin (changes requested):
Generally looks good to me. 
I wonder whether all combinations of elastic kernel and padding are covered in the unit tests now: 
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/nas/test_elastic_kernel.py#L46
at least need a test for `padding value != kernel size // 2`

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -168,7 +168,13 @@ class ElasticKernelConv2DOp(ElasticKernelOp, nn.Module):
     and modifies in the way that kernel size is changing to a given value.
     """
 
-    def __init__(self, max_kernel_size: KernelSizeType, node_name: NNCFNodeName, params: ElasticKernelParams):
+    def __init__(
+        self,
+        max_kernel_size: KernelSizeType,
+        node_name: NNCFNodeName,
+        params: ElasticKernelParams,
+        original_padding_value: Optional[int] = 0,
```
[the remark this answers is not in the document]
yzheng124: Done. Thanks!

nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_kernel.py
```diff
@@ -193,7 +200,9 @@ def __init__(self, max_kernel_size: KernelSizeType, node_name: NNCFNodeName, par
         for name, param in scale_params.items():
             self.register_parameter(name, param)
 
-    def generate_kernel_size_list(self, max_kernel_size: KernelSizeType) -> List[KernelSizeType]:
+    def generate_kernel_size_list(
+        self, max_kernel_size: KernelSizeType, original_padding_value: int
```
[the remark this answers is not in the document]
yzheng124: Done. Thanks!

yzheng124:
> test_elastic_kernel_output_shape

Added a test "test_elastic_kernel_output_shape". Thanks!

PR #1917, opened 2023-06-14
[ONNX] Add support of correct quantization of MatMul

kshpv:
### Changes

Add correct handling MatMul ops during quantization for every case with only activations and activation with weight.

- Introduce ONNXLayerAttributes, which are assigned to every NNCFNode.
- Split weight_port_ids to constan_port_ids and possible_weight_ports. possible_weight_ports are used to determine weight dynamically.
- Add logic to determine whether a node has weight (_get_weight_edge_name)
- Add transpose attribute for GEMM node

### Reason for changes

To get the most optimized performance after quantization.

### Related tickets

112530
95156

### Tests

Synthetic models were added in graph tests
gpt-2 and bertsquad were added in graph test
Add test on scales after quantization
Add test on transpose_axis func


kshpv:
run openvino pre-commit tests

kshpv:
run pylint pre-commit tests

kshpv:
run pytorch pre-commit tests

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -34,14 +34,14 @@ def get_subtypes(cls) -> List[Type[OperatorMetatype]]:
         return cls.subtypes
 
     @classmethod
-    def matches(cls, model: onnx.ModelProto, node: onnx.NodeProto) -> Optional[bool]:
+    def matches(cls, onnx_graph: ONNXGraph, node: onnx.NodeProto) -> Optional[bool]:
```
alexsu52, line 37: Could you explain the benefits of migration to ONNXGraph?

nncf/onnx/graph/onnx_graph.py
```diff
@@ -15,10 +15,6 @@
 import onnx
 from onnx import numpy_helper
 
-from nncf.onnx.graph.metatypes.onnx_metatypes import ONNX_OPERATION_METATYPES
-from nncf.onnx.graph.metatypes.onnx_metatypes import WEIGHT_LAYER_METATYPES
-from nncf.onnx.graph.metatypes.onnx_metatypes import OpWeightDef
-
 
 # pylint: disable=too-many-public-methods
 class ONNXGraph:
```
alexsu52, line 20: My opinion is that ONNXGraph is a redundant abstraction to access ONNX model attributes. I suggest minimizing the use ONNXGraph instead of increasing the use of ONNXGraph to access ONNX model attributes.

alexsu52 (commented):
Could you provide results of the e2e test and conformance test for ONNX backend?

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,183 @@
...
+    def get_bias_port_ids(self) -> List[int]:
+        """
+        Returns indices of input ports corresponding to the constant nodes.
+
+        :returns: List of input port indices with constants.
+        """
+        return list(self.bias_attrs.keys())
+
+
+def _get_weight_edge_name(onnx_graph: ONNXGraph, node: onnx.NodeProto, port_id: int) -> Optional[str]:
```
kshpv, line 67: Not sure whether the logic of ` _get_weight_edge_name` is fully correct

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,183 @@
...
+            return None
+        if node.op_type in WEIGHT_CONSUMING_NODES:
+            if onnx_graph.has_initializer(node.input[weight_index]):
+                return node.input[weight_index]
+            return None
+        queue.extend(onnx_graph.get_parent(node, 0))
+    return None
+
+
+def _get_layer_name(weight_edge_names: List[str]) -> str:
```
kshpv, line 116: Want to discuss the layer name attribute of NNCFNode in case the node has several shared weights.

nncf/onnx/graph/metatypes/onnx_metatypes.py
kshpv: To be consistent that we use ONNXGraph as an interface to interact with ONNX model

nncf/onnx/graph/onnx_graph.py
kshpv: Yes, good suggestion

nncf/onnx/graph/metatypes/onnx_metatypes.py
alexsu52: Then the name of the `ONNXGraph` abstraction confuses me.

nncf/quantization/algorithms/bias_correction/onnx_backend.py
```diff
@@ -114,12 +114,13 @@ def get_output_name(model: onnx.ModelProto, node_name: str) -> List[str]:
...
-        if not hasattr(node.metatype, "weight_definitions"):
-            return False
-        weight_port_id = node.metatype.weight_definitions.weight_port_id
-        weight_node = input_nodes[weight_port_id]
-        return weight_node.metatype == ONNXDequantizeLinearMetatype
+        weight_port_ids = node.metatype.weight_port_ids
+        for weight_port_id in weight_port_ids:
+            weight_node = input_nodes[weight_port_id]
+            is_quantized_weights = is_quantized_weights or weight_node.metatype == ONNXDequantizeLinearMetatype
+        return is_quantized_weights
```
l-bat, line 123: I think we should move implementation to `nncf/onnx/graph/node_utils.py` and reuse it in BC and FBC algorithms

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,183 @@
...
+        + ONNXIdentityMetatype.get_all_aliases()
+        + ONNXReshapeMetatype.get_all_aliases()
+        + ONNXTransposeMetatype.get_all_aliases()
+        + ONNXQuantizeLinearMetatype.get_all_aliases()
+    )
+    ALLOWED_OP_TYPES = WEIGHT_CONSUMING_NODES + ONNXDequantizeLinearMetatype.get_all_aliases()
+
+    weight_index = 0
+    if onnx_graph.has_initializer(node.input[port_id]):
+        return node.input[port_id]
```
l-bat, line 102: should be updated after https://github.com/openvinotoolkit/nncf/pull/1940

nncf/onnx/graph/nncf_graph_builder.py
kshpv: Updated

nncf/onnx/graph/nncf_graph_builder.py
kshpv: Done

nncf/quantization/algorithms/bias_correction/onnx_backend.py
kshpv: Done

tests/onnx/quantization/test_ptq_params.py
```diff
@@ -78,7 +78,7 @@ def target_point(self, target_type: TargetType, target_node_name: str, port_id:
     def metatypes_mapping(self):
         return {
             Conv2dTestMetatype: ONNXConvolutionMetatype,
-            LinearTestMetatype: ONNXLinearMetatype,
+            LinearTestMetatype: MATMUL_METATYPES[0],
```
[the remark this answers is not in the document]
kshpv: Done

nncf/onnx/graph/nncf_graph_builder.py
kshpv: Done

kshpv:
E2E ONNX build - 376
Conformance build ONNX - 105

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,167 @@
...
+    def __init__(
+        self,
+        weight_attrs: Optional[Dict[int, Dict]] = None,
+        bias_attrs: Optional[Dict[str, Any]] = None,
+        node_attrs: Optional[Dict[str, Any]] = None,
+    ):
+        self.weight_attrs = weight_attrs if weight_attrs is not None else {}
+        self.bias_attrs = bias_attrs if bias_attrs is not None else {}
+        self.node_attrs = node_attrs if node_attrs is not None else {}
+
```
l-bat, line 53: As far as I remember it was fixed in OV backend
```suggestion
        self.weight_attrs = weight_attrs
        self.bias_attrs = bias_attrs
        self.node_attrs = node_attrs

```

l-bat (approved):

nncf/onnx/graph/metatypes/onnx_metatypes.py
kshpv: We have a ticket on removing ONNXGraph - 96982

kshpv:
E2E ONNX build - 380
Conformance build - 34

l-bat (approved):

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -153,7 +155,17 @@ def _get_axis(
...
-        return node.metatype.weight_definitions.weight_channel_axis
+
+        weight_channel_axis = node.metatype.weight_channel_axis
+        if node.layer_attributes.has_node_attrs():
+            if node.metatype == ONNXGemmMetatype:
+                weight_shape = node.layer_attributes.weight_attrs[target_point.port_id]["shape"]
+                if target_point.port_id == 0 and node.layer_attributes.node_attrs["transA"] == 1:
+                    weight_channel_axis = 1 - weight_channel_axis
+                elif target_point.port_id == 1 and node.layer_attributes.node_attrs["transB"] == 1:
+                    weight_channel_axis = 1 - weight_channel_axis
```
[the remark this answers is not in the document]
kshpv: Done

kshpv:
run pylint pre-commit tests

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+    """
+    Returns bias tensor attributes.
+
+    :param node: ONNX node.
+    :param onnx_graph: ONNXGraph.
+    :return: Bias tensor attributes.
+    """
+    bias_attrs = {}
+    metatype = get_metatype(onnx_graph, node)
+    if _is_node_with_bias(node, onnx_graph):
```
KodiaqQ, line 208: Why don't we have this check outside the method? I mean, for each node, even without bias, it would contain some `bias_attrs`.

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+        bias_tensor_port_id = get_bias_tensor_port_id(metatype)
+        bias_edge_name = onnx_graph.get_node_edge_names(node.name)["input"][bias_tensor_port_id]
+        edge = onnx_graph.get_edge(bias_edge_name)
+        bias_attrs["name"] = edge.name
```
KodiaqQ, line 212: Why can't we just have this attribute with the `weight_attr`? It contains only `name` inside. Also, weights can exist without biases, but biases can't exist without weights, as I know.

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -153,26 +344,28 @@ def create_nncf_graph(onnx_model: onnx.ModelProto) -> NNCFGraph:
...
-                layer_attributes = ONNXExtendedLayerAttributes(node.input, node.output, weight_shape)
-            else:
-                is_shared, weight_edge_name, layer_attributes = None, None, None
+            metatype = get_metatype(onnx_graph, node)
+            port_ids = _get_weight_port_ids(node, onnx_graph)
+            is_shared = None
+            weight_attrs = {}
+            node_attrs = _get_node_attrs(node, onnx_graph)
+            bias_attrs = _get_bias_attr(node, onnx_graph)
+            if port_ids:  # If node has weight
```
KodiaqQ, line 353: Let's call it `weight_port_ids`, because `port_ids` also exists for nodes with activations, I would say.

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -153,26 +344,28 @@ def create_nncf_graph(onnx_model: onnx.ModelProto) -> NNCFGraph:
...
+                weight_edge_names = []
+                for port_id in port_ids:
+                    weight_edge_names.append(_get_weight_edge_name(onnx_graph, node, port_id))
+                    weight_attrs.update(_get_weight_attr(node, onnx_graph, port_id))
+                    if not is_shared and onnx_graph.is_node_has_shared_weight(node, port_id):
+                        is_shared = True
+
+            layer_attributes = ONNXLayerAttributes(
```
KodiaqQ, line 361: The same comment about having `ONNXLayerAttributes` for each node. Not all nodes have `weight_attrs` and `bias_attrs`.

nncf/onnx/graph/node_utils.py
```diff
@@ -48,12 +44,11 @@ def get_bias_value(node_with_bias: NNCFNode, model: onnx.ModelProto) -> np.ndarr
...
     onnx_graph = ONNXGraph(model)
-    onnx_node = onnx_graph.get_node_by_name(node_with_bias.node_name)
-    bias_port_id = onnx_graph.get_bias_tensor_port_id(onnx_node)
-    bias_input_name = onnx_node.input[bias_port_id]
-    if onnx_graph.has_tensor(bias_input_name):
-        return onnx_graph.get_tensor_value(bias_input_name)
-    node = onnx_graph.get_nodes_by_output(bias_input_name)[0]
+    assert node_with_bias.layer_attributes.has_bias()
+    bias_name = node_with_bias.layer_attributes.bias_attrs["name"]
+    if onnx_graph.has_tensor(bias_name):
```
KodiaqQ, line 49: What if it hasn't this tensor? Is it saying about incorrect `NNCFGraph`?

nncf/onnx/graph/node_utils.py
```diff
@@ -93,3 +88,34 @@ def get_input_edge(input_node_name: str, input_edges_mapping: Dict[str, Tuple[st
...
+    """
+    Returns True if any weight port id of node is quantized,
+    False - if all weights are not quantized or the node can not have weight.
+
+    :param node: NNCFNode.
+    :param nncf_graph: NNCGraph.
+    :return: True if any weight port id of node is quantized,
+    False - if all weights are not quantized or the node can not have weight.
+    """
+    is_quanitzed_weight = False
```
KodiaqQ, line 103:
```suggestion
    is_quantized_weight = False
```

nncf/onnx/graph/node_utils.py
```diff
@@ -93,3 +88,34 @@ def get_input_edge(input_node_name: str, input_edges_mapping: Dict[str, Tuple[st
         input_edges.add(onnx_graph.get_node_edge_names(name)["input"][port_id])
     assert len(input_edges) == 1
     return input_edges.pop()
+
+
+def is_any_weight_quantized(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
```
KodiaqQ, line 93: IMHO, it's better to call this method `is_weights_quantized` since `any` adds some kind of uncertainty.

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -155,7 +157,20 @@ def _get_axis(
         if not target_point.is_weight_target_point():
             return 1
         node = nncf_graph.get_node_by_name(target_point.target_node_name)
-        return node.metatype.weight_definitions.weight_channel_axis
+
+        weight_channel_axis = node.metatype.weight_channel_axis
+        if node.layer_attributes.has_node_attrs():
+            if node.metatype == ONNXGemmMetatype:
```
KodiaqQ, line 163: Why should the `MinMaxAlgoBackend` calculate this? Why not calculate it while building `NNCFGraph`?

tests/onnx/data/reference_graphs/quantization/synthetic/gemm_weight_transpose_model.dot
```diff
@@ -0,0 +1,19 @@
...
+"1 DequantizeLinear_X_1" [id=1, type=DequantizeLinear];
+"2 Identity" [id=2, type=Identity];
+"3 QuantizeLinear_W_1" [id=3, type=QuantizeLinear];
+"4 DequantizeLinear_W_1" [id=4, type=DequantizeLinear];
+"5 Gemm" [id=5, type=Gemm];
+"6 Softmax" [id=6, type=Softmax];
+"7 nncf_model_input_0" [id=7, type=nncf_model_input];
+"8 nncf_model_output_0" [id=8, type=nncf_model_output];
+"0 QuantizeLinear_X_1" -> "1 DequantizeLinear_X_1"  [label="[1, 10]", style=dashed];
+"1 DequantizeLinear_X_1" -> "2 Identity"  [label="[1, 10]", style=solid];
```
KodiaqQ, line 12: Is the `Identity -> Gemm` the combined node? I mean, why do we place the `QuantizeLinear + DequantizeLinear` pair before it, not after?

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -99,17 +98,33 @@ class ONNXConvolutionTransposeMetatype(ONNXOpWithWeightsMetatype):
...
 @ONNX_OPERATION_METATYPES.register()
-class ONNXLinearMetatype(ONNXOpWithWeightsMetatype):
-    name = "LinearOp"
+class ONNXGemmMetatype(ONNXOpWithWeightsMetatype):
+    name = "GemmOp"
     op_names = ["Gemm"]
     hw_config_names = [HWConfigOpName.MATMUL]
-    # TODO(kshpv): ticket:95156
-    weight_definitions = OpWeightDef(weight_channel_axis=0, weight_port_id=1, bias_port_id=2)
+    weight_channel_axis = -1
```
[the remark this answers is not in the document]
kshpv: In metatype `weight_channel_axis` is not changeable.
Transpose attribute influences only in choosing an axis for quantization

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -99,17 +98,33 @@ class ONNXConvolutionTransposeMetatype(ONNXOpWithWeightsMetatype):
...
+    weight_port_ids = None
+    bias_port_id = 2
```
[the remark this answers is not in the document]
kshpv: Now they are not used for FBC and BC. As far as I remember there was a todo to include all nodes with biases for FBC and BC. Don't you remember?

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -593,20 +604,72 @@ def get_operator_metatypes() -> List[Type[OperatorMetatype]]:
...
+
+    :param metatype: Metatype.
+    :return: Port ids.
+    """
+    if metatype in CONSTANT_WEIGHT_LAYER_METATYPES:
+        return metatype.weight_port_ids
+    return []
+
+
+def get_possible_weight_port_ids(metatype: ONNXOpMetatype) -> List[int]:
```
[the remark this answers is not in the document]
kshpv: They do different things. `get_possible_weight_port_ids` returns ports which we are using to dynamically determine whether node has weights.

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+
+    if node.op_type not in PROPAGATING_ONLY_TYPES and onnx_graph.has_tensor(node.input[port_id]):
+        if node.op_type in ONNXReshapeMetatype.get_all_aliases():
+            # Only Reshape nodes returns output edge.
+            return node.output[0]
+        return node.input[port_id]
+
+    parent = onnx_graph.get_parent(node, port_id)
+    if parent and parent.op_type in (WEIGHT_CONSUMING_TYPES + PROPAGATING_ONLY_TYPES):
+        return _get_weight_edge_name(onnx_graph, parent, 0)
```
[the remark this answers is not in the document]
kshpv: This function returns a name of a tensor which is a node's weight. 
There are nodes that can consume weight tensor on 0-input port:
```
        ONNXConstantMetatype
        ONNXIdentityMetatype
        ONNXReshapeMetatype
        ONNXTransposeMetatype
        ONNXQuantizeLinearMetatype
```
There are two actions:
1) Getting weight tensor
2) Propagating

There are two node types:
1) Node that can have weight tensor and can be propagated further.
2) Node that is only for propagation.

If the node in the type 1, then we try to get weight tensor from it. If there is no tensor, we take a parent on that input port and if the type lays in these two, do the same for this node.
If the node in the type 2, we take a parent on that input port and if the type lays in these two, do the same for this node.

> Also, It's not clear, why we always walk through 0-port.

Because the nodes that consume weight tensor they must consume it only on 0 input port

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+    :param onnx_graph: ONNXGraph.
+    :return: Port ids with weights.
+    """
+    port_ids = set()
+    metatype = get_metatype(onnx_graph, node)
+    constant_port_ids = get_constant_weight_port_ids(metatype)
+    port_ids.update(constant_port_ids)
+    possible_port_ids = get_possible_weight_port_ids(metatype)
+    for port_id in possible_port_ids:
+        if _get_weight_edge_name(onnx_graph, node, port_id):
```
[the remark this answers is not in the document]
kshpv: This check needs to understand whether the node has weight on this port or its None meaning there is an activation tensor

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+    :param node: ONNX node.
+    :param onnx_graph: ONNXGraph.
+    :param weight_port_ids: Port ids with weights location.
+    :return: Weight attributes.
+    """
+    weight_attrs = {}
+    weight_edge_name = _get_weight_edge_name(onnx_graph, node, weight_port_id)
+    edge = onnx_graph.get_edge(weight_edge_name)
+    weight_shape = ONNXGraph.get_edge_shape(edge)
+    weight_attrs[weight_port_id] = {"name": weight_edge_name, "shape": weight_shape}
```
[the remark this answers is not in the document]
kshpv: This is the same as OV does. I assume we should do that when we will unite LayerAttributes for backends.

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+    return weight_attrs
+
+
+def _get_gemm_attrs(node: onnx.NodeProto) -> Dict[str, int]:
```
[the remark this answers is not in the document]
kshpv: This particular function returns `Dict[str, int]`

alexsu52 (approved):
LGTM

nncf/onnx/graph/nncf_graph_builder.py
kshpv: There is a possibility that some nodes (MatMul, Gemm) don't have weights but have biases

nncf/onnx/graph/nncf_graph_builder.py
kshpv: If nodes have no weights or no biases these attrs will be an empty `Dict`'s

nncf/onnx/graph/node_utils.py
kshpv: This function relies that `NNCFGraph` is correct

nncf/onnx/graph/node_utils.py
kshpv: `any` adds understanding that if there are different weights it is enough to only one of them to be quantized to return True.

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: Because NNCFGraph is used not only for MinMax

tests/onnx/data/reference_graphs/quantization/synthetic/gemm_weight_transpose_model.dot
kshpv: We place QDQ on the edge where the weight is. This is because of propagation till the weigth tensor

nncf/onnx/graph/nncf_graph_builder.py
kshpv: And this is the same for `node_attrs`. The logic is that if it is an empty `Dict` then there is no such attribute.

nncf/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
kshpv: Done

nncf/onnx/graph/nncf_graph_builder.py
kshpv: Done

nncf/onnx/graph/node_utils.py
kshpv: Applied

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+    """
+    metatype = get_metatype(onnx_graph, node)
+    if metatype in OPERATIONS_WITH_BIAS_METATYPES:
+        bias_tensor_port_id = get_bias_tensor_port_id(metatype)
+        if len(node.input) > bias_tensor_port_id:
+            return True
+    return False
+
+
+def _get_weight_attr(node: onnx.NodeProto, onnx_graph: ONNXGraph, weight_port_id: int) -> Dict[str, str]:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -21,11 +21,198 @@
...
+def _is_node_with_bias(node: onnx.NodeProto, onnx_graph: ONNXGraph) -> bool:
+    """
+    Returns True if node has bias tensor, otherwise - False.
+
+    :param node: ONNX node.
+    :param onnx_graph: ONNXGraph.
+    :return: True if node has bias tensor, otherwise - False.
+    """
+    metatype = get_metatype(onnx_graph, node)
+    if metatype in OPERATIONS_WITH_BIAS_METATYPES:
```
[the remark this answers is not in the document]
kshpv: Done

tests/onnx/data/reference_graphs/quantization/synthetic/gemm_weight_transpose_model.dot
kshpv: Here the weight is an input of `Identity`

nncf/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
KodiaqQ: For which purposes is `weight_channel_axis` used? I mean, it's more efficient to calculate the correct `weight_channel_axis` for each `transpose` position while building `NNCFGraph`, cause it affects correct statistics collection for weights optimization.

nncf/onnx/graph/node_utils.py
KodiaqQ: Not agree.

tests/onnx/data/reference_graphs/quantization/synthetic/gemm_weight_transpose_model.dot
KodiaqQ: Not sure that it will correctly work for the `Reshape` layer, for example.
IMHO, it's better when we put the `QDQ` pair right before the `Convolution`/`MatMul`/etc.

nncf/onnx/graph/nncf_graph_builder.py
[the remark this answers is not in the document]
kshpv: In general - yes. But we seek for an attribute which can be only `int`

nncf/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
kshpv: Ok, what do you suggest?

nncf/onnx/graph/nncf_graph_builder.py
kshpv: Yes, it is.
Sorry this is not applicable for MatMul, only for Gemm.
https://onnx.ai/onnx/operators/onnx__Gemm.html

tests/onnx/data/reference_graphs/quantization/synthetic/gemm_weight_transpose_model.dot
kshpv: Reshape layers are a special case when we return output. Please, take a look at the examples in the tests.

kshpv: I am not sure whether we can apply weight compression to int8 in your case

nncf/onnx/graph/node_utils.py
kshpv: True, I will remove this check

nncf/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
KodiaqQ: Highlight it in the ticket (maybe about refactoring), we are able now to correct biases for `MatMul` layers. And extend the FBC/BC algorithms as soon as possible.

nncf/quantization/algorithms/min_max/onnx_backend.py
KodiaqQ: I don't see any problem to calculate it for `NNCFGraph` while building. For example, the `SmoothQuant` algorithm also uses information about the `weight_channel_axis`.

nncf/onnx/graph/nncf_graph_builder.py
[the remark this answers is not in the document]
kshpv: I don't see any problems to have a method that can return empty data meaning that this node has no such attribute.

nncf/onnx/graph/metatypes/onnx_metatypes.py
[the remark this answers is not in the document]
kshpv: I got your point

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: This is the same as it is done in OpenVINO. I aligned with this logic and don't see any problem with that.
If you thnk that it should be redone you concern only about ONNX?

nncf/onnx/graph/node_utils.py
kshpv: Done

andrey-churkin (approved):

KodiaqQ (approved):
With the notice that Gemm layer have no biases, but it was added.

PR #1920, opened 2023-06-15
Add parameters `strict_check_scopes`

AlexanderDokuchaev:
### Changes

- Add parameters to NNCFConfig `validate_scopes` to disable raise runtime error on existing not matched to model.
- Add `validate` attribute to `IgnoredScope`.
- Disable `validate_scopes` for ModelType.TRANSFORMER.
- spellcheck

### Ticket 
110630

### Tests
test_raise_runtimeerror_for_not_matched_scope_names
TestPTQParams::test_strict_check_ignored_scope


nncf/tensorflow/quantization/quantize_model.py
```diff
@@ -126,6 +126,8 @@ def _create_nncf_config(
     if advanced_parameters is not None:
         compression_config = apply_advanced_parameters_to_config(compression_config, advanced_parameters)
 
+    compression_config["strict_check_scopes"] = False
```
alexsu52, line 129: I would suggest change behavior of `nncf.quantize()` for `torch` and `tensorflow` backends more granularity:
- `if model_type==None then compression_config["strict_check_scopes"] = True`
- `if model_type==ModelType.TRANSFORMER then compression_config["strict_check_scopes"] = False`, because the algorithm specified additional `IgnoredScope`.

In this case we will obtain more align behavior with other backends.

alexsu52 (commented):
Thanks for the contribution!

nncf/tensorflow/quantization/quantize_model.py
AlexanderDokuchaev: Update, "strict_check_scopes"= False only for ModelType.TRANSFORMER.

AlexKoff88: `strict_check_scopes` -> `validate_scopes` if it work for both ignored and target scopes. Otherwise, it is better to call it `validate_ignored_scopes`

nncf/quantization/advanced_parameters.py
```diff
@@ -147,6 +150,7 @@ class AdvancedQuantizationParameters:
     quantize_outputs: bool = False
     inplace_statistics: bool = True
     disable_bias_correction: bool = False
+    strict_check_scopes: bool = True
```
alexsu52, line 153: I suggest to add strict parameter in the IgnoredScope class as following: 

```
@api(canonical_alias="nncf.IgnoredScope")
@dataclass
class IgnoredScope:
    """
    Provides an option to specify portions of model to be excluded from compression.

    The ignored scope defines model sub-graphs that should be excluded from the compression process such as
    quantization, pruning and etc.

    Example:

    ..  code-block:: python

            import nncf

            # Exclude by node name:
            node_names = ['node_1', 'node_2', 'node_3']
            ignored_scope = nncf.IgnoredScope(names=node_names)

            # Exclude using regular expressions:
            patterns = ['node_\\d']
            ignored_scope = nncf.IgnoredScope(patterns=patterns)

            # Exclude by operation type:

            # OpenVINO opset https://docs.openvino.ai/latest/openvino_docs_ops_opset.html
            operation_types = ['Multiply', 'GroupConvolution', 'Interpolate']
            ignored_scope = nncf.IgnoredScope(types=operation_types)

            # ONNX opset https://github.com/onnx/onnx/blob/main/docs/Operators.md
            operation_types = ['Mul', 'Conv', 'Resize']
            ignored_scope = nncf.IgnoredScope(types=operation_types)

    **Note:** Operation types must be specified according to the model framework.

    :param names: List of ignored node names.
    :type names: List[str]
    :param patterns: List of regular expressions that define patterns for names of ignored nodes.
    :type patterns: List[str]
    :param types: List of ignored operation types.
    :type types: List[str]
    :param strict: If set to True, then a RuntimeError will be raised if any ignored scope does not match in the model graph.
    :type types: List[str]
    """

    names: List[str] = field(default_factory=list)
    patterns: List[str] = field(default_factory=list)
    types: List[str] = field(default_factory=list)
    strict: bool = True
```

alexsu52 (changes requested):

AlexanderDokuchaev:
run pytorch pre-commit tests

AlexKoff88 (approved):

nncf/quantization/advanced_parameters.py
```diff
@@ -147,6 +150,7 @@ class AdvancedQuantizationParameters:
     quantize_outputs: bool = False
     inplace_statistics: bool = True
     disable_bias_correction: bool = False
+    validate_scopes: bool = True
```
alexsu52, line 153: Please, remove these parameter.

AlexanderDokuchaev: done

nncf/torch/quantization/quantize_model.py
```diff
@@ -169,10 +169,14 @@ def _create_nncf_config(
             compression_config["ignored_scopes"].extend(_ignored_scope)
         else:
             compression_config["ignored_scopes"] = _ignored_scope
+        compression_config["validate_scopes"] = ignored_scope.validate
```
alexsu52, line 172: It looks like the same changes is needed to the https://github.com/openvinotoolkit/nncf/blob/444ca7da363d8dac82641a0e7851b7b8e7e2e2f4/nncf/tensorflow/quantization/quantize_model.py#L91

AlexanderDokuchaev: Yeap, added

alexsu52 (approved):
LGTM

PR #1931, opened 2023-06-22
Remove CUDA spec from torch installation in tests

vshampor:
### Changes
As stated in the title

### Reason for changes
TPS installation tests cannot find the torch version with the CUDA-specific index URL. Also removed the associated test parameter since we don't need this kind of disambiguation for supported torch versions anymore.

### Related tickets
N/A

### Tests
third_party_sanity run


vshampor:
TPS build 30 shows that the original issue is gone. Another, known failure is now visible, which will be covered in a separate PR.

PR #1932, opened 2023-06-22
Dispatch bfloat along with the rest of the types for the kernels

vshampor:
### Changes
As stated in the title

### Reason for changes
Otherwise the autocast to BF16 does not work with quantization.

### Related tickets
N/A

### Tests
test_external_autocast on CPU-only hosts

### DIFF
--- a/nncf/torch/extensions/include/common_cpu_funcs.h
+++ b/nncf/torch/extensions/include/common_cpu_funcs.h
@@ -2,6 +2,7 @@
 #define _COMMON_CPU_FUNCS_H_
 
 #include <torch/torch.h>
+#include "dispatch.h"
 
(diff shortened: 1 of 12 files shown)


vshampor:
Quantization function benchmark results TBA

vshampor:
CUDA kernels measured to have an average of 3.4% perf boost as a result of this change, CPU kernels have a 0.89% perf degradation. Safe to say that the performance is largely unchanged.

alexsu52 (commented):
Thanks for the contribution. Could you provide sota_eval test results?

vshampor:
sota_pytorch_eval run 411 shows no significant changes when compared to the nightly run from develop.

alexsu52 (approved):
LGTM

vshampor:
run pytorch pre-commit tests

PR #1945, opened 2023-07-03
Skip smooth quant algorithm by default

daniil-lyakhov:
### Changes

Disable SmoothQuant by default

### Reason for changes

Current implementation of SmoothQuant brings accuracy degradation for transformer models from conformance tests


daniil-lyakhov:
Manual conformance test 102 showed accuracy for transformer models is recovered

alexsu52 (approved):

PR #1948, opened 2023-07-04
[POC] Add tensor operators

AlexanderDokuchaev:


nncf/common/tensor_numpy.py
```diff
@@ -0,0 +1,56 @@
...
+
+import numpy as np
+
+from nncf.common.tensor import NNCFTensorExt
+
+TensorType = TypeVar("TensorType")
+DeviceType = TypeVar("DeviceType")
+
+
+class NUMPYNNCFTensor(NNCFTensorExt):
```
alexsu52, line 22: The major comment form my side:
- nncf.Tensor should be like a torch.Tensor or tf.Tensor.
- nncf.Tensor should be provided semiless integration torch.Tensor, tf.Tensor and numpy array.
- nncf.Tensor should be easily extended to another data representation.
- math functions should be provided for nncf.Tensor

I would recommend to think how to encapsulate all logic work with different data representation into nncf.Tensor abstraction.

alexsu52 (changes requested):

PR #1952, opened 2023-07-07
Added redefined __class__ method for ProxyModule.

andreyanufr:
### Changes

Redefined __class__ in ProxyModule

### Reason for changes

If wrapped (user defined module added with "nncf.torch.register_module") module call super() in forward method then this change fix problem "super(type, obj): obj must be an instance or subtype of type".
For example: https://github.com/mosaicml/llm-foundry/blob/ef350d9e64d13cb1db35ab7941bf9039b1b499fd/llmfoundry/models/layers/custom_embedding.py#L15
### Tests
No tests.

### DIFF
--- a/nncf/torch/layer_utils.py
+++ b/nncf/torch/layer_utils.py
@@ -24,6 +24,9 @@ class ProxyModule:
     def __getattr__(self, name):
         return getattr(self._module, name)
 
+    def __class__(self):
+        return type(self._module)
+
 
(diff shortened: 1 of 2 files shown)


alexsu52 (commented):
Please add a unit test for these changes.

alexsu52 (approved):
LGTM

ljaljushkin (approved):

PR #1955, opened 2023-07-07
Add strict parameter for find_matching_subgraphs

kshpv:
### Changes

Add strict with default value True parameter to find_matching_subgraphs.

### Reason for changes

IgnoredPatterns shouldn't be checked on strict matches. This is specific only for HWFusedPatterns.

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -711,7 +711,7 @@ class NNCFGraph:
         for node_key, node in self._nx_graph.nodes.items():
             self._node_id_to_key_dict[node["id"]] = node_key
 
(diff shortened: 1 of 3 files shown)


alexsu52 (commented):
Please, check this changes on the wide models scope.

kshpv:
run pytorch pre-commit tests

kshpv:
> Please, check this changes on the wide models scope.

Checked on model cache ww27_static. The models that have ignored patterns on `develop` and `this PR` are the same. Moreover, the matched ignored subgraphs are the same.

alexsu52 (approved):

PR #1962, opened 2023-07-10
Use nncf_graph for FBC

AlexanderDokuchaev:
### Changes

- Use compressed graph for torch backend
- Add `with_bias` for `WeightedLayerAttributes` what means that layer contain bias inside module (will true only for torch and tf)
- Removed `bias` from `LinearLayerAttributes`, use `with_bias`.
- Remove `model` from functions backend of `fast_bias_correction`
 
### Related tickets
111576

### DIFF
--- a/nncf/torch/model_transformer.py
+++ b/nncf/torch/model_transformer.py
@@ -139,7 +139,8 @@ def update_fused_bias(target_node_name: str, new_bias: Tensor, model: NNCFNetwor
     :param new_bias: New bias value.
     :param model: The model.
     """
-    fused_node = get_potential_fused_node(target_node_name, model)
+    nncf_graph = model.get_graph()
+    fused_node = get_potential_fused_node(target_node_name, nncf_graph)
     if fused_node:
         target_node_name = fused_node.node_name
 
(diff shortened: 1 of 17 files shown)


AlexanderDokuchaev:
job post_training_quantization: 113

nncf/torch/model_transformer.py
```diff
@@ -139,7 +139,8 @@ def update_fused_bias(target_node_name: str, new_bias: Tensor, model: NNCFNetwor
     :param new_bias: New bias value.
     :param model: The model.
     """
-    fused_node = get_potential_fused_node(target_node_name, model)
+    nncf_graph = model.get_graph()
```
KodiaqQ, line 142: Wouldn't it be very expensive to `get_graph` for every `update_fused_bias` call?

nncf/torch/quantization/quantize_functions.py
```diff
@@ -20,6 +20,8 @@
 from nncf.torch.quantization.extensions import QuantizedFunctionsCUDA
 from nncf.torch.utils import add_domain
 
+QUANTIZE_OPERATION_NAMES = ["symmetric_quantize", "asymmetric_quantize"]
```
KodiaqQ, line 23: Why is it here? Where is it used?

nncf/torch/model_transformer.py
AlexanderDokuchaev: No, it's nncfnetwrok contain `_compressed_graph` to save graph, and does not build graph on each call. https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/nncf_network.py#L425

KodiaqQ (approved):

AlexanderDokuchaev:
run pylint pre-commit tests

alexsu52 (approved):
Cool staff! Thanks!

PR #1981, opened 2023-07-17
Speed-up calculation of FQ important scores

alexsu52:


nncf/data/dataset.py
```diff
@@ -73,6 +96,63 @@ def get_inference_data(self, indices: Optional[List[int]] = None) -> Iterable[Mo
         return DataProvider(self._data_source, self._transform_func, indices)
 
 
+class CountingDatasetWrapper(IDataset):
```
andrey-churkin, line 99: I think it is better to extend `Evaluator` class instead of creation wrapper here. I suggest to use this class
```python
class IterationCounter:
    def __init__(self, iterable):
        self._iterable = iterable
        self._num_iterations = 0

    @property
    def num_iterations(self) -> int:
        return self._num_iterations

    def __iter__(self):
        for x in self._iterable:
            self._num_iterations += 1
            yield x
```
inside `Evaluator` class.

Also add property to `Evaluator` class that returns number of passed iterations during last validation process. For example,
```python
@property
def num_passed_iterations(self) -> int:
    """
    Returns number of passed iterations during last validation process.
    """
```

nncf/quantization/algorithms/accuracy_control/algorithm.py
```diff
@@ -12,6 +12,8 @@
 import sys
 from typing import Any, Callable, Iterable, List, Optional, Tuple, TypeVar, Union
 
+from attr import dataclass
```
andrey-churkin, line 15:
```suggestion
from dataclasses import dataclass
```

nncf/openvino/engine.py
```diff
@@ -74,3 +70,43 @@ def infer(
...
+    ) -> Dict[str, np.ndarray]:
+        """
+        Runs model on the provided input via OpenVINO Runtime.
+        Returns the dictionary of model outputs by node names.
+
+        :param input_data: Inputs for the model.
+        :return output_data: Model's output.
+        """
+        if self.engine is None:
+            compiled_model = self._get_compiled_model()
```
andrey-churkin, line 110: Do we really need `_get_compiled_model()` method? Why it is better to use method instead of
```python
compiled_model = ov.Core().compile_model(self.model, self.target_device.value)
```

alexsu52: Can I assign the same question to you?

nncf/data/dataset.py
alexsu52: In this case, I need to enter a flag that will enable or disable the collection of the number of iterations passed. Do you still support your proposal?

nncf/common/utils/backend.py
```diff
@@ -22,58 +23,115 @@ class BackendType(Enum):
...
 
-def get_backend(model) -> BackendType:
+def get_available_backends() -> List[BackendType]:
     """
-    Returns the NNCF backend name string inferred from the type of the model object passed into this function.
+    Returns a list of available backends.
 
-    :param model: The framework-specific object representing the trainable model.
-    :return: A BackendType representing the correct NNCF backend to be used when working with the framework.
+    :return: A list of avauilable backends.
```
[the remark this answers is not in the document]
alexsu52: Done

nncf/common/utils/backend.py
```diff
@@ -22,58 +23,115 @@ class BackendType(Enum):
...
-        import openvino.runtime as ov
 
-        available_frameworks.append("OpenVINO")
-    except ImportError:
-        ov = None
+def get_backend(model: TModel) -> BackendType:
+    """
+    Returns the NNCF backend name string inferred from the type of the model object passed into this function.
+
+    :param model: The framework-specific object representing the trainable model.
```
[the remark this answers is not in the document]
alexsu52: Done

nncf/common/utils/os.py
```diff
@@ -37,3 +40,23 @@ def is_windows():
 
 def is_linux():
     return "linux" in sys.platform
+
+
+def available_cpu_count() -> int:
```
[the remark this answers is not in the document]
alexsu52: Done

nncf/quantization/algorithms/accuracy_control/algorithm.py
alexsu52: Done

nncf/common/utils/backend.py
```diff
@@ -22,58 +23,115 @@ class BackendType(Enum):
...
-        torch = None
+    return isinstance(model, ov.Model)
 
-    try:
-        import tensorflow
 
-        available_frameworks.append("Tensorflow")
-    except ImportError:
-        tensorflow = None
+def is_openvino_compiled_model(model: TModel):
```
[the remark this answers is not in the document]
alexsu52: Done

nncf/data/dataset.py
andrey-churkin: Why not collect them all the time?

nncf/openvino/engine.py
andrey-churkin: Sure! I don't see any reason why this method is required here. So I would like to understand your point of view. This is the reason why I asked the question.

nncf/data/dataset.py
alexsu52: It looks like it is not needed every time. I don't want to add overhead.

alexsu52: I have updated the PR, please take a look.

nncf/quantization/algorithms/accuracy_control/algorithm.py
```diff
@@ -128,6 +134,23 @@ def reverted_operations(self) -> List[NNCFNode]:
...
 
+@dataclass
+class MetricResults:
+    """
+    Results of metrics collection.
+
+    :param metric_value: Aggregated metric value.
+    :param values_for_each_item: Metric values for each data item.
+    :param preperation_time: Time that it takes to prepare model for validation.
+    :param validation_time: Time that it takes to validate model.
```
andrey-churkin, line 145: Could you please specify what is it milliseconds or seconds?

nncf/quantization/algorithms/accuracy_control/algorithm.py
```diff
@@ -221,19 +294,32 @@ def apply(
...
+        if num_ranking_processes is None:
+            model_size = algo_backend.get_model_size(quantized_model)
+            num_ranking_processes = self._calculate_number_ranker_parallel_proc(
+                model_size,
+                quantized_metric_results.preperation_time,
+                quantized_metric_results.validation_time,
+                validation_dataset_size,
+            )
+
+        nncf_logger.info(f"Number of parallel processes to rank quantized operations: {num_ranking_processes}")
```
andrey-churkin, line 308:
```suggestion
        nncf_logger.info(f"Number of parallel processes to rank quantization operations: {num_ranking_processes}")
```

andrey-churkin (approved):

PR #1989, opened 2023-07-19
Shift+Scale and Input+Shift+Scale pattern for PT

ljaljushkin:
### Changes

Introduced Shift + Scale fused pattern and Input+Shift+Scale pattern to insert Fake Quantize operations optimally for CPU.

### Reason for changes

Customer has a model that is quantized not optimally:
image

FakeQuantize between subtraction and division is redundant and introduces additional cost in runtime.
FakeQuantize between input and pre-processing is not needed in case of single edge from input, because pre-processing can be fused to the FQ after pre-processing.
image
When there are multiple edges from input and one edge is going to pre-processing, it's optimal to have a common fake quantize for all edges. 
image

If pre-processing represented via normalize op from torchvision (e.g. like here https://github.com/PeterL1n/RobustVideoMatting/blob/master/model/mobilenetv3.py#L37), NNCF doesn't insert FQ between subtraction and division and between input and pre-processing.
It happens because pre-processing is implemented via in-place operations, and since FQ is not in-place it can't be applied (see more details here: https://github.com/openvinotoolkit/nncf/pull/1565)
image

### Related tickets

112934

### Tests

synthetic tests for pre-processing

### DIFF
--- a/nncf/onnx/hardware/fused_patterns.py
+++ b/nncf/onnx/hardware/fused_patterns.py
@@ -44,6 +44,23 @@ def create_scale_shift() -> GraphPattern:
     return pattern
 
 
+@ONNX_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SHIFT_SCALE)
+def create_shift_scale() -> GraphPattern:
+    pattern = GraphPattern()
+    add_node = pattern.add_node(
+        **{
+            GraphPattern.LABEL_ATTR: "ADD, SUBTRACT",
+            GraphPattern.METATYPE_ATTR: [om.ONNXAddLayerMetatype, om.ONNXSubMetatype],
+        }
+    )
+    mul_node = pattern.add_node(
+        **{GraphPattern.LABEL_ATTR: "MULTIPLY", GraphPattern.METATYPE_ATTR: om.ONNXMulLayerMetatype}
+    )
+
+    pattern.add_edge(add_node, mul_node)
+    return pattern
+
+
 @ONNX_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SWISH_WITH_SIGMOID)
 def create_swish_with_sigmoid() -> GraphPattern:
     pattern = GraphPattern()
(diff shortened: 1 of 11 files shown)


nncf/openvino/hardware/fused_patterns.py
```diff
@@ -128,6 +128,22 @@ def create_scale_shift() -> GraphPattern:
     return pattern
 
 
+@OPENVINO_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SHIFT_SCALE)
```
ljaljushkin, line 131: @KodiaqQ please review OV part

nncf/onnx/hardware/fused_patterns.py
```diff
@@ -44,6 +44,28 @@ def create_scale_shift() -> GraphPattern:
     return pattern
 
 
+@ONNX_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SHIFT_SCALE)
```
ljaljushkin, line 47: @kshpv please review ONNX part

nncf/onnx/hardware/fused_patterns.py
```diff
@@ -44,6 +44,28 @@ def create_scale_shift() -> GraphPattern:
...
+def create_shift_scale() -> GraphPattern:
+    pattern = GraphPattern()
+    pattern_input_node = pattern.add_node(
```
kshpv, line 50: you don't need this node here

kshpv (changes requested):

nncf/onnx/hardware/fused_patterns.py
ljaljushkin: removed, thanks

nncf/onnx/hardware/fused_patterns.py
```diff
@@ -44,6 +44,23 @@ def create_scale_shift() -> GraphPattern:
...
+def create_shift_scale() -> GraphPattern:
+    pattern = GraphPattern()
+    add_node = pattern.add_node(
+        **{
+            GraphPattern.LABEL_ATTR: "ADD, SUBTRACT",
+            GraphPattern.METATYPE_ATTR: [om.ONNXAddLayerMetatype, om.ONNXSubMetatype],
+        }
+    )
+    mul_node = pattern.add_node(
+        **{GraphPattern.LABEL_ATTR: "MULTIPLY", GraphPattern.METATYPE_ATTR: om.ONNXMulLayerMetatype}
```
kshpv, line 57: Could be here also `Div` layer?

kshpv (changes requested):

KodiaqQ (approved):

nncf/onnx/hardware/fused_patterns.py
ljaljushkin: Extended list of types for the pattern

kshpv:
Thanks!

kshpv (approved):

alexsu52 (approved):
LGTM

PR #1973, opened 2023-07-13
[PTQ] Support for matmul [x,y]x[y]

alexsu52:
### Changes
Fix calculation channel axes for matmul operation [x,y]x[y]

### Reason for changes

Wrong channel axes calculation

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/1936

### Tests

test_get_weight_channel_axes_for_matmul

PR #1993, opened 2023-07-20
[ONNX] Fix calculation of reduction_shape when quantized tensor has 1 dim

kshpv:
### Changes

Fix the calculation of reduction_shape.

### Reason for changes

the same as https://github.com/openvinotoolkit/nncf/pull/1973

### Related tickets

N/A

### Tests

test_get_reduction_shape


nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -171,6 +171,8 @@ def _get_reduction_shape_and_use_abs_max(
         assert node.layer_attributes.has_weight()
         weight_shape = node.layer_attributes.weight_attrs[target_point.port_id]["shape"]
         reduction_shape = list(range(len(weight_shape)))
+        if len(reduction_shape) == 1:  # If only one channel
```
alexsu52, line 174: Please, add a test.

kshpv: Done

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -132,49 +135,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
         return get_input_edges_mapping(nncf_graph)
 
     @staticmethod
-    def _get_axis(
-        nncf_graph: NNCFGraph, target_point: ONNXTargetPoint, quantizer_config: QuantizerConfig
+    def _get_quantized_channel_axis(
```
KodiaqQ, line 130: What does this method do? Which axis would be used while conversion in case of a `None` result?

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -132,49 +135,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
...
         return weight_channel_axis
 
     @staticmethod
-    def _get_reduction_shape_and_use_abs_max(
-        nncf_graph: NNCFGraph, target_point: ONNXTargetPoint, quantizer_config: QuantizerConfig
-    ) -> Tuple[Optional[Tuple[int, ...]], bool]:
-        use_abs_max = quantizer_config.mode == QuantizationMode.SYMMETRIC
-        if not quantizer_config.per_channel:
-            return None, use_abs_max
+    def _get_reduction_shape_for_weight(
```
KodiaqQ, line 165: Wouldn't we need this method in `node_utils.py`? For the `ChannelAlignment` or `SmoothQuant` algorithms, for example.
Also, no description of method.

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -125,49 +127,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
...
-    ) -> Tuple[ReductionShape, bool]:
-        use_abs_max = quantizer_config.mode == QuantizationMode.SYMMETRIC
-        if not quantizer_config.per_channel:
-            return None, use_abs_max
+    def _get_reduction_shape_for_weight(
+        node: NNCFNode, weight_port_id: int, is_per_channel: bool
+    ) -> Optional[ReductionShape]:
+        if not is_per_channel:  # Per-Tensor
+            return None
+        weight_shape = node.layer_attributes.weight_attrs[weight_port_id]["shape"]
```
KodiaqQ, line 170: Are we sure that `layer_attributes` exists for every node?

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -125,49 +127,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
...
+            return tuple(reduction_shape)
+        axis = ONNXMinMaxAlgoBackend._get_weight_channel_axis(node, weight_port_id)
+        reduction_shape.pop(axis)
+        return tuple(reduction_shape)
 
-        if not target_point.is_weight_target_point():
-            # TODO: support reduction shapes for 3D-5D conv cases
-            return (0, 2, 3), use_abs_max
+    @staticmethod
+    def _get_reduction_shape_for_activation(
```
KodiaqQ, line 179: Missed description.

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -125,49 +127,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
...
+            if target_point.type == TargetType.PRE_LAYER_OPERATION:
+                nncf_logger.info(
+                    f"The shape of input edge of a node {node.node_name} is unkown. Therefore per-tensor quantizaiton is applied."
+                )
+            elif target_point.type == TargetType.POST_LAYER_OPERATION:
+                nncf_logger.info(
+                    f"The shape of output edge of a node {node.node_name} is unkown. Therefore per-tensor quantizaiton is applied."
+                )
+            nncf_logger.info(f"Please consider to run pre-processing before quantization.")
+            # TODO: add preprocessing tool for ONNX model.
```
KodiaqQ, line 201: Which tool needs to be added and why we don't add it now?

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -125,49 +127,85 @@ def _get_input_edges_mapping(nncf_graph: NNCFGraph):
...
+            return None
 
-        axis = ONNXMinMaxAlgoBackend._get_axis(nncf_graph, target_point, quantizer_config)
-        reduction_shape.pop(axis)
-        return tuple(reduction_shape), use_abs_max
+        # TODO (l-bat): Disable quantizer propogation through layout changing operations
+        channel_axis = 1  # Activations have channel first layout: [N, C, Z, Y, X]
```
KodiaqQ, line 205: Is it correct for `PRE_LAYER_OPERATION` and `POST_LAYER_OPERATION`? For example, for `MatMul` layers.
Why don't we use the `output_channel_axis` attribute, for example?

tests/onnx/quantization/test_min_max.py
```diff
@@ -0,0 +1,180 @@
...
+    ids=[test_case.nncf_node.node_name for test_case in test_cases],
+)
+def test_get_reduction_shape(test_case):
+    """Checks the correct return reduction shape in ONNXMinMaxAlgo.
+    Edge cases:
+    1) per-tensor.
+    2) transpose axis of GEMM node.
+    3) one dimensional weight tensor.
+    """
+    reduction_shape = ONNXMinMaxAlgoBackend()._get_reduction_shape_for_weight(
```
KodiaqQ, line 177: Where is the test for activation?

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: This tool is out-of-scope of this PR. This processing toll should be added in the following PR not to mix up things

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: For ONNX this is correct

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: Please take a look at the updated methods

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: All descriptions are filled now

nncf/quantization/algorithms/min_max/onnx_backend.py
kshpv: This is a copy past from OV - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/openvino_backend.py#L142

tests/onnx/quantization/test_min_max.py
kshpv: Test updated

KodiaqQ (approved):

PR #1995, opened 2023-07-22
Add torch.set_num_threads for ptq tests

AlexanderDokuchaev:
### Changes

Set number of thread for torch backends for `torch.set_num_threads(int(cpu_threads_num))` by set `CPU_THREADS_NUM` env variable.

Example of reducing quantization time for TORCH backend:

Model | Before | After  
--- | --- | --- 
deit3_small_patch16_224 | 2:34:09  | 0:00:46
dla34 | 0:16:38  | 0:00:38

Build: manual/post_training_quantization/122

Renamed tests to `test_ptq_quantization`

### Reason for changes

Quantization time is dramatically slow for TORCH backends on CI.

### DIFF
--- a/tests/post_training/pipelines/base.py
+++ b/tests/post_training/pipelines/base.py
@@ -8,6 +8,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
 import time
 from abc import ABC
 from abc import abstractmethod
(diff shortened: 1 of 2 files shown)


AlexanderDokuchaev:
run openvino pre-commit tests
run pylint pre-commit tests

AlexanderDokuchaev:
run openvino pre-commit tests

tests/post_training/pipelines/base.py
```diff
@@ -212,6 +213,12 @@ def quantize(self) -> None:
         Run quantization of the model and collect time and memory usage information.
         """
         print("Quantization...")
+
+        if self.backend in [BackendType.TORCH, BackendType.OLD_TORCH]:
+            cpu_threads_num = os.environ.get("CPU_THREADS_NUM")
```
alexsu52, line 218: What do you think about getting number of CPU processes using the following function https://github.com/openvinotoolkit/nncf/blob/ffa15d92032543afeac457376a4db120f8872f6e/nncf/common/utils/os.py#L45?

AlexanderDokuchaev: The root cause of slow quantization is limit of cpu usage by configuration of Kubernetes but torch try use all available threads. `get_available_cpu_count` returns the number of cores in the processor despite the config limits.

alexsu52 (approved):
LGTM

PR #2002, opened 2023-07-25
[PTQ] Update BC for Split nodes

KodiaqQ:
### Changes

- Updated BC algorithm pipeline for specific Split cases.

### Reason for changes

- Accuracy degradation
- Wrong statistics calculation

### Related tickets

- 111542

### Tests

- Conformance run - post_training_quantization/124/


alexsu52 (commented):
Does the conformance test include a model with split? I would recommend to add unit test for this changes as well.

KodiaqQ:
> Does the conformance test include a model with split? I would recommend to add unit test for this changes as well.

It's not included. Do we really need such a huge model in the conformance?

alexsu52:
> > Does the conformance test include a model with split? I would recommend to add unit test for this changes as well.
> 
> It's not included. Do we really need such a huge model in the conformance?

I think, the test will be enough.

alexsu52 (approved):
LGTM

KodiaqQ:
Waiting for updates from @kshpv, for ONNX tests.

PR #2016, opened 2023-08-01
Use new property for telemetry and fix to actually send events

vshampor:
### Changes
Use a new measurement ID for `openvino_telemetry` instead of the old one. Fixed the `skip_if_raises` wrapper which silently broke down the telemetry measurements all this time since #1493.

### Reason for changes
The old property has been deprecated by the provider and will no longer work. 

### Related tickets

N/A

### Tests

test_telemetry, test_skip_if_raises

### DIFF
--- a/nncf/telemetry/wrapper.py
+++ b/nncf/telemetry/wrapper.py
@@ -13,7 +13,7 @@ import os
 import sys
 from abc import ABC
 from abc import abstractmethod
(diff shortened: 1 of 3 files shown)


alexsu52 (approved):
LGTM

KodiaqQ (approved):

PR #2017, opened 2023-08-01
[ONNX] Fix bug in Constant subgraph traversing

kshpv:
### Changes

Fix traversing 

### Reason for changes

Bug in test for BiasCorrection - https://github.com/openvinotoolkit/nncf/blob/develop/tests/onnx/quantization/test_bias_correction.py#L87

### Related tickets

N/A

### Tests

Added new test cases


tests/onnx/data/reference_graphs/quantization/synthetic/weight_propagation_conv_model.dot
```diff
@@ -0,0 +1,49 @@
+strict digraph  {
+"0 reshape" [id=0, type=Reshape];
+"1 transpose" [id=1, type=Transpose];
+"2 identity1" [id=2, type=Identity];
+"3 QuantizeLinear_input_1" [id=3, type=QuantizeLinear];
+"4 DequantizeLinear_input_1" [id=4, type=DequantizeLinear];
+"5 QuantizeLinear_identity_output1_1" [id=5, type=QuantizeLinear];
```
KodiaqQ, line 7: The same question is about propagation through layers. The so-called "constant" subgraph starts with the `Reshape`, not the `Constant` node as I can see. Thus, propagation is needed. Or the graph is invalid.

KodiaqQ (approved):

PR #2020, opened 2023-08-01
[PTQ] Pass that allows quantize inputs for special nodes

KodiaqQ:
### Changes

- Added a new step to the `transform_to_inference_graph` pass

### Reason for changes

- Correct quantization scheme for special nodes

### Related tickets

- 113396

### Tests

- tests/common/quantization/test_passes.py

Graph before update:
image
Graph after update:
image


alexsu52 (commented):
What do you think about insertion const_subgraph node as noop for all weights operations?

nncf/quantization/passes.py
```diff
@@ -24,6 +26,7 @@ def transform_to_inference_graph(
     nncf_graph: NNCFGraph,
     shapeof_metatypes: List[OperatorMetatype],
     read_variable_metatypes: Optional[List[OperatorMetatype]] = None,
+    metatypes_to_insert_noop: Optional[List[OperatorMetatype]] = None,
```
kshpv, line 29: There is a need to update the docstring accordingly.

nncf/quantization/passes.py
```diff
@@ -34,8 +37,10 @@ def transform_to_inference_graph(
         that also can be interpreted as inputs (ReadValue).
     :return: NNCFGraph in the inference style.
     """
-    inference_nncf_graph = remove_shapeof_subgraphs(nncf_graph, shapeof_metatypes, read_variable_metatypes)
-    inference_nncf_graph = filter_constant_nodes(nncf_graph, read_variable_metatypes)
+    inference_nncf_graph = deepcopy(nncf_graph)
```
kshpv, line 40: Why calling `deepcopy` here?

nncf/quantization/passes.py
```diff
@@ -136,3 +141,33 @@ def insert_null_biases_pass(model: TModel, graph: NNCFGraph) -> TModel:
 
         return insert_null_biases(model, graph)
     return model
+
+
+def insert_noops_instead_constants(
+    nncf_graph: NNCFGraph,
+    original_nncf_graph: NNCFGraph,
+    metatypes_to_insert_noop: Optional[List[OperatorMetatype]] = None,
+) -> NNCFGraph:
```
kshpv, line 150: There is no docstring

nncf/quantization/passes.py
```diff
@@ -34,8 +37,10 @@ def transform_to_inference_graph(
         that also can be interpreted as inputs (ReadValue).
     :return: NNCFGraph in the inference style.
     """
-    inference_nncf_graph = remove_shapeof_subgraphs(nncf_graph, shapeof_metatypes, read_variable_metatypes)
-    inference_nncf_graph = filter_constant_nodes(nncf_graph, read_variable_metatypes)
+    inference_nncf_graph = deepcopy(nncf_graph)
+    inference_nncf_graph = remove_shapeof_subgraphs(inference_nncf_graph, shapeof_metatypes, read_variable_metatypes)
+    inference_nncf_graph = filter_constant_nodes(inference_nncf_graph, read_variable_metatypes)
+    inference_nncf_graph = insert_noops_instead_constants(inference_nncf_graph, nncf_graph, metatypes_to_insert_noop)
```
kshpv, line 43: If we do `insert_noops_instead_constants` first and then apply `filter_constant_nodes` then `insert_noops_instead_constants` can have a more scalable signature. At least you will not need the original `nncf_graph`

kshpv (changes requested):

nncf/quantization/algorithms/min_max/backend.py
```diff
@@ -91,6 +91,13 @@ def group_conv_metatypes(self) -> List[OperatorMetatype]:
         Property for the backend-specific Grouped Convolution metatypes.
         """
 
+    @property
+    @abstractmethod
+    def keep_const_metatypes(self) -> List[OperatorMetatype]:
+        """
+        Property for the backend-specific metatypes which allowed to have pre-hooks.
```
kshpv, line 98:
```suggestion
        Metatypes which all inputs subgraphs should be treated as not constant.
```

nncf/quantization/passes.py
```diff
@@ -109,15 +116,27 @@ def filter_constant_nodes(
...
-    visited_nodes = set()
-    nodes_queue = collections.deque(start_nodes)
-    while nodes_queue:
-        node = nodes_queue.pop()
-        if node in visited_nodes:
-            continue
-        visited_nodes.add(node)
-        nodes_queue.extend(nncf_graph.get_next_nodes(node))
-    constant_nodes = [node for node in nncf_graph.get_all_nodes() if node not in visited_nodes]
+    def fill_nodes_list(initial_list: List[NNCFNode], fill_list: Set[NNCFNode], extend_call: callable):
```
kshpv, line 119:
```suggestion
    def visit_nodes(start_nodes: List[NNCFNode], visited_nodes: Set[NNCFNode], extend_call: Callable[NNCFNode, List[NNCFNode]]):
```

tests/openvino/native/data/reference_scales/IntegerModel_mixed.json
```diff
@@ -3877,10 +3877,10 @@
             ]
         ]
     },
-    "Gather_2/fq_output_0": {
+    "Gather_2_data/fq_output_0": {
         "input_low": 0.0,
-        "input_high": 0.997209906578064,
```
kshpv, line 3882: I am not sure, but I didn't fully get why the value is changed.

tests/common/quantization/test_passes.py
```diff
@@ -0,0 +1,192 @@
...
+            Edge(6, 0, 18, 0),
+            Edge(20, 0, 19, 0),
+            Edge(19, 0, 18, 1),
+        ],
+    )
+    return create_nncf_graph(graph)
+
+
+def graph_with_constants() -> NNCFGraph:
+    graph = Graph(
```
kshpv, line 94: This is really hard to understand the graph. Maybe you can save it in .dot format and then load, or, probably, add here a visualization.

tests/common/quantization/test_passes.py
```diff
@@ -0,0 +1,192 @@
+# Copyright (c) 2023 Intel Corporation
```
kshpv, line 1: Before we put such tests in `tests/post_training`

kshpv (changes requested):

tests/openvino/native/data/reference_scales/IntegerModel_mixed.json
KodiaqQ: Previously, FQ was placed after the `Gather` layer. For now, it puts before.

nncf/quantization/passes.py
KodiaqQ: Don't see any improvement in suggestions.

tests/common/quantization/test_passes.py
KodiaqQ: Do we have any mechanisms to restore the `NNCFGraph` instance from the .dot file?

kshpv: I am not aware of such mechanisms.

nncf/quantization/passes.py
kshpv: Take a look at the typehints

KodiaqQ: Done

alexsu52 (commented):
Please, check your approach for PyTorch and ONNX backends and add tests for mentioned backends.

tests/common/quantization/test_passes.py
kshpv: Can you put the tests in `tests/post_training` because passes for quantization are used only in `MinMax` (PTQ)?

PR #2025, opened 2023-08-02
Plot NAS search progression non-interactively

vshampor:
### Changes
As stated in the title

### Reason for changes
Windows pre-commits fail because they cannot find tk during the sanity tests. The plotting in sanity tests doesn't actually need interactivity, so agg can be used instead of tk as a backend which does not pose cross-platform problems.

### Related tickets
N/A

### Tests
windows/precommit_torch_cpu


vshampor:
run openvino pre-commit tests

vshampor:
windows/precommit_torch_cpu build 92 shows no tk-related errors

PR #2030, opened 2023-08-04
Optimize NNCFGraph

alexsu52:
### Changes

Optimize NNCFGraph. This changes shows speed-up in 10.8x for building NNCFGraph by "hf-internal-testing/tiny-random-GPTNeoXForCausalLM" model from optimum.

### Reason for changes

Speed-up NNCF compression algorithms

### Related tickets

ref: 113245

### Tests

N/A

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -187,7 +187,7 @@ class NNCFGraph:
     def __init__(self):
         self._nx_graph = nx.DiGraph()
         self._node_id_to_key_dict = {}
-        self._nodes = {}  # type: Dict[str, NNCFNode]
+        self._nodes: Dict[str, NNCFNode] = {} 
         self._input_nncf_nodes = {}  # type: Dict[int, NNCFNode]
         self._output_nncf_nodes = {}  # type: Dict[int, NNCFNode]
 
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -33,51 +33,73 @@ class NNCFNode:
     Class describing nodes used in NNCFGraph.
     """
 
-    def __init__(self, node_id: int, node_name: NNCFNodeName, data: dict = None):
-        self.node_id = node_id
-        self.data = data if data else {}
-        self.data[NNCFGraph.NODE_NAME_ATTR] = node_name
+    ID_NODE_ATTR = "id"
+    NODE_NAME_ATTR = "node_name"
+    KEY_NODE_ATTR = "key"
+    NODE_TYPE_ATTR = "type"
+    METATYPE_ATTR = "metatype"
+    LAYER_NAME_ATTR = "layer_name"
+    LAYER_ATTRIBUTES = "layer_attributes"
+    IGNORED_ALGOS_ATTR = "ignored_algos"
+    IS_IN_ITERATION_SCOPE_NODE_ATTR = "is_in_iteration_scope"
+    IS_INTEGER_INPUT_NODE_ATTR = "is_integer_input"
+    IS_SHARED_ATTR = "is_shared"
+
+    def __init__(self, attributes: Dict[str, Any]):
+        self._attributes = attributes
+
+    @property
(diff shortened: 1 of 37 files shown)


nncf/common/graph/graph.py
```diff
@@ -33,51 +33,73 @@ class NNCFNode:
     Class describing nodes used in NNCFGraph.
     """
 
-    def __init__(self, node_id: int, node_name: NNCFNodeName, data: dict = None):
-        self.node_id = node_id
-        self.data = data if data else {}
-        self.data[NNCFGraph.NODE_NAME_ATTR] = node_name
+    ID_NODE_ATTR = "id"
```
AlexanderDokuchaev, line 36: May be add dataclass for layer attribute instead of use names?
```python
@dataclass
class NodeAttributes:
    id: int
    name: str
    type: str
    ....
```

Or more, it's really need to save node attribute in `attribute` may be save node attribute in the class attributes?
```python
@dataclass
class NNCFNode:
    id: int
    name: str
    type: str
    ....
    data # for extra data like `output_mask`
```

nncf/common/graph/graph.py
```diff
@@ -162,32 +178,26 @@ class NNCFGraph:
...
-    IS_IN_ITERATION_SCOPE_NODE_ATTR = "is_in_iteration_scope"
-    IS_INTEGER_INPUT_NODE_ATTR = "is_integer_input"
     DTYPE_EDGE_ATTR = "dtype"
-    IS_SHARED_ATTR = "is_shared"
     PARALLEL_INPUT_PORT_IDS_ATTR = "parallel_input_ports"
 
     def __init__(self):
         self._nx_graph = nx.DiGraph()
         self._node_id_to_key_dict = {}
+        self._nodes = {}  # type: Dict[str, NNCFNode]
```
AlexanderDokuchaev, line 190:
```suggestion
        self._nodes: Dict[str, NNCFNode] = {} 
```

`type comments` use to support python2. 


`self._nodes` is the main change that improves performance, isn't it?

nncf/common/graph/graph.py
```diff
@@ -33,51 +33,73 @@ class NNCFNode:
...
 
     @property
     def node_type(self) -> str:
-        return self.data.get(NNCFGraph.NODE_TYPE_ATTR)
+        return self._attributes[NNCFNode.NODE_TYPE_ATTR]
 
     @property
-    def layer_name(self) -> LayerName:
-        return self.data.get(NNCFGraph.LAYER_NAME_ATTR)
+    def layer_name(self) -> Optional[LayerName]:
```
AlexanderDokuchaev, line 76: Is it really need here type aliases `LayerName` and `NNCFNodeName`. How about use `str` instead of create extra 2 names of `str` type?

nncf/common/graph/graph.py
alexsu52: This is necessary to avoid copying attributes from the networkx representation to NNCFNode. This is main optimization in this PR.

nncf/common/graph/graph.py
alexsu52: @vshampor, could you answer of this comment, because you introduced `LayerName`?

vshampor: The two entities are different in purpose and must have different types. That both types can both currently be sufficiently represented by `str` is a coincidence and does not bring any overhead at all.

KodiaqQ (approved):

nncf/common/graph/graph.py
AlexanderDokuchaev: But now it's `str`. In this logic we can create own aliases for each argument, but it's looks like create aliases to create alias. And this can confuse when you expect `LayerName` class but got `str`. 

I think that we should avoid creation nonfunctional entities. 
To describe different between `LayerName` and `NNCFNodeName` better use name of functions and arguments and docstring.

andrey-churkin (approved):
LGTM

AlexanderDokuchaev (approved):

PR #2032, opened 2023-08-04
NNCFTensor/NNCFTensorBackend-based algo code unification

vshampor:
### Changes
* Introduced `NNCFTensor` and `NNCFTensorBackend` abstract base classes for backend-powered tensor manipulation in common code
* Implemented corresponding backend-specific derived classes (e.g. `PTNNCFTensor`, `TFNNCFTensor`, `ONNXNNCFTensor` etc.)
* Rewritten the common algos for calculation code reuse powered by above design
* Removed duplicate code (e.g. common algo parts from backend-specific classes, almost all `TensorProcessor`s) and unified classes for statistics collection

### Reason for changes
Providing a way to write more common code and less backend-specific duplicate code

### Related tickets
N/A

### Tests
precommit, nightly


nncf/common/tensor.py
```diff
@@ -8,40 +8,358 @@
...
-TensorElementsType = TypeVar("TensorElementsType")
 
 
-class NNCFTensor:
+class TensorDtype(IntEnum):
+    FLOAT32 = 0
+    INT64 = 1
+
+
+class NNCFTensor(Generic[TensorType], abc.ABC):
```
vshampor, line 28: Backend implementations only need to implement this class and the `NNCFTensorBackend`. The methods in both classes are supposed to mirror the `numpy` syntax in terms of meaning, method names, parameter names, default values, so that the developers could intuitively use numpy-like reasoning and get consistent results.

nncf/common/tensor_impl_np.py
```diff
@@ -0,0 +1,248 @@
...
+from nncf.common.tensor import NNCFTensorBackend
+from nncf.common.tensor import TensorDtype
+
+_DTYPE_MAP: Dict[TensorDtype, Any] = {TensorDtype.FLOAT32: np.float32, TensorDtype.INT64: np.int64}
+
+_INV_DTYPE_MAP: Dict[Any, TensorDtype] = {v: k for k, v in _DTYPE_MAP.items()}
+_INV_DTYPE_MAP[dtype("float32")] = TensorDtype.FLOAT32
+
+
+class NPNNCFTensor(NNCFTensor[np.ndarray]):
```
vshampor, line 29: This is one such implementation that is currently reused for OV and ONNX, and also can be used for common NNCF-powered code testing with correct functional results.

nncf/openvino/tensor.py
```diff
@@ -8,24 +8,20 @@
...
+from nncf.common.tensor_impl_np import NPNNCFTensorBackend
 from nncf.parameters import TargetDevice
 
 
-class OVNNCFTensor(NNCFTensor):
-    """
-    A realisation of OpenVINO tensor wrapper for common NNCF algorithms.
-    """
+class OVNNCFTensor(NPNNCFTensor):
+    pass
```
vshampor, line 24: Reusing the `NPNNCFTensor` as-is for OV for the time being is as easy as this.

nncf/openvino/engine.py
```diff
@@ -68,7 +70,7 @@ def infer(
         output_data = {}
         for tensor, value in model_outputs.items():
             for tensor_name in tensor.get_names():
-                output_data[tensor_name] = value
+                output_data[tensor_name] = OVNNCFTensor(value)
```
vshampor, line 73: This is one such point for OV. It is enough to wrap the raw tensor value with a backend-specific derived class version of `NNCFTensor` in order for the resulting value to be usable in all common code based on `NNCFTensor` interface.

nncf/torch/quantization/init_range.py
```diff
@@ -250,8 +253,11 @@ def __init__(
         self.batch_size = batch_size
 
     def _get_fwd_hook(self, collector: TensorStatisticCollectorBase) -> Callable:
+        from nncf.torch import no_nncf_trace
+
         def fwd_hook(module, input_, output):
-            collector.register_input(input_[0])
+            with no_nncf_trace():
+                collector.register_input(PTNNCFTensor(input_[0]))
```
vshampor, line 260: For Torch the tensor wrapping is also done at the hooks collecting the function/module inputs.

nncf/quantization/algorithms/fast_bias_correction/backend.py
```diff
@@ -121,7 +114,7 @@ def create_input_data(
 
     @staticmethod
     @abstractmethod
-    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: TModel) -> np.ndarray:
+    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: TModel) -> NNCFTensor:
```
vshampor, line 117: "Backend entities" etc. are constrained to return `NNCFTensors` whenever the corresponding methods are supposed to be used from common code

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -199,10 +206,22 @@ def reshape_bias_shift(self, bias_shift: TTensor, bias_value: TTensor, channel_a
         if bias_value.ndim > 1:
             new_shape = [1] * bias_value.ndim
             new_shape[channel_axis] = bias_shift.shape[0]
-            bias_shift = self._backend_entity.reshape_tensor(bias_shift, new_shape)
+            bias_shift = bias_shift.reshape(*new_shape)
         return bias_shift
 
-    def _get_fp_inputs(self, statistic_points: StatisticPointsContainer, node_name: str) -> Tuple[List, List]:
+    @staticmethod
+    def _get_bias_shift_magnitude(current_bias_value: NNCFTensor, updated_bias_value: NNCFTensor) -> float:
```
vshampor, line 213: Once proper `NNCFTensor`s are passed into the algorithm, the code involving the tensor manipulations can be written once and be working out of the box for all backends (i.e. backend-specific implementations of `NNCFTensor`). The ex-backend-specific methods for the same were straightforward to move into the common algo part.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -199,10 +206,22 @@ def reshape_bias_shift(self, bias_shift: TTensor, bias_value: TTensor, channel_a
...
+        backend = current_bias_value.backend
```
vshampor, line 214: A reference to the backend interface (of abstract, well-defined type `NNCFTensorBackend`) is accessible from each NNCFTensor. Accessing this backend adds an extra line of code indeed, but it is not much more different than an extra line of `import numpy as np` that one had to do before.

Furthermore, if one does not want to do explicit calls to the backend when trying to call an operation on an `NNCFTensor`, he or she can almost always implement an equivalent method on the `NNCFTensor` itself and use that instead, e.g. if the following syntax seems to verbose:
```python
assert isinstance(nncf_tensor, NNCFTensor)
backend = nncf_tensor.backend
m = backend.mean(nncf_tensor, axis=(0, 2))
```
one could do:
```
assert isinstance(nncf_tensor, NNCFTensor)
m = nncf_tensor.mean(axis=(0, 2))
```

tests/openvino/native/quantization/test_fq_configurations.py
```diff
@@ -1,18 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-from nncf.openvino.statistics.statistics import OVMinMaxTensorStatistic
-from tests.post_training.test_templates.test_calculate_quantizer_parameters import TemplateTestFQParams
-
-
-class TestFQParams(TemplateTestFQParams):
-    @property
-    def tensor_statistic(self):
-        return OVMinMaxTensorStatistic
```
vshampor, line 18: Patterns such as this are no longer required to test the common code. In fact, the common code could be tested in the `tests/common` scope, by instantiating and running the `TemplateTestFQParams` as an actual test class there, powered by the `NPNNCFTensor`, and the backend-specific tests in `tests/openvino`, `tests/torch` etc. should instead be written to concentrate on running the backend-specific code for testing.

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -166,7 +165,38 @@ def _aggregate_impl(self) -> Any:
...
+    @abstractmethod
+    def _reset_sample_container(self):
+        pass
+
+    def __eq__(self, __o: object) -> bool:
+        return isinstance(__o, self.__class__) and self._num_samples == __o.num_samples
+
+    def __hash__(self) -> int:
+        return hash(self.__class__.__name__)
+
```
daniil-lyakhov, line 179: Please don't do this as this methods are not needed for aggregators but increases responsibility for implementation and testing

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -31,15 +30,22 @@ class TensorReducerBase(ABC):
     the specified rule. Could handle tensors inplace or out of place.
     """
 
-    def __init__(self, reduction_shape: Optional[ReductionShape] = None, inplace: bool = False):
+    def __init__(
+        self, reduction_axes: Optional[ReductionAxes] = None, channel_axis: Optional[int] = None, inplace: bool = False
+    ):
```
daniil-lyakhov, line 35: This change should be discussed as potentially ambiguous

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -408,195 +450,228 @@ def __init__(self, tensor_collectors: List[TensorCollector]) -> None:
...
 
 
 ##################################################Aggregators##################################################
 
 
 class NoopAggregator(TensorAggregatorBase):
+    # TODO (vshampor): this seems to be a bad design since this class does not actually aggregate anything
+    #  and therefore cannot be an `Aggregator` instance. In particular this is visible in the interface violation
+    #  for the _aggregate_impl function, which is supposed to return a single tensor - the aggregate - but has to
+    #  return a list of tensors instead in order for the purpose of the class to be served.
```
daniil-lyakhov, line 560: I believe overhead for empty aggregators support in a semantically correct way will be way to big

nncf/experimental/common/tensor_statistics/collectors.py
vshampor: It seems that I accidentally included the implementation for these methods twice in the class hierarchy. I removed the second implementations, but the first ones remain in the base `Aggregator` class, and it's your own design. Existing code will break if I simply remove thiese methods at all.

nncf/experimental/common/tensor_statistics/collectors.py
vshampor: Please state exactly what is ambiguous about this.

nncf/experimental/common/tensor_statistics/collectors.py
vshampor: Consistency is worth the effort, and the overhead for checking the aggregator list for a `NoopAggregator` to handle its inconsistent output locally is worse.

nncf/experimental/common/tensor_statistics/collectors.py
daniil-lyakhov: `channel_axis` is redundant, only `MeanPerChReducer` needs `channel_axis`

nncf/experimental/common/tensor_statistics/collectors.py
daniil-lyakhov: As far as I understand aggregators output signature is `Any`, as aggregator object can output anything. Do you think we need to make it strict?

vshampor: The stricter the boundaries, and types, the easier it is to integrate the new code into the existing code, and to understand it limitations. `Any` is not an interface at all, you could have omitted the type hint entirely then; and if there is no definition of a more or less concrete return value type here, you cannot write any meaningful code that utilizes this default value without resorting to `isinstance` checks, everywhere where you use this function.

nncf/experimental/common/tensor_statistics/collectors.py
vshampor: The `MeanPerChReducer` defines a use case when the reduction axes are not completely known at the moment of instantiating the reducer, instead the requirement is "to reduce all axes except for a given axis (i.e. the channel axis)". Evidently, the code that instantiates reducers in general currently has no requirement that it should be always possible to get the shape of the input tensors to the reducer so that, in combination with a known channel axis, the equivalent reduction axes could be calculated and the reducer initialized with these. If you can suggest an exact way to refactor the existing code that instantiates `MeanPerChReducer` so that I could calculate the `reduction_axes` unambiguously there, I would be happy to remove the `channel_axis` init argument. I would also be able then to remove the `MeanPerChReducer` at all, since a regular `MeanReducer` with correct reduction axes would fulfill the role.

vshampor:
After much discussion, a communal decision has been made to continue with the dispatching approach in `nncf/experimental/common/tensor.py` instead of this one. The existing approach allows to save a line of code when invoking package-function-like syntax for tensor manipulation and works slightly faster for tensor-method calls due to the fact that Python's `abc.ABC` is poorly optimized and method calls on abstract class implementations incur substantial overhead. (This could probably be optimized, but we would be optimizing Python by sacrificing readability). The existing approach, however, is slower than the one proposed here for package-function invocations, is limited in defining functions that don't take tensors as 0-th arg, more complex in the mechanism of functioning and forces an implicit import of all available backends whenever a `Tensor` or a function from `functions.py` is used. 

I will refactor the code base in this PR to use the existing approach, but will open a separate PR to do so.

PR #2034, opened 2023-08-07
[PTQ] Reshape models to original shapes after quantization in calibrate.py

daniil-lyakhov:
### Changes

Models are reshaped back after quantization

### Reason for changes

To return quantized models with the same shapes as it was before the quantization

### Related tickets

115226

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -689,7 +689,10 @@ def maybe_reshape_model(model, dataset, subset_size, input_to_tensor_name):
     model_inputs_shapes = {}
     for input_output in model.inputs:
         input_node = input_output.get_node()
-        model_inputs_shapes[input_to_tensor_name[input_node.friendly_name]] = tuple(input_node.partial_shape)
+        partial_shape = []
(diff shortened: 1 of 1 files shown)


tests/openvino/tools/calibrate.py
```diff
@@ -689,7 +689,10 @@ def maybe_reshape_model(model, dataset, subset_size, input_to_tensor_name):
     model_inputs_shapes = {}
     for input_output in model.inputs:
         input_node = input_output.get_node()
-        model_inputs_shapes[input_to_tensor_name[input_node.friendly_name]] = tuple(input_node.partial_shape)
+        partial_shape = []
+        for dim in input_node.partial_shape:
+            partial_shape.append(Dimension(str(dim)))
```
daniil-lyakhov, line 694: This is needed to make a copy of original model input shapes as original model reshape mutates partial shapes dimentions

daniil-lyakhov:
Build 10470 is finished successfully. No regressions are introduced by this PR, it's safe to merge it

alexsu52 (approved):
LGTM

PR #2047, opened 2023-08-16
[TF] Added a warning when training without pre-trained weights

nikita-savelyevv:
Added a warning when starting training object detection model without provided weights

#### Tickets
116938


examples/tensorflow/object_detection/main.py
```diff
@@ -384,6 +380,8 @@ def run(config):
     test_step = create_test_step_fn(strategy, compress_model, predict_post_process_fn)
 
     if "train" in config.mode:
+        if config.weights is None:
+            logger.warning("Pretrained checkpoint is not provided. This may lead to poor training results!")
```
alexsu52, line 384: What is about resume mode? Will the user see this warning?

nikita-savelyevv: The user will see this warning in resume mode

alexsu52: It seems like this behavior is wrong, doesn't it?

nikita-savelyevv: Seems like it, because in resume mode the model and compression state are loaded from `ckpt_path` argument and not `weights` argument. Added a check for resume mode to the if.

alexsu52 (approved):

PR #2052, opened 2023-08-16
Switch to pyproject.toml for specifying dependencies

vshampor:
### Changes
Most of the metadata has been moved out from setup.py to pyproject.toml.

### Reason for changes
setup.py is increasingly deprecated, and tools turn to pyproject.toml for specifying dependencies and other static configuration. setup.py cannot be removed completely, though, at this time, since we use it for dynamic version generation (if installed from source) and there are no good and compatible alternatives for this.

### Related tickets
N/A

### Tests
install tests


vshampor:
In its current state, and with reference to the current state of OS distributions and Python build tools, this change breaks a lot. `pip install -e .` is broken due to https://github.com/pypa/pip/issues/7953, and on Ubuntu 22 (although not Ubuntu 20), `pip install .` is broken due to https://github.com/pypa/pip/issues/6264 (pip can't tell the package's name and requirements and builds an `UNKNOWN` package). Both seem to have been fixed in newer versions of `setuptools` and `pip`, but no idea when the fixes will become available in an actual distro. Could work around this by adding a line mentioning system-wide setuptools upgrade into the installation notes and/or FAQ for Ubuntu 22 users who want the NNCF to be installable with `pip install -e .`

vshampor:
pyproject.toml is the future of the Python packaging, but we will only be able to do the switch once the distros we claim support for ship with setuptools/pip versions that do not exhibit breaking issues above.

PR #2057, opened 2023-08-17
Revert "Run coverage with pytest (#2038)"

AlexanderDokuchaev:
Disable coverage for internal precommit runs


vshampor:
I had run the pytest for torch locally (without sanity test cases) and can see that without --cov, the run takes ~65 minutes, and with --cov - 79 minutes. Will do some more time measurements, but the overhead seems to be substantial at these runtimes, so for now will merge this and disable coverage runs internally until the release is done and we can see if reducing the precommit scope on the most affected tests is possible.

PR #2058, opened 2023-08-18
Tensor for PTQ

AlexanderDokuchaev:
### Changes

- Update `MinMax` and `FastBiasCorrection` to use common Tensor.
- Remove converting torch -> numpy -> torch of data.
- `FakeQuantizeParameters` collect data wrapped by tensor.
- Add support cuda for torch backend.
- Add new functions for `Tensor`:
  - stack
  - unstack
  - moveaxis
  - mean
  - round 
- Removed `__all__` from function.py, it's works like default behavior.
- Add `statistical_functions.py` for high level functions that used only function from `functions.py` and have no backend specific implementations:
  - mean_per_channel
-  Disable warnings for divide operators of numpy

### Related tickets

113315


AlexanderDokuchaev:
post_training_quantization/141/

nncf/experimental/tensor/functions.py
```diff
@@ -48,7 +48,7 @@ def device(a: TTensor) -> TensorDeviceType:
 
 @functools.singledispatch
 @_tensor_guard
-def squeeze(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) -> TTensor:
+def squeeze(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) -> Tensor:
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Functions in this file always return Tensor or list of Tensor

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -181,12 +184,31 @@ def apply(
...
+        """
+        bias_shift_magnitude = inf
+        if functions.count_nonzero(current_bias_value == 0) == 0:
+            bias_shift_magnitude = functions.max(
+                functions.abs((updated_bias_value - current_bias_value) / current_bias_value)
+            )
+        return bias_shift_magnitude
+
+    @staticmethod
+    def _reshape_bias_shift(bias_shift: Tensor, bias_value: Tensor, channel_axis: int) -> TTensor:
```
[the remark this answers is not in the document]
AlexanderDokuchaev: It should be Tensor

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -223,7 +245,7 @@ def input_filter_func(point):
             node_name, input_filter_func, self._algorithm_key
         ):
             statistics = tensor_collector.get_statistics()
-            input_fp.extend(statistics.mean_values)
+            input_fp.extend(Tensor(statistics.mean_values))
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Yes, but i dont want to refactor all in one PR. In this PR fixed ptq for cuda and shows how to use Tensor on real code.

nncf/experimental/tensor/functions.py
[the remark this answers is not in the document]
andrey-churkin: I think this is not entirely correct. I mean these functions return a `Tensor` object only if there is no registered version for the type of the passed argument.

nncf/experimental/tensor/functions.py
```diff
@@ -324,25 +350,119 @@ def zeros_like(a: TTensor) -> TTensor:
...
+
+    :param a: Input data.
+    :param decimals: Number of decimal places to round to (default: 0). If decimals is negative,
+      it specifies the number of positions to the left of the decimal point.
+    :return: An array of the same type as a, containing the rounded values.
+    """
+    return Tensor(round(a.data, decimals))
+
+
+def mean_per_channel(x: Tensor, axis: int) -> Tensor:
```
alexsu52, line 426: Please move this function to the more relevant python module.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -26,6 +27,8 @@
 from nncf.common.tensor_statistics.statistic_point import StatisticPointsContainer
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
+from nncf.experimental.tensor import Tensor
+from nncf.experimental.tensor import functions
```
alexsu52, line 31:
```suggestion
from nncf.experimental.tensor import functions as fns
```

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -79,25 +80,24 @@ def get_sub_input_output_names(subgraph: onnx.ModelProto) -> Tuple[str, str]:
     def create_input_data(
         shape: Tuple[int], data: List[np.ndarray], input_name: str, channel_axis: int
```
alexsu52, line 81: Please adjust type hints here and below.

alexsu52 (changes requested):

nncf/experimental/tensor/functions.py
AlexanderDokuchaev: Moved to math_functions.py

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
AlexanderDokuchaev: done

AlexanderDokuchaev:
run pytorch pre-commit tests

nncf/experimental/tensor/functions.py
```diff
@@ -88,7 +88,20 @@ def max(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) -> TTensor:
 
 @functools.singledispatch
 @_tensor_guard
-def min(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) -> TTensor:  # pylint: disable=redefined-builtin
+def amax(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) -> Tensor:  # pylint: disable=redefined-builtin
```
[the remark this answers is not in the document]
alexsu52: Thanks, I would suggest sticking to numpy notation.

nncf/experimental/tensor/functions.py
```diff
@@ -294,7 +320,7 @@ def ones_like(a: TTensor) -> TTensor:
 
 @functools.singledispatch
 @_tensor_guard
-def where(condition: TTensor, x: TTensor, y: TTensor) -> TTensor:
+def where(condition: TTensor, x: TTensor, y: TTensor) -> Tensor:
```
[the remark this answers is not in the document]
alexsu52: Is it difficult to support `bool`?

nncf/experimental/tensor/functions.py
alexsu52: I interpret this function as a statistical function, and I would recommend moving this function to the "nncf/experimental/common/tensor_statistics" package, for a example, into statistical_functions.py.

AlexanderDokuchaev: done

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -77,27 +78,26 @@ def get_sub_input_output_names(subgraph: onnx.ModelProto) -> Tuple[str, str]:
...
         for j, idx in enumerate(np.ndindex(blob.shape[channel_axis])):
             index = tuple(slice(None) if i != channel_axis else idx for i in range(blob.ndim))
-            blob[index] = data[j]
-        blob = blob.astype(data[0].dtype)
+            blob[index] = data[j].data
         input_data = {input_name: blob}
         return input_data
 
     @staticmethod
     def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: onnx.ModelProto) -> np.ndarray:
```
alexsu52, line 91:
```suggestion
    def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: onnx.ModelProto) -> Tensor:
```

Please correct the type hints here and elsewhere.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -77,27 +78,26 @@ def get_sub_input_output_names(subgraph: onnx.ModelProto) -> Tuple[str, str]:
...
-        return get_bias_value(node, model)
+        return Tensor(get_bias_value(node, model))
 
     @staticmethod
     def get_activation_port_ids_for_bias_node(node: NNCFNode) -> Tuple[int, int]:
         return 0, 0
 
     @staticmethod
     def process_model_output(raw_data: Dict, output_name: str) -> ONNXNNCFTensor:
```
alexsu52, line 99:
```suggestion
    def process_model_output(raw_data: Dict, output_name: str) -> Tensor:
```

Please correct the type hints here and elsewhere.

nncf/experimental/tensor/functions.py
```diff
@@ -218,7 +244,7 @@ def count_nonzero(a: TTensor, axis: Optional[Union[int, Tuple[int]]] = None) ->
 
 @functools.singledispatch
 @_tensor_guard
-def isempty(a: TTensor) -> TTensor:
+def isempty(a: TTensor) -> Tensor:
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Some comments about bool type.
Comparison operator can return `bool` too in case of comparing numbers, but return np.array or torch.tensor if compare arrays.
```python
>>> np.array([0,0]) < np.array([1,0])
array([ True, False])
>>> np.array(0) < np.array(1)
True
```

May be better to return Tensor in all case, include `isempty` and `allclose`? Or return bool too for comparison operator  in case of numbers?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -300,8 +319,8 @@ def _get_bias_shift(
         engine = EngineFactory.create(model)
         raw_output = engine.infer(input_blob)
         q_outputs = self._backend_entity.process_model_output(raw_output, output_name)
-        q_outputs = self._backend_entity.tensor_processor.mean_per_channel(q_outputs, channel_axis).tensor
```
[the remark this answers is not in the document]
AlexanderDokuchaev: yes, removed

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
AlexanderDokuchaev: Hope i fixed all

nncf/experimental/tensor/functions.py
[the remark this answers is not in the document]
alexsu52: My comment was only about `isempty` function. I agree that `allclose` should return a Tensor. But I do not understand why isempty should return a Tensor.

nncf/experimental/common/tensor_statistics/statistical_functions.py
```diff
@@ -0,0 +1,28 @@
...
+
+    :param x: Tensor to reduce.
+    :param axis: The channel dimensions to reduce.
+    :return: Reduced Tensor.
+    """
+    if len(x.shape) < 3:
+        return fns.mean(x, axis=0)
+    x = fns.moveaxis(x, axis, 1)
+    t = x.reshape([x.shape[0], x.shape[1], -1])
+    return fns.mean(t, axis=(0, 2))
```
vshampor, line 28: You can do this in a shorter fashion:
```python
axis = tuple(i for i in range(x.ndim) if i != channel_axis)
return fns.mean(t, axis=axis)
```

nncf/experimental/tensor/functions.py
```diff
@@ -324,28 +322,105 @@ def zeros_like(a: TTensor) -> TTensor:
...
+        unwrapped_x = [i.data for i in x]
+        # singledispatch cannot dispatch function by element in a list
+        res = stack.dispatch(type(unwrapped_x[0]))(unwrapped_x, axis=axis)
+        return Tensor(res)
+    raise NotImplementedError(f"Function `stack` is not implemented for {type(x)}")
+
+
+@functools.singledispatch
+@_tensor_guard
+def unstack(a: TTensor, axis: int = 0) -> List[Tensor]:
```
vshampor, line 344: I guess that here, and everywhere, the type hint for the function args should be `Tensor`, not `TTensor`

nncf/experimental/tensor/README.md
```diff
@@ -120,28 +120,25 @@ tensor_a[0:2]  # Tensor(array([[1],[2]]))
...
-
-4. Add backend specific implementation of method to:
+3. Add backend specific implementation of method to:
 
     - [numpy_function.py](numpy_function.py)
 
         ```python
-        @functions.foo.register(np.ndarray)
-        @functions.foo.register(np.number)
+        @registry_numpy_types(fns.foo)
```
vshampor, line 128: `@register_numpy_types`

nncf/experimental/tensor/numpy_functions.py
```diff
@@ -158,7 +154,49 @@ def _(
...
+def _(a: Union[np.ndarray, np.number], decimals: int = 0) -> np.ndarray:
+    return np.round(a, decimals=decimals)
+
+
+@registry_numpy_types(fns.binary_operator)
+def _(
+    a: Union[np.ndarray, np.number], b: Union[np.ndarray, np.number], operator_fn: Callable
+) -> Union[np.ndarray, np.number]:
+    # Run operator with disabled warning
+    with np.errstate(invalid="ignore", divide="ignore"):
```
vshampor, line 192: Why exactly do we need warnings disabled during these ops? Don't we want to know about places where we try to divide by zero, and try to handle this at the common code level by ensuring that there are no places where we unexpectedly try to divide by zero?

nncf/experimental/tensor/functions.py
```diff
@@ -324,28 +322,105 @@ def zeros_like(a: TTensor) -> TTensor:
...
+    """
+    Stacks a list or deque of Tensors rank-R tensors into one Tensor rank-(R+1) tensor.
+
+    :param x: List or deque of Tensors.
+    :param axis: The axis to stack along.
+    :return: Stacked Tensor.
+    """
+    if isinstance(x, List):
+        unwrapped_x = [i.data for i in x]
+        # singledispatch cannot dispatch function by element in a list
```
vshampor, line 336: Same limitation of the `singledispatch` mechanism necessitates complications such as this. I imagine that you will run into this yet, for example, when implementing an equivalent of `np.concatenate` required [here](https://github.com/openvinotoolkit/nncf/blob/c07810beeb2a638ba8c1d1800167f2f0bb97cabe/nncf/common/tensor_statistics/reduction.py#L57), so might as well pull the logic here out into a helper function.

Notice that the OOP approach allowed to immediately write such functions ([interface](https://github.com/openvinotoolkit/nncf/pull/2032/files#diff-ec72fc8ffe89978a911fbca7e7f6a1bc30ba838d79b3b4b59a43d3b81aec910fR223), [implementation](https://github.com/openvinotoolkit/nncf/pull/2032/files#diff-a4ab3aa944f7ba33bcea038694de8d08784f5124668ba7a08e585fc3e47e5ff6R109))

nncf/experimental/tensor/torch_functions.py
```diff
@@ -37,112 +37,151 @@ def _(a: torch.Tensor) -> TensorDeviceType:
...
 
 
-@functions.zeros_like.register(torch.Tensor)
+@fns.zeros_like.register(torch.Tensor)
 def _(a: torch.Tensor) -> torch.Tensor:
     return torch.zeros_like(a)
+
+
+@fns.stack.register(torch.Tensor)
+def _(x: List[torch.Tensor], axis: int = 0) -> List[torch.Tensor]:
```
vshampor, line 154: The type passed to `.register` will now be different from the type of the first arg, which is a new precedent among these functions. You might want to add a comment in code about that, or an explanation in the README.md

nncf/experimental/tensor/numpy_functions.py
```diff
@@ -28,68 +28,72 @@
...
+def registry_numpy_types(singledispatch_fn):
+    """
+    Decorator to register function to singledispatch for numpy classes.
+
+    :param singledispatch_fn: singledispatch function.
+    """
+
+    def inner(func):
+        singledispatch_fn.register(np.ndarray)(func)
+        singledispatch_fn.register(np.generic)(func)
```
vshampor, line 40: This function refers to `np.generic`, while the type hint in the implementations is `Union[np.ndarray, np.number]`. Should establish consistency here.

nncf/experimental/tensor/functions.py
andrey-churkin: > I guess that here, and everywhere, the type hint for the function args should be `Tensor`, not `TTensor`

I think this is not entirely correct. I mean these functions return a `Tensor` object only if there is no registered version for the type of the passed argument.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -26,6 +27,9 @@
 from nncf.common.tensor_statistics.statistic_point import StatisticPointsContainer
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
+from nncf.experimental.common.tensor_statistics import statistical_functions as s_fns
+from nncf.experimental.tensor import Tensor
+from nncf.experimental.tensor import functions as fns
```
KodiaqQ, line 32: What is the difference between `fns` and `s_fns`? Why should we use all of them?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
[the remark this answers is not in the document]
KodiaqQ: IMO, we should provide the changes in case of UX improvements or developing/code support improvements.
Providing changes only `for demonstration` is not good.
So, this is why I still can't understand why should we add a new wrapper for all tensors in this PR.

tests/torch/ptq/test_calculation_quantizer_params.py
```diff
@@ -343,8 +342,8 @@ def test_quantizer_parameters_export(tmp_path: Path):
 
     for name, param in fq_params.items():
         assert name in torch_ptq_params
-        assert np.allclose(param["input_low"], torch_ptq_params[name]["input_low"])
-        assert np.allclose(param["input_high"], torch_ptq_params[name]["input_high"])
+        assert np.allclose(param["input_low"].data.numpy(), torch_ptq_params[name]["input_low"])
```
KodiaqQ, line 345: Why `numpy` and not `fns.allclose`?

nncf/quantization/fake_quantize.py
```diff
@@ -21,6 +21,9 @@
 from nncf.common.quantization.structs import QuantizerConfig
 from nncf.common.quantization.structs import QuantizerGroup
 from nncf.common.tensor_statistics.statistics import MinMaxTensorStatistic
+from nncf.experimental.tensor import Tensor
+from nncf.experimental.tensor import TensorDataType
+from nncf.experimental.tensor import functions as fns
```
KodiaqQ, line 26: Looks like we may try to implement all these entities in the one namespace as done in the `numpy`.
Because for now, we should always keep in mind that we can't do anything with the `Tensor` without importing `fns` everywhere, which is not convenient as may be.

KodiaqQ:
Also, are there any performance improvement numbers while doing quantization for `PyTorch` with the `CUDA` device?

AlexanderDokuchaev:
Speedup of quantization time 


Model | cpu | cuda | speedup
-- | -- | -- | --
timm/crossvit_9_240 | 0:00:44 | 0:00:35 | 1.26
timm/darknet53 | 0:00:38 | 0:00:25 | 1.52
timm/deit3_small_patch16_224 | 0:00:31 | 0:00:21 | 1.48
timm/dpn68 | 0:00:43 | 0:00:33 | 1.30
timm/efficientnet_b0 | 0:00:37 | 0:00:31 | 1.19
timm/efficientnet_lite0 | 0:00:25 | 0:00:20 | 1.25
timm/hrnet_w18 | 0:03:46 | 0:02:55 | 1.29
timm/levit_128 | 0:00:39 | 0:00:33 | 1.18
timm/mobilenetv2_050 | 0:00:27 | 0:00:21 | 1.29
timm/mobilenetv3_small_050 | 0:00:24 | 0:00:20 | 1.20
timm/regnetx_002 | 0:00:25 | 0:00:20 | 1.25
timm/resnest14d | 0:00:19 | 0:00:14 | 1.36
timm/resnet18 | 0:00:15 | 0:00:09 | 1.67
timm/swin_base_patch4_window7_224 | 0:01:28 | 0:00:54 | 1.63
timm/vgg11 | 0:00:28 | 0:00:07 | 4.00
timm/visformer_small | 0:00:35 | 0:00:22 | 1.59
timm/wide_resnet50_2 | 0:00:42 | 0:00:22 | 1.91

tests/torch/ptq/test_calculation_quantizer_params.py
AlexanderDokuchaev: Updated

nncf/quantization/fake_quantize.py
AlexanderDokuchaev: Yes, i have suggest to move `nncf` scope (ex. `nncf.Tensor`, `nncf.DataType`, `nncf.mean`), but not sure that we should do it while Tensor in `experimental`.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
AlexanderDokuchaev: It's suggestion of @alexsu52 https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1312954941

Idea to split functions that override functions under Tensor (like mean, min and etc) and functions that we use to calculate statistics that operating only wrapped Tensors.

nncf/experimental/common/tensor_statistics/statistical_functions.py
```diff
@@ -0,0 +1,28 @@
...
+def mean_per_channel(x: Tensor, axis: int) -> Tensor:
+    """
+    Computes the mean of elements across given channel dimension of Tensor.
+
+    :param x: Tensor to reduce.
+    :param axis: The channel dimensions to reduce.
+    :return: Reduced Tensor.
+    """
+    if len(x.shape) < 3:
+        return fns.mean(x, axis=0)
```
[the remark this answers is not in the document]
AlexanderDokuchaev: hm.. dont know
It's just copypast from https://github.com/openvinotoolkit/nncf/blob/3b4ba6a8140cf08457df6ed8adc4b1305dee6028/nncf/openvino/statistics/collectors.py#L98-L103
For other backends the function looks like same https://github.com/search?q=repo%3Aopenvinotoolkit%2Fnncf+def+mean_per_channel%28x%3A+NNCFTensor%2C+axis%3A+int%29+-%3E+NNCFTensor%3A&type=code

nncf/experimental/tensor/torch_functions.py
AlexanderDokuchaev: Dont think that need it to readme.md until we encounter more precedents like this, for one function doc string and type hints looks enough.

nncf/experimental/tensor/tensor.py
```diff
@@ -132,13 +137,13 @@ def min(self, axis: Optional[TTensor] = None) -> "Tensor":
...
 
-    def isempty(self) -> "Tensor":
+    def isempty(self) -> bool:
         return _call_function("isempty", self)
 
     def astype(self, dtype: TensorDataType):
         return _call_function("astype", self, dtype)
 
-    def reshape(self, shape: TTensor) -> "Tensor":
+    def reshape(self, shape: List) -> "Tensor":
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Fixed

nncf/experimental/tensor/tensor.py
```diff
@@ -32,7 +33,11 @@ def data(self) -> TTensor:
 
     @property
     def shape(self) -> List[int]:
-        return list(self.data.shape)
+        return tuple(self.data.shape)
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Updated typehint for `shape` and `axis` arguments to `Tuple[int, ...]`

nncf/experimental/tensor/README.md
AlexanderDokuchaev: Fixed

nncf/experimental/tensor/tensor.py
```diff
@@ -81,16 +86,16 @@ def __pow__(self, other: TTensor) -> "Tensor":
...
-        return Tensor(unwrap_tensor_data(other) / self.data)
+        return _call_function("binary_reverse_operator", self, other, operator.truediv)
 
     def __floordiv__(self, other: TTensor) -> "Tensor":
-        return Tensor(self.data // unwrap_tensor_data(other))
+        return _call_function("binary_operator", self, other, operator.floordiv)
 
     def __rfloordiv__(self, other: TTensor) -> "Tensor":
-        return Tensor(unwrap_tensor_data(other) // self.data)
+        return _call_function("binary_reverse_operator", self, other, operator.floordiv)
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Yeap, `_` good idea, renamed to `_binary_op_nowarn` and `_binary_reverse_op_nowarn`

nncf/experimental/tensor/functions.py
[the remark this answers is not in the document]
AlexanderDokuchaev: Update all return type to `Tensor` to correctly works "pop-up suggestions" (or how it names when editor suggest functions by first symbols?).
So set Tensor type to first argument. But not sure about second, added like Union[torch.Tensor, float], because it can be used like `torch.tensor(1) + 1` or `fns.min(torch.tensor(1), 0)`. May be do you have anny suugestion about it?

nncf/experimental/common/tensor_statistics/statistical_functions.py
AlexanderDokuchaev:
```python
    pos_axis = axis + x.ndim if axis < 0 else axis
    if not 0 <= pos_axis < x.ndim:
        raise ValueError(f"axis {axis} is out of bounds for array of dimension {x.ndim}")
    axis = tuple(i for i in range(x.ndim) if i != pos_axis)
    return fns.mean(x, axis=axis)
```
Not shorter, but it looks clearer 👍 


About performance, looks the same in most cases but in case of `shape: 1, 256, 16, 16` and `axis=1` speedup = `1.18`
```

    x = np.random.rand(1, i, 256, 256)
    axis = 1
-------
Shape: 1, 16, 256, 256
New: 3.24s  Old: 3.3s  Speedup: 1.0177
-------
Shape: 1, 48, 256, 256
New: 12.8s  Old: 12.8s  Speedup: 0.99951
-------
Shape: 1, 96, 256, 256
New: 31.8s  Old: 31.5s  Speedup: 0.99251
-------
Shape: 1, 256, 256, 256
New: 94.9s  Old: 94.9s  Speedup: 1.0

    x = np.random.rand(1, 256, i, i)
    axis = 1
-------
Shape: 1, 256, 16, 16
New: 0.283s  Old: 0.335s  Speedup: 1.1816
-------
Shape: 1, 256, 48, 48
New: 2.02s  Old: 2.09s  Speedup: 1.0323
-------
Shape: 1, 256, 96, 96
New: 9.57s  Old: 9.63s  Speedup: 1.0058
-------
Shape: 1, 256, 256, 256
New: 95.7s  Old: 95.8s  Speedup: 1.0009
```
I will update after https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1322651908

nncf/experimental/tensor/torch_functions.py
vshampor: There will be at least [three more](https://github.com/openvinotoolkit/nncf/pull/2120/files#diff-7746d0efc38f06575ddc5ce118d41c5e80850aa03a470ef2adb6954f87a57c15R212) "precedents" per backend. And even one is enough for this to have to be documented.

nncf/experimental/common/tensor_statistics/statistical_functions.py
AlexanderDokuchaev: updated

nncf/experimental/tensor/functions.py
AlexanderDokuchaev: Updated all typhints `TTensor` now only used for .data

nncf/experimental/tensor/torch_functions.py
AlexanderDokuchaev: Added to section "Add new method or function" and added helper function `_dispatch_list`

nncf/experimental/tensor/functions.py
AlexanderDokuchaev: Added function `_dispatch_list`

nncf/experimental/tensor/numpy_functions.py
AlexanderDokuchaev: It's need to sync behavior for numpy and torch, torch has not analog of `errstate` to raise error or warnings. 
Default behavior in both backends is that dividing by zero return `inf`, as result in common methods requires addition checks and it's not depend from warnings.
And in NNCF no any places where used `np.errstate(divide="raise"):` only "ignore".

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
KodiaqQ: I still don't get when the `fns` and `s_fns` would be merged.

nncf/quantization/algorithms/min_max/torch_backend.py
```diff
@@ -141,10 +134,10 @@ def create_quantizer_insertion_command(
...
         max_values, min_values = [], []
         for statistic in statistics:
-            max_values.append(torch.tensor(statistic.max_values).flatten())
-            min_values.append(torch.tensor(statistic.min_values).flatten())
-        max_values = torch.max(torch.tensor(max_values))
-        min_values = torch.min(torch.tensor(min_values))
+            max_values.append(statistic.max_values.flatten())
+            min_values.append(statistic.min_values.flatten())
+        max_values = torch.amax(torch.stack(max_values), dim=0)
+        min_values = torch.amin(torch.stack(min_values), dim=0)
```
KodiaqQ, line 140: I don't get why this moment wasn't placed into the common (algo) part, while the `FastBiasCorrection` parts were.

KodiaqQ (commented):
One of the major changes that are not clear:

- Why shouldn't we place all tensor-related entities in the one namespace (`Tensor`, `functions`);
- Why the one algorithm updated with the `Tensor` and the others are not;
- Why the developers should use multiple imports to start working with the `Tensor` instead of `import numpy as np`.

andrey-churkin (approved):

KodiaqQ (approved):

nncf/experimental/tensor/functions.py
```diff
@@ -324,28 +326,113 @@ def zeros_like(a: TTensor) -> TTensor:
...
+def stack(x: List[Tensor], axis: int = 0) -> Tensor:
+    """
+    Stacks a list of Tensors rank-R tensors into one Tensor rank-(R+1) tensor.
+
+    :param x: List of Tensors.
+    :param axis: The axis to stack along.
+    :return: Stacked Tensor.
+    """
+    if isinstance(x, List):
+        unwrapped_x = [i.data for i in x]
```
vshampor, line 339: Can this go into `_dispatch_list` too?

AlexanderDokuchaev: `_dispatch_list` have one task to run `dispatch` by type of element from first argument that has type List. 

Why `_dispatch_list` should contain extra logic to unwrap data? 
Should it unwrap only first argument or all? What about kwargs?  
I think unwrapping data should be done where we know names to avoid call `isinstance` check for all arguments of function.

vshampor: Limit the `_dispatch_list` to taking lists of tensors as its first arg. You don't need to build the most generic API here.

AlexanderDokuchaev: `_dispatch_list` already limited to get function and args to dispatch by first element in the first argument.

nncf/experimental/tensor/__init__.py
```diff
@@ -12,5 +12,6 @@
 from nncf.experimental.tensor.enums import TensorBackendType
 from nncf.experimental.tensor.enums import TensorDataType
 from nncf.experimental.tensor.enums import TensorDeviceType
+from nncf.experimental.tensor.functions import *  # pylint: disable=redefined-builtin
```
alexsu52, line 15: Could you briefly explain why you need this import?

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
alexsu52: What is reason to merge them?

nncf/experimental/tensor/functions.py
vshampor: It is an implicit limitation. Make it explicit by separating the corresponding argument from *args in the signature and setting up a corresponding type hint.

AlexanderDokuchaev: Please write the justification to your suggestion.

vshampor: That would be code reuse and explicitness through typing.

nncf/experimental/tensor/__init__.py
AlexanderDokuchaev: removed, import will be discussed in next PR when the Tensor module will be retired from experimental status

nncf/experimental/tensor/functions.py
AlexanderDokuchaev: It is not a reason to unwrap data in `_dispatch_list` function.
Is it blocking the merge?

vshampor: Yes, it is a reason, since you have to take every opportunity at making the code better, especially when you are pointed toward such opportunity. Now that you have asked - yes, I want to see this change in this PR, unless *you* provide a reason as to why your solution is better than what I suggest.

vshampor (changes requested):
I'd like to see the `_dispatch_list` improved as directed - after that I don't see major blockers for merging this PR (other than the already visible deficiencies of this approach code-wise as compared to the [proper OOP approach](https://github.com/vshampor/nncf/blob/nncf_tensor_saved/nncf/common/tensor.py))

nncf/experimental/tensor/functions.py
vshampor: I am requesting the changes in accordance with the arguments I've provided in [two](https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1345596652) [comments](https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1345882442) now.

AlexanderDokuchaev: @AlexKoff88 Need you help to resolve it. Does this comment block the merge of this PR or not?
@alexsu52

alexsu52: @vshampor, please provide a code of your proposal.

vshampor: In the `_dispatch_list` definition, you would have:
```python
def _dispatch_list(fn: "functools._SingleDispatchCallable", tensor_list: List[Tensor], *args, **kwargs):
    unwrapped_list = [i.data for i in tensor_list]
    return fn.dispatch(type(unwrapped_list[0]))(unwrapped_list, *args, **kwargs)
```
and the `stack` implementation then becomes simply:

```python
@functools.singledispatch
def stack(x: List[Tensor], axis: int = 0) -> Tensor:
    return Tensor(_dispatch_list(stack, x, axis=axis))
```
and this is exactly what I wrote [here](https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1343002474) and [here](https://github.com/openvinotoolkit/nncf/pull/2058#discussion_r1345467003), in good English. I would expect at this point in time of the NNCF devs to be able to do this kind of primitive, obvious refactoring on their own without code examples and instead to use their reading comprehension skills. As it is, note that I had to spend (a quite marginal) amount of time to write this code on my own to demonstrate this.

AlexanderDokuchaev: ok, added

nncf/quantization/algorithms/fast_bias_correction/backend.py
```diff
@@ -31,13 +31,6 @@
 
 
 class FastBiasCorrectionAlgoBackend(ABC):
```
alexsu52, line 33: Please update the type hints to take into account the migration from the tensor framework to nncf.Tensor.

alexsu52 (approved):
Please, address my minor comment

PR #2040, opened 2023-08-10
[PTQ][ONNX][OV] Quantized embedding weights

KodiaqQ:
### Changes

- Added embedding weights quantization

### Reason for changes

- Proper quantization scheme

### Related tickets

- 113692

### Tests

- Added graph for `tests/torch/ptq/test_graphs.py`
- Added graph for `tests/onnx/quantization/test_graphs.py`
- Updated graph for `tests/openvino/native/quantization/test_graphs.py`

Old (no-) FQ placement (OV):
image

New FQ placement (OV):
image

New FQ placement (ONNX):
image

PR #2066, opened 2023-08-18
Update QPSG logic

KodiaqQ:
### Changes

- Updated QPSG logic for Embedding quantization case

### Reason for changes

- Proper quantization logic for QPSG in the specific case

### Related tickets

- Post-merge of #2040 (CVS-113692)

### Tests

- Updated `tests/common/quantization/test_quantizer_propagation_graph.py`


nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -1422,6 +1422,27 @@ def _handle_output_quantizers_for_weights_as_outputs_ops(
...
+        """
+        act_qconfig_extend_list = []
+        for act_qconfig in activation_qconfig_options:
+            if act_qconfig.signedness_to_force is None:
+                for signedness_to_force_position in [True, False]:
+                    act_qconfig_updated = deepcopy(act_qconfig)
+                    act_qconfig_updated.signedness_to_force = signedness_to_force_position
+                    act_qconfig_extend_list.append(act_qconfig_updated)
+        act_qconfig_extend_list += activation_qconfig_options
+        return [qconf for qconf in weight_qconfig_options if qconf in act_qconfig_extend_list]
```
vshampor, line 1444: Make sure that this logic covers the following case:
let `weight_qconfig_options = [(*partial_configuration*, signedness_to_force=None), ]`

let `activation_qconfig_options = [(*same_partial_configuration*, signedness_to_force=None), ]`

then the intersection should be:
`intersection == [(*same_partial_configuration*, signedness_to_force=None),]`

and currently it looks like your logic will produce the following instead:
`intersection == [(*same_partial_configuration*, signedness_to_force=True), (*same_partial_configuration*, signedness_to_force=False)]`

While we are not likely to encounter this case realistically, I think that the current behaviour in this case would be confusing.

KodiaqQ:
run pytorch pre-commit tests

vshampor (approved):

PR #2076, opened 2023-08-22
[PTQ][OV] Support GroupConvolution with ShapeOf case

KodiaqQ:
### Changes

- Added support for weights quantization of GroupConvolution with ShapeOf

### Reason for changes

- Proper quantization scheme. Currently, weights for this case not quantizable.

### Related tickets

- 113647

### Tests

- Added test case

Currently, it's possible that model from the **TF** framework can contain a **ShapeOf** subgraph for weights of **GroupConvolution**:
image

Updated quantization scheme. Weights quantized (not decomposed):
image


l-bat (approved):

KodiaqQ:
Still validating locally. Found an issue with the FQ constants shape.

PR #2078, opened 2023-08-23
Fix torch.jit.script wrapper

vshampor:
### Changes
`torch.jit.script` wrapper fixed to catch exceptions in the original `torch.jit.script` call in order to keep torch state patched if an exception from the user's `torch.jit.script` invocation was expected to fail (by handling the original exception) and the process should have continued.

### Reason for changes
If the user has the following code:
```python
_, nncf_model = create_compressed_model(...)
try:
    _ = torch.jit.script(nncf_model)
except:
    pass
```
then, since `torch.jit.script` will fail for NNCF models and the wrapper had to unpatch torch ops in the meanwhile, we end up in a state where the script continues, but torch ops are unpatched, and further interaction with `nncf_model` will fail.

### Related tickets
118468

### Tests
tests.torch.test_pytorch_patch.test_jit_script_exception_preserves_patching

### DIFF
--- a/nncf/torch/dynamic_graph/patch_pytorch.py
+++ b/nncf/torch/dynamic_graph/patch_pytorch.py
@@ -188,7 +188,7 @@ def torch_jit_script_wrapper(*args, **kwargs):
     # so at call of torch.jit.script function we need to
     # un-patch the torch operators
 
-    with disable():
+    with disable_patching():
         signature = inspect.signature(_ORIG_JIT_SCRIPT)
         bound_args = signature.bind(*args, **kwargs).arguments
         # Process the case when the object-to-script is a class as in the original jit.script logic
@@ -218,7 +218,7 @@ def torch_jit_script_wrapper(*args, **kwargs):
 
 
 def torch_jit_trace_make_module_wrapper(*args, **kwargs):
-    with disable():
+    with disable_patching():
         retval = _ORIG_JIT_TRACE_MAKE_MODULE(*args, **kwargs)
     return retval
 
@@ -408,7 +408,7 @@ def unpatch_torch_operators():
 
 
 @contextmanager
-def disable():
+def disable_patching():
     was_patched = _OPERATORS_ALREADY_WRAPPED
     if was_patched:
         unpatch_torch_operators()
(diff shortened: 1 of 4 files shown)


nncf/torch/dynamic_graph/patch_pytorch.py
```diff
@@ -201,7 +201,12 @@ def torch_jit_script_wrapper(*args, **kwargs):
             frames_up = bound_args.get("_frames_up", 0)
             rcb = createResolutionCallbackFromFrame(frames_up + 1)
             kwargs["_rcb"] = rcb
-        retval = _ORIG_JIT_SCRIPT(*args, **kwargs)
+        try:
+            retval = _ORIG_JIT_SCRIPT(*args, **kwargs)
+        except Exception as e:
```
nikita-savelyevv, line 206: Could you please add a small comment here for when this is needed? May be similar to the PR description

nncf/torch/dynamic_graph/patch_pytorch.py
```diff
@@ -201,7 +201,12 @@ def torch_jit_script_wrapper(*args, **kwargs):
...
+            if apply_unpatch:
+                patch_torch_operators()
+            raise e
     else:
```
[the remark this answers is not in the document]
vshampor: I implemented a context manager for this purpose since the code seems to be recurring, check it out.

nncf/torch/dynamic_graph/patch_pytorch.py
vshampor: Commented in the contextmanager

nncf/torch/dynamic_graph/patch_pytorch.py
```diff
@@ -415,3 +405,17 @@ def unpatch_torch_operators():
 
     for orig_op_info in ORIGINAL_OPERATORS:
         setattr(orig_op_info.namespace, orig_op_info.name, orig_op_info.op)
+
+
+@contextmanager
+def disable():
```
nikita-savelyevv, line 411: The name is kind of ambiguous, moreover if we make it available from `nncf.torch`. I suggest `patching_disabled` or something like that.

vshampor: Renamed. Note that this context manager is not yet part of the API (although I imagine it will be in the future) since it is not decorated as such via `@api`.

nikita-savelyevv (approved):

vshampor:
run pytorch pre-commit tests

PR #2082, opened 2023-08-24
[PTQ][OV] Weights layout in layer attributes is introduced

daniil-lyakhov:
### Changes

1) Introduce Layout parameters for Convolutional / MM layers for OV backend
2) Refactor ChannelAlignment/SmoothQuant/MinMax/Wegihts compression algorithms to use layout parameters from layer attributes
3) Fix `get_matmul_channel_axes` from openvino node utils

### Reason for changes

1-2) To separate weights layout collection code in layer attributes instead of separate collection in each algorithm
3) To work properly with 1d constants in PTQ algorithms

### Related tickets

114328
114583

### Tests

* tests/openvino/native/test_layer_attributes.py are updated
* tests/openvino/native/quantization/test_channel_alignment.py
* tests/openvino/native/test_smooth_quant.py
* tests/openvino/native/test_node_utils.py


daniil-lyakhov:
TODO: fix MM case with 1d weight

alexsu52:
Please, rebase PR.

tests/openvino/native/test_node_utils.py
```diff
@@ -60,10 +62,10 @@ def test_is_node_with_bias(model_to_create, is_with_bias, node_name):
 @pytest.mark.parametrize(
     "weights_port_id, transpose, shape, expected_channel_axes",
     [
-        (0, False, (1,), [0]),
+        (0, False, (1,), []),
         (0, True, (1,), []),
         (1, False, (1,), []),
-        (1, True, (1,), [0]),
+        (1, True, (1,), []),
         (0, False, (1, 1), [0]),
```
daniil-lyakhov, line 69: As we found, ["Transpose attributes are ignored for 1D tensors."](https://docs.openvino.ai/2023.0/openvino_docs_ops_matrix_MatMul_1.html), so tests were adjusted accordingly

nncf/common/graph/layer_attributes.py
```diff
@@ -30,6 +37,9 @@ class BaseLayerAttributes(ABC):
     def __eq__(self, __o: object) -> bool:
         return isinstance(__o, self.__class__) and self.__dict__ == __o.__dict__
 
+    def get_backend_agnostic_attributes(self) -> "BaseLayerAttributes":
+        return self
```
daniil-lyakhov, line 41: For future: move all backend-specific layer attributes e.g. `activation_attributes`, `weights_attributes` to common attributes to minimize differences between backends

daniil-lyakhov:

nncf/common/graph/layer_attributes.py
daniil-lyakhov, line 1: @vshampor, please review changes in common layer attributes

nncf/common/graph/layer_attributes.py
```diff
@@ -13,14 +13,21 @@
...
-from typing import Any, List, Tuple, Union
+from typing import Any, List, Optional, Tuple, Union
 
 
 class Dtype(Enum):
     FLOAT = "float"
     INTEGER = "int"
 
 
+class LayoutElem(Enum):
```
vshampor, line 24: Needs a docstring describing exactly what each element means, with examples from relevant backends

vshampor (changes requested):

nncf/openvino/graph/layer_attributes.py
```diff
@@ -79,51 +81,71 @@ def get_weighted_layer_attributes(
...
+            "padding_values": tuple(node_attrs["pads_begin"] + node_attrs["pads_end"]),
+        }
+
+        weights_layout_map = {
+            OVConvolutionMetatype: [LayoutElem.C_OUT, LayoutElem.C_IN],
+            OVGroupConvolutionMetatype: [LayoutElem.GROUPS, LayoutElem.C_OUT, LayoutElem.C_IN],
+            OVDepthwiseConvolutionMetatype: [LayoutElem.GROUPS, LayoutElem.C_OUT, LayoutElem.C_IN],
+            OVConvolutionBackpropDataMetatype: [LayoutElem.C_IN, LayoutElem.C_OUT],
+            OVGroupConvolutionBackpropDataMetatype: [LayoutElem.GROUPS, LayoutElem.C_IN, LayoutElem.C_OUT],
+        }
```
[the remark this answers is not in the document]
daniil-lyakhov: > Also, it's not clear why the layout for `Convolution` contains only `out` and `in` channels. It also has other dimensions. Shouldn't we introduce a full layout for consistency? It might be useful to determine the full layout while building `NNCFGraph` instead of calculating in algorithms like `ChannelAlignment`.

Full layout is being calculated below in this function for both convolutions and matmuls 
Convs:
https://github.com/openvinotoolkit/nncf/blob/0982ce472dfb2f05c70f044710695796f3e835b8/nncf/openvino/graph/layer_attributes.py#L125

MM:
https://github.com/openvinotoolkit/nncf/blob/0982ce472dfb2f05c70f044710695796f3e835b8/nncf/openvino/graph/layer_attributes.py#L131

nncf/openvino/graph/layer_attributes.py
```diff
@@ -79,51 +81,71 @@ def get_weighted_layer_attributes(
...
+
+        return ConvolutionLayerAttributes(**kwargs)
+    if ov_metatype == OVMatMulMetatype:
+        weights_shape = attrs["shape"]
+
+        weights_layout = [LayoutElem.SPATIAL] * (len(weights_shape) - 2)
+        if len(weights_shape) > 1:
+            transpose = attrs.get("transpose", False)
+            if (transpose and port_id == 0) or (not transpose and port_id == 1):
+                weights_layout += [LayoutElem.C_IN, LayoutElem.C_OUT]
```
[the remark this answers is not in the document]
daniil-lyakhov: Could you please give an example where `C_OUT` or `C_IN` is processed differently for convolutions and MM?

nncf/openvino/graph/node_utils.py
```diff
@@ -321,35 +323,27 @@ def get_weight_channel_axes(node: NNCFNode, weights_port_id: int) -> List[int]:
...
-        channel_axes.append(matmul_channel_axis)
-    return channel_axes
+    assert isinstance(node.layer_attributes, OVLayerAttributes)
+    layer_attributes = node.layer_attributes.get_backend_agnostic_attributes()
+    assert isinstance(layer_attributes, LinearLayerAttributes)
+    return [
+        idx
+        for idx, elem in enumerate(layer_attributes.weights_layout)
+        if elem in [LayoutElem.SPATIAL, LayoutElem.C_OUT]
+    ]
```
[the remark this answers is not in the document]
daniil-lyakhov: Sure, transpose logic is moved to layer attributes creation
https://github.com/openvinotoolkit/nncf/blob/0982ce472dfb2f05c70f044710695796f3e835b8/nncf/openvino/graph/layer_attributes.py#L132-L139

nncf/openvino/quantization/weights_compression.py
```diff
@@ -56,7 +58,8 @@ def insert_pre_compression_operations(model: ov.Model, bits: int = 8) -> None:
                 continue
 
             weight = get_const_value(weight_node)
-            axes = _get_reduction_axes(metatype, node, const_port_id)
+            weight_shape = weight_node.get_output_shape(0)
```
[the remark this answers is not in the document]
daniil-lyakhov: As weights node is the result of `get_operation_const_op`, constant should be located correctly in specified case. As constant value has only one output port it is fair to take shape of constant from the zero output port

nncf/openvino/graph/layer_attributes.py
```diff
@@ -79,51 +81,71 @@ def get_weighted_layer_attributes(
...
+        return None
+
+    port_id, attrs = constant_attributes.copy().popitem()
+    if ov_metatype in [
+        OVConvolutionMetatype,
+        OVDepthwiseConvolutionMetatype,
+        OVGroupConvolutionMetatype,
+        OVConvolutionBackpropDataMetatype,
+        OVGroupConvolutionBackpropDataMetatype,
+    ]:
```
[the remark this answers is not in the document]
daniil-lyakhov: `OPERATIONS_WITH_WEIGHTS` already exist in OV backend, but it has LSTM, GRU and Embedding layers as well 
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/openvino/graph/metatypes/groups.py#L165-L175

nncf/common/graph/layer_attributes.py
daniil-lyakhov: Done

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
daniil-lyakhov: > Can we place this mapping into the metatype attributes?

Done

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
KodiaqQ: It might be the marker that our graph builder is not applicable for all types from the `OPERATIONS_WITH_WEIGHTS` or `OPERATIONS_WITH_WEIGHTS` contains not correct meta types, isn't it?
At least, I'm not sure that it's good practice to hard-code such lists in the methods.

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
KodiaqQ: `SmoothQuant` and `MinMax` have different understandings of the question of the `channel axis`, for example.

nncf/openvino/quantization/weights_compression.py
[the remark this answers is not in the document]
KodiaqQ: The question is not about the `constant axis`, but about the correctness of the shape that we looking into.
Shapes after the `Reshape` node in this pattern and after the `Constant` node may be different.

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
daniil-lyakhov: Yes, but they are using different dims: `C_in` in the `SmoothQuant` and `Spatial` + `C_out` in min max

nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -359,3 +360,10 @@ def _create_scale_node_name(self, source_name: str, source_port_id: int) -> str:
         unique_index = self._cached_multiply_names[scale_node_name]
         self._cached_multiply_names[scale_node_name] += 1
         return f"{scale_node_name}_{unique_index}/sq_multiply"
+
+    @staticmethod
+    def _get_weight_channel_axis(node: NNCFNode) -> int:
+        layer_attributes = node.layer_attributes.get_backend_agnostic_attributes()
+        if layer_attributes is None or layer_attributes.weights_layout is None:
+            return 1
```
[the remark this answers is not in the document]
daniil-lyakhov: Moved it to OV backend

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,68 @@ def get_const_port_ids(self) -> List[int]:
...
+        weights_shape=constant_layer_attrs["shape"],
+        transpose=constant_layer_attrs.get("transpose", False),
+        port_id=port_id,
+    )
+
+
+def _get_port_id_from_layer_attributes(layer_attributes):
+    port_ids = list(layer_attributes.constant_attributes.keys())
+    assert len(port_ids) == 1
+    return port_ids[0]
```
l-bat, line 108: Is it possible to call this function for `LSTM` or `GRU` nodes?

nncf/quantization/algorithms/channel_alignment/algorithm.py
```diff
@@ -373,12 +387,13 @@ def get_statistic_points(self, model: TModel, graph: NNCFGraph) -> StatisticPoin
         statistic_container = StatisticPointsContainer()
         for conv_in, add_in, _ in self._get_node_pairs(graph):
             target_point, node_in = self._get_target_point_and_node_in(conv_in, add_in)
+
             channel_axis = conv_in.metatype.output_channel_axis
-            reduction_axes = list(range(len(graph.get_output_edges(node_in)[0].tensor_shape)))
-            reduction_axes.remove(channel_axis)
+            activation_shape = list(range(len(graph.get_output_edges(node_in)[0].tensor_shape)))
+            reduction_shape = self._backend_entity.get_channel_agnostic_reduction_axes([channel_axis], activation_shape)
```
l-bat, line 393: I'm guessing that `get_channel_agnostic_reduction_axes` should return `reduction_axes`

nncf/quantization/algorithms/channel_alignment/algorithm.py
```diff
@@ -112,10 +113,23 @@ def filter_func(point: StatisticPoint) -> bool:
...
             conv_in_cont = ConvParamsContainer(conv_in, model, graph, self._backend_entity)
             conv_out_cont = ConvParamsContainer(conv_out, model, graph, self._backend_entity)
+            if (
+                conv_in_cont.dims.conv_weight_out_channels_dim is None
+                or conv_out_cont.dims.conv_weight_out_channels_dim is None
+            ):
+                nncf_logger.debug(
+                    f"Skipping channel alignment for pairs {conv_in.node_name}, {conv_out.node_name} "
+                    " because one of the node is 1D MatMul, 1D Matmuls are not supported by CA algortihm yet."
+                )
```
l-bat, line 131: Could you please explain this message? Is `conv_in` a Convolution or MatMul node?

nncf/quantization/algorithms/channel_alignment/openvino_backend.py
```diff
@@ -102,52 +106,34 @@ def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
...
+            weights_layout = get_conv_weights_layout_from_node(node=node)
+        elif node.metatype == OVMatMulMetatype:
+            weights_layout = get_linear_weights_layout_from_node(node=node)
+        else:
+            raise RuntimeError(
+                f"Metatype {node.metatype} of node {node.node_name} dimensions description retrieving is not supported"
             )
-        if node.metatype in [OVGroupConvolutionMetatype, OVDepthwiseConvolutionMetatype]:
+
+        if OVConvLayoutElem.GROUPS in weights_layout:
```
l-bat, line 119: Should it work with `OVGroupConvolutionBackpropDataMetatype`?

l-bat:
Could we test these changes on several DLB (~5) models? I suggest including models with `LSTM`, `GroupConv` and `GroupConvBackprop` operations.

nncf/openvino/graph/node_utils.py
```diff
@@ -350,35 +351,23 @@ def get_weight_channel_axes(node: NNCFNode, weights_port_id: int) -> List[int]:
...
-    if (weights_port_id == 1) == transpose:
-        matmul_channel_axis -= 1
-    matmul_channel_axis = max(ndims, 2) + matmul_channel_axis
-    channel_axes = list(range(ndims - 2))
-    if matmul_channel_axis < ndims:
-        channel_axes.append(matmul_channel_axis)
-    return channel_axes
+    weights_layout = get_linear_weights_layout_from_node(node)
+    return [
+        idx for idx, elem in enumerate(weights_layout) if elem in [OVConvLayoutElem.SPATIAL, OVConvLayoutElem.C_OUT]
```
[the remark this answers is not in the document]
daniil-lyakhov: I named them this way as Matmul is a special case of a convolution layer and has the same layout elements in OV backend. I don't mind to add `OVMatmulLayoutElem`, but I'm not sure it is needed

nncf/openvino/graph/layer_attributes.py
daniil-lyakhov: This is axillary function applicable only for Conv and MatMul layers. LSTM and GRU are using `GenericWeightedLayerAttributes` for now

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,68 @@ def get_const_port_ids(self) -> List[int]:
...
+    )
+
+
+def get_linear_weights_layout_from_node(node: NNCFNode) -> OVConvLayout:
+    layer_attributes = node.layer_attributes
+    port_id = _get_port_id_from_layer_attributes(layer_attributes)
+    constant_layer_attrs = layer_attributes.constant_attributes[port_id]
+    return get_linear_weights_layout(
+        weights_shape=constant_layer_attrs["shape"],
+        transpose=constant_layer_attrs.get("transpose", False),
```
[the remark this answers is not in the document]
daniil-lyakhov: From the NNCFGraph building standpoint it looks like transpose attribute is always present, right. Should I refactor it to use `constant_layer_attrs["transpose"]`?

nncf/quantization/algorithms/channel_alignment/algorithm.py
daniil-lyakhov: The condition is that at least one of the matched operation is a 1d MM, which means it has no output dimension. We can solve it by reshaping weights, but I suggest to keep it for future PRs.

nncf/quantization/algorithms/channel_alignment/openvino_backend.py
daniil-lyakhov: It should work correctly as ` get_conv_weights_layout` function supports `OVGroupConvolutionBackpropDataMetatype`

nncf/quantization/algorithms/channel_alignment/algorithm.py
daniil-lyakhov: Done

nncf/openvino/quantization/weights_compression.py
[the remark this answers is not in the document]
daniil-lyakhov: Outdated

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,94 @@ def get_const_port_ids(self) -> List[int]:
...
+        port_id=port_id,
+    )
+
+
+def _get_port_id_from_layer_attributes(layer_attributes) -> int:
+    port_ids = list(layer_attributes.constant_attributes.keys())
+    assert len(port_ids) == 1
+    return port_ids[0]
```
l-bat, line 120: Please add a docstring that states this function should be used for MatMul and Convs.

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
l-bat: Yes

nncf/openvino/graph/layout.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class OVConvLayoutElem(Enum):
+    """
+    Layout elements descriptor for convolutional and linear openvino layers:
+        C_IN: Input channels dimension.
+        C_OUT: Output channels dimension.
+        SPATIAL: Spatial dimension.
+        GROUPS: Groups dimention.
+    """
```
l-bat, line 22: I think we can use general name - `OVLayoutElem`

nncf/quantization/algorithms/channel_alignment/algorithm.py
l-bat: O, but `conv_in` or `conv_out` name for a MatMul node is confusing.

nncf/openvino/graph/layout.py
daniil-lyakhov: Agree, done

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,100 @@ def get_const_port_ids(self) -> List[int]:
...
+    :return: Target convolution Node weights layout.
+    """
+    layer_attributes = node.layer_attributes
+    port_id = _get_constant_port_id_from_layer_attributes(layer_attributes)
+    return get_conv_weights_layout(
+        ov_metatype=node.metatype, weights_shape=layer_attributes.constant_attributes[port_id]["shape"]
+    )
+
+
+def get_linear_weights_layout_from_node(node: NNCFNode) -> OVConvLayout:
```
l-bat, line 100: Again, why are you returning `OVConvLayout` for linear ops? For me `List[OVLayoutElem]` is less confusing

daniil-lyakhov: That is fair, I've fixed it, thanks

l-bat (approved):

nncf/openvino/graph/layer_attributes.py
```diff
@@ -79,51 +81,71 @@ def get_weighted_layer_attributes(
...
         else:
-            common_layer_attr = GenericWeightedLayerAttributes(
-                weight_requires_grad=False, weight_shape=attrs.get("shape", None)
-            )
-        retval[port_id] = common_layer_attr
-    return retval
+            weights_layout += [LayoutElem.C_IN]
+
+        kwargs = {
+            "weight_requires_grad": False,
```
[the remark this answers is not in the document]
daniil-lyakhov: This is Torch QAT legacy parameter, I don't think we need to refactor this. If you think it worth refactoring let's add this to the refactoring list

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
daniil-lyakhov: This is how it's done in our QAT backends:
Torch: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/dynamic_graph/layer_attributes_handlers.py#L68
TF: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/tensorflow/graph/converter.py#L745

We need a subset of metatypes that have the same type of layer attributes, and this subset is not always equal to `OPERATIONS_WITH_WEGIHTS` metatypes subset

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,97 @@ def get_const_port_ids(self) -> List[int]:
...
+    Calculates weights layout for a target convolution node.
+
+    :param ov_metatype: Target convolution node OpenVINO metatype.
+    :param weights_shape: Shape of the target convolution node weight.
+    :return: Target convolution node weights layout.
+    """
+    weights_layout = ov_metatype.const_layout
+    kernel_size = weights_shape[len(weights_layout) :]
+    weights_layout += [OVLayoutElem.SPATIAL] * len(kernel_size)
+    return tuple(weights_layout)
```
KodiaqQ, line 137: Can we be sure that weight layout calculation is possible? For example, it is possible that between weights and the target layer, some reshapes are placed.

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
KodiaqQ: I meant that we shouldn't define some parameters in the OpenVINO part that do not relate to OpenVINO itself.

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,97 @@ def get_const_port_ids(self) -> List[int]:
...
+    :param weights_shape: Shape of the target linear node weight.
+    :param port_id: Port id of the target liner node weights.
+    :return: Target linear node weight layout.
+    """
+    weights_layout = [OVLayoutElem.SPATIAL] * (len(weights_shape) - 2)
+    if len(weights_shape) > 1:
+        if (transpose and port_id == 0) or (not transpose and port_id == 1):
+            weights_layout += [OVLayoutElem.C_IN, OVLayoutElem.C_OUT]
+        else:
+            weights_layout += [OVLayoutElem.C_OUT, OVLayoutElem.C_IN]
```
KodiaqQ, line 153: Minor. We have an example in the `calculate_port_based_channel_axis` method that looks more reliable than a bunch of the `and/or` conditions.

nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -359,7 +359,7 @@ def _process_weight_statistics(self, node: NNCFNode, weights: TTensor, port_id:
         """
         channel_axis = 0
         if len(weights.shape) > 1:
-            channel_axis = self._backend_entity.get_weight_channel_axis(node, port_id)
+            channel_axis = self._backend_entity.get_weight_channel_axis(node)
```
KodiaqQ, line 362: We may drop the `port_id` method's attribute then.

tests/openvino/native/test_smooth_quant.py
```diff
@@ -79,18 +80,47 @@ def test_get_activation_channel_axis(self, node_metatype, layer_attributes, port
...
+            (
+                OVMatMulMetatype,
+                (
+                    OVLayoutElem.SPATIAL,
+                    OVLayoutElem.SPATIAL,
+                    OVLayoutElem.SPATIAL,
+                    OVLayoutElem.C_IN,
+                    OVLayoutElem.C_OUT,
+                ),
+                3,
```
KodiaqQ, line 104: Why 3 `SPATIAL`? Why 2 was changed to 3?

nncf/quantization/algorithms/smooth_quant/algorithm.py
daniil-lyakhov: Thank you, done

nncf/openvino/graph/layer_attributes.py
```diff
@@ -67,9 +67,97 @@ def get_const_port_ids(self) -> List[int]:
...
+        return self._layer_attributes
+
+
+OV_CONV_METATYPES = [
+    OVConvolutionMetatype,
+    OVDepthwiseConvolutionMetatype,
+    OVGroupConvolutionMetatype,
+    OVConvolutionBackpropDataMetatype,
+    OVGroupConvolutionBackpropDataMetatype,
+]
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/openvino/graph/layer_attributes.py
daniil-lyakhov: I would leave it as it is, looks like this if is covering all possible cases for the matmul operation

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
KodiaqQ: @alexsu52, what is the definition of the `backend-agnostic layer attributes`? Is it a mix of all possible backend attributes or the minimal common, but backend-enough set of them? Based on the definition, we may decide if the changes are legit or not.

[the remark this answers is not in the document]
alexsu52: `backend-agnostic layer attributes` are attributes that presented into all backends.

KodiaqQ (approved):

nncf/openvino/graph/layer_attributes.py
```diff
@@ -68,8 +70,84 @@ def get_const_port_ids(self) -> List[int]:
...
+        if (transpose and port_id == 0) or (not transpose and port_id == 1):
+            weights_layout += [OVLayoutElem.C_IN, OVLayoutElem.C_OUT]
+        else:
+            weights_layout += [OVLayoutElem.C_OUT, OVLayoutElem.C_IN]
+    else:
+        weights_layout += [OVLayoutElem.C_IN]
+    return tuple(weights_layout)
+
+
 def get_weighted_layer_attributes(
```
alexsu52, line 149: Is my statement correct that `OVLayoutElem` is not part of Layer_attributes but is only used for calculation attributes like "in_channels", "out_channels", "kernel_size", "in_features", "out_features"...? What are the benefits of using OVLayoutElem? Could you give some code examples where the code is simpler and more extensible?

nncf/openvino/graph/metatypes/openvino_metatypes.py
```diff
@@ -99,7 +104,8 @@ class OVGroupConvolutionBackpropDataMetatype(OVOpMetatype):
     name = "GroupConvolutionBackpropDataOp"
     op_names = ["GroupConvolutionBackpropData"]
     hw_config_names = [HWConfigOpName.CONVOLUTION]
-    const_channel_axis = [0, 2]  # const layout: [GROUPS, C_IN / GROUPS,  C_OUT / GROUPS, Z, Y, X]
+    const_channel_axis = [0, 2]
+    const_layout = [OVLayoutElem.GROUPS, OVLayoutElem.C_IN, OVLayoutElem.C_OUT]
```
alexsu52, line 108: It looks like you add new abstraction but don't remove the previous representation `const_channel_axis`. What's the plan here?

nncf/quantization/algorithms/channel_alignment/backend.py
```diff
@@ -152,3 +151,14 @@ def create_bias_tensor(node: NNCFNode, nncf_graph: NNCFGraph, value: Any) -> np.
         :param value: Value to fill bias constant array.
         :return: Bias value constant array filled by given value.
         """
+
+    @staticmethod
+    @abstractmethod
+    def get_channel_agnostic_reduction_axes(channel_axis: int, shape: Tuple[int]) -> Tuple[int]:
+        """
+        Returns filtered reduction shape without axes that corresponds channels.
```
alexsu52, line 159: An example will be helpful here.

nncf/openvino/graph/layer_attributes.py
```diff
@@ -68,8 +70,84 @@ def get_const_port_ids(self) -> List[int]:
...
+    """
+    Calculates weights layout for a target convolution node.
+
+    :param ov_metatype: Target convolution node OpenVINO metatype.
+    :param weights_shape: Shape of the target convolution node weight.
+    :return: Target convolution node weights layout.
+    """
+    weights_layout = ov_metatype.const_layout
+    kernel_size = weights_shape[len(weights_layout) :]
+    weights_layout += [OVLayoutElem.SPATIAL] * len(kernel_size)
```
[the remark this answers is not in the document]
daniil-lyakhov: Because layout depends on kernel size:
Conv 1D: [C_OUT, C_IN, X]
Conv 2D: [C_OUT, C_IN, X, Y]
Conv 3D: [C_OUT, C_IN, X, Y, Z] and etc.
But we have only one metatype for all cases: `OVConvolutionMetatype`. Is there any way we can calculate layout inside conv metatypes I don't aware of?

nncf/openvino/graph/layer_attributes.py
daniil-lyakhov: It is unifying search for specific axes, output/input channel axis for example.
Place that were unified:

https://github.com/openvinotoolkit/nncf/pull/2082/files#diff-d216bff84a765d765e444e276cb1440fbc3f085257d2fc3b71bcb56c3a71c467L113-L140 
https://github.com/openvinotoolkit/nncf/pull/2082/files#diff-7bc5389e936e911a4b240a4617740c5549a27386437a0508e56099d3a5dfaa1cL170-L181

there were more unified places 2 month before when this pr was introduced, but it was unified already

nncf/openvino/graph/layer_attributes.py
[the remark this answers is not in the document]
alexsu52: Essentially that's what I mean. You introduced the abstraction that can not fully specify the constant layout of `OVConvolutionMetatype`. However, you just need C_OUT, C_IN and GROUPS dimensions.

nncf/openvino/graph/layer_attributes.py
alexsu52: Would it be easier to introduce `get_linear_weights_channel_input_dimension`?

nncf/openvino/graph/metatypes/openvino_metatypes.py
daniil-lyakhov: I've removed `const_channel_axis` for each metatype which have `const_layout`

alexsu52 (approved):
LGTM

PR #2083, opened 2023-08-24
Deprecate some methods for NNCF 2.6 release

nikita-savelyevv:
### Changes

Some methods were left for backward compatibility after #1584 and are due for deletion before NNCF 2.6 release.


nikita-savelyevv:
run pytorch pre-commit tests

PR #2090, opened 2023-08-30
Skip SmoothQuant creation if alpha is negative

nikita-savelyevv:
### Changes

Don't create Smooth Quant algorithm if alpha parameter is negative.

### Reason for changes

When running AA with `tune_hyperparams=True` and `smooth_quant_alpha<0`, it fails here https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/post_training/algorithm.py#L168 because SQ had been created. This fix makes it so SQ is not created if it won't be used.

### DIFF
--- a/nncf/quantization/algorithms/smooth_quant/algorithm.py
+++ b/nncf/quantization/algorithms/smooth_quant/algorithm.py
@@ -63,6 +63,10 @@ class SmoothQuant(Algorithm):
         :param alpha: The parameter that regulates the calculation of the scale.
             The default value is 0.95. Negative value switches off the algorithm.
         """
+
+        if alpha < 0:
+            raise RuntimeError("Smooth Quant algorithm does not support negative alpha parameter!")
+
         super().__init__()
         self._subset_size = subset_size
         self._inplace_statistics = inplace_statistics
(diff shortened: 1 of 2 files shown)


nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -63,6 +63,10 @@ def __init__(self, subset_size: int = 300, inplace_statistics: bool = True, alph
         :param alpha: The parameter that regulates the calculation of the scale.
             The default value is 0.95. Negative value switches off the algorithm.
         """
+
+        if alpha < 0:
+            raise RuntimeError("Smooth Quant algorithm does not support negative alpha parameter!")
```
KodiaqQ, line 68: I don't know, maybe should we keep `if self._alpha < 0:` conditions or do not add `RuntimeError`. @alexsu52, what do you think?

nikita-savelyevv:
run pytorch pre-commit tests

nncf/quantization/algorithms/smooth_quant/algorithm.py
alexsu52: LGTM. @AlexKoff88, do you have a comment?

AlexKoff88: I am ok with this too.

AlexKoff88 (approved):

alexsu52 (approved):

KodiaqQ (approved):

PR #2094, opened 2023-08-31
Reworking the placement of metatype groups

andrey-churkin:
### Changes

Create a separate module to store metatype groups in one location.

### Reason for changes

Ref: 113580

### Related tickets

Ref: 113580

### Tests

N/A


nncf/openvino/graph/metatypes/groups.py
```diff
@@ -46,34 +47,38 @@
 ]
 
 
-QUANTIZABLE_OPERATIONS = [
+INPUTS_QUANTIZABLE_OPERATIONS = [
```
andrey-churkin, line 50: @KodiaqQ @alexsu52 I noticed that `ov_metatypes.OVDeformableConvolutionMetatype` is missing here, but it is present in `LINEAR_OPERATIONS` list and consequently used in fused patterns. So, my question is whether this is expected behavior and we should not quantize the `DeformableConvolution` operation, or if it is a bug that should be fixed?"

nncf/onnx/graph/metatypes/groups.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+MATMUL_METATYPES = [onnx_metatypes.ONNXGemmMetatype, onnx_metatypes.ONNXMatMulMetatype]
+
+
+INPUTS_QUANTIZABLE_OPERATIONS = [
+    onnx_metatypes.ONNXConvolutionMetatype,
+    onnx_metatypes.ONNXDepthwiseConvolutionMetatype,
+    onnx_metatypes.ONNXConvolutionTransposeMetatype,
+    *MATMUL_METATYPES,
```
KodiaqQ, line 54: IMHO, it's better to place this at the end of the list.

nncf/openvino/graph/metatypes/groups.py
KodiaqQ: Never met `DeformableConvolution` in the OpenVINO models. Need to an example.

KodiaqQ (approved):

nncf/openvino/graph/metatypes/groups.py
l-bat: It isn't a bug. `DeformableConvolution` should be fused but runtime doesn't support INT8 execution (ticket: 106633)

l-bat (approved):

kshpv (approved):

alexsu52 (approved):
Thanks for the contribution!

PR #2098, opened 2023-09-01
Constrain filelock package version

nikita-savelyevv:
### Changes

Add package constraint `filelock<3.12.3`.

### Reason for changes

Installing tensorflow test environment produces the error:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tensorflow 2.12.1 requires typing-extensions<4.6.0,>=3.6.6, but you have typing-extensions 4.7.1 which is incompatible.
```

Starting from `filelock==3.12.3`, it requires `typing-extensions>=4.7.1` so downgrading to `filelock==3.12.2` resolves the issue.

This is temporary measure unless we upgrade to TF 2.13 which does not have `typing-extensions<4.6.0` constraint.

### DIFF
--- a/tests/tensorflow/requirements.txt
+++ b/tests/tensorflow/requirements.txt
@@ -10,5 +10,9 @@ pydot
 tensorflow_hub
 virtualenv
 
+# filelock 3.12.3 requires typing-extensions>=4.7.1, but tensorflow 2.12.1 requires typing-extensions<4.6.0,>=3.6.6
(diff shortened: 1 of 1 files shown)


alexsu52:
run pytorch pre-commit tests

alexsu52 (approved):
Thanks!

PR #2101, opened 2023-09-01
[OV] Introduce support of quantization If operation

kshpv:
### Changes

Add support of quantization for OpenVINO models with If operation - https://docs.openvino.ai/2023.0/openvino_docs_ops_infrastructure_If_8.html

### Reason for changes

Customer request / Obtain the maximum performance for models with inner subgraphs under If operation

### Related tickets

113826

### Tests

Tested on a model attached to a ticket
Add synthetic model with If operation


nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -86,75 +87,86 @@ def __init__(
...
+        self.target_device = target_device
+        self.subset_size = subset_size
+        self.fast_bias_correction = fast_bias_correction
+        self.model_type = model_type
+        self.ignored_scope = ignored_scope
+        self.dump_intermediate_models = dump_intermediate_models
+        self.advanced_parameters = advanced_parameters
+        self.reset()
+
+    def reset(self):
```
alexsu52, line 100: Why do you need to reset algorithms?

nncf/quantization/algorithms/post_training/openvino_backend.py
```diff
@@ -0,0 +1,101 @@
...
+            {"if_op": if_op, "if_op_subgraph_port_id": port_id},
+        )
+
+    def make_tasks(
+        self, model: ov.Model, calibration_dataset: Dataset, subset_size: int
+    ) -> Deque[Tuple[ov.Model, Dataset, Dict[str, Any]]]:
+        tasks = deque()
+        for op in model.get_ops():
+            if op.get_type_name() == "If":
+                subgraph_dataset = self._collect_dataset(model, op, calibration_dataset, subset_size)
```
alexsu52, line 98: It looks like not optimal solution by memory footprint.

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,6 @@
+from abc import ABC
+from abc import abstractmethod
+
+
+class PostTrainingBackend(ABC):
```
alexsu52, line 5: Could you specify `PostTrainingBackend` methods?

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: Reverted. Added a logic into MinMax to reset inner state when a new model comes for quantization

nncf/quantization/algorithms/post_training/backend.py
kshpv: Update `PostTrainingBackend` methods, while it mostly was inspired by OpenVINO backend, so, probably there is a better way to separate backend methods from the common part.

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: To reduce memory footprint the solution through DFS was implemented

KodiaqQ (commented):
Need to have explanations for each method that was added. Also, the changes for the `PostTrainingQuantization` are not clear, including adding the backend entity for it.
The general approach and reasons for it are not clear.

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+
+        :param model: Model to infer to collect dataitems.
+        :param calibration_dataset: Dataset is used to collect new dataitems.
+        :param subset_size: Size of dataitems to collect
+        :param model_cnt: Global model number.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def make_dataset_for_child_models(dataitems: Iterable[DataItem], backend_params: Dict[str, Any]) -> Dataset:
```
alexsu52, line 39: Is it possible to make this method general?

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Chechks whether a model has inner subgraphs to quantize.
+
+        :param model: Model to check.
+        :return: True if the model has no inner subgraphs, otherwise - False.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def get_child_models(model: TModel) -> List[Tuple[TModel, Dict[str, Any]]]:
```
alexsu52, line 59: Is it possible to make this method general via using ModelTransformer abstraction?

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Returns all child models of passed model.
+
+        :param model: Model to seek for child models.
+        :return: Models with backend specific parameters.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def add_additional_outputs(model: TModel) -> TModel:
```
alexsu52, line 69: It looks like only OpenVINO specific.

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+
+from nncf import Dataset
+from nncf.data.dataset import DataItem
+from nncf.quantization.algorithms.post_training.algorithm import TModel
+
+
+class PostTrainingBackend(ABC):
+    @staticmethod
+    @abstractmethod
+    def collect_dataitems_for_children_models(
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+        """
+        Return dataset for child models.
+
+        :param dataitems: Data items to collect into dataset.
+        :param backend_params: Backend-specific parameters.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def is_single_model(model: TModel) -> bool:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,96 @@
...
+        Save a model to a directory. Backend params are used to determine the model name to dump.
+
+        :param model: Model to dump.
+        :param dir: Directory path.
+        :param backend_params: Backend specific parameters.
+        """
+
+    @staticmethod
+    @abstractmethod
+    def set_child_model(child_model: TModel, backend_params: Dict[str, Any]) -> None:
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -199,31 +236,76 @@ def apply(
...
+        model: TModel,
+        graph: NNCFGraph,
+        statistic_points: Optional[StatisticPointsContainer] = None,
+        dataset: Optional[Dataset] = None,
+    ) -> TModel:
+        model_copy = copy_model(model)
+        if self._is_single_model(model_copy):
+            return self._apply(model_copy, graph, statistic_points, dataset)
+        self._set_backend_entity(model)
+        nncf_logger.info("The model consists of child submodels. The iteratively each submodel will be quantized.")
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+        :return: Updated model with a new submodel.
+        """
+        transformation_layout = TransformationLayout()
+        command = self._backend_entity.create_update_subgraph_command(if_node, if_submodel_condition, submodel)
+        transformation_layout.register(command)
+        return model_transformer.transform(transformation_layout)
+
+    def _add_outputs_before_if_node(
+        self, model_transformer: ModelTransformer, model: TModel, if_node: NNCFNode
+    ) -> TModel:
```
KodiaqQ, line 342: Why do we add outputs before the `if` node?

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+                    )
+
+                    child_dataset = self._make_dataset_for_child_model(
+                        factory.EngineFactory.create(parent_model_with_additional_outputs),
+                        parent_dataset,
+                        if_cond_input_name,
+                        child_model_input_names,
+                        if_submodel_condition,
+                        model_cnt,
+                    )
```
KodiaqQ, line 396: Can't we collect the outputs once for all subgraphs? It looks like we can.

nncf/openvino/graph/model_transformer.py
```diff
@@ -526,3 +530,42 @@ def _apply_multiply_insertion_transformations(
...
+        :param transformations: Transformations with information from which
+        If node and input port extract a model subgraph.
+        :return: Model subgraph.
+        """
+        transformation = transformations[-1]
+        name_to_node_mapping = OVModelTransformer._get_name_to_node_mapping(model)
+        ov_node = name_to_node_mapping[transformation.if_node.node_name]
+        if transformation.if_submodel_condition:
+            return ov_node.get_function(0)
+        return ov_node.get_function(1)
```
KodiaqQ, line 571: Is it the new `ov.Model` instance or just a link to the object?

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+
+                    child_quantized_model, model_cnt = self._dfs_quantize_models(
+                        child_model, NNCFGraphFactory.create(child_model), child_dataset, None, model_cnt
+                    )
```
KodiaqQ, line 400: IMO, it would be better to separate the `if` subgraphs extraction and quantization process to avoid recursion. As the positive effects of this: 

- Removing double checking of the `if` subgraphs' existence, since it's not possible to have `if` in the `if` (or possible?);
- The `_dfs_quantize_models` method looks difficult now, but separated methods are more convenient for understanding;
- It's not necessary to keep not-applicable to the child models options like `parent_statistic_points`;

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+                    parent_model = self._update_if_submodel(
+                        model_transformer, if_node, if_submodel_condition, child_quantized_model
+                    )
+                    if self.intermediate_model_dir:
+                        nncf_logger.info(
+                            f"Save quantized model number {model_cnt} to dir {self.intermediate_model_dir}"
+                        )
+                        self._backend_entity.dump_submodel(
+                            child_quantized_model, self.intermediate_model_dir, if_node, if_submodel_condition
+                        )
```
KodiaqQ, line 413: From my understanding, `intermediate_model_dir` is used for the whole model saving during AAQ. Not for the subgraph's dumping while quantization.

nncf/quantization/algorithms/post_training/backend.py
```diff
@@ -0,0 +1,99 @@
...
+
+from nncf.common.graph.graph import NNCFNode
+from nncf.common.graph.transformations.commands import Command
+from nncf.quantization.algorithms.post_training.algorithm import TModel
+
+
+class PostTrainingBackend(ABC):
+    @property
+    @abstractmethod
+    def if_node_metatype(self):
```
KodiaqQ, line 25: Maybe it's better to use `metatypes` and a list of them. But, it's minor.

nncf/quantization/algorithms/post_training/openvino_backend.py
```diff
@@ -0,0 +1,81 @@
...
+    IF_OP_MODEL_INPUT_PORTS = (0, 1)
+
+    @property
+    def if_node_metatype(self):
+        return OVIfMetatype
+
+    @staticmethod
+    def get_if_subgraph_input_names(model: ov.Model, if_node: NNCFNode, if_submodel_condition: bool) -> List[str]:
+        input_names = []
+        subgraph_port_id = 0 if if_submodel_condition else 1
```
KodiaqQ, line 37: `int(if_submodel_condition)`?

nncf/quantization/algorithms/post_training/openvino_backend.py
```diff
@@ -0,0 +1,81 @@
...
+from nncf.openvino.graph.metatypes.openvino_metatypes import OVIfMetatype
+from nncf.openvino.graph.transformations.commands import OVExtractIfSubgraphCommand
+from nncf.openvino.graph.transformations.commands import OVOutputInsertionCommand
+from nncf.openvino.graph.transformations.commands import OVTargetPoint
+from nncf.openvino.graph.transformations.commands import OVUpdateIfSubgraphCommand
+from nncf.quantization.algorithms.post_training.backend import PostTrainingBackend
+
+
+class OVPostTrainingBackend(PostTrainingBackend):
+    IF_OP_MODEL_INPUT_PORTS = (0, 1)
```
KodiaqQ, line 28: Since we provide the `IF_OP_MODEL_INPUT_PORTS` variable, we should use it instead/or inside methods.

nncf/quantization/algorithms/post_training/openvino_backend.py
```diff
@@ -0,0 +1,81 @@
...
+        ov_node = name_to_node_mapping[if_node.node_name]
+        input_indices = [desc.input_index for desc in ov_node.get_input_descriptions(subgraph_port_id)]
+        input_names.extend([ov_node.input_values()[index].any_name for index in input_indices])
+        return input_names
+
+    @staticmethod
+    def get_if_cond_input_name(model: ov.Model, if_node: NNCFNode) -> str:
+        name_to_node_mapping = {op.get_friendly_name(): op for op in model.get_ops()}
+        ov_node = name_to_node_mapping[if_node.node_name]
+        return ov_node.input_values()[0].any_name
```
KodiaqQ, line 48: Does the `if` node have only one input? I don't think so.
Why do we use `[0]` value?

nncf/quantization/algorithms/post_training/openvino_backend.py
```diff
@@ -0,0 +1,81 @@
...
+        return OVExtractIfSubgraphCommand(if_node, if_submodel_condition)
+
+    @staticmethod
+    def create_output_insertion_commands(model: ov.Model, if_node: NNCFNode) -> List[OVOutputInsertionCommand]:
+        commands = []
+        name_to_node_mapping = {op.get_friendly_name(): op for op in model.get_ops()}
+        ov_node = name_to_node_mapping[if_node.node_name]
+        for port_id in range(len(ov_node.inputs())):
+            commands.append(
+                OVOutputInsertionCommand(OVTargetPoint(TargetType.PRE_LAYER_OPERATION, if_node.node_name, port_id))
```
KodiaqQ, line 71: It's not consistent that we call the `*_output_*` command that uses the `PRE_LAYER_*` approach.

nncf/data/dataset.py
```diff
@@ -72,6 +72,15 @@ def get_inference_data(self, indices: Optional[List[int]] = None) -> Iterable[Mo
         """
         return DataProvider(self._data_source, self._transform_func, indices)
 
+    def get_length(self) -> Optional[int]:
```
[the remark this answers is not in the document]
kshpv: Please, take a look at https://github.com/openvinotoolkit/nncf/pull/2114

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+        for input_data in tqdm(
+            islice(calibration_dataset.get_inference_data(), calibration_dataset_size),
+            total=calibration_dataset_size,
+            desc=f"Collect dataset for {model_cnt} model:",
+        ):
+            data_item = []
+            results = engine.infer(input_data)
+            if (if_submodel_condition and results[if_cond_input_name]) or (
+                not if_submodel_condition and not results[if_cond_input_name]
+            ):
```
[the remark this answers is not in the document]
kshpv: This check is about checking the IF condition bool value on the particular sample. Based on that we collect data or not for particular then/else submodel

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: We need extra outputs of original (parent) model to collect dataitems for then/else submodels

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+        Update submodel of If node.
+
+        :param model_transformer: ModelTransformer instance.
+        :param if_node: If node.
+        :param if_submodel_condition: Condition of If node submodel.
+        :param submodel: New submodel.
+        :return: Updated model with a new submodel.
+        """
+        transformation_layout = TransformationLayout()
+        command = self._backend_entity.create_update_subgraph_command(if_node, if_submodel_condition, submodel)
```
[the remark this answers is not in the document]
kshpv: We can, but in this logic, we should store both submodels in memory. While introduced approach vanishes this issue.
We quantize then update the model, and go to the next. No need to store two quantized models at memory at one time.

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -188,28 +211,207 @@ def apply(
...
+            transformation_layout.register(command)
+        return model_transformer.transform(transformation_layout)
+
+    def _dfs_quantize_models(
+        self,
+        parent_model: TModel,
+        parent_graph: NNCFGraph,
+        parent_dataset: Dataset,
+        parent_statistic_points: Optional[StatisticPointsContainer],
+        parent_model_cnt: int,
```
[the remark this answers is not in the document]
kshpv: It is used only for logging purposes, but, probably, it is not the best experience for the user.

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: 0-input is the condition input for IF node. We use this to correctly collect dataset

nncf/openvino/graph/model_transformer.py
kshpv: According to the docs https://docs.openvino.ai/2023.0/classov_1_1op_1_1util_1_1MultiSubGraphOp.html#methods
`return pointer to Model with sub-graph`

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: We can - yes. But should we store all datatitems for all subgraphs at one time?

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: I am not sure what you mean by 

> to separate the if subgraphs extraction and quantization process to avoid recursion

extraction of IF subgraph for now is just one function. Please, give a suggestion by code or more detailed explanation.

> Removing double checking of the if subgraphs' existence, since it's not possible to have if in the if (or possible?);

Yes, it is possible that if node exists inside of IF else or then bodies.

> It's not necessary to keep not-applicable to the child models options like parent_statistic_points;

It is a good suggestion. It can be avoided by first quantizing the parent model and then quantizing its subgraphs. I will update its logic.

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: I want to use it to have the ability to dump subgraphs, because it is convenient for debugging purposes.

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: The word output here is inside the context of Model's output. The same function naming used in `nncf/openvino/statistics/aggregator.py` `_get_transformation_layout_extra_outputs`

nncf/quantization/advanced_parameters.py
```diff
@@ -170,6 +172,8 @@ class AdvancedQuantizationParameters:
     # Backend specific parameters
     backend_params: Dict[str, Any] = field(default_factory=dict)
 
+    intermediate_model_dir: Optional[str] = None
```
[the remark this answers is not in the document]
kshpv: There is no changes about intermediate_model_dir onto `develop`

nncf/openvino/graph/metatypes/openvino_metatypes.py
```diff
@@ -673,6 +673,12 @@ class OVAbsMetatype(OVOpMetatype):
     op_names = ["Abs"]
 
 
+@OV_OPERATOR_METATYPES.register()
+class OVIfMetatype(OVOpMetatype):
+    name = "IfOp"
+    op_names = ["If"]
```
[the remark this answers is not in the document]
kshpv: Done

nncf/openvino/graph/transformations/commands.py
```diff
@@ -191,3 +193,40 @@ def __init__(
...
+        # Have a look at nncf/torch/graph/transformations/commands/PTInsertionCommand
+        raise NotImplementedError()
+
+
+class OVExtractIfSubgraphCommand(Command):
+    """
+    Extracts If node subgraph.
+    """
+
+    def __init__(self, if_node: NNCFNode, if_submodel_condition: bool):
```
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: Done. But only 
`int(not if_submodel_condition)`

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: Removed

nncf/common/hardware/configs/cpu.json
KodiaqQ, line 1: Not necessary code bloat, IMO.

nncf/openvino/graph/model_transformer.py
```diff
@@ -526,3 +530,42 @@ def _apply_multiply_insertion_transformations(
...
+        """
+        Update model subgraph for IF node.
+
+        :param model: Model to update and insert a new subgraph.
+        :param transformations: Transformations with information of If node and an updated subgraph.
+        :return: Original model with an updated subgraph.
+        """
+        name_to_node_mapping = OVModelTransformer._get_name_to_node_mapping(model)
+        for transformation in transformations:
+            subgraph_model = transformation.subgraph_model
```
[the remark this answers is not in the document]
KodiaqQ: Since `_apply_extract_if_subgraph_transformation` returns `return pointer to Model with sub-graph` I don't see any reason to change the sub-graph with the new one instead of updating it.

nncf/openvino/graph/model_transformer.py
KodiaqQ: So that means that we can change the sub-graph `in-place` in the original model.

nncf/quantization/algorithms/post_training/algorithm.py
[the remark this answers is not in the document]
KodiaqQ: The question was not about `if_submodel_condition`, but about `if_cond_input_name` since `results` should contain (based on the code) it and not `child_model_input_names` that are really needed. Why so?

nncf/quantization/algorithms/post_training/algorithm.py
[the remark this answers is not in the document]
KodiaqQ: Still can't get it since the `if` subgraph is a link to the set of nodes in the original model.

nncf/quantization/algorithms/post_training/algorithm.py
KodiaqQ: I don't see any reason to increase pipeline time using the loop with the model execution at each step instead of collecting statistics for all models at one time.
Does it really explode memory consumption? Are there any numbers of it based on experiments with a few `if` sub-graphs?

nncf/quantization/algorithms/post_training/algorithm.py
KodiaqQ: > Please, give a suggestion by code or more detailed explanation.

Please, separate the `_dfs_quantize_models` method into pieces to avoid using recursion.

> Yes, it is possible that if node exists inside of IF else or then bodies.

Any examples? How deep can it be? I'm not sure that there is no limit to the recursion deep in this case.

nncf/quantization/algorithms/post_training/algorithm.py
KodiaqQ: `intermediate_model_dir` is not for debugging purposes at all.
It was introduced to save the fully quantized model during the AAQ process and give the ability to evaluate the different model states.
Dumping the N subgraphs instead of the 1 model is not a good experience, IMO. Even the `BiasCorrection` doesn't have such option, but it's really useful to dump a subgraph in each step since it's a more complex algorithm.

nncf/quantization/algorithms/post_training/openvino_backend.py
KodiaqQ: Let's name methods more clearly in the future even in case of the mistakes in the past.

kshpv:
run pylint pre-commit tests
run pytorch pre-commit tests
run openvino pre-commit tests
run onnx pre-commit tests

nncf/quantization/algorithms/post_training/algorithm.py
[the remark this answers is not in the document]
kshpv: Removed `model_counter`

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: Update the main method body

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: Removed

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -95,6 +95,8 @@ class MinMaxQuantization(Algorithm):
     It is expected that the inference of the obtained model in the int8 mode would be faster than the original model.
     """
 
+    MIN_STATISTICS_SAMPLES = 10
```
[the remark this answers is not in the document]
kshpv: Removed this part

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -690,6 +696,20 @@ def filter_func(point: StatisticPoint) -> bool:
...
                     raise RuntimeError(f"Statistics were not collected for the node {target_node_name}")
+                min_collected_num_samples = min(
+                    [
+                        tensor_collector.aggregators[aggregator].collected_num_samples
+                        for aggregator in tensor_collector.aggregators
+                    ]
+                )
+                if (
+                    min_collected_num_samples < self.statistics_threshold
+                    and not quantization_target_point.is_weight_target_point()  # Weights will be quantized anyway
```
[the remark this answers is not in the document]
kshpv: Also removed

nncf/common/hardware/configs/cpu.json
kshpv: This was don by auto formatting. But can be removed indeed

kshpv:
After the new round of discussion the new proposed approach was implemented. The logic for quantization of IF operation was placed into `nncf/openvino/quantization/quantize_model_if_op.py`. The changes to `PostTrainingQuantization` were removed.

nncf/openvino/graph/node_utils.py
```diff
@@ -49,6 +51,15 @@ def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
     return bias_constant is not None
 
 
+def has_if_op(model: ov.Model) -> bool:
+    """ """
```
KodiaqQ, line 55: Minor. Missed description.

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,226 @@
...
+from nncf.openvino.graph.transformations.commands import OVUpdateIfSubgraphCommand
+from nncf.quantization.algorithms.algorithm import Algorithm
+
+
+def _make_dataset_for_children_model(
+    engine: Engine,
+    calibration_dataset: Dataset,
+    if_cond_input_name: str,
+    then_model_input_names,
+    else_model_input_names,
```
KodiaqQ, line 33: Minor. Missed type hints and descriptions as well.

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,226 @@
...
+    :param if_submodel_condition: If node submodel condition.
+    :return Dataset: Dataset for child model.
+    """
+    then_dataset, else_dataset = [], []
+    calibration_dataset_size = (
+        min(subset_size, calibration_dataset.get_length())
+        if calibration_dataset.get_length() is not None
+        else subset_size
+    )
+    for input_data in tqdm(
```
KodiaqQ, line 53: Minor. Shouldn't we update it with the `rich` progress bar?

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,226 @@
...
+    parent_statistic_points: Optional[StatisticPointsContainer] = None,
+) -> ov.Model:
+    """
+
+    :param parent_model:
+    :param parent_graph:
+    :param parent_dataset:
+    :param parent_statistic_points:
+    :param parent_model_cnt:
+    :return:
```
KodiaqQ, line 138: Minor. Missed descriptions.

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    nncf_logger.info(f"Quantize a new submodel")
```
KodiaqQ, line 140: IMO, such logging is unnecessary for the user since the number of the `if` blocks is not limited and thus it might not look good in case of the huge number of these blocks. Especially with the logging like `The length of dataset for ..`.
I suggest moving it into the `debug` section.

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,226 @@
...
+        else_model = _extract_if_submodel(model_transformer_fp32, if_node, False)
+        then_quantized_model = apply_algorithm_if_bodies(
+            algorithm, then_model, NNCFGraphFactory.create(then_model), then_dataset, subset_size, None
+        )
+        else_quantized_model = apply_algorithm_if_bodies(
+            algorithm, else_model, NNCFGraphFactory.create(else_model), else_dataset, subset_size, None
+        )
+        model_transformer_int8 = factory.ModelTransformerFactory.create(quantized_model)
+        quantized_model = _update_if_submodel(model_transformer_int8, if_node, True, then_quantized_model)
+        model_transformer_int8 = factory.ModelTransformerFactory.create(quantized_model)
```
KodiaqQ, line 170: Why not use one instance of the `ModelTransformer` since we create it based on the main model?

nncf/openvino/graph/model_transformer.py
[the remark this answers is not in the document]
kshpv: This is the current limitation of the implementation of `OVModelTransformer`. `OVModelTransforemer` returns a new model for every applied transformation. The model after inserting FakeQuantizers is a copy, so there is a need to set the new model instead of an old one

nncf/openvino/quantization/quantize_model_if_op.py
kshpv: Done

nncf/openvino/graph/node_utils.py
kshpv: Done

nncf/openvino/quantization/quantize_model_if_op.py
kshpv: Done

nncf/openvino/quantization/quantize_model_if_op.py
kshpv: Done

nncf/quantization/algorithms/post_training/backend.py
kshpv: Done

nncf/quantization/algorithms/post_training/openvino_backend.py
kshpv: Please, add a suggestion if you have

nncf/openvino/quantization/quantize_model_if_op.py
kshpv: Updated. Removed 
`nncf_logger.info(f"Quantize a new submodel")`.
Also, put into the debug section the logging of dataset length.

KodiaqQ (commented):
General comments:

- The insertion of the `subgraph` instead of the in-place model change is not good (but this is because of the `ModelTransformer` limitations);
- The recursion while quantization of the subgraphs is not the best approach;

kshpv:
> General comments:
> 
> * The insertion of the `subgraph` instead of the in-place model change is not good (but this is because of the `ModelTransformer` limitations);

Agree. This is how it works for now.

> * The recursion while quantization of the subgraphs is not the best approach;

Do you think that it is better to use a loop with a queue? I don't see recursion as a bad practice. 

> * The dataset collection for subgraphs in each step is not good due to pipeline performance;

At each step, the dataset for then and else bodies are collected by one inference of a parent model on a full dataset. A model is inferred once for each If node and keeps in memory only datasets needed for the quantization of these nodes. It is a tradeoff between performance and memory footprint.

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -80,6 +80,7 @@ def __init__(
         """
         super().__init__()
         self.algorithms = []
+        self.subset_size = subset_size
```
alexsu52, line 83: Could you explain this change?

nncf/common/hardware/configs/cpu.json
```diff
@@ -264,6 +264,12 @@
                 "activations": "q8_a"
             }
         },
+        {
+            "type": "If",
```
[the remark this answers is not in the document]
kshpv: Discussed offline and decided to revert these changes.

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: Outdated. Thanks for noticing!

nncf/openvino/quantization/quantize_model_if_op.py
```diff
@@ -0,0 +1,290 @@
+# Copyright (c) 2023 Intel Corporation
```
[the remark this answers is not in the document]
kshpv: Done

nncf/common/hardware/configs/cpu.json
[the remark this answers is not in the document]
kshpv: Done

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -133,7 +133,8 @@ def apply(
             for node in graph.get_all_nodes()
             if self._backend_entity.is_node_with_bias(node, graph)
         ]
-
+        if not node_and_bias_value:
+            return model
```
andrey-churkin, line 137: It is not clear for me why this is required. Let's look:

- If `node_and_bias_value` is empty, the `for` statement is not executed.
- The `node_and_new_bias_value` is empty because the `for` statement was not executed.
- The `transformation_layout` is empty because the `node_and_new_bias_value` is empty
- The `model_transformer.transform(transformation_layout)` returns the same model because `transformation_layout` is empty

So, this special check is not required. Everything works as expected.

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -202,7 +202,9 @@ def apply(
             for algorithm in self.algorithms:
                 algo_statistic_points = algorithm.get_statistic_points(modified_model, modified_model_graph)
                 statistics_aggregator.register_statistic_points(algo_statistic_points)
-
+            if not statistics_aggregator.statistic_points:
+                nncf_logger.info("The model has no operations to apply quantization.")
+                return modified_model
```
andrey-churkin, line 207: Could you please explain why this is required?

nncf/openvino/quantization/quantize_model.py
```diff
@@ -91,6 +93,57 @@ def dump_parameters(model: ov.Model, parameters: Dict, path: Optional[List] = No
...
+        subset_size=subset_size,
+        fast_bias_correction=fast_bias_correction,
+        model_type=model_type,
+        ignored_scope=ignored_scope,
+        advanced_parameters=advanced_parameters,
+    )
+
+    graph = GraphConverter.create_nncf_graph(model)
+
+    quantized_model = apply_algorithm_if_bodies(quantization_algorithm, model, graph, calibration_dataset, subset_size)
```
andrey-churkin, line 127: By the way, why don't we want to call this method inside the `native_quantize_impl()` method? It would allow us to reduce the number of lines of code and I think it would be easier to maintain.
@alexsu52 @KodiaqQ

nncf/openvino/quantization/quantize_model.py
```diff
@@ -91,6 +93,57 @@ def dump_parameters(model: ov.Model, parameters: Dict, path: Optional[List] = No
...
+    ignored_scope: Optional[IgnoredScope] = None,
+    advanced_parameters: Optional[AdvancedQuantizationParameters] = None,
+) -> ov.Model:
+    """
+    Implementation of the `quantize()` method for the OpenVINO backend via the OpenVINO Runtime API.
+    """
+    if not fast_bias_correction:
+        raise NotImplementedError(
+            "The BiasCorrection algorithm is not supported for OpenVINO models with If operation."
+        )
```
andrey-churkin, line 114: Does the `FastBiasCorrection` algorithm work for the `ifmodel`? If not, then another condition should be used.
```python
if not advanced_parameters.disable_bias_correction:
    ...
```

Do the SmoothQuant and ChannelAlignment algorithms work for `ifmodel`?

nncf/openvino/graph/node_utils.py
```diff
@@ -49,6 +51,20 @@ def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
     return bias_constant is not None
 
 
+def has_if_op(model: ov.Model) -> bool:
```
andrey-churkin, line 54: I suggest using the name `is_ifmodel()` for this method.

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
kshpv: This is because of the printing logs even in cases when  `node_and_bias_value` has no elements.

nncf/quantization/algorithms/post_training/algorithm.py
kshpv: This is an early exit:
1) to not do extra work.
2) for better logging purposes. Also, in case of having If node, the log will be informative.
image

nncf/openvino/graph/node_utils.py
kshpv: Updated to a new method

nncf/openvino/quantization/quantize_model.py
kshpv: Updated the `native_quantize_if_op_impl`

nncf/quantization/algorithms/fast_bias_correction/algorithm.py
andrey-churkin: I got it. But incorrect logging is not the problem of the FBC algorithm, is it? It should be fixed inside the `track` class.

https://github.com/openvinotoolkit/nncf/blob/50feea44ac6d705e7d0752e1f1b0f1f1bfb01c01/nncf/common/logging/track_progress.py#L50

nikita-savelyevv: I plan to make changes to `track` class in the near future in order not to display progress bar for empty sequence.

kshpv: I will take the changes from this commit- https://github.com/nikita-savelyevv/nncf/commit/0485f6f1c3c089f5d031675222dcefbd70c2af9a and revert changes inside of FBC

kshpv: Done

andrey-churkin: @kshpv @nikita-savelyevv Thanks!

nncf/quantization/algorithms/post_training/algorithm.py
andrey-churkin: Once again, you are trying to solve the logging problem by adding these checks.

In my opinion, there should be no checks here. If there are no statistics (i.e., it is empty), the MMQ and FBC should not be applied.

kshpv: @andrey-churkin should the user somehow be notified that the model was not quantized?

kshpv: After discussion decided to put logging inside `MinMaxQuantization`

KodiaqQ (approved):

andrey-churkin (approved):

alexsu52 (approved):
LGTM

PR #2107, opened 2023-09-05
[NNCFGraph] Fix `get_input_edges` for parallel edges

daniil-lyakhov:
### Changes

* `parallel_input_port_ids` parameter plugged in to `OVNNCFGraph`
* `NNCFGraph.get_input_edges` unwraps `parallel_input_port_ids` and returns separate entities of `NNCFGraphEdge` for each parallel edge

### Reason for changes

* To make possible to insert statistics collectors on parallel edges in OV backend

### Related tickets

119666

### Tests

* tests/common/graph/test_nncf_graph.py is updated
* tests/openvino/native/test_nncf_graph_builder.py is updated


nncf/common/graph/graph.py
```diff
@@ -331,7 +331,24 @@ def get_input_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
...
         """
         input_nodes = self.get_previous_nodes(node)
-        edges = [self.get_edge(from_node, node) for from_node in input_nodes]
+        edges = []
+        for from_node in input_nodes:
+            edge = self.get_edge(from_node, node)
+            if not edge.parallel_input_port_ids:
+                edges.append(edge)
+                continue
+            for input_port_id in edge.parallel_input_port_ids:
```
nikita-savelyevv, line 340:
```suggestion
            if len(edge.parallel_input_port_ids) == 0:
                edges.append(edge)
            for input_port_id in edge.parallel_input_port_ids:
```

Minor, according to my taste :)

daniil-lyakhov:
run pytorch pre-commit tests

daniil-lyakhov:

nncf/common/graph/graph.py
daniil-lyakhov: Please check

daniil-lyakhov: According to [PEP8](https://peps.python.org/pep-0008/#programming-recommendations) it's better to use `not` instead of `len(...) == 0`

nncf/common/graph/graph.py
```diff
@@ -148,17 +148,23 @@ def __init__(
...
+            (
+                self.from_node,
+                self.to_node,
+                self.input_port_id,
+                self.output_port_id,
+                tuple(self.tensor_shape),
+                self.dtype,
+                tuple(self.parallel_input_port_ids),
+            )
+        )
```
nikita-savelyevv, line 164: Are all these properties required for unique identification of an edge? I'm not sure how frequently edges are used as dict keys, but such a big list of things to be hashed may worsen the performance.

nncf/common/graph/graph.py
```diff
@@ -148,17 +148,23 @@ def __init__(
...
 
     def __eq__(self, other):
-        return (
-            self.from_node == other.from_node
-            and self.to_node == other.to_node
-            and self.tensor_shape == other.tensor_shape
-        )
+        return self.__dict__ == other.__dict__
```
nikita-savelyevv, line 167: Same concern as for hash

nncf/common/graph/graph.py
```diff
@@ -148,17 +148,23 @@ def __init__(
         self.parallel_input_port_ids = parallel_input_port_ids
```
alexsu52, line 148: What do you think using MultiDiGraph can resolve the issue with `parallel_input_port_ids` https://networkx.org/documentation/stable/reference/classes/multidigraph.html?

daniil-lyakhov: @vshampor, could you please share a ticket number for this issue?

nncf/common/graph/graph.py
daniil-lyakhov: That could affect performance. I think the best solution will be dataclass

nncf/common/graph/graph.py
vshampor: I could find the 110944 to that accord.

nncf/common/graph/graph.py
```diff
@@ -148,17 +148,23 @@ def __init__(
...
 
     def __str__(self):
-        return str(self.from_node) + " -> " + str(self.tensor_shape) + " -> " + str(self.to_node)
+        return f"{self.from_node}:{self.output_port_id} -> {self.tensor_shape} -> {self.to_node}:{self.input_port_id}"
 
     def __hash__(self):
-        return hash(str(self))
+        return hash(
```
vshampor, line 154: So what's the difference here exactly?

nncf/common/graph/graph.py
vshampor: Why? Iterating over a dict and comparing is O(N) where N is the number of fields

nncf/common/graph/graph.py
```diff
@@ -331,7 +337,9 @@ def get_input_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
         :return: List of input edges for the node sorted by input port ID.
         """
         input_nodes = self.get_previous_nodes(node)
-        edges = [self.get_edge(from_node, node) for from_node in input_nodes]
+        edges = []
+        for from_node in input_nodes:
+            edges.extend(self._get_edges(from_node, node))
```
vshampor, line 342: What is the point of this change? Keep a single-liner.

nncf/common/graph/graph.py
```diff
@@ -343,9 +351,30 @@ def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
         """
 
         output_nodes = self.get_next_nodes(node)
-        edges = [self.get_edge(node, to_node) for to_node in output_nodes]
+        edges = []
+        for to_node in output_nodes:
+            edges.extend(self._get_edges(node, to_node))
         return sorted(edges, key=lambda x: x.output_port_id)
```
vshampor, line 357: Same

nncf/common/graph/graph.py
```diff
@@ -343,9 +351,30 @@ def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
...
 
+    def _get_edges(self, from_node: NNCFNode, to_node: NNCFNode) -> NNCFGraphEdge:
```
vshampor, line 359: Your type hint is wrong,

nncf/openvino/graph/nncf_graph_builder.py
```diff
@@ -88,19 +89,28 @@ def _add_edges_to_nncf_graph(model: ov.Model, graph: NNCFGraph) -> None:
...
+
+                for out_node, inputs in node_vs_target_inputs.items():
                     tensor_shape = list(out.partial_shape.get_max_shape())
                     output_node_id = graph.get_node_by_name(out_node.get_friendly_name()).node_id
                     ov_dtype = out.get_element_type().get_type_name()
                     nncf_dtype = GraphConverter.convert_to_nncf_dtype(ov_dtype)
+
+                    parallel_inputs_kwarg = {}
+                    if len(inputs) > 1:
+                        parallel_inputs_kwarg = {"parallel_input_port_ids": [inp.get_index() for inp in inputs]}
```
vshampor, line 104: Stop building `kwargs` dicts in your pull requests, this is not a refactoring-friendly, readability-friendly or type-checking-friendly solution. If params are optional, and shouldn't be used in a given case, set these to `None` values.

tests/common/graph/test_nncf_graph.py
```diff
@@ -48,3 +49,46 @@ def test_find_matching_subgraphs():
...
         assert match == nodes[:2]
+
+
+def test_parallel_edges():
+    nncf_graph = NNCFGraph()
+    nodes = []
+    for node in "abc":
+        nodes.append(nncf_graph.add_nncf_node(node, f"type_{node}", f"metatype_{node}"))
+
+    kwargs = {
```
vshampor, line 60: Same about `kwargs`

tests/openvino/native/test_nncf_graph_builder.py
```diff
@@ -47,3 +51,20 @@ def test_compare_nncf_graph_omz_models(tmp_path, model_name):
...
     compare_nncf_graphs(model, path_to_dot)
+
+
+def test_parallel_edges():
+    model = ParallelEdgesModel().ov_model
+    nncf_graph = GraphConverter.create_nncf_graph(model)
+    input_node = nncf_graph.get_node_by_name("Input")
+    mm_node = nncf_graph.get_node_by_name("Mm")
+    add_node = nncf_graph.get_node_by_name("Add")
+    kwargs = {"dtype": Dtype.FLOAT, "tensor_shape": [1, 3, 3], "parallel_input_port_ids": []}
```
vshampor, line 62: Same

vshampor (changes requested):

nncf/common/graph/graph.py
nikita-savelyevv: My concern is that before there were 3 comparisons, but now there are more -- since `self.__dict__` has more values.

nncf/common/graph/graph.py
daniil-lyakhov: The point is that _get_edges returns list, and this for statement makes flatten list of edges

nncf/openvino/graph/nncf_graph_builder.py
daniil-lyakhov: Ok, I need to refactor `graph.add_edge_between_nncf_nodes` then

tests/common/graph/test_nncf_graph.py
daniil-lyakhov: Ok, I need to refactor graph.add_edge_between_nncf_nodes then

tests/openvino/native/test_nncf_graph_builder.py
daniil-lyakhov: You mean I should repeat this twice in `NNCFGraphEdge` constructor? And after that to refactor this peace one need to change value in two places instead of one, is it "refactoring friendly"?

daniil-lyakhov: Or should I create a local variable for each param? What is better?

nncf/common/graph/graph.py
daniil-lyakhov: What is the definition of  equal edges? I believe two edges are equal iff all attributes are the same. Do you think it's wrong?

nncf/common/graph/graph.py
daniil-lyakhov: https://github.com/openvinotoolkit/nncf/pull/2107/files/d950c520a6cdeefaccfb56cad4ffbd1c5ceff03a#r1324503677

nncf/common/graph/graph.py
vshampor: It doesn't matter that much whether you are hashing the string representation of all the params or the params themselves. If you delegate the hashing to hashing the string, then you only have one place to adjust when adding/removing params, and that is the `__str__` method.

tests/openvino/native/test_nncf_graph_builder.py
vshampor: Repeating would be OK here since the entities you are initializing are not related, and have this param set identical only accidentally. Making a local var for this would work too, but wouldn't save anything in terms of line space.

nncf/common/graph/graph.py
daniil-lyakhov: Strange behavior encountered: in case hash is
```
hash(str(self.from_node) + " -> " + str(self.tensor_shape) + " -> " + str(self.to_node))
```
test `test_graph_analysis.test_graph_pattern_io_building` passed without a problem. But for hash
```
hash(f"{self.from_node}:{self.output_port_id} -> {self.tensor_shape} -> {self.to_node}:{self.input_port_id}")
```

the same test fails due to one of the edge has wrong input port_id (1 instead of 0)
I'm trying to investigate the root cause as there could be a potential bug in NNCFGraph code

daniil-lyakhov: There is a bug in `test_graph_analysis.test_graph_pattern_io_building` references, I'll fix it

nncf/common/graph/graph.py
daniil-lyakhov: Thanks, done

tests/openvino/native/test_nncf_graph_builder.py
daniil-lyakhov: Done

tests/common/graph/test_nncf_graph.py
daniil-lyakhov: Done

nncf/openvino/graph/nncf_graph_builder.py
daniil-lyakhov: Done

nncf/common/graph/graph.py
daniil-lyakhov: Done

nikita-savelyevv (approved):

nncf/common/graph/graph.py
```diff
@@ -343,9 +351,31 @@ def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
...
+                    from_node=edge.from_node,
+                    to_node=edge.to_node,
+                    input_port_id=input_port_id,
+                    output_port_id=edge.output_port_id,
+                    tensor_shape=edge.tensor_shape,
+                    dtype=edge.dtype,
+                    parallel_input_port_ids=[],
+                )
+            )
+        return edges
```
daniil-lyakhov, line 377: I inspected [this code](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/dynamic_graph/graph.py#L342-L357) and found previous implementation of `_get_edges` was wrong, first node should be always added to `edges` list

vshampor (approved):

alexsu52 (approved):
LGTM

PR #2116, opened 2023-09-07
[DOCS] OV PTQ Model Zoo

nikita-savelyevv:
### Changes

Added OV PTQ section to Model Zoo. The results for OV PTQ are gathered from the latest OV validation.
Also added a small demo portion of results for weight compression, will need to expand in the future.

Result: https://github.com/nikita-savelyevv/nncf/blob/ov-ptq-model-zoo/docs/ModelZoo.md

### Reason for changes

Documentation improvement

### DIFF
--- a/docs/ModelZoo.md
+++ b/docs/ModelZoo.md
@@ -30,6 +30,7 @@ Additional PTQ results may be found on an OpenVINO Performance Benchmarks [page]
 ### OpenVINO PTQ
 
 #### OpenVINO PTQ Classification
+
 <table>
   <thead>
     <tr style="text-align: center;">
@@ -134,6 +135,7 @@ Additional PTQ results may be found on an OpenVINO Performance Benchmarks [page]
 </table>
 
 #### OpenVINO PTQ Object Detection
+
 <table>
   <thead>
     <tr style="text-align: center;">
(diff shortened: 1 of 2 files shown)


docs/ModelZoo.md
```diff
@@ -1,23 +1,380 @@
...
+Below we present weight compression results for LLMs from Hugging Face.
 
-## PyTorch
+| Model           | Accuracy (<em>drop</em>) | Size reduction |
+|-----------------|--------------------------|----------------|
+| open-llama-3b   | 66.49 (0.02)             | 4.33           |
+| llama-7b        | 73.39 (-0.22)            | 4.17           |
+| llama-2-7b-chat | 70.85 (-0.23)            | 3.83           |
+| gpt-j-6b        | 68.12 (-0.19)            | 4.00           |
+| dolly-v2-12b    | 64.29 (-0.06)            | 4.05           |
```
nikita-savelyevv, line 369: This is not much for now, plan to improve in the future.

alexsu52 (commented):
Some comments from my side:
1. Looks like the list is quite outdated
2. How can I take a model and reproduce result?
3. What about performance numbers?
4. What do you think about using models from notebooks and nncf examples?

nikita-savelyevv:
Decided to close due to reasons highlighted above

PR #2118, opened 2023-09-08
Add message about deprecation `export_to_onnx_standard_ops` 

AlexanderDokuchaev:
### Changes

Add message about deprecation of  `export_to_onnx_standard_ops`  option in NNCFConfig

### Reason 

Recommended way to export to onnx with QuantizeLinear-DequantizeLinear node pairs is `nncf.strip(quantized_model)`.

### DIFF
--- a/nncf/config/schemata/algo/quantization.py
+++ b/nncf/config/schemata/algo/quantization.py
@@ -526,7 +526,7 @@ QUANTIZATION_SCHEMA = {
         },
         "export_to_onnx_standard_ops": with_attributes(
             BOOLEAN,
-            description="(Deprecated) Determines how should the additional quantization "
+            description="[Deprecated] Determines how should the additional quantization "
             "operations be exported into the ONNX format. Set "
             "this to true to export to ONNX "
             "standard QuantizeLinear-DequantizeLinear "
--- a/nncf/config/schemata/algo/quantization.py
+++ b/nncf/config/schemata/algo/quantization.py
@@ -526,7 +526,7 @@ QUANTIZATION_SCHEMA = {
         },
         "export_to_onnx_standard_ops": with_attributes(
             BOOLEAN,
-            description="Determines how should the additional quantization "
+            description="(Deprecated) Determines how should the additional quantization "
             "operations be exported into the ONNX format. Set "
             "this to true to export to ONNX "
             "standard QuantizeLinear-DequantizeLinear "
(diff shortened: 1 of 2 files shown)


alexsu52 (approved):
LGTM

nncf/torch/quantization/algo.py
```diff
@@ -1361,6 +1362,11 @@ def __init__(
             "export_to_onnx_standard_ops", QUANTIZATION_EXPORT_TO_ONNX_STANDARD_OPS
         )
         if should_export_to_onnx_qdq:
+            warning_deprecated(
+                "The config option `export_to_onnx_standard_ops` is deprecated and will be removed in a future version. "
+                "Please use the `nncf.strip(quantized_model)` method instead to export to ONNX "
```
vshampor, line 1367: `.strip()` does not do export on its own. Specify just when and in what context the export should occur, i.e. "Call `quantized_model.nncf.strip()` before calling `compression_ctrl.export_model(...)` or `torch.onnx.export(...)` to export to ONNX standard ops"

nncf/config/schemata/algo/quantization.py
```diff
@@ -526,7 +526,7 @@
         },
         "export_to_onnx_standard_ops": with_attributes(
             BOOLEAN,
-            description="Determines how should the additional quantization "
+            description="(Deprecated) Determines how should the additional quantization "
```
vshampor, line 529: Use brackets instead of parentheses as for "disable_shape_matching" so that the deprecation is visible in the visualization of the JSON schema, as in:
https://openvinotoolkit.github.io/nncf/schema/#disable_shape_matching

See also:
https://pypi.org/project/json-schema-for-humans/0.3.1/
(option --deprecated-from-description)

vshampor (changes requested):

nncf/config/schemata/algo/quantization.py
AlexanderDokuchaev: Updated

nncf/torch/quantization/algo.py
AlexanderDokuchaev: Yes, thanks. Rewrite as `Please use the `nncf.strip(quantized_model)` method before export to ONNX to get model with QuantizeLinear-DequantizeLinear node pairs.`
But still write `nncf.strip` because it's common api for PTQ and QAT model.

vshampor (approved):

AlexKoff88 (approved):

PR #2120, opened 2023-09-09
Reference for algo unification with existing tensor dispatch

vshampor:


vshampor:
@AlexanderDokuchaev @daniil-lyakhov FYI

vshampor:
Will be closed once the common algo code completely switches to the unified tensor approach.

nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -360,9 +400,10 @@ def _process_weight_statistics(self, node: NNCFNode, weights: TTensor, port_id:
         channel_axis = 0
         if len(weights.shape) > 1:
             channel_axis = self._backend_entity.get_weight_channel_axis(node, port_id)
-        reduction_shape = [i for i, _ in enumerate(weights.shape)]
-        reduction_shape.pop(channel_axis)
-        return self._backend_entity.process_weight_statistics(weights, tuple(reduction_shape))
+            base_axes = list(range(weights.ndim - 2))
+            transpose_axes = base_axes + [-1, -2]
+            weights = fns.transpose(weights, axes=transpose_axes)
+        return fns.max(fns.abs(weights), axis=channel_axis)
```
daniil-lyakhov, line 406: NB: this transpose was removed after support of convolutions in SQ

nncf/experimental/tensor/torch_functions.py
```diff
@@ -178,25 +178,165 @@ def _(x: torch.Tensor, axis: int = 0) -> List[torch.Tensor]:
...
+def _(
+    a: torch.Tensor,
+    q: Union[float, List[float]],
+    axis: Union[int, List[int]] = None,
+    keepdims: Optional[bool] = None,
+) -> Union[float, torch.Tensor]:
+    if not isinstance(axis, list):
+        return torch.quantile(a, q=q, dim=axis, keepdim=keepdims)
+    if len(axis) == 1:
+        return torch.quantile(a, q=q, dim=axis[0], keepdim=keepdims)
```
daniil-lyakhov, line 307: `torch.quantile` does not work with big tensors as well
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/tensor_statistics/collectors.py#L168-L189

vshampor:
The original version of the PR was ready for merge with CI passing at the time of preparation. As I had to rebase this against a heavily changed code base where presuppositions of the original PR were not holding, I had to remove more and more collateral improvements from my PR just to make CI pass without touching the rest of the code. I still cannot reach the passing CI state due to this or that inconsistency in the current code state I have to rebase against (no invariants on shape handling, bad design of the `TensorCollector` and related classes, no decoupling of the common, out-of-place reduction/aggregation logic from the inplace versions etc.). In the meanwhile, the base code state keeps advancing, leading to a frequent necessity of rebase.

Taking all of the above into account, I have decided that I will not be continuing work on this version of the PR. I suppose, as the entire idea of moving statistics to the common tensor representation was not a priority originally, that this would not be a big issue anyway from the team's perspective. Further efforts of enforcing the common tensor approach are better to be started from the clean slate, e.g. the current repository state, rather than from this version of the PR.

PR #2137, opened 2023-09-14
[PTQ] Fix non-optional algos in calibrate.py

KodiaqQ:
### Changes

- Fixed behaviour in the `calibrate.py` for algos without options

### Reason for changes

- Bugfix

### Related tickets

- 120295


alexsu52 (approved):
LGTM

PR #2141, opened 2023-09-16
Add PTQuantizerInsertionCommand

AlexanderDokuchaev:
### Changes

- Add `PTQuantizerInsertionCommand` to correctly add quantizers in ptq algorithm.
- Move some classes from `nncf/torch/nncf_network.py` and `nncf/torch/quntization/algo.py` to separates files to avoid circular import. 
- Update reference files because names was changed.
- Added 'NNCFNetworkInterface:is_compression_module_registered' method to check that extra module are registered without raising Runtime exception.

### Reason for changes

Activation quantizes does not collect in `model.nncf.external_quantizer` as result methods `.cpu()` and `.cuda()` does not change device for parameters of quantizers.

### Related tickets

120350

### Tests

test_quantizer_insertion_transformations


nncf/torch/checkpoint_loading.py
```diff
@@ -337,13 +337,13 @@ def run(self) -> Dict[str, torch.Tensor]:
...
-                "The loader will try to match these entries to the correspoindig `relu` and `relu_` op "
+                "The loader will try to match these entries to the corresponding `relu` and `relu_` op "
                 "names. The newly exported checkpoints will be adjusted to the new format."
             )
 
         if normalized_keys_to_load.has_legacy_storage_keys:
-            from nncf.torch.nncf_network import CURRENT_EXTERNAL_QUANTIZERS_STORAGE_PREFIX
-            from nncf.torch.nncf_network import LEGACY_EXTERNAL_QUANTIZERS_STORAGE_PREFIX
+            from nncf.torch.quantization.external_quantizer import CURRENT_EXTERNAL_QUANTIZERS_STORAGE_PREFIX
+            from nncf.torch.quantization.external_quantizer import LEGACY_EXTERNAL_QUANTIZERS_STORAGE_PREFIX
```
vshampor, line 346: Remove the legacy prefix altogether, we've supported it long enough.

nncf/torch/extra_compression_module_type.py
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+
+
+class ExtraCompressionModuleType(Enum):
```
vshampor, line 15: No need to extract this single enum into an own file, this will be only used along with `NNCFNetwork` and thus can stay in `nncf_network.py`. If importing this from the `nncf_network.py` only causes circular imports in your new debug interface file, just do a local import in the associated function.

nncf/quantization/algorithms/min_max/torch_backend.py
```diff
@@ -299,14 +298,13 @@ def _create_quantizer_insertion_command(
...
-    ) -> PTInsertionCommand:
+    ) -> PTQuantizerInsertionCommand:
         _, scale_shape, _ = PTMinMaxAlgoBackend._get_input_scale_shape(nncf_graph, target_point, quantizer_config)
 
         quantizer = PTMinMaxAlgoBackend._create_quantizer(
             quantizer_config, scale_shape, parameters, target_point.target_type
         )
-
-        return PTInsertionCommand(target_point, quantizer, TransformationPriority.QUANTIZATION_PRIORITY)
+        return PTQuantizerInsertionCommand(target_point, quantizer)
```
vshampor, line 308: What's wrong with the regular `PTInsertionCommand` that it had to be changed? Even if you need to bring the generated commands to a specific interface, I would expect that `PTQuantizerInsertionCommand` would ultimately get turned into the old `PTInsertionCommand`. Otherwise you are breaking the priority functionality of insertions into the PT models.

vshampor (changes requested):

nncf/torch/model_transformer.py
```diff
@@ -97,6 +102,42 @@ def _apply_insertion_transformations(model: NNCFNetwork, transformations: List[P
...
+        :param transformations: List of the OVQuantizerInsertionCommand transformations.
+        :return: Model with inserted FakeQuantize nodes.
+        """
+        node_to_op_address_mapping = model.nncf.get_node_to_op_address_mapping()
+
+        model.nncf.register_compression_module_type(ExtraCompressionModuleType.EXTERNAL_QUANTIZER)
+
+        for transformation_command in transformations:
+            target_point: PTTargetPoint = transformation_command.target_point
+            target_node_name = target_point.target_node_name
```
daniil-lyakhov, line 122: Do you have any thoughts on how we can incorporate shared quantizers logic into here?

nncf/quantization/algorithms/min_max/torch_backend.py
daniil-lyakhov: Our goal is to move quantizer insertion logic into model transformer as quantizer insertion requires quite a lot of additional steps (in some specific cases we need to insert quantizer into the NNCFNetwork, register its address in the External quantizer hook and put this hook into the NNCFNetwork as well).  Other backends don't need all this, so `PTQuantizerInsertionCommand` is aligning torch in following way: to insert a quantizer you need to 
* Create a quantizer instance
* Create insertion command
* Call model transformer with the command

daniil-lyakhov: About priority functionality: we are planning to completely migrate to PTQ interface as PTQ has more algorithms for initialization, aligned with other backends and models after torch PTQ are trainable.

nncf/torch/extra_compression_module_type.py
AlexanderDokuchaev: done

nncf/torch/checkpoint_loading.py
AlexanderDokuchaev: It's out of scope of this PR and ticket

nncf/quantization/algorithms/min_max/torch_backend.py
AlexanderDokuchaev: > What's wrong with the regular PTInsertionCommand that it had to be changed?

Additional steps before call transformation for `PTInsertionCommand` brakes logic of work with common algorithms.
From `ModelTransformer` expected that all logic about insertion FQ will be done inside, without call another backend-specific functions. 

> Otherwise you are breaking the priority functionality of insertions into the PT models.

Please provide example of incorrect work of PT PTQ.

nncf/torch/checkpoint_loading.py
vshampor: You had to touch this to implement this PR, which makes this relevant, and since you had to touch this, you can now do better - make it right.

nncf/quantization/algorithms/min_max/torch_backend.py
vshampor: > Other backends don't need all this

Which doesn't mean that you have to break existing functionality in torch in order to do this.

> About priority functionality: we are planning to completely migrate to PTQ interface as PTQ has more algorithms for initialization, aligned with other backends and models after torch PTQ are trainable.

Just what are you calling a PTQ interface? Whatever a PTQ interface is, it must follow and execute existing code that inserts quantizers in PT. If you need to issue `QuantizerInsertionCommand`s instead of `InsertionCommand`s - do it, but ultimately this must be translated into a regular `PTInsertionCommand` with a quantization priority.

tests/torch/test_model_transformer.py
```diff
@@ -505,3 +507,39 @@ def test_rebuild_graph_after_insert_transformation():
...
+    model_transformer = PTModelTransformer(model)
+
+    target_point = PTTargetPoint(target_type, node_name, input_port_id=input_port_id)
+    command = PTQuantizerInsertionCommand(target_point, BaseOp(lambda x: x))
+
+    transformation_layout = PTTransformationLayout()
+    transformation_layout.register(command)
+    transformed_model = model_transformer.transform(transformation_layout)
+
+    assert hasattr(transformed_model.nncf, "external_quantizers")
```
vshampor, line 536: BTW, this should reference the corresponding constant. I did a quick search on "external_quantizers" in the codebase and noticed other places where you use it verbatim instead of referring to the constant (e.g. in strip functionality). Replace those usage with reference to the constant as well.

nncf/quantization/algorithms/min_max/torch_backend.py
vshampor: > Additional steps before call transformation for `PTInsertionCommand` brakes logic of work with common algorithms. From `ModelTransformer` expected that all logic about insertion FQ will be done inside, without call another backend-specific functions.

It is impossible to insert FQs without calling backend-specific functions.

> 
> > Otherwise you are breaking the priority functionality of insertions into the PT models.
> 
> Please provide example of incorrect work of PT PTQ.

You won't be able to apply PTQ as an extra step on top of a pruned or sparsified model, for instance.

In general - just why are you repeating the code that is already there in the NNCF PT code base to insert quantizers? Why aren't you reusing the existing functionality? If the existing code is not convenient to use for the PTQ right now - refactor it so that it becomes convenient both for PTQ and QAT.

tests/torch/test_model_transformer.py
AlexanderDokuchaev: Good point. Updated. 
And added 'NNCFNetworkInterface:is_compression_module_registered' method to check that extra module are registered without raising Runtime exception.

nncf/quantization/algorithms/min_max/torch_backend.py
AlexanderDokuchaev: `ModelTransfomer` refactoring is already planned, ticket 102414.

nncf/torch/model_transformer.py
vshampor: I think that the shared quantizer handling is already there in the NNCF PT [code](https://github.com/openvinotoolkit/nncf/blob/631a7b8f2990a80b9b9473f999651b1808a7e7d4/nncf/torch/quantization/algo.py#L955C5-L966), you are just continuously deciding not to use it and roll your own code instead.

nncf/torch/checkpoint_loading.py
AlexanderDokuchaev: Please create ticket for this task.

vshampor: 121535

nncf/quantization/algorithms/min_max/torch_backend.py
AlexanderDokuchaev: > I'm pretty sure you can reuse most of the code in nncf.torch.model_transformer.PTModelTransformer._apply_insertion_transformations by calling it inside your new _apply_quantizer_insertion_transformations.

Done

daniil-lyakhov (approved):

AlexanderDokuchaev:
Fixed merge conflicts, fails only API check

MaximProshin:
@vshampor , I think it can be merged as API check fails due to other reasons

PR #2149, opened 2023-09-20
Use torch for CUDA 11.8 in tests

vshampor:


alexsu52 (approved):
LGTM

Makefile
```diff
@@ -117,11 +117,11 @@ test-examples-tensorflow:
 # PyTorch backend
 install-torch-test:
 	pip install -U pip
-	pip install -e .[torch]
-	pip install -r tests/torch/requirements.txt
+	pip install -e .[torch] --index-url https://download.pytorch.org/whl/cu118 --extra-index-url=https://pypi.org/simple  # ticket 119128
```
AlexanderDokuchaev, line 120: Tried to `pip install -e .[torch]  --extra-index-url=https://download.pytorch.org/whl/cu118` it's still install `torch-2.0.1%2Bcu118-cp310-cp310-linux_x86_64.whl`.
Looks like it can be replaced `pip install -e .[torch] --extra-index-url=https://download.pytorch.org/whl/cu118`

With `torch==2.0.1+cu118`  bug is not reproduced 👍

vshampor: That's what I tried too, but it didn't work in precommit. At the same time it works locally, maybe pip version is at fault. This is a temporary solution anyway until default torch starts shipping for CUDA 11.8.

PR #2013, opened 2023-07-28
Algorithm.py refactoring

alexsu52:

PR #2154, opened 2023-09-25
Add SmoothQuant and ChannelAlignment to HyperparameterTuner

andrey-churkin:
### Changes

- Add the `Pipeline` class. To use the algorithm with the `HyperparameterTuner`, it should be wrapped within a `Pipeline` object.

### Reason for changes

- Add the SmoothQuant and ChannelAlignment algorithms to the parameter optimization process.

### Related tickets

Ref: 117471

### DIFF
--- a/nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
+++ b/nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
@@ -333,12 +333,16 @@ class HyperparameterTuner:
         best_settings,
     ) -> None:
         """
-        TODO:
+        Creates a separate pipeline for each combination from step_combination.
+        Each combination only changes the parameters of the `step_index`-th pipeline
+        step. After that, combines the statistics required to execute the `step_index`-th
+        pipeline step and collects them using `step_model`, `step_graph`, and the calibration
+        dataset.
 
-        :param step_index:
-        :param step_model:
-        :param step_graph:
-        :param step_combinations:
+        :param step_index: Zero-based index of pipeline step that should be prepared.
+        :param step_model: A model.
+        :param step_graph: A graph assosiated with a model.
+        :param step_combinations: Combinations that change parameters only for the step_index-th pipeline step.
         """
         # Create a separate pipeline for each combination
 
--- a/nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
+++ b/nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
@@ -14,12 +14,12 @@ import dataclasses
 import functools
 import itertools
 import operator
-from typing import Any, Callable, Dict, Iterable, List, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, Iterable, List, Tuple, TypeVar, Union
 
 from nncf.common.factory import NNCFGraphFactory
-from nncf.common.factory import StatisticsAggregatorFactory
 from nncf.common.graph.graph import NNCFGraph
 from nncf.common.logging import nncf_logger
+from nncf.common.tensor_statistics.statistic_point import StatisticPointsContainer
 from nncf.common.utils.backend import get_backend
 from nncf.common.utils.timer import timer
 from nncf.data.dataset import Dataset
@@ -27,7 +27,8 @@ from nncf.quantization.algorithms.accuracy_control.evaluator import Evaluator
 from nncf.quantization.algorithms.accuracy_control.evaluator import MetricResults
 from nncf.quantization.algorithms.accuracy_control.rank_functions import create_normalized_mse_func
 from nncf.quantization.algorithms.accuracy_control.subset_selection import select_subset
-from nncf.quantization.algorithms.algorithm import Algorithm
+from nncf.quantization.algorithms.pipeline import Pipeline
+from nncf.quantization.algorithms.pipeline import collect_statistics
 
 TModel = TypeVar("TModel")
 TTensor = TypeVar("TTensor")
@@ -186,7 +187,7 @@ class HyperparameterTuner:
             "param_name": [0.1, 0.2],
         }
 
-    The parameters names should be same as in `algorithm_cls.__init__()` method.
+    The parameters names should be same as in `pipeline_fn()` method.
     In case when "param_name" parameter is a dataclass object there is a way to specify settings
     to try for his fields using marker ":"
 
@@ -214,9 +215,9 @@ class HyperparameterTuner:
 
     def __init__(
         self,
-        algorithm_cls: Type[Algorithm],
+        pipeline_fn: Callable[..., Pipeline],
         init_params: Dict[str, Any],
-        param_grid: Dict[str, List[Any]],
+        param_grids: List[Dict[str, List[Any]]],
         calibration_dataset: Dataset,
         validation_fn: Callable[[Any, Iterable[Any]], Tuple[float, Union[None, List[float], List[List[TTensor]]]]],
         subset_size: int,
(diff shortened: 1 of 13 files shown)


nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
```diff
@@ -70,5 +74,38 @@ def get_quantization_param_grid() -> Dict[str, Any]:
             for min_v, max_v in itertools.product(min_param_values, max_param_values)
         ],
     }
-
     return param_grid
+
+
+def _get_smooth_quant_param_grid() -> ParamGrid:
+    return {"advanced_parameters:smooth_quant_alpha": [0.2, 0.5, 0.9, 0.95]}
```
andreyanufr, line 81: Is it possible to set first parameter 0.15 and make more "symmetric" grid: [0.15, 0.25, 0.5, 0.75, 0.95] ?

alexsu52 (commented):
I'm OK with this solution. Please clean code and write doc strings.

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
andrey-churkin: Sure! I used these values only for testing purposes. I updated them.

nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
```diff
@@ -275,58 +276,101 @@ def apply(self, model: TModel, validation_dataset: Dataset) -> TModel:
...
+            step_model = pipeline.run_step(
+                step_index, self._step_index_to_statistics[step_index], step_model, step_graph
+            )
 
-        algorithm = self._algorithms[best_combination_key]
-        result_model = algorithm.apply(model, initial_graph, self._statistic_points)
+        # TODO(andrey-churkin): Show final best settings
 
-        return result_model
+        return step_model
```
KodiaqQ, line 325: When does the `best_settings` apply to the model? I see that at each step `best_settings` updates with the new combination.

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
```diff
@@ -70,5 +74,38 @@ def get_quantization_param_grid() -> Dict[str, Any]:
...
-
     return param_grid
+
+
+def _get_smooth_quant_param_grid() -> ParamGrid:
+    return {"advanced_parameters:smooth_quant_alpha": [0.15, 0.25, 0.5, 0.75, 0.95]}
+
+
+def _get_channel_alignment_param_grid() -> ParamGrid:
+    return {}
```
KodiaqQ, line 85: Why is it empty since we add `ChannelAlignment` in this PR to the Tuner?

nncf/quantization/algorithms/post_training/algorithm.py
```diff
@@ -175,41 +84,14 @@ def apply(
...
-                nncf_logger.debug(f"{backend.name} does not support ChannelAlignment algorithm yet.")
-                continue
-
-            statistics_aggregator = StatisticsAggregatorFactory.create(modified_model, dataset)
-            algo_statistic_points = algorithm.get_statistic_points(modified_model, modified_model_graph)
-            statistics_aggregator.register_statistic_points(algo_statistic_points)
-            statistics_aggregator.collect_statistics(modified_model, modified_model_graph)
-            modified_model = algorithm.apply(
-                modified_model, modified_model_graph, statistics_aggregator.statistic_points
+        if dataset is None and len(self._pipeline.pipeline_steps) > 1:
```
KodiaqQ, line 87: What would happen if the `len(self._pipeline.pipeline_steps)` be 1? What should the user do?

nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
andrey-churkin: The algorithm updates `step_model` iteratively. It starts with `step_model` equal to `model`. After that, the algorithm searches for the best parameters for the current step of the pipeline. Once it finds them, it applies the step (with the found best parameters) to `step_model` to obtain the next `step_model`. So, when the `for` statement is finished, it ends up with the final model obtained using the best settings.

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
andrey-churkin: The `ChannelAlignment` algorithm does not have any parameters to tune, but it is still applied to the model as a pipeline step.

nncf/quantization/algorithms/post_training/algorithm.py
andrey-churkin: There are two scenarios

1. When `len(pipeline_steps)` equals `1`, in this case, we can pass only a value for the `statistic_points` parameter to the `PostTrainingQuantization.apply()` method, and there's no need to pass a value for the `dataset` parameter.
   ```python
   quantized_model = PostTrainingQuantization.apply(model, graph, statistic_points=statistic_points)
   ```
   It is also possible to pass only a value for the `dataset` parameter to call this method.
   ```python
   quantized_model = PostTrainingQuantization.apply(model, graph, dataset=calibration_dataset)
   ```
2. When `len(pipeline_steps)` is greater than `1`, in this case, a value for the `dataset` parameter is always required because it is necessary to collect statistics for intermediate models. So, the `PostTrainingQuantization.apply()` method can be called the following ways
   ```python
   # Pass the statistic_points collected for the first step and the dataset
   quantized_model = PostTrainingQuantization.apply(model, graph, statistic_points, dataset)

   # Pass the dataset only
   quantized_model = PostTrainingQuantization.apply(model, graph, dataset=calibration_dataset)
   ```
I understand that this doesn't appear user-friendly. I guess this API will be improved here https://github.com/openvinotoolkit/nncf/pull/2013

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
KodiaqQ: I thought that we should tune `ChannelAlignment` as the algorithm itself: apply it or not. Otherwise, I don't get why should we have a grid for this algo.
@alexsu52

nncf/quantization/algorithms/hyperparameter_tuner/algorithm.py
KodiaqQ: Does it mean that we always keep two models in the memory: one with the current `best configuration` and the second with the `current configuration` that would be compared to the `best`?

andrey-churkin: When the algorithm optimizes parameters for the (i + 1)-th step, there is the `step_model_{i}` model for which 0 to i steps were applied with optimal parameters.

For the current configuration of parameters, the algorithm applies the (i + 1)-th step with this configuration to `step_model_{i}` model to obtain `current_step_model_{i+1}` model and calculates the score. After that, `current_step_model_{i+1}` model is destroyed. See `HyperparameterTuner._calculate_combination_score()` method for more details.

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
andrey-churkin: I have just one idea for how to handle this. I propose always including the `ChannelAlignment` algorithm in the Pipeline and adding an additional parameter, `disable_algorithm`, to the `ChannelAlignment.__init__()` method. This parameter would allow us to disable this algorithm when needed. What do you think?

KodiaqQ: Looks like it's the same parameter with the `disable_channel_alignment`, but the new one keeps the `ChannelAlignment` instance alive. Since we had a problem with this behaviour, but with the `SmoothQuant` (see #2090), I propose to keep this case in mind.

andrey-churkin: I've filed the ticket (121767) to support this feature.

KodiaqQ (approved):

alexsu52 (approved):
LGTM

andreyanufr (approved):

andrey-churkin:
Build: 10740

PR #2156, opened 2023-09-25
Using ov.save_model(...) in PTQ examples

alexsu52:
### Changes

Using ov.save_model(...) in PTQ examples

### Reason for changes

ov.save_model(...) is introduced in OpenVINO==2023.1.0

### Related tickets

N/A

### Tests

examples

### DIFF
--- a/examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
+++ b/examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
@@ -19,7 +19,7 @@ from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import numpy as np
-import openvino.runtime as ov
+import openvino as ov
 import torch
 from anomalib.data.mvtec import MVTec
 from anomalib.data.utils import download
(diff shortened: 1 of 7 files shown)


MaximProshin:
As we agreed offline, let's switch to fp16 in examples.

MaximProshin:
`import openvino.runtime as ov`
can now be replaced by 
`import openvino as ov`

examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
```diff
@@ -165,12 +165,12 @@ def transform_fn(data_item):
 # Benchmark performance, calculate compression rate and validate accuracy
 
 fp32_ir_path = f"{ROOT}/stfpm_fp32.xml"
-ov.serialize(ov_model, fp32_ir_path)
+ov.save_model(ov_model, fp32_ir_path, compress_to_fp16=False)
```
MaximProshin, line 168: add a comment why this is required here

MaximProshin:
@AlexKoff88 , please provide your feedback. As by default convert() doesn't return fp16 we agreed to keep as it's in the PR and add a comment why it's required to use 'compress_to_fp16=False' in the examples.

alexsu52:
> `import openvino.runtime as ov` can now be replaced by `import openvino as ov`

done

examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
alexsu52: done

MaximProshin (approved):

PR #2157, opened 2023-09-25
[PTQ] Parallel quantizers merge

KodiaqQ:
### Changes

- Updated `Solver` and `QPSG` to save branched quantizers group for further merge in the PTQ

### Reason for changes

- Better quantization scheme

### Related tickets

- 106193

### Tests

TBD


nncf/onnx/graph/transformations/commands.py
```diff
@@ -22,18 +22,26 @@
...
+        destination_node_names: Optional[List[str]] = None,
+    ):
         """
         Constructor.
 
         :param target_type: Target type of the target point.
         :param target_node_name: ONNX node name.
         :param port_id: Number of port id in case target_type is
             TargetType.PRE_LAYER_OPERATION or TargetType.POST_LAYER_OPERATION.
+        :param destination_node_names: List of successor the node names.
```
vshampor, line 39: I understand your need to store the affected node set in a `TargetPoint`, but the way you are doing it right now is too messy. Evidently we need at least a new `TargetType` here, something like `POST_BRANCH_WITH_PARTIAL_MERGE`, the additional field `branch_node_names` that would only be valid for this new type (can set up some runtime checks in `__init__` to make sure that we don't instantiate a wrong `TargetPoint`; by the way, is there an opportunity to do this in the base `TargetPoint` class, and maybe move `target_node_name` and `port_id` there too?). It will also then be possible to implement a check in the `PTTargetPoint` so that a target point with this type cannot be instantiated with a `RuntimeError("Torch does not support target type TargetType.POST_BRANCH_WITH_PARTIAL_MERGE")`, and maybe do the same for TF right now, but with a `NotImplementedError` instead.

vshampor (changes requested):

nncf/common/quantization/quantizer_setup.py
```diff
@@ -514,6 +541,24 @@ def select_first_qconfig_for_each_point(self) -> SingleConfigQuantizerSetup:
             qp_id_vs_qconfig_dict[qp_id] = qp.possible_qconfigs[0]
         return self.select_qconfigs(qp_id_vs_qconfig_dict)
 
+    def filter_branching_quantizers(self, retval: SingleConfigQuantizerSetup) -> SingleConfigQuantizerSetup:
+        """
+        Filters brancging quantizert groups based on the quantizer configuration.
```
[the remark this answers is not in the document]
KodiaqQ: Done.

nncf/common/quantization/quantizer_setup.py
```diff
@@ -299,6 +311,12 @@ def get_unified_scale_group_id(self, qp_id: QuantizationPointId) -> Optional[int
                 return gid
         return None
 
+    def get_branch_merge_group_id(self, qp_id: QuantizationPointId) -> Optional[int]:
```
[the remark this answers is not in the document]
KodiaqQ: Renamed.

nncf/common/quantization/quantizer_propagation/grouping.py
```diff
@@ -9,44 +9,32 @@
...
-
-    def _get_next_gid(self) -> int:
-        retval = self._next_gid
-        self._next_gid += 1
-        return retval
+        self._group_vs_prop_quants_dict = {}  # type: Dict[Any, Set[PropagatingQuantizer]]
 
-    def register_group(self, prop_quants: Set[PropagatingQuantizer]) -> int:
+    @abstractmethod
+    def register_group(self, prop_quants: Set[PropagatingQuantizer], **kwargs) -> Any:
```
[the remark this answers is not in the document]
KodiaqQ: Done.

nncf/onnx/graph/transformations/commands.py
KodiaqQ: Added new options into the `TargetPoint`.
The new `TargetType.POST_BRANCH_WITH_PARTIAL_MERGE` looks too heavy in comparison to the existing ones, IMO.

vshampor (changes requested):

vshampor (approved):

nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -1510,3 +1518,15 @@ def traverse_fn(curr_node_key: str, unused) -> Tuple[bool, Any]:
...
                 assert pq in op_node[QuantizerPropagationStateGraph.AFFECTING_PROPAGATING_QUANTIZERS_ATTR]
+
+    def add_branching_quantizer_to_group(self, propagation_quantizer: PropagatingQuantizer, group_id: str) -> None:
+        """ "
+        Adds branching quantizer into the group associated by provided group ID.
+
+        :param propagation_quantizer: PropagatingQuantizer instance.
+        :param group_id: Group ID.
+        """
+        if group_id in self._partial_branch_merge_group_manager.get_group_vs_prop_quants_dict().keys():
```
daniil-lyakhov, line 1529: Looks like pylint missed it because couldn't assign dict type to function output
```suggestion
        if group_id in self._partial_branch_merge_group_manager.get_group_vs_prop_quants_dict():
```

nncf/common/graph/transformations/commands.py
```diff
@@ -133,18 +139,39 @@ class TargetPoint:
 
     _state_names = TargetPointStateNames
 
-    def __init__(self, target_type: TargetType):
+    def __init__(
+        self,
+        target_type: TargetType,
+        target_node_name: Optional[str] = None,
+        port_id: Optional[int] = None,
+        branch_node_names: Optional[List[str]] = None,
```
daniil-lyakhov, line 147: Finally! I was waiting for that for some time 💯

daniil-lyakhov: Don't forget to update docstring please

nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -1510,3 +1518,15 @@ def traverse_fn(curr_node_key: str, unused) -> Tuple[bool, Any]:
                     == QuantizerPropagationStateGraphNodeType.OPERATOR
                 )
                 assert pq in op_node[QuantizerPropagationStateGraph.AFFECTING_PROPAGATING_QUANTIZERS_ATTR]
+
+    def add_branching_quantizer_to_group(self, propagation_quantizer: PropagatingQuantizer, group_id: str) -> None:
+        """ "
```
daniil-lyakhov, line 1523:
```suggestion
        """
```

daniil-lyakhov:
@KodiaqQ, did you check that `POST_BRANCH_WITH_PARTIAL_MERGE` works properly in `_get_merged_statistic_points`?

KodiaqQ:
> @KodiaqQ, did you check that `POST_BRANCH_WITH_PARTIAL_MERGE` works properly in `_get_merged_statistic_points`?

I thought that it was done in our tests for the new types automatically, was it?

daniil-lyakhov:
> > @KodiaqQ, did you check that `POST_BRANCH_WITH_PARTIAL_MERGE` works properly in `_get_merged_statistic_points`?
> 
> I thought that it was done in our tests for the new types automatically, was it?

Unfortunately it wasn't, and I don't thing there is an easy way to enable new TargetPoint types automatically. it To enable tests one need to implement a model with `POST_BRANCH_WITH_PARTIAL_MERGE` type and enable it in the [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/common/test_statistics_aggregator.py#L739)

KodiaqQ:
> Unfortunately it wasn't, and I don't think there is an easy way to enable new TargetPoint types automatically. it To enable tests one need to implement a model with `POST_BRANCH_WITH_PARTIAL_MERGE` type and enable it in the [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/common/test_statistics_aggregator.py#L739). Merging algorithm should be updated as well to maintain target point merging with new type

To be honest, there is no new behaviour with this `TargetType` in comparison with the `POST_LAYER_OPERATION`, for example.
The test model was added to the OpenVINO backend section. Isn't it enough?

nncf/common/graph/transformations/commands.py
```diff
@@ -63,6 +63,8 @@ class TargetType(IntEnum):
                           specific module attributes - N/A in TF
     `OPERATOR_POST_HOOK` - a location after a function call in PT without access to
                            specific module attributes - N/A in TF
+    `POST_BRANCH_WITH_PARTIAL_MERGE` - a location after the associated model layer,
```
alexsu52, line 66: Did you implement support for POST_BRANCH_WITH_PARTIAL_MERGE in the ONNX and Torch?

nncf/common/graph/transformations/commands.py
```diff
@@ -63,6 +63,8 @@ class TargetType(IntEnum):
...
+                                       that represents the connection with the specific branches
```
alexsu52, line 67: By dentition, TargetType describes **the types of locations in the model that can be modified** using NNCF in order to create a compressed model. For example, Torch will insert the same FQ before branch_node_names and ONNX and OpenVINO I hope will do the same.

In my opinion, you should use special command or extension of existing for this that was implemented via OPERATOR_PRE_HOOK in Torch backend.

nncf/common/graph/transformations/commands.py
KodiaqQ: This is not applicable to the Torch as I understand.
For the ONNX it's implemented as well.

nncf/common/graph/transformations/commands.py
KodiaqQ: @vshampor, could you please comment on the possibility of the proposal for the Torch backend?

vshampor: It does not seem possible to me that torch backend would be able produce the kind of topology that we need here (with the possibilities of graph modification in torch that we have now).

alexsu52: @vshampor can we discuss this offline?

KodiaqQ:
The question of the current solution and alternative proposals should be discussed with @alexsu52 and @vshampor before the merge.

nncf/common/graph/transformations/commands.py
alexsu52: Please, open a ticket for Torch implementation.

alexsu52 (approved):
LGTM

kshpv (approved):

PR #2132, opened 2023-09-12
Added rich-based progress bar

nikita-savelyevv:
### Changes

Added `nncf.common.logging.track_progress.track()` method to replace `tqdm` for quantization algorithms bars.
This method is an almost exact copy of `rich.progress.track()` method, but with an addition of an iteration counter. This is kind of a hack, but `rich` does not provide a way to add custom `Column` objects to the `track()` method.

By default `rich.progress.track()` renders progress bar as 
image

With an addition of customizations this becomes
ac437dfc-de24-46fc-9538-3bdfee2e3ac9

With this change the quantization pipeline looks like
image

For iterables without length the progress bar displays as 
image

How it looks in a notebook:
image

### Reason for changes

- User experience improvement
- Avoiding multi-line logs in CI produced by `tqdm`, for example:
Statistics collection:   0%|          | 0/300 [00:00<?, ?it/s]
Statistics collection:   0%|          | 1/300 [00:00<01:13,  4.09it/s]
Statistics collection:   1%|          | 2/300 [00:00<00:50,  5.87it/s]
Statistics collection:   1%|▏         | 4/300 [00:00<00:33,  8.79it/s]
Statistics collection:   2%|▏         | 6/300 [00:00<00:26, 11.30it/s]
Statistics collection:   3%|▎         | 8/300 [00:00<00:24, 11.88it/s]

PR #2162, opened 2023-09-26
Progress bar visuals enhancement 

nikita-savelyevv:
### Changes

- Progress bar now extends to screen width
- Iteration count, time spent and time remaining text components are now Intel blue (RGB [0,104,181])

How it looks in terminal (Windows Terminal):
image

How it looks in Jupyter:
image

Please see how it looked before at #2132

Note: I've tried changing color for progress percent too, but couldn't, most probably due to a bug in rich https://github.com/Textualize/rich/issues/3133

### Reason for changes

- Improving user experience

Special thanks to @AlexanderDokuchaev and @kshpv for suggesting these changes.

### DIFF
--- a/nncf/common/logging/track_progress.py
+++ b/nncf/common/logging/track_progress.py
@@ -50,13 +50,13 @@ class SeparatorColumn(ProgressColumn):
 class TimeElapsedColumnWithStyle(TimeElapsedColumn):
     def render(self, task: "Task") -> Text:
         text = super().render(task)
-        return Text(text._text[0], style=INTEL_BLUE_COLOR)
+        return Text(text._text[0], style=INTEL_BLUE_COLOR)  # pylint: disable=protected-access
 
 
 class TimeRemainingColumnWithStyle(TimeRemainingColumn):
     def render(self, task: "Task") -> Text:
         text = super().render(task)
-        return Text(text._text[0], style=INTEL_BLUE_COLOR)
+        return Text(text._text[0], style=INTEL_BLUE_COLOR)  # pylint: disable=protected-access
 
 
 class track:
--- a/nncf/common/logging/track_progress.py
+++ b/nncf/common/logging/track_progress.py
@@ -23,7 +23,6 @@ from rich.progress import TextColumn
 from rich.style import StyleType
 from rich.text import Text
 
-# INTEL_BLUE_COLOR = (0, 174, 239)
 INTEL_BLUE_COLOR = (0, 113, 197)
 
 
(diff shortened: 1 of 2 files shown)


nncf/common/logging/track_progress.py
```diff
@@ -47,6 +49,73 @@ def render(self, task: Task) -> Text:
...
+    def render(self, task: "Task") -> Text:
+        """Show time elapsed."""
+        elapsed = task.finished_time if task.finished else task.elapsed
+        if elapsed is None:
+            return Text("-:--:--", style=self.style)
+        delta = timedelta(seconds=max(0, int(elapsed)))
+        return Text(str(delta), style=self.style)
+
+
+class TimeRemainingColumnWithStyle(ProgressColumn):
```
AlexanderDokuchaev, line 72: Is there a way to override default styles?
Maybe it would be better to inherit from `TimeRemainingColumn`?
```python
class TimeRemainingColumnWithStyle(TimeRemainingColumn):

    def render(self, task: "Task") -> Text:
        text = super().render(task)
        return Text(text._text[0], style=INTEL_BLUE_COLOR_STYLE)
```
Ofc it's access to protected attribute but looks shorter and without copypast.

nncf/common/logging/track_progress.py
```diff
@@ -20,20 +20,22 @@
 from rich.progress import Task
 from rich.progress import TaskProgressColumn
 from rich.progress import TextColumn
-from rich.progress import TimeElapsedColumn
-from rich.progress import TimeRemainingColumn
 from rich.style import StyleType
 from rich.text import Text
 
+INTEL_BLUE_COLOR = (0, 113, 197)
```
AlexanderDokuchaev, line 26: `INTEL_BLUE_COLOR_STYLE = "rgb(0,113,197)"`
To avoid `text_style = f"rgb({INTEL_BLUE_COLOR[0]},{INTEL_BLUE_COLOR[1]},{INTEL_BLUE_COLOR[2]})"`

AlexanderDokuchaev: image
On intel.com background of logo is `0, 104, 181` 😉

nncf/common/logging/track_progress.py
```diff
@@ -105,13 +176,14 @@ def __init__(
                     complete_style=complete_style,
                     finished_style=finished_style,
                     pulse_style=pulse_style,
+                    bar_width=None,
                 ),
                 TaskProgressColumn(show_speed=show_speed),
```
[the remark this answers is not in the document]
nikita-savelyevv: I would personally like to keep percent because it's a clear indication on the fraction of progress, otherwise one would need to estimate division of two numbers

nncf/common/logging/track_progress.py
nikita-savelyevv: Haven't found a way to override default style. Changed custom columns to the way you suggested.

nncf/common/logging/track_progress.py
nikita-savelyevv: Done

AlexanderDokuchaev (approved):

MaximProshin (approved):

kshpv (approved):

PR #1975, opened 2023-07-14
Move model type processing to Solver

l-bat:
### Changes

Mark nodes as ignored in Solver to reduce the log for the automatically generated ignored scope.

### Reason for changes

To improve user experience 

### Tests

conformance current log: 114
conformance with PR changes: 118

PR #1598, opened 2023-02-24
[PTQ] Add passes for PTQ to create inference NNCFGraph

KodiaqQ:
### Changes

- Added pass for PTQ to create inference NNCFGraph

### Reason for changes

- Deletion of ShapeOf subgraphs will allow Solver to insert quantizers correctly

### Related tickets

- 103931

PR #2166, opened 2023-09-29
[PTQ][Torch][KQV self attention] Align FQ placement between OV and Torch backend

daniil-lyakhov:
### Changes

* "unbing" and "__matmul__" ops are added to torch patterns
* Dropout removing pass is added to function `transform_to_inference_graph`
*  LayerNorm and GroupNorm metatypes are added to ignored metatypes in MinMax algorithm

### Reason for changes

To align quantization placement between OV and Torch backends for the following models:
* timm/crossvit_9_240
* timm/deit3_small_patch16_224
* timm/swin_base_patch4_window7_224

### Related tickets

121647

### Tests
*  test_metatypes_to_ignore for quantization propagation solver


AlexanderDokuchaev:
Please run post_training build and update metrics if needed.

nncf/quantization/algorithms/min_max/torch_backend.py
```diff
@@ -315,6 +319,10 @@ def get_ignored_metatypes(model_type: ModelType, device: TargetDevice) -> List[O
                 om.PTDivMetatype,
                 om.PTMaxMetatype,
                 om.PTSqueezeMetatype,
+                om.PTLayerNormMetatype,
+                om.PTModuleLayerNormMetatype,
+                om.PTGroupNormMetatype,
+                om.PTModuleGroupNormMetatype,
```
daniil-lyakhov, line 325: Test were missed it this pr https://github.com/openvinotoolkit/nncf/pull/1975 thus ignored metatypes solver logic is not tested now.
CC: @l-bat

nncf/quantization/passes.py
```diff
@@ -23,35 +24,38 @@
 def transform_to_inference_graph(
     nncf_graph: NNCFGraph,
     shapeof_metatypes: List[OperatorMetatype],
```
AlexanderDokuchaev, line 26: metatypes can be `Optional`, to avoid call with empty args.

daniil-lyakhov: Yeah, but it is always a list entity. In case there are no target metatypes metatypes list is empty, so there is no signature violation here

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -505,7 +505,10 @@ def _get_quantization_target_points(
         hw_patterns = PatternsManager.get_full_hw_pattern_graph(backend=backend, device=device, model_type=model_type)
 
         inference_nncf_graph = transform_to_inference_graph(
-            deepcopy(nncf_graph), self._backend_entity.shapeof_metatypes, self._backend_entity.read_variable_metatypes
+            nncf_graph,
+            self._backend_entity.shapeof_metatypes,
```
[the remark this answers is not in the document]
daniil-lyakhov: Done, please check

nncf/quantization/passes.py
daniil-lyakhov: Outdated

daniil-lyakhov:
PTQ conformance test finished successfully (build 175), metrics are updated accordingly

AlexanderDokuchaev (approved):

nncf/common/graph/graph.py
```diff
@@ -594,12 +594,22 @@ def get_graph_for_structure_analysis(self, extended: bool = False) -> nx.DiGraph
...
                 else:
                     attrs_edge["style"] = "solid"
-                attrs_edge["label"] = edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR]
+                label["shape"] = edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR]
+
+            if label:
+                if len(label) == 1 and extended:
+                    attrs_edge["label"] = label.popitem()[1]
+                else:
+                    attrs_edge["label"] = ", ".join((f"{k}:{v}" for k, v in label.items()))
```
KodiaqQ, line 612: Why do we need this?

nncf/quantization/algorithms/accuracy_control/ranker.py
```diff
@@ -98,8 +98,9 @@ def find_groups_of_quantizers_to_rank(self, quantized_model_graph: NNCFGraph) ->
             if x.metatype in self._algo_backend.get_quantizer_metatypes()
         ]
 
-        quantized_model_graph_without_shapeof = remove_shapeof_subgraphs(
-            deepcopy(quantized_model_graph), self._algo_backend.get_shapeof_metatypes()
+        quantized_model_graph_without_shapeof = deepcopy(quantized_model_graph)
```
KodiaqQ, line 101: This is just the graph copy, not `quantized_model_graph_without_shapeof`.

nncf/quantization/algorithms/accuracy_control/ranker.py
```diff
@@ -98,8 +98,9 @@ def find_groups_of_quantizers_to_rank(self, quantized_model_graph: NNCFGraph) ->
...
+        remove_shapeof_subgraphs_inplace(
+            quantized_model_graph_without_shapeof, self._algo_backend.get_shapeof_metatypes()
```
KodiaqQ, line 103: Why do we need to remove it in-place?

nncf/quantization/passes.py
```diff
@@ -85,20 +65,65 @@ def remove_shapeof_subgraphs(
...
+        for each dropout node.
+
+    :param nncf_graph: NNCFGraph instance for the transformation.
+    :param dropout_metatypes: List of backend-specific Dropout metatypes.
+    """
+    if not dropout_metatypes:
+        return
+
+    nodes_to_drop = []
+    for node in nncf_graph.get_all_nodes():
```
KodiaqQ, line 86: Why not to use `nncf_graph.get_nodes_by_metatypes`?

nncf/quantization/passes.py
```diff
@@ -85,20 +65,65 @@ def remove_shapeof_subgraphs(
...
+                assert input_edge.tensor_shape == output_edge.tensor_shape
+                nncf_graph.add_edge_between_nncf_nodes(
+                    from_node_id=prev_node.node_id,
+                    to_node_id=output_edge.to_node.node_id,
+                    tensor_shape=input_edge.tensor_shape,
+                    input_port_id=output_edge.input_port_id,
+                    output_port_id=input_edge.output_port_id,
+                    dtype=input_edge.dtype,
+                    parallel_input_port_ids=output_edge.parallel_input_port_ids,
+                )
```
KodiaqQ, line 113: What is the difference between removing `dropout` and any other nodes?

nncf/quantization/algorithms/min_max/torch_backend.py
```diff
@@ -188,6 +183,13 @@ def get_statistic_collector(
             collector.register_statistic_branch(container_key, reducer, aggregator)
         return collector
 
+    @staticmethod
+    def transform_to_inference_graph(graph: NNCFGraph) -> NNCFGraph:
```
alexsu52, line 187: It looks like the right place of `transform_to_inference_graph` into `passes.py`

nncf/quantization/passes.py
```diff
@@ -20,38 +20,18 @@
 TModel = TypeVar("TModel")
 
 
-def transform_to_inference_graph(
+def remove_shapeof_subgraphs_inplace(
     nncf_graph: NNCFGraph,
     shapeof_metatypes: List[OperatorMetatype],
     read_variable_metatypes: Optional[List[OperatorMetatype]] = None,
-) -> NNCFGraph:
+) -> None:
```
KodiaqQ, line 27: Since we drop the `transform_to_inference_graph` method, it may be more useful to add a backend-specific method that does the same, but to the `model_utils`, for example, since this method may be useful not only for `MinMax`.

tests/common/quantization/test_passes.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestModes(Enum):
+    VALID = "valid"
+    WRONG_TENSOR_SHAPE = "wrong_dropout_node"
+    WRONG_PARALLEL_EDGES = "wrong_parallel_edges"
+
+
+@pytest.mark.parametrize("mode", [TestModes.VALID, TestModes.WRONG_TENSOR_SHAPE, TestModes.WRONG_PARALLEL_EDGES])
+def test_remove_dropout_nodes_inplace(mode: TestModes):
```
KodiaqQ, line 32: Why do we not add tests for each pass since we introduce this file in this PR?

tests/post_training/test_templates/test_ptq_params.py
```diff
@@ -288,7 +276,8 @@ def test_quantization_points_overflow_fix(self, overflow_fix, affected_target_po
     @pytest.mark.parametrize("validate_scopes", (True, False))
     def test_validate_scope(self, test_params, validate_scopes):
         nncf_graph = test_params["test_model_type_pass"]["nncf_graph"]
-        inference_nncf_graph = transform_to_inference_graph(deepcopy(nncf_graph), [])
+        inference_nncf_graph = deepcopy(nncf_graph)
+        filter_constant_nodes_inplace(inference_nncf_graph)
```
KodiaqQ, line 280: This is the place where we will start to mix up every time since we need to understand what really we want to do: `transform_to_inference_graph` using backend-specific entity or just `filter_constant_nodes` using just `NNCFGraph`.
I suggest thinking about current implementation in case of usability.

tests/common/data/reference_graphs/passes/dropout_synthetic_model.dot
```diff
@@ -0,0 +1,19 @@
+strict digraph  {
```
KodiaqQ, line 1: Where is the original `.dot` for comparison of the test correctness?

KodiaqQ (changes requested):

nncf/quantization/algorithms/min_max/torch_backend.py
daniil-lyakhov: We should decide where is the best place for that. @AlexanderDokuchaev suggest algorithm backend is the best way for this method

nncf/quantization/algorithms/accuracy_control/ranker.py
```diff
@@ -98,8 +98,9 @@ def find_groups_of_quantizers_to_rank(self, quantized_model_graph: NNCFGraph) ->
             if x.metatype in self._algo_backend.get_quantizer_metatypes()
         ]
 
-        quantized_model_graph_without_shapeof = remove_shapeof_subgraphs(
-            deepcopy(quantized_model_graph), self._algo_backend.get_shapeof_metatypes()
+        quantized_model_graph_without_shapeof = deepcopy(quantized_model_graph)
+        remove_shapeof_subgraphs_inplace(
```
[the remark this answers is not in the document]
daniil-lyakhov: I was not aware of the fact that this function is modifying original model, I made a mistake because of that and decided that it would be clearer to add `inplace` to function name. Should I remove it from the name?

nncf/common/graph/graph.py
daniil-lyakhov: To add information about parallel edges into the dot graph representation. I've added this if to prevent modifications of all dot graph: only graphs with parallel edges information will have additional information in it

nncf/quantization/algorithms/accuracy_control/ranker.py
daniil-lyakhov: The function was working inplace, I make it clear by naming them `..._inplace`. Should I refactor all this functions to be out of place?

nncf/quantization/passes.py
daniil-lyakhov: Do you suggest to make this function common? With metatypy as an attribute? We don't need such function yet, that's why I made a specific function for that

tests/common/data/reference_graphs/passes/dropout_synthetic_model.dot
daniil-lyakhov: The target NNCFGraph is built in test code, should we really check its dot representation? Should we add graph test on test NNCFGraph? Or is it only for debugging?

tests/common/quantization/test_passes.py
daniil-lyakhov: Because there was no tests for passes before, `get_inference_nncf_graph` was not tested but rather used in tests as an helper

nncf/quantization/algorithms/accuracy_control/ranker.py
daniil-lyakhov: Please see the next line

nncf/common/graph/graph.py
KodiaqQ: Can we do it in a more general way? The hard-code like `label.popitem()[1]` looks weird.

nncf/quantization/algorithms/accuracy_control/ranker.py
KodiaqQ: I see and the `deepcopy(quantized_model_graph)` is still deepcopy, not `quantized_model_graph_without_shapeof`.

nncf/quantization/algorithms/accuracy_control/ranker.py
KodiaqQ: It's right, before, this method was doing its job in place. I suggest placing graph copying into this method and making it return the new `NNCFGraph` instance.

tests/post_training/test_templates/test_ptq_params.py
KodiaqQ: Let's do it in this PR.

tests/common/quantization/test_passes.py
KodiaqQ: Can we add tests for all passes then?

tests/common/data/reference_graphs/passes/dropout_synthetic_model.dot
KodiaqQ: IMO. The code representation without visual reference takes some time to imagine the graph that exists before the pass was applied. In the case of huge and complicated graphs, it would be very inappropriate not to provide such `.dot`s.
We have `original_nncf_graph` sub-folders for 3/4 backends for some reason, don't we?

nncf/quantization/passes.py
KodiaqQ: I don't see any reason to implement such a method instead of using `remove_nodes_from` yet. This is why I was asking for the difference.

daniil-lyakhov: In comparison with `NNCFGraph.remove_nodes_from`, this method connects previous nodes with following one to keep NNCFGraph connected

nncf/common/graph/graph.py
daniil-lyakhov: What would you recommend? is `list(label.values())[0]` looks any better?

nncf/quantization/algorithms/accuracy_control/ranker.py
daniil-lyakhov: It will affect performance as all steps during `get_inference_graph` will be coping models inside. Is that acceptable?

nncf/quantization/passes.py
daniil-lyakhov: I returned `transform_to_inference_graph` function, hope it's better now

nncf/quantization/algorithms/accuracy_control/ranker.py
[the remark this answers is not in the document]
daniil-lyakhov: I reverted the function renaming

nncf/quantization/algorithms/min_max/torch_backend.py
daniil-lyakhov: I reverted `transform_to_inference_graph` function back

nncf/quantization/passes.py
KodiaqQ: > this method connects previous nodes with following one to keep NNCFGraph connected

Does it mean that currently the functionality of the  `remove_nodes_from` is not correct and needs to be fixed?

nncf/quantization/passes.py
daniil-lyakhov: Done

tests/post_training/test_templates/test_ptq_params.py
daniil-lyakhov: I reverted transform_to_inference_graph function back

nncf/quantization/algorithms/accuracy_control/ranker.py
KodiaqQ: I would suggest that you implement copying only once as it is done for now. Not for each step as default.

nncf/quantization/passes.py
daniil-lyakhov: I don't know purpose of the `remove_nodes_from` function, I don't know is it correct or not

nncf/common/graph/graph.py
KodiaqQ: I recommend reducing hard-coding and implementing it more in a more general way, if possible.

nncf/quantization/passes.py
KodiaqQ: > I don't know purpose of the remove_nodes_from function, I don't know is it correct or not

Sorry, but if you're not familiar with the `remove_nodes_from`, but trying to implement your own functionality that `connects previous nodes with ...`, doesn't it mean that basic functionality does not as expected = not correct? 

Then, why do we need the separate functionality for `dropout` nodes? It's still not clear.

tests/common/quantization/test_passes.py
daniil-lyakhov: Aren't tests presented in https://github.com/openvinotoolkit/nncf/pull/1598, https://github.com/openvinotoolkit/nncf/pull/1642 cover enough?

nncf/common/graph/graph.py
daniil-lyakhov: Done

nncf/quantization/algorithms/accuracy_control/ranker.py
daniil-lyakhov: This comment is outdated, `transform_to_inference_graph` is reverted back

tests/common/quantization/test_passes.py
KodiaqQ: There were no tests for passes at all since they were used in the default pipeline.

KodiaqQ (approved):

tests/common/quantization/test_passes.py
daniil-lyakhov: AFAIK, passes `remove_shapeof_subgraphs` and `filter_constant_nodes` were checked implicitly by graph tests. For my pass it was easier to create a synthetic model to check the pass

PR #2171, opened 2023-10-04
Add `cleanup_torchscript_cache` to test_quantize_conformance.py

AlexanderDokuchaev:
### Changes

Add `cleanup_torchscript_cache` function to test_quantize_conformance.py

### Reason for changes

After run torch.jit.trace in convert_model, PyTorch does not clear the trace cache automatically.

Same function in torch test and openvino notebok:
https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/jit_utils.py#L59
https://github.com/openvinotoolkit/openvino_notebooks/blob/1932d4b4e99116bdedaa620c9dc92069fbb1f05e/notebooks/236-stable-diffusion-v2/implementation/conversion_helper_utils.py#L8

### DIFF
--- a/tests/post_training/pipelines/base.py
+++ b/tests/post_training/pipelines/base.py
@@ -278,6 +278,18 @@ class BaseTestPipeline(ABC):
         self.save_quantized_model()
         self.get_num_fq()
         self.validate()
+        self.cleanup_torchscript_cache()
+
+    @staticmethod
+    def cleanup_torchscript_cache():
+        """
+        Helper for removing cached model representation.
+
+        After run torch.jit.trace in convert_model, PyTorch does not clear the trace cache automatically.
+        """
+        torch._C._jit_clear_class_registry()
+        torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
+        torch.jit._state._clear_class_state()
 
     def get_run_info(self) -> RunInfo:
         return self.run_info
--- a/tests/post_training/pipelines/base.py
+++ b/tests/post_training/pipelines/base.py
@@ -287,6 +287,7 @@ class BaseTestPipeline(ABC):
 
         After run torch.jit.trace in convert_model, PyTorch does not clear the trace cache automatically.
         """
+        # pylint: disable=protected-access
         torch._C._jit_clear_class_registry()
         torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
         torch.jit._state._clear_class_state()


tests/post_training/test_quantize_conformance.py
```diff
@@ -104,6 +105,22 @@ def test_ptq_quantization(test_case_name, data, output, result):
...
 
     run_info.time_total = time.perf_counter() - start_time
+    process_connection.send((run_info, err_msg))
+
+
+@pytest.mark.parametrize("test_case_name", TEST_CASES.keys())
+def test_ptq_quantization(test_case_name, data, output, result):
+    main_connection, process_connection = Pipe()
+
+    # Use multiprocessing to clean torch cache after trace model to convert in IR
```
vshampor, line 115: What exactly is meant as a "mem leak", and how exactly does using multiprocessing help in this case? If the culprit is known to be torch, why could this have not been fixed by invoking torch-specific functions for cache clearing?

AlexanderDokuchaev: https://en.wikipedia.org/wiki/Memory_leak

You can reproduced it by 
```python
dummy_tensor = torch.rand([1, 3, 224, 224])
for i in range(42):
    timm_model = timm.create_model("hrnet_w18", num_classes=1000, in_chans=3, pretrained=True, checkpoint_path="")
    ov_model = ov.convert_model(timm_model, example_input=dummy_tensor)
```
It helps because torch cache cleaned on removing `Process`. 
Pytorch does not have public methods to clean cache like `torch.cuda.empty_cache()`

AlexanderDokuchaev: So maybe you should wait until reviews will be requested?

Just for interesting, how memory leaks depends from size of RAM?

vshampor: > So maybe you should wait until reviews will be requested?
> 
> Just for interesting, how memory leaks depends from size of RAM?

I am automatically requested for review as a member of nncf-maintainers. If the PR is not ready for review, you should have left it in draft state.

Memory leaks depend on the size of RAM in a sense that the more RAM you have, the more time it takes for a leak to crash a process. The 42 iterations in your script weren't enough in my case to produce a crash, neither did they lead to GPU memory increase, so had I not observed the resident memory increase in `htop` in parallel, I could have missed the entire symptom.

AlexanderDokuchaev:
Reproducer: 
```python
import openvino as ov
import timm
import torch

dummy_tensor = torch.rand([1, 3, 224, 224])

def cleanup_torchscript_cache():
    torch._C._jit_clear_class_registry()
    torch.jit._recursive.concrete_type_store = torch.jit._recursive.ConcreteTypeStore()
    torch.jit._state._clear_class_state()

for i in range(10):
    timm_model = timm.create_model("hrnet_w18", num_classes=1000, in_chans=3, pretrained=True, checkpoint_path="")
    ov_model = ov.convert_model(timm_model, example_input=dummy_tensor)
    # cleanup_torchscript_cache()
```

Run script with `mprof`
```bash
pip install memery_profiler
mprof run --python python3 mem.py
mprof plot -o plot.png
```

Without `cleanup_torchscript_cache`:
image
With `cleanup_torchscript_cache`:
image


In `test_quantize_conformance.py` result looks like: 


Model | RAM MiB
-- | --
hf/bert-base-uncased | 2198
timm/crossvit_9_240 | 4778
timm/darknet53 | 5264
timm/deit3_small_patch16_224 | 5259
timm/dla34 | 5345
timm/dpn68 | 5339
timm/efficientnet_b0 | 5476
timm/efficientnet_lite0 | 5626
timm/hrnet_w18 | 6224
timm/inception_resnet_v2 | 7009
timm/levit_128 | 7294
timm/mobilenetv2_050 | 7289
timm/mobilenetv3_small_050 | 7348
timm/regnetx_002 | 7482
timm/resnest14d | 7631
timm/resnet18 | 7694
timm/swin_base_patch4_window7_224 | 8186
timm/tf_inception_v3 | 8440
timm/vgg11 | 8817
timm/visformer_small | 8462
timm/wide_resnet50_2 | 8634

tests/post_training/test_quantize_conformance.py
vshampor: Now that you have introduced actual, specific code that solves the problem, the solution has become much better and it is no longer necessary to add extended explanations in-code (although it wouldn't hurt too).

vshampor (approved):
Good job tracking this down to actual source.

PR #2173, opened 2023-10-04
[ONNX] Remove ONNXGraph

kshpv:
### Changes

Remove ONNXGraph and place all its methods into module onnx_helper.py
To optimize performance four mappings are introduced for ONNX model manipulation.

`get_node_mapping(model: onnx.ModelProto) -> Dict[str, onnx.NodeProto]`
Mapping from node name to the corresponding node. This needs to not iterate through all nodes of a model.
`get_edge_info_mapping(model: onnx.ModelProto) -> Dict[str, onnx.ValueInfoProto]`
Mapping from edge name to corresponding edge information. This needs to not iterate through all edge infos of a model.
`get_children_node_mapping(model: onnx.ModelProto) -> Dict[str, List[onnx.NodeProto]]`
Mapping from edge name and corresponding nodes that consume that edge as an input. Used to traverse forwardwith the optimal performance.
`get_parents_node_mapping(model: onnx.ModelProto) -> Dict[str, onnx.NodeProto]`
Mapping from edge name to node which outputs this edge. Used to traverse backward with the optimal performance.

Locally measured perf difference after removing ONNXGraph.

Model | PR time (sec) | develop time (sec) | SpeedUp
-- | -- | -- | --
swinv2_cr_tiny_224| 91.434 | 105.73 | 15.64%
visformer_small | 57.265 | 59.097 | 3.2%
deit3_small_patch16_224 | 52.31 | 55.503 | 6.1%

### Reason for changes

Code refactor

### Related tickets

96982

### Tests

N/A


vshampor:
Exactly why was the speedup gained? It couldn't have been due to the usage of the `ONNXGraph` class per se, since in its `__init__` method nothing performance-impacting is being done. Breaking the class into free functions is a step back - if the perf slowdowns in the previous implementations were not due to unoptimizable Python class behaviour, then keep the class and the functions in it and instead optimize the real problems.

kshpv:
> Exactly why was the speedup gained? It couldn't have been due to the usage of the `ONNXGraph` class per se, since in its `__init__` method nothing performance-impacting is being done. Breaking the class into free functions is a step back - if the perf slowdowns in the previous implementations were not due to unoptimizable Python class behaviour, then keep the class and the functions in it and instead optimize the real problems.

Speedup was obtained because of reimplementing the function - `get_nodes_by_input()`. This function could be called a thousand times depending on the model. The current implementation is not optimized as it iterates through all nodes, while the proposed implementation has a mapping that reduces performance complexity.

The main goal of removing this class was due to complaints from developers. Developers don't like to create an instance of ONNXGraph for any purpose to work with ONNX model, especially in the case when they need to use one simple function.

vshampor:
> > Exactly why was the speedup gained? It couldn't have been due to the usage of the `ONNXGraph` class per se, since in its `__init__` method nothing performance-impacting is being done. Breaking the class into free functions is a step back - if the perf slowdowns in the previous implementations were not due to unoptimizable Python class behaviour, then keep the class and the functions in it and instead optimize the real problems.
> 
> Speedup was obtained because of reimplementing the function - `get_nodes_by_input()`. This function could be called a thousand times depending on the model. The current implementation is not optimized as it iterates through all nodes, while the proposed implementation has a mapping that reduces performance complexity.
> 
> The main goal of removing this class was due to complaints from developers. Developers don't like to create an instance of ONNXGraph for any purpose to work with ONNX model, especially in the case when they need to use one simple function.

The developers should then readjust their expectations and accept the fact that software development almost always benefits from OOP and properly designed classes. You have now turned a well-defined class with developer-oriented methods into a bunch of functions without immediately visible connection with each other and put them into a file that has "helper" in its name, which means that this file will now become a dump for every ONNX-related function. 

If the devs complain that they need to write a single extra line of code in scopes where they have a raw `ModelProto`, then the surrounding code should probably be refactored so that the scope already has the `ONNXGraph` - for instance, a function that accepted `ModelProto` as parameter should now be accepting `ONNXGraph` references, and the wrapping of `ModelProto` into `ONNXGraph` should be happening higher up the stack.

nncf/onnx/graph/model_transformer.py
```diff
@@ -397,11 +407,12 @@ def _apply_qdq_node_removing_transformations(
...
         """
-        onnx_graph = ONNXGraph(model)
         for transformation in transformations:
-            node = onnx_graph.get_node_by_name(transformation.target_point.target_node_name)
+            node_mapping = get_node_mapping(model)
+            edge_node_mapping = get_edge_node_mapping(model)
+            node = node_mapping[transformation.target_point.target_node_name]
 
-            node_children = onnx_graph.get_children(node)
+            node_children = get_children(node, edge_node_mapping)
```
vshampor, line 415: Really, how is the new approach better than the previous one? The raw line count has increased in this particular case, and barely changed if we look at the line count of the PR as a whole. 

You are now exposing the implementation details (numerous `mapping`s) while the user doesn't really need to know about this - instead the user needs to get children of a node with a given name from a graph, which is exactly what the previous implementation was about, and the code itself was readable and consistent with the purpose. Now the code is cluttered with extra calls and dicts that aren't used per se but are necessary to fit to the new approach.

vshampor (changes requested):

alexsu52:
> > > Exactly why was the speedup gained? It couldn't have been due to the usage of the `ONNXGraph` class per se, since in its `__init__` method nothing performance-impacting is being done. Breaking the class into free functions is a step back - if the perf slowdowns in the previous implementations were not due to unoptimizable Python class behaviour, then keep the class and the functions in it and instead optimize the real problems.
> > 
> > 
> > Speedup was obtained because of reimplementing the function - `get_nodes_by_input()`. This function could be called a thousand times depending on the model. The current implementation is not optimized as it iterates through all nodes, while the proposed implementation has a mapping that reduces performance complexity.
> > The main goal of removing this class was due to complaints from developers. Developers don't like to create an instance of ONNXGraph for any purpose to work with ONNX model, especially in the case when they need to use one simple function.
> 
> The developers should then readjust their expectations and accept the fact that software development almost always benefits from OOP and properly designed classes. You have now turned a well-defined class with developer-oriented methods into a bunch of functions without immediately visible connection with each other and put them into a file that has "helper" in its name, which means that this file will now become a dump for every ONNX-related function.
> 
> If the devs complain that they need to write a single extra line of code in scopes where they have a raw `ModelProto`, then the surrounding code should probably be refactored so that the scope already has the `ONNXGraph` - for instance, a function that accepted `ModelProto` as parameter should now be accepting `ONNXGraph` references, and the wrapping of `ModelProto` into `ONNXGraph` should be happening higher up the stack.

I would suggest focusing on implementing a common interface to extend the model object ref:116745 and avoid using ONNXGraph as a model in the common implementation. This refactoring prevents the kind of usage that occurred in some PR requests previously.

alexsu52 (approved):
LGTM

nncf/onnx/graph/model_transformer.py
```diff
@@ -193,7 +200,8 @@ def _apply_quantizer_insertion_transformations(
         """
         self._added_target_edges = Counter()
         for transformation in transformations:
-            model = self._insert_quantizer_dequantizer(model, transformation)
+            children_node_mapping = get_children_node_mapping(model)
```
andrey-churkin, line 203: I am not clear on why we should do this every iteration.

nncf/onnx/graph/model_transformer.py
```diff
@@ -336,7 +348,7 @@ def _insert_quantizer_dequantizer(
         )
         model.graph.initializer.extend([onnx_scale_tensor, onnx_zero_point_tensor])
         model.graph.value_info.extend([onnx_scale_value_info, onnx_zero_point_info])
-        insert_index = onnx_graph.get_node_index(input_nodes[0].name)
+        insert_index = get_node_index(model, input_nodes[0].name)
```
andrey-churkin, line 351: Could you please briefly explain why we need an insertion index? Is it not possible to insert at the end?

nncf/onnx/graph/onnx_helper.py
```diff
@@ -0,0 +1,290 @@
...
+# limitations under the License.
+from collections import defaultdict
+from typing import Dict, Iterator, List, Optional, Union
+
+import numpy as np
+import onnx
+from onnx import numpy_helper
+
+
+def get_node_mapping(model: onnx.ModelProto) -> Dict[str, onnx.NodeProto]:
```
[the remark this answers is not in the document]
kshpv: Good idea why not

nncf/onnx/graph/model_transformer.py
andrey-churkin: Since this comment is not related to removing the ONNX graph, let's investigate this in a follow-up PR.

kshpv: After offline discussion we decided to make these changes in the follow up PR

nncf/onnx/graph/model_transformer.py
```diff
@@ -274,51 +282,55 @@ def _get_scale_zero_point_tensors(
...
 
         :param model: Model to insert new nodes.
         :param transformation: QuantizeLinear-DequantizeLinear insertion transformation.
+        :param children_node_mapping: Mapping from edge name to nodes which consume this edge as an input.
         :return: Updated model with inserted QuantizeLinear-DequantizeLinear pair.
         """
-        onnx_graph = ONNXGraph(model)
-        target_edge_name = self._get_quantizer_dequantizer_edge_name(transformation, onnx_graph)
+        node_mapping = get_node_mapping(model)
+        target_edge_name = self._get_quantizer_dequantizer_edge_name(transformation, node_mapping)
```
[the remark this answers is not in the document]
andrey-churkin: I've added this finding to the refactoring roadmap to discuss with the team.

andrey-churkin (approved):

PR #2174, opened 2023-10-04
Update calibrate tool

andrey-churkin:
### Changes

- Add support for models that don't have per-sample metrics.

### Reason for changes

Some models don't have per-sample metrics. An error occurred with these models.

### Related tickets

N/A


andrey-churkin:
run openvino pre-commit tests

PR #2175, opened 2023-10-04
[ONNX, Torch, OpenVINO] Add MVN, GELU, LINEAR_SHIFT_SCALE patterns

kshpv:
### Changes

Add three new patterns. With these patterns, performance of the model from the ticket is aligned with quantized OpenVINO IR.
Left is before; right is after
image
image
image

### Reason for changes

To get the best performance

### Related tickets

117169

### Tests

N/A


nncf/onnx/hardware/fused_patterns.py
```diff
@@ -372,6 +459,16 @@ def create_linear_arithmetic_activations() -> GraphPattern:
     return linear
 
 
+@ONNX_HW_FUSED_PATTERNS.register(HWFusedPatternNames.LINEAR_ARITHMETIC_ARITHMETIC)
+def create_linear_arithmetic_arithmetic() -> GraphPattern:
```
l-bat, line 463: I think it could be `linear_shift_scale` and we can use `create_shift_scale()`

tests/openvino/native/test_pattern_manager.py
```diff
@@ -32,6 +32,9 @@
     HWFusedPatternNames.LINEAR_BATCH_NORM_ACTIVATIONS: "Not relevant for OpenVINO.",
     HWFusedPatternNames.LINEAR_BATCH_NORM_SCALE_SHIFT_ACTIVATIONS: "Not relevant for OpenVINO.",
     HWFusedPatternNames.LINEAR_SCALE_SHIFT_ACTIVATIONS: "Not relevant for OpenVINO.",
+    HWFusedPatternNames.LINEAR_ARITHMETIC_ARITHMETIC: "Not relevant for OpenVINO.",
```
alexsu52, line 35: Could you explain why LINEAR_ARITHMETIC_ARITHMETIC is not relevant to OpenVINO?

tests/torch/test_pattern_manager.py
```diff
@@ -65,6 +65,9 @@
     HWFusedPatternNames.LINEAR_SQUEEZE_ACTIVATIONS: "Not relevant for Torch.",
     HWFusedPatternNames.LINEAR_SQUEEZE_ARITHMETIC_ACTIVATIONS: "Not relevant for Torch.",
     HWFusedPatternNames.LINEAR_ACTIVATIONS_UNSQUEEZE_BN_SQUEEZE: "Not relevant for Torch.",
+    HWFusedPatternNames.LINEAR_ARITHMETIC_ARITHMETIC: "Not relevant for Torch.",
```
alexsu52, line 68: Could you explain why LINEAR_ARITHMETIC_ARITHMETIC is not relevant to Torch?

tests/openvino/native/test_pattern_manager.py
kshpv: Added this pattern to OpenVINO as well

nncf/onnx/hardware/fused_patterns.py
kshpv: Good idea! Done

andrey-churkin:
My general comment pertains to tests. Is it possible to include some tests to ensure that everything will be okay in the future?

kshpv:
> My general comment pertains to tests. Is it possible to include some tests to ensure that everything will be okay in the future?

That's a good question. For now, we don't have a robust mechanism for testing our recently added patterns except manual testing.

andrey-churkin (approved):

PR #2176, opened 2023-10-05
Fix OV precommit - update graphs with new names

kshpv:
### Changes

Update graphs with new names.
Freeze torch version for OV tests.

### Reason for changes

The names of the nodes are changed, while the topologies are the same.

### Related tickets

N/A

### Tests

N/A


alexsu52 (approved):
LGTM

PR #2177, opened 2023-10-05
[Torch] SE block HW fusing pattern is presented

daniil-lyakhov:
### Changes

* SE block ignored pattern is presented for Torch backend
image
* SE block ignored pattern for OV backend is adjusted
* NON_PATTERN_NODE_WITH_TYPE is introduced in the graph matcher

### Reason for changes

To align FQ layout between TORCH and OV backend for efficientnet_b0 model
To fix FQ placement for SE block:
Before: 
image
After:
image

### Related tickets

121647

### Tests
* test_non_pattern_node_with_type for the graph matcher

Post training quantization manual build 180

| Model | Backend              | Metric name | Metric value | Metric diff | Num FQ  | RAM MiB | Quant. time | Total time | Status  |
|-------|----------------------|-------------|--------------|-------------|---------|---------|-------------|------------|---------|
| 1     | timm/efficientnet_b0 | OV          | Acc@1        | 0.7688      | -0.0042 | 92      | 1033        | 0:00:25    | 1:03:58 |   |
| 2     | timm/efficientnet_b0 | TORCH       | Acc@1        | 0.7680      | -0.0049 | 92      | 892         | 0:00:48    | 1:04:27 |

* pattern manager tests


nncf/common/graph/patterns/patterns.py
```diff
@@ -296,7 +307,7 @@ class HWFusedPatternNames(Enum):
     NORMALIZE_L2_MULTIPLY = PatternDesc("normalize_l2_multiply")
     SCALE_SHIFT = PatternDesc("scale_shift")
     SHIFT_SCALE = PatternDesc("shift_scale")
-    SE_BLOCK = PatternDesc("se_block")
+    SE_BLOCK = PatternDesc("se_block", ignored_algorithms=[AlgorithmType.NAS])
```
ljaljushkin, line 310: Ignoring pattern for NAS looks good to me

tests/post_training/reference_data.yaml
```diff
@@ -104,7 +104,7 @@ timm/efficientnet_b0_backend_OLD_TORCH:
     metric_value: 0.75538
     metric_value_fp32: 0.77292
 timm/efficientnet_b0_backend_TORCH:
-    metric_value: 0.76476
+    metric_value: 0.7680
```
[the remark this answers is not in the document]
daniil-lyakhov: Whoops, that's right. Fixed

ljaljushkin (approved):

nncf/torch/hardware/fused_patterns.py
```diff
@@ -320,3 +320,120 @@ def create_h_sigmoid_act() -> GraphPattern:
     main_pattern.add_pattern_alternative(pattern)
 
     return main_pattern
+
+
+# pylint:disable=too-many-statements
+@PT_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SE_BLOCK)
```
alexsu52, line 326: As far as I remember, we agreed to move the SE pattern to ignore patterns.

alexsu52 (changes requested):
As far I remember we agreed to move SE pattern to ignore pattern

nncf/torch/hardware/fused_patterns.py
daniil-lyakhov: Should we do this in a separate PR as this change will affect both OV and Torch backends or should we do this in this one?

nncf/common/graph/patterns/patterns.py
daniil-lyakhov: This is outdated now, I removed this functional, SE Block moved to ignored patters

nncf/torch/hardware/fused_patterns.py
daniil-lyakhov: Done

nncf/common/graph/patterns/patterns.py
```diff
@@ -80,6 +80,7 @@ class GraphPattern:
     NODE_TYPE_ATTR = "metatype"
     ANY_PATTERN_NODE_TYPE = "ANY_PATTERN_NODE"
     NON_PATTERN_NODE_TYPE = "NON_PATTERN_NODE"
+    NON_PATTERN_NODE_WITH_TYPE = "NON_PATTERN_NODE_WITH_TYPE"
```
kshpv, line 83: I am not sure that this name looks good. `NON_PATTERN_NODE_TYPE` is a special name which means that this node is not included in the pattern but has edges leading to that pattern. Your proposal attribute means that the node is actually inside the pattern but should be excluded after matching.
What do you think about `PATTERN_NODE_TO_EXCLUDE`?

kshpv (changes requested):

nncf/openvino/quantization/ignored_patterns.py
```diff
@@ -108,3 +109,53 @@ def create_equal_logicalnot() -> GraphPattern:
 
     pattern.add_edge(equal_node, logical_not_node)
     return pattern
+
+
+@OPENVINO_IGNORED_PATTERNS.register(IgnoredPatternNames.SE_BLOCK)
+def create_se_block() -> GraphPattern:
+    pattern = GraphPattern()
+    any_node = pattern.add_node(
+        **{GraphPattern.LABEL_ATTR: "ANY", GraphPattern.METATYPE_ATTR: GraphPattern.NON_PATTERN_NODE_TYPE}
```
kshpv, line 118: Minor: I would change the label from `ANY` to `NON_PATTERN_NODE` to explicitly demonstrate the node is an outside pattern and not mixing up with `ANY_PATTERN_NODE_TYPE`

daniil-lyakhov: Done

nncf/common/graph/patterns/patterns.py
daniil-lyakhov: Done

kshpv (approved):

alexsu52 (approved):
LGTM

PR #2183, opened 2023-10-09
[OV] Apply output insertion transformation first for model with If node 

kshpv:
### Changes

Apply output transformation first to correctly take the tensor names.

### Reason for changes

This allows to use of these names to collect data from If inputs

### Related tickets

122100

### Tests

Manually model from the ticket was tested.

### DIFF
--- a/nncf/openvino/quantization/quantize_ifmodel.py
+++ b/nncf/openvino/quantization/quantize_ifmodel.py
@@ -246,7 +246,7 @@ class OVBackend:
         ]
         for index in input_indices:
             if not ov_node.inputs()[index].get_tensor().get_names():
-                ov_node.inputs()[index].get_tensor().set_names(set([f"if_{index}_input"]))
+                ov_node.inputs()[index].get_tensor().set_names(set([f"{if_node.node_name}_{index}_input"]))
             input_names.append(ov_node.inputs()[index].get_tensor().get_any_name())
         return input_names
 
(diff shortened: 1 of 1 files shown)


nncf/openvino/quantization/quantize_ifmodel.py
```diff
@@ -258,7 +261,9 @@ def get_if_cond_input_name(model: ov.Model, if_node: NNCFNode) -> str:
         """
         name_to_node_mapping = {op.get_friendly_name(): op for op in model.get_ops()}
         ov_node = name_to_node_mapping[if_node.node_name]
-        return ov_node.input_values()[0].any_name
+        if not ov_node.inputs()[0].get_tensor().get_names():
+            ov_node.inputs()[0].get_tensor().set_names(set(["if_0_input"]))
```
ilyachur, line 265: Why do we need to add some additional hardcoded names?

OpenVINO allows to work with model inputs/outputs without names.

kshpv: This PR is a hot fix of ticket 122100. Current implementation fully relies on the fact that tensors have names. These names are used to map outputs to corresponding tensors. 
The ticket with updating the current approach to not using tensor names will be created.

ilyachur: Ok.

kshpv:
Updated fix.
Also noticed that implementation of `_add_outputs_before_if_node` was not common - maed it common reusing `create_output_insertion_commands_if_node()` from `OVBackend`.

alexsu52 (approved):

PR #2190, opened 2023-10-11
Optimize NNCFGraph.get_node_by_name() method

andrey-churkin:
### Changes

Linear search was removed inside the `NNCFGraph.get_node_by_name()` method.

### Reason for changes

The implementation of the `NNCFGraph.get_node_by_name()` method is very slow. The total time spent on its execution (several calls during quantization) is 216.887 seconds for the databricks/dolly-v2-3b model. This method currently operates at O(N) time complexity, where N is the number of nodes in the `NNCFGraph`. I expected it to have O(1) complexity.

### Related tickets

Ref: 119299

### Tests

pre-commit scope


nncf/common/graph/graph.py
```diff
@@ -635,16 +640,11 @@ def _get_graph_for_visualization(self) -> nx.DiGraph:
         return out_graph
 
     def get_node_by_name(self, name: NNCFNodeName) -> NNCFNode:
-        matches = [node for node in self.get_all_nodes() if node.node_name == name]
-        if not matches:
+        node_id = self._node_name_to_node_id_map.get(name, None)
```
alexsu52, line 643: The original ticket is about removing `get_all_nodes()` from NNCFGraph.  `get_all_nodes()` uses in many places such as algorithms, NNCFGraph and etc. Speeding up `get_all_nodes()` accelerates NNCF pipeline more than speeding up only `get_node_by_name`. My opinion that this PR is workaround for `get_node_by_name` because new entity was added to handle it and it does not fix the main problem.

andrey-churkin: > Speeding up get_all_nodes() accelerates NNCF pipeline more than speeding up only get_node_by_name. 

I ran `cProfile` after my fix, and I got the following results.

| ncalls | tottime | percall | cumtime | percall | filename:lineno(function) |
|:------:|:---------:|:-------:|:----------:|:--------:|:----------------------------:|
|13      | 0.004     | 0.000   | 0.004      | 0.000    | nncf/nncf/common/graph/graph.py:271(get_all_nodes)|

According to this results, I don't consider that removing `get_all_nodes()` significantly improves execution time. The main problem was in the `get_node_by_name()` method.

alexsu52 (commented):
Please, provide results e2e and PTQ conformance tests

andrey-churkin:
Okay, some tests have failed. It seems that the `node_name` may not be unique. Due to this, I have a question: Why do we use `node_name` instead of `node_key` in the PTQ algorithm?
@vshampor Could you please help me answer the question: Why is `node_name` used inside `QuantizationInsertionPointBase` instead of `node_key`?
https://github.com/openvinotoolkit/nncf/blob/ff28bce1f0321784a966761b9a9a73f532bef7ac/nncf/common/quantization/quantizer_setup.py#L43-L47

vshampor:
Node key, of type `str`, is a networkx-specific implementation detail. Node name (in the form of `NNCFNodeName`) is the entity that the users of NNCFGraph should be dealing with. In NNCF PT in particular, the node key is basically the same as the node name, but also prefixed with a numerical ID for historical reasons.

The `NNCFNodeName` is currently just a typedef for a `str`, but it needs a separate type exactly to separate the node names that are part of the graph interaction API from other `str` objects. "Part of graph interaction API" means that the node names should have a stable format, so that they could be specified in configs or ignored scopes, should be the same names that are visible to the user in visualization tools such as `.dot` graphs etc., and should actually be unique and probably have a one-to-one mapping to the node keys of the underlying graph implementation (but not necessarily, one could theoretically do optimization of the storage space required for nodes by only storing unique data dicts for nodes). I'm not sure that all of these requirements are fulfilled by now, but this seems to be the way to go if we want users to understand just where such names are expected by the API.

When I introduced `NNCFNodeName`, I meant for it to be used as a marker for the implementation-decoupled, contract-bound "node names" as described above - unfortunately at that time I didn't seek out all the places in code where instead of the `str` the `NNCFNodeName` type hint should have been specified. So basically the "node keys" should not leave the internals of `NNCFGraph` if possible; for simplicity, the one-to-one mapping of node names to node keys should be enforced, but the node name is still not the same as node key in general, hence the typedef. If you see a `str` typehint in code where in fact a `NNCFNodeName` should have been specified, then these should be replaced to improve the understanding of where the visual, user-facing node names are utilized in our code.

andrey-churkin:
@vshampor Thank you for your reply! It was very helpful. I have one more question. As I understand, node names are not necessarily unique in both TF and PT. How do you think it is possible to make node names unique in the PT backend?

vshampor:
We control the node name creation process so the node names should be unique everywhere, otherwise it's a bug. For PT the node name is built from the `OperationAddress` which is unique since this is how we are addressing the operations in the torch models. I presume that for TF models there is already a TF-native name or a user-visible identifier for each node in TF's graph, and if it's unique, we can use that; if there is no user-visible node identifier in TF's native approach, then we are again free to choose our own format, and we can for simplicity just use the most visible of the TF internal identifiers for the nodes.

andrey-churkin:
post_training_quantization: 177 (176 build failed due to CI problems)

PR #2191, opened 2023-10-11
Skipping CUDA tests for CPU setup

AlexanderDokuchaev:
### Changes

Add check `torch.cuda.is_available` to `TestTorchCudaFBCAlgorithm` test.

### Related tickets

122537

### DIFF
--- a/tests/torch/ptq/test_fast_bias_correction.py
+++ b/tests/torch/ptq/test_fast_bias_correction.py
@@ -62,17 +62,14 @@ class TestTorchFBCAlgorithm(TemplateTestFBCAlgorithm):
         raise ValueError("Not found node with bias")
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="Skipping for CPU-only setups")
 class TestTorchCudaFBCAlgorithm(TestTorchFBCAlgorithm):
     @staticmethod
     def list_to_backend_type(data: List) -> torch.Tensor:
-        if not torch.cuda.is_available():
-            pytest.skip("Skipping for CPU-only setups")
         return torch.Tensor(data).cuda()
 
     @staticmethod
     def backend_specific_model(model: bool, tmp_dir: str):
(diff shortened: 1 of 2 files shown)


vshampor:
@AlexanderDokuchaev post a number of the corresponding passing torch CPU precommit build.

vshampor (commented):
I disapprove of the entire "templating" pattern for these kinds of tests. Because of this pattern, the common algo part is being run in each of the backend tests, leading to at least (N -1) * (time_in_common_algo_code) extra computing time spent in precommits. Also because of this pattern, you now have to put `pytest.skip` not into the test cases, where it would have been a very visible, local marker that this exact test case may be skipped, but into the overridden functions hosting the backend-specific functionality of the template, where it is not immediately visible which test cases the `.skip` will impact.

Anyway, if you have to continue with the current approach for now, I would expect a `pytest.skip` in the `def fn_to_type` method of this class as well.

After the algo unification, these tests should be split into tests for common code, tests for backend-specific parts of the algo (which probably won't share enough code to warrant a template), and maybe interface conformance tests between the backend and the common code.

AlexanderDokuchaev:
> Anyway, if you have to continue with the current approach for now, I would expect a pytest.skip in the def fn_to_type method of this class as well.

@vshampor, please provide a code of your proposal. As you describe, tests will still fail with same cuda error.

AlexanderDokuchaev:
precommit_torch_cpu/154/

vshampor:
The new solution is much more explicit, thanks. My previous statement still holds, though, although I naturally don't expect it to be applied in this PR. Please make sure to re-run the build and post the updated build number.

AlexanderDokuchaev:
`precommit_torch_cpu/154/` on last commit
```
[2023-10-12T13:46:01.604Z] * | 5bafdab3 check cuda in class decorator
```

vshampor (approved):

PR #2194, opened 2023-10-12
Added documentation for NF4 mode

ljaljushkin:


docs/compression_algorithms/CompressWeights.md
```diff
@@ -6,17 +6,44 @@
...
 
-- Compress weights of linear layers and embeddings to int8
+- Compress weights to 8-bit integer data type.
 
 ```python
 from nncf import compress_weights
 compressed_model = compress_weights(model)
 ```
 
+- Compress internal weights to nf4 data type with group size = 128.
```
MaximProshin, line 26: Why do you use 'internal' here and below but not in int8 case? Please align it.

docs/compression_algorithms/CompressWeights.md
```diff
@@ -6,17 +6,44 @@
...
+from nncf import CompressWeightsMode
+compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4, group_size=64, ratio=0.9)
+```
+
 ##### Limitations
 
 - The algorithm is supported for OpenVINO and PyTorch models.
 - The compression applies in-place.
 - The compressed model is not trainable.
+- NF4 support is experimental - models quantized to nf4 should not be faster models quantized to 8-bit integer on client CPUs.
```
MaximProshin, line 47: I think it's confusing. I would remove it as we don't know when CPU will support it or at lease rephrase it saying that models with nf4 compressed weights should not be faster than models with int8 compressed weights on client CPUs.

docs/compression_algorithms/CompressWeights.md
MaximProshin: I see, but I believe it's still confusing. Why do you want to explicitly use 'internal'? I think it's enough to explain that first and last layers are quantized to int8

docs/compression_algorithms/CompressWeights.md
ljaljushkin: corrected

docs/compression_algorithms/CompressWeights.md
ljaljushkin: Removed mentioning of internal weight to reduce confusion, explicitly say about first and last layers

ljaljushkin (commented):
Thanks for review, @MaximProshin! Please take a look at the corrected document once more.

docs/compression_algorithms/CompressWeights.md
```diff
@@ -6,17 +6,44 @@
...
 
+- Compress weights to nf4 data type with group size = 128, except first embedding and last linear layers - they are compressed to 8-bit integer data type.
+
+```python
+from nncf import compress_weights
+from nncf import CompressWeightsMode
+compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4)
+```
+
+- Compress part of weights to nf4 with a different group size and the rest layers to 8-bit integer data type.
```
AlexKoff88, line 35:
```suggestion
- Compress weights of 90% of layers to nf4 with the group size 64, and the rest of layers to 8-bit integer data type.
```

AlexKoff88 (approved):

MaximProshin (approved):

PR #2196, opened 2023-10-13
Graph build from dataloader

vshampor:
### Changes
Extends `ModelInputInfo` mechanism used to specify inputs to `NNCFNetwork` for graph building/exporting - now the input info can be specified either as `FillerInputInfo`, which functions pretty much the same as before and uses NNCF config file as the source of specification for the input tensors, or as `ExactInputInfo`, which allows to specify exact forward arguments for graph building. The latter is used to build the model graph based on outputs of dataloaders attached to `NNCFConfig` in the QAT API if the "input_info" field is not specified in `NNCFConfig`, and also in the PTQ API flow to build the graph based on the output of the calibration dataset.

### Reason for changes
Previously the PTQ API had to specify own `wrap_inputs_fn`, `wrap_outputs_fn`, `dummy_forward_fn` to make NNCFNetwork build its graph based on the outputs of the calibration dataloader - these functions had to be mostly copy-pasted from the QAT approach to preserve basic NNCF PT functionality such as traced tensor expiry, same tensor replication etc. The new approach allows code reuse. Also the QAT use cases where the init dataloaders are specified are made easier since "input_info" fields in the NNCFConfig may now be omitted.

### Related tickets
N/A

### Tests
tests.torch.test_graph_building.test_input_info_args_are_passed_into_forward
tests.torch.test_graph_building.test_filler_input_info_arg_generation
tests.torch.test_graph_building.test_compressed_model_creation_can_build_exact_input_infos_from_dataloader_in_config
tests.torch.ptq.test_quantize_model_helpers.test_create_nncf_network_with_nncf_dataset


vshampor:
@alexsu52 FYI

nncf/common/exporter.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         self._model = model
         self._input_names = input_names
         self._output_names = output_names
-        self._model_args = model_args if model_args else ({},)
+        self._model_args = model_args
```
alexsu52, line 50: What is the reason for these changes?

nncf/common/graph/graph.py
```diff
@@ -143,7 +143,7 @@ def __init__(
         self.to_node = to_node
         self.input_port_id = input_port_id
         self.output_port_id = output_port_id
-        self.tensor_shape = tensor_shape
+        self.tensor_shape: Tuple[int] = tuple(tensor_shape)
```
alexsu52, line 146: What is the reason for these changes?

nncf/torch/quantization/quantize_model.py
```diff
@@ -26,45 +26,41 @@
...
 from nncf.torch.quantization.weights_compression import insert_pre_compression_operations
-from nncf.torch.utils import get_model_device
-from nncf.torch.utils import is_tensor
 
 DEFAULT_RANGE_TYPE = "mean_min_max"
 
 
 # TODO(alexsu52): It is a workaround and should be removed.
-class CalibrationDataLoader(PTInitializingDataLoader):
+class CalibrationDataLoaderAdapter(PTInitializingDataLoader):
```
alexsu52, line 38: What is the reason for these changes?

nncf/common/exporter.py
vshampor: `None` should be the marker that the `model_args` were not provided, because it's easier to write a conditional checking that (compare `self._model_args is None` vs `self._model_args == ({},)`

nncf/common/graph/graph.py
vshampor: Type hinting, and through experience I have found that using tuples to represent shapes works better since there is no way to accidentally store a shape by reference where you expected it to be stored by copy.

nncf/torch/quantization/quantize_model.py
vshampor: This class doesn't seem to be a data loader that is functional on its own (e.g. instantiated with arbitrary data sources), its only purpose is to serve as an adapter from the `nncf.Dataset` to a `PTInitializingDataLoader`, so I renamed it accordingly.

nncf/common/graph/graph.py
andrey-churkin: I think it makes sense to use a tuple here because we probably don't want to change the shape at the edge. However, I think it's better to change the type in the signature of the `NNCFGraphEdge.__init__()` method.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -74,44 +79,164 @@ def replicate_fn(tensor: torch.Tensor) -> torch.Tensor:
...
+    FILLER_TYPE_RANDOM = "random"
+    FILLER_TYPES = [FILLER_TYPE_ONES, FILLER_TYPE_ZEROS, FILLER_TYPE_RANDOM]
+
+    def __init__(self, shape: List[int], type_str: str = "float", keyword=None, filler=None):
+        super().__init__(keyword=keyword)
+        self.shape = shape
+        self.type = self._string_to_torch_type(type_str)
+        if filler is None:
+            self.filler = self.FILLER_TYPE_ONES
+        else:
```
andrey-churkin, line 113: Are other types of filler used anywhere?

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -124,25 +249,33 @@ def wrap_inputs(self, model_args, model_kwargs):
...
                     device = get_model_device(self._module_ref_for_device)
-                dummy_tensor = create_mock_tensor(info_for_missing_input, device)
-                _ = nncf_model_input(dummy_tensor)
+                dummy_input_copy = dummy_input.clone().to(device)
+                _ = nncf_model_input(dummy_input_copy)
 
         return bound_model_params.args, bound_model_params.kwargs
+
+
+class HasDataloader(Protocol):
```
andrey-churkin, line 275: Could you please explain the necessity of this class?

nncf/torch/model_creation.py
```diff
@@ -156,6 +160,34 @@ def create_compressed_model(
...
+        "initialization dataloader to the NNCFConfig by calling "
+        "`NNCFConfig.register_extra_structs(...)` with one of the following extra "
+        f"structures:\n"
+        f"{EXTRA_STRUCTS_WITH_DATALOADERS}\n"
+        f"or by calling `nncf.torch.register_default_init_args`"
+    )
+
+
+def create_nncf_network_with_inputs_from_config(model: torch.nn.Module, config: NNCFConfig):
+    return create_nncf_network(model, config=config)
```
andrey-churkin, line 188: Could you please provide more information about why this method is required?

nncf/common/exporter.py
alexsu52: Ok, How did you check that this changes does not break something?

nncf/torch/quantization/quantize_model.py
alexsu52: This changes are not aligned with the PR description. I suggest to do it in a separate PR.

nncf/common/exporter.py
vshampor: I tend to rely on existing unit tests for that. In particular, `test_can_export_with_model_args` does that for PyTorch, and the TF doesn't use that arg at all. 

After you pointed this out, I actually discovered that I marked the corresponding test as xfail-ed at the same time when I added it. I removed the xfail and the test showed a bug in the implementation, which I fixed right now to make the test passing. So thanks for spotting this!

nncf/common/graph/graph.py
vshampor: I can do this if you insist, but I think that the most points in code right now try to pass a list to the __init__ method, so changing the __init__ signature would require changes in all of those places too if we want strict type conformance in this spot.

nncf/torch/quantization/quantize_model.py
vshampor: This is a related improvement, so you can't dismiss it on these grounds alone. But this name is not exposed, so there's no good reason for me to push it.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -74,44 +79,164 @@ def replicate_fn(tensor: torch.Tensor) -> torch.Tensor:
...
-        "Inputs mismatch - could not find kwarg '{}' in NNCF-wrapped model input "
-        "kwargs! " + INPUTS_MISMATCH_WARNING_TEXT
-    )
+class ModelInputInfo(abc.ABC):
+    @abc.abstractmethod
+    def get_forward_inputs(self, device: str = None) -> Tuple[Tuple, Dict]:
+        pass
+
+
+class ModelInputElement(abc.ABC):
```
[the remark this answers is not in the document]
vshampor: Leftover from a previous idea, removed this, thanks.

nncf/torch/dynamic_graph/io_handling.py
vshampor: It's not us who decide this, but the model topology and its treatment of inputs. For example, you can't use zeros and random for LM inputs, but one can think of a model that would require non-uniform inputs (random will cover this) to trigger the main control flow path. I haven't encountered a model that would require zeros exactly, but it seems to be a natural complement to ones.

nncf/torch/dynamic_graph/io_handling.py
vshampor: I need it in `ExactInputsInfo.from_nncf_config_dataloaders` in the line `extra_struct: HasDataloader` to tell the future type checker or developer about what is expected from an extra struct to be considered for providing a dataloader. We usually do this by subclassing, but it doesn't look like that we need an extra explicit class in the inheritance hierarchy of `NNCFExtraConfigStruct`s for this purpose alone. Looks like `Protocol`s solve the problem of having to specify required methods for purposes of typing without explicit inheritance. Actually this is the first time I encountered a place where one would need a `Protocol`, and I'm not entirely sure that this is the right way to use it.

I noticed that in the previous version the `EXTRA_STRUCTS_WITH_DATALOADER` was typed as `List[Type[NNCFExtraConfigStruct]]`. I changed it to `List[Type[HasDataloader]]`, maybe it looks better now. I do not yet know how to type a list of classes which both should subclass from one class (`NNCFExtraConfigStruct` in our case) and implement a protocol defined by another class (`HasDataloader`), perhaps you know of a solution.

nncf/torch/model_creation.py
vshampor: Another leftover, thanks for spotting.

nncf/torch/dynamic_graph/io_handling.py
andrey-churkin: It is not clear for me why this approach doesn't work
```python

if isinstance(extra_struct, (QuantizationRangeInitArgs, BNAdaptationInitArgs)):
    # We know that `data_loader` property exist

# INSTEAD OF

if extra_struct.__class__ in EXTRA_STRUCTS_WITH_DATALOADERS:
    extra_struct: HasDataloader
```

nncf/torch/quantization/quantize_model.py
```diff
@@ -26,44 +26,40 @@
...
-    This is required for proper initialization of certain compression algorithms.
+    An adapter from the nncf.Dataset to the PTInitializingDataLoader.
     """
 
-    def __init__(self, data_loader: Dataset):
-        super().__init__(data_loader)
+    def get_target(self, dataloader_output: Any) -> Any:
+        raise NotImplementedError
+
+    def __init__(self, dataset: Dataset):
```
andrey-churkin, line 46: Should be first method in the class

nncf/common/graph/graph.py
andrey-churkin: I'm okay with it. I don't think it is necessary to do this here.

examples/torch/object_detection/dataset.py
```diff
@@ -45,8 +45,8 @@ def get_training_dataset(dataset_name, path_to_annotations, path_to_imgs, config
     # for VOC path_to_imgs = path_to_annotations = voc_root
     assert dataset_name in ["voc", "coco"]
     preprocessing = get_preprocessing(config)
-    input_info_list = create_input_infos(config)
-    image_size = input_info_list[0].shape[-1]
+    input_info_list = FillerInputInfo.from_nncf_config(config)
```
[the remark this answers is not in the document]
vshampor: Done, thanks

nncf/torch/dynamic_graph/graph.py
```diff
@@ -650,6 +652,11 @@ def add_node(
 
         if node.op_exec_context.operator_name == MODEL_INPUT_OP_NAME:
             self._input_nncf_nodes.append(node)
+            assert inputs.op_args
```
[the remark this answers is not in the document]
vshampor: I use asserts for something that I know to be true, or must be true in all cases, something which must be statically universally true for any given state of code but which I cannot ensure by using typing. This is in comparison to RuntimeErrors, raising which may be a desired solution due to something only determined by the runtime. I removed the assertion and added the comment explaining the piece of static universal truth I expect here for the code that follows to work.

nncf/torch/dynamic_graph/graph_tracer.py
```diff
@@ -119,11 +49,11 @@ def trace_graph(
...
 def create_dummy_forward_fn(
-    input_infos: List[ModelInputInfo],
-    with_input_tracing=False,
-    wrap_inputs_fn=None,
-    wrap_outputs_fn=None,
-    with_output_tracing=False,
+    input_info: ModelInputInfo,
+    with_input_tracing: bool = False,
+    wrap_inputs_fn: Callable = None,
+    wrap_outputs_fn: Callable = None,
```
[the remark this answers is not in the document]
vshampor: Narrowed the type somewhat.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -74,44 +78,151 @@ def replicate_fn(tensor: torch.Tensor) -> torch.Tensor:
...
-    )
-    ARGS_INPUTS_MISMATCH_FORMAT_STRING = (
-        "Inputs mismatch - could not find arg with idx {} in NNCF-wrapped model "
-        "input args! " + INPUTS_MISMATCH_WARNING_TEXT
-    )
-    KWARGS_INPUTS_MISMATCH_FORMAT_STRING = (
-        "Inputs mismatch - could not find kwarg '{}' in NNCF-wrapped model input "
-        "kwargs! " + INPUTS_MISMATCH_WARNING_TEXT
-    )
+class ModelInputInfo(abc.ABC):
```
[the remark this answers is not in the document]
vshampor: Done, thanks.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -74,44 +78,151 @@ def replicate_fn(tensor: torch.Tensor) -> torch.Tensor:
...
+    def __init__(self, shape: List[int], type_str: str = "float", keyword=None, filler=None):
+        self.shape = shape
+        self.type = self._string_to_torch_type(type_str)
+        self.keyword = keyword
+        if filler is None:
+            self.filler = self.FILLER_TYPE_ONES
+        else:
+            self.filler = filler
+            if self.filler not in self.FILLER_TYPES:
+                raise RuntimeError("Unknown input filler type: {}".format(filler))
```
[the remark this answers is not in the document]
vshampor: Done

nncf/torch/quantization/quantize_model.py
vshampor: Done

vshampor:
>It is not clear for me why this approach doesn't work
>
>if isinstance(extra_struct, (QuantizationRangeInitArgs, BNAdaptationInitArgs)):
>if extra_struct.__class__ in EXTRA_STRUCTS_WITH_DATALOADERS:
>    extra_struct: HasDataloader

I forgot about the tuple version of `isinstance`, and I will use this now, thanks, but it seems that PyCharm at least still> cannot parse the type hint with the protocol type and deduce that the matching object has a dataloader-providing method:
Screenshot from 2023-10-27 14-05-57

nncf/torch/dynamic_graph/io_handling.py
andrey-churkin: Could you please check if PyCharm still shows an error for this exact code?
```python
if isinstance(extra_struct, (QuantizationRangeInitArgs, BNAdaptationInitArgs)):
    ...
```
Visual Studio Code doesn't show any warnings for the provided code. Nevertheless, I don't think adding a class to resolve IDE issues is a good idea.

vshampor: PyCharm doesn't show warnings for your variant too, but I need a list of compatible types as a "constant" in one place. It's not that pycharm has an issue, it's more likely that the static type deduction system in Python does not allow to handle this case reliably.

andrey-churkin (approved):
LGTM

alexsu52 (approved):
LGTM

Issue #2195, opened 2023-10-13
Concat scales not being grouped


basioli-k:
I am trying to quantize a pytorch model using NNCF.
The output of my model is a concatenation of two tensors.

To quantize my outputs I set:
`advanced_parameters = AdvancedQuantizationParameters(quantize_outputs=True)`

When I quantize the model I get a separate quantizer for each input:
```
ModuleDict(
  (/nncf_model_input_0|OUTPUT): AsymmetricQuantizer(bit=8, ch=False)
  (/nncf_model_input_1|OUTPUT): AsymmetricQuantizer(bit=8, ch=False)
)
``` 

Based on what I saw in NNCF I would expect to get something like this.
```
ModuleDict(
  (/nncf_model_input_0|OUTPUT;/nncf_model_input_1|OUTPUT): AsymmetricQuantizer(bit=8, ch=False)
)
```

I am guessing it's an edge case which comes up due to `AdvancedQuantizationParameters`.

**NNCF version:** 2.6.0

Run the following to reproduce:

```[python]
import nncf
import torch
import numpy as np
from nncf.quantization.advanced_parameters import AdvancedQuantizationParameters

class DummyDataset(torch.utils.data.Dataset):
    """ Loads images from a folder """
    def __init__(self, input_shapes, input_names):
        self.input_shapes = input_shapes
        self.input_names = input_names

    def __len__(self):
        return 1

    def __getitem__(self, index):
        return { self.input_names[i]: np.random.rand(*input_shape) for i, input_shape in enumerate(self.input_shapes) }

class DummyModel(torch.nn.Module):

    def __init__(self):
        super().__init__()
        self.conv = torch.nn.Conv2d(6, 6, 3, 1, 1)
        
    def forward(self, x, y):
        x_cat_y = torch.cat((x,y), dim=1)
        
        return x_cat_y
        # return self.conv(x_cat_y) # use this to verify that quantizers get grouped if concat isn't the output
    
def quantize_model(model, input_shapes, input_names):
    def transform_fn(data):
        data_dict = data
        return tuple(data_dict[key][0].to(torch.float32) for key in data_dict)
    
    dummy_dl = torch.utils.data.DataLoader(DummyDataset(input_shapes, input_names))
    calibration_dataset = nncf.Dataset(dummy_dl, transform_fn)
    
    advanced_parameters = AdvancedQuantizationParameters(quantize_outputs=True)
    
[...]

PR #1778, opened 2023-05-05
[PTQ] Update scales unification rules based on types

KodiaqQ:
### Changes

- Updated logic for scales unification based on the types

### Reason for changes

- Correctness of the unification functionality

### Related tickets

- 108258

### Tests

- Conformance run - 25

PR #2199, opened 2023-10-13
Reenable scale unification

vshampor:
### Changes
Restored the original, pre-#1778 logic of scale unification and added missing tests for the logic. Added `BatchNorm` as a quantizable operation (asymmetric, per-channel) to the CPU HW config to handle cases like densenet where batch norm is the first operation in a branch.

### Reason for changes
Scales are currently not correctly unified in cases such as #2195.

### Related tickets
N/A

### Tests
tests/common/quantization/test_quantizer_setup.py
tests/**/quantization/test_unified_scales.py

Fixes: #2195


vshampor:
I put this out of the draft state to trigger the CI runs first.

vshampor:
ONNX E2E 489 shows accuracy improvement for densenet12 (+0.77% acc.) and resnet50-v2-7 (+0.1% acc) and no difference for other models.

KodiaqQ (commented):
General questions:

1. How does scale unification work in this PR? Based on the code, I assume that all possible layers unify now. Is that correct?
2. Why were most Tensorflow models updated in tests? What about other backends?
3. Were any conformance tests run on the final version? Are there any degradations?

vshampor:
@KodiaqQ 

> General questions:
> 
> 1. How does scale unification work in this PR? Based on the code, I assume that all possible layers unify now. Is that correct?
> 2. Why were most Tensorflow models updated in tests? What about other backends?
> 3. Were any conformance tests run on the final version? Are there any degradations?

1. See the tests. The activation scales are unified based on the marker in HW config, and does not take into account metatypes.
2. I think your original PR with the changes in the same place also lead to TF model updates, so there should be no surprise here. 
3. See comment above ("PTQ build 240").

KodiaqQ:
> 1. See the tests. The activation scales are unified based on the marker in HW config, and does not take into account metatypes.
> 2. I think your original PR with the changes in the same place also lead to TF model updates, so there should be no surprise here.
> 3. See comment above ("PTQ build 240").

Thanks! Let's run the validation again (with the latest changes in NNCF & OV) and verify the results. Then merge.

vshampor:
@KodiaqQ PTQ build 285 shows a single regression vs build 286 on timm/visformer_small. Overall metric is still within 99% FP32.

KodiaqQ (approved):

KodiaqQ:
> @KodiaqQ PTQ build 285 shows a single regression vs build 286 on timm/visformer_small. Overall metric is still within 99% FP32.

If the PTQ build is red, then we need to update the reference for the **timm/visformer_small** model as well. We can do it in this PR or follow-up, doesn't matter. But we should do it to keep builds green.
Also, is there any observation of why this model shows lower accuracy than the OV and ONNX versions?

PR #2204, opened 2023-10-17
[PTQ][OV] SQ for Convolutions

nikita-malininn:
### Changes

- Updated SmoothQuant algorithm to work with Convolution layers;

### Reason for changes

- Better accuracy results in some cases;

### Related tickets

- 113591

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -434,7 +434,8 @@ class NNCFGraph:
         :param layer_name: The name of the framework-specific "layer" object that houses the operation represented by
             the node and associated trainable weights, if any.
         :param ignored_algorithms: A list of compression algorithm names (from the same set of strings that are
-            specified in the `"algorithm": ...` section of the .json NNCF config) which should ignore this operation.
+            specified in the `"algorithm": ...` section of the .json NNCF config or `ptq_quantization`)
+            which should ignore this operation.
         :param is_in_iteration_scope: Whether the node to be currently added corresponds to an iteration of an RNN
             cycle (where the number of iterations is determined dynamically based on the RNN input shape).
         :param is_integer_input: Only valid for input nodes - whether the input node corresponds to an integer input.
--- a/nncf/openvino/graph/nncf_graph_builder.py
+++ b/nncf/openvino/graph/nncf_graph_builder.py
@@ -142,7 +142,7 @@ class GraphConverter:
         """
         ignored_algorithms = []
         if "nncf_smooth_quant" in node.get_friendly_name():
-            ignored_algorithms.append("quantization")
+            ignored_algorithms.append("ptq_quantization")
         return ignored_algorithms
 
     @staticmethod
--- a/nncf/quantization/algorithms/min_max/algorithm.py
+++ b/nncf/quantization/algorithms/min_max/algorithm.py
@@ -378,7 +378,7 @@ class MinMaxQuantization(Algorithm):
         """
         nncf_node_names = []
         for nncf_node in inference_nncf_graph.get_all_nodes():
-            if "quantization" in nncf_node.ignored_algorithms:
+            if "ptq_quantization" in nncf_node.ignored_algorithms:
                 nncf_node_names.append(nncf_node.node_name)
         return IgnoredScope(names=nncf_node_names)
 


nncf/openvino/graph/metatypes/openvino_metatypes.py
```diff
@@ -191,10 +191,21 @@ class OVSubtractMetatype(OVOpMetatype):
     hw_config_names = [HWConfigOpName.SUBTRACT]
 
 
+@OV_OPERATOR_METATYPES.register()
+class OVSQMultiplyMetatype(OVOpMetatype):
```
alexsu52, line 195: This is not good idea to introduce new metatype for a specific algorithm. I hope you can implement this check "if "nncf_smooth_quant" in node.get_friendly_name():" into algorithm.

nncf/quantization/advanced_parameters.py
```diff
@@ -154,7 +169,6 @@ class AdvancedQuantizationParameters:
     inplace_statistics: bool = True
     disable_channel_alignment: bool = True
     disable_bias_correction: bool = False
-    smooth_quant_alpha: float = 0.95
```
alexsu52, line 157: This is not backward compatible changes. I suggest to depreciate  smooth_quant_alpha in this release and remove in the next release.

nncf/quantization/advanced_parameters.py
```diff
@@ -111,6 +111,25 @@ class AdvancedBiasCorrectionParameters:
...
+    this parameter may be adjusted in the range from 0 to 1 or set -1 to disable SmoothQuant algorithm.
+
+    :param convolution: Whether to apply smoothing for Convolution layers.
+    :type convolution: float
+    :param matmul: Whether to apply smoothing for MatMul layers.
+    :type matmul: float
+    """
+
+    convolution: float = -1
+    matmul: float = 0.95
```
alexsu52, line 130:
```suggestion
    convolution_alpha: float = -1
    matmul_alpha: float = 0.95
```

alexsu52 (changes requested):

nncf/openvino/graph/metatypes/openvino_metatypes.py
KodiaqQ: The original problem looks like this: after the SmoothQuant inserts scales, we need to avoid these scale quantizations.
I've solved this problem by introducing a new type that handles as agnostic for insertion but blocks FakeQuantize propagation.
I wouldn't implement checking `nncf_smooth_quant` in the MinMax algorithm since algorithms are independent.
Do you have any other ideas to solve the problem?

nncf/quantization/advanced_parameters.py
KodiaqQ: Do you have any ideas on how to keep new and old arguments together since they are called almost the same?

nncf/quantization/advanced_parameters.py
KodiaqQ: Wouldn't be better to rename `smooth_quant_params` into `smooth_quant_alphas` and keep these arguments as is?

nncf/openvino/graph/metatypes/openvino_metatypes.py
alexsu52: You already introduced implicit dependent via nncf graph. You check `nncf_smooth_quant` name during building nncf graph. I would recommend explicit check this condition into the algorithm and add all such node into ignored scope for the propagation solver algorithm.

nncf/quantization/advanced_parameters.py
alexsu52: I hope, the deprecation warning and mapping smooth_quant_alph to AdvancedSmoothQuantParameters. alpha will be enough.

nncf/quantization/advanced_parameters.py
alexsu52: Sounds good.

nncf/openvino/graph/metatypes/openvino_metatypes.py
KodiaqQ: It's not implicitly dependent, since NNCFGraph is another entity.
Checking `nncf_smooth_quant` will break the statement that algorithms don't know about each other and can exist independently, this is not a solution.

nncf/quantization/advanced_parameters.py
KodiaqQ: Done

nncf/quantization/advanced_parameters.py
KodiaqQ: Added warnings.

nncf/openvino/pot/quantization/quantize_model.py
```diff
@@ -339,10 +340,19 @@ def quantize_impl(
     if advanced_parameters is None:
         advanced_parameters = AdvancedQuantizationParameters()
 
-    if model_type == ModelType.TRANSFORMER and advanced_parameters.smooth_quant_alpha > 0:
+    if advanced_parameters.smooth_quant_alpha is not None:
+        warning_deprecated(
+            "`AdvancedQuantizationParameters(smooth_quant_alpha=..)` is deprecated."
+            "Please, use `AdvancedQuantizationParameters(smooth_quant_alphas)` option "
+            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..)) as value instead."
```
l-bat, line 347:
```suggestion
            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..) as value instead."
```

nncf/openvino/pot/quantization/quantize_model.py
```diff
@@ -443,10 +453,19 @@ def quantize_with_accuracy_control_impl(
     if advanced_quantization_parameters is None:
         advanced_quantization_parameters = AdvancedQuantizationParameters()
 
-    if model_type == ModelType.TRANSFORMER and advanced_quantization_parameters.smooth_quant_alpha > 0:
+    if advanced_quantization_parameters.smooth_quant_alpha is not None:
+        warning_deprecated(
+            "`AdvancedQuantizationParameters(smooth_quant_alpha=..)` is deprecated."
+            "Please, use `AdvancedQuantizationParameters(smooth_quant_alphas)` option "
+            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..)) as value instead."
```
l-bat, line 460:
```suggestion
            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..) as value instead."
```

nncf/openvino/pot/quantization/quantize_model.py
```diff
@@ -443,10 +453,19 @@ def quantize_with_accuracy_control_impl(
...
+        )
+
+    sq_params = advanced_quantization_parameters.smooth_quant_alphas
+
+    if model_type == ModelType.TRANSFORMER and (sq_params.convolution > 0 or sq_params.matmul > 0):
         nncf_logger.warning(
-            'IMPORTANT. The advanced parameter "smooth_quant_alpha > 0" IS NOT SUPPORTED for the POT backend!'
-            'Please, use "smooth_quant_alpha = -1".'
+            'IMPORTANT. The AdvancedSmoothQuantParameters parameter value > 0" IS NOT SUPPORTED for the POT backend!'
```
l-bat, line 467:
```suggestion
            'IMPORTANT. The AdvancedSmoothQuantParameters parameter value > 0 IS NOT SUPPORTED for the POT backend!'
```

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
```diff
@@ -79,7 +80,12 @@ def _get_minmax_quantization_param_grid() -> ParamGrid:
 
 
 def _get_smooth_quant_param_grid() -> ParamGrid:
-    return {"advanced_parameters:smooth_quant_alpha": [0.15, 0.25, 0.5, 0.75, 0.95]}
+    alpha_values = [0.15, 0.25, 0.5, 0.75, 0.95]
```
l-bat, line 83: Are you planning to add alpha for convolutions later?

nncf/quantization/algorithms/post_training/pipeline.py
```diff
@@ -80,10 +80,23 @@ def create_ptq_pipeline(
...
-    if model_type == ModelType.TRANSFORMER and advanced_parameters.smooth_quant_alpha >= 0:
-        pipeline_steps.append(
-            [SmoothQuant(subset_size, advanced_parameters.inplace_statistics, advanced_parameters.smooth_quant_alpha)]
+    sq_params = advanced_parameters.smooth_quant_alphas
+    sq_alpha = advanced_parameters.smooth_quant_alpha
+    if sq_alpha is not None:
+        warning_deprecated(
+            "`AdvancedQuantizationParameters(smooth_quant_alpha=..)` is deprecated."
+            "Please, use `AdvancedQuantizationParameters(smooth_quant_alphas)` option "
+            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..)) as value instead."
```
l-bat, line 89:
```suggestion
            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..) as value instead."
```

nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
KodiaqQ: Yes. I think that it's better to do some experiments with Convolution's alpha and AAQ as a separate activitiy.

nncf/openvino/graph/metatypes/openvino_metatypes.py
KodiaqQ: Done as discussed.

nncf/openvino/pot/quantization/quantize_model.py
```diff
@@ -339,10 +340,19 @@ def quantize_impl(
...
+            "with AdvancedSmoothQuantParameters(convolution=.., matmul=..) as value instead."
+        )
+
+    sq_params = advanced_parameters.smooth_quant_alphas
+
+    if model_type == ModelType.TRANSFORMER and (sq_params.convolution > 0 or sq_params.matmul > 0):
         nncf_logger.warning(
-            'IMPORTANT. The advanced parameter "smooth_quant_alpha > 0" IS NOT SUPPORTED for the POT backend!'
-            'Please, use "smooth_quant_alpha = -1".'
+            'IMPORTANT. The AdvancedSmoothQuant parameter value > 0" IS NOT SUPPORTED for the POT backend!'
```
l-bat, line 354:
```suggestion
            'IMPORTANT. The AdvancedSmoothQuant parameter value > 0 IS NOT SUPPORTED for the POT backend!'
```

l-bat (approved):

alexsu52 (approved):
LGTM

PR #2205, opened 2023-10-18
Remove pylint

vshampor:
### Changes
Pylint was removed from the make targets and from code. Added `.pylintrc` to `.gitignore` so that developers could have their own `.pylintrc` if they still want to use pylint in informational capacity locally.

### Reason for changes
Following through with the decision to use ruff as the main linter. Most of the pylint rules not covered by ruff will be covered by mypy in the future.

### Related tickets
N/A

### Tests
Existing test scope


.gitignore
```diff
@@ -108,6 +108,9 @@ ENV/
 # snapshots
 *.tar
 
+# pylint config is left at dev's discretion, CI uses ruff/isort/black instead
+.pylintrc
```
AlexanderDokuchaev, line 112: Is it really need? If run locally pylint without disable tags, pylint will show many errors.

CONTRIBUTING.md
```diff
@@ -34,9 +34,12 @@ If your testing code is more extensive than unit tests (in terms of test executi
 
 Changes to NNCF Python code should conform to [Python Style Guide](./docs/styleguide/PyGuide.md)
 
-Pylint is used throughout the project to ensure code cleanliness and quality.
-A Pylint run is also done as part of the pre-commit scope - the pre-commit `pytest` scope will not be run if your code fails the Pylint checks.
-The Pylint rules and exceptions for this repository are described in the standard [.pylintrc](./.pylintrc) format - make sure your local linter uses these.
+Basic code style and static checks are enforced using a `pre-commit` Github action.
```
AlexanderDokuchaev, line 37: Looks like it can be shortly and details describe in PyGuide.md
```
Changes to NNCF Python code should conform to [Python Style Guide](./docs/styleguide/PyGuide.md).
You can run the code style checks locally using the make pre-commit command.
For more details, see [Automating Code Formatting](./docs/styleguide/PyGuide.md#2-automating-code-formatting) and [Ruff](TBD).
```
 
I think all mentions about pylint can be deleted. Developers can use any tools while checks passes by default, don't need write about that.

docs/styleguide/PyGuide.md
```diff
@@ -102,39 +101,19 @@ arr2 = [
 
 ## 3 Python Language Rules
 
-<a id="s3.1-pylint"></a>
-<a id="31-pylint"></a>
-<a id="pylint"></a>
-
-### 3.1 PyLint
```
AlexanderDokuchaev, line 109: Please add description about ruff instead of pylint, with links and useful tags.

AlexanderDokuchaev (changes requested):

.gitignore
vshampor: I have pylint integration in my IDE and it respects my local .pylintrc for highlighting only the stuff I find to be helpful - this is the use case I'm enabling. A global pylint run will be broken indeed.

CONTRIBUTING.md
vshampor: Ok, didn't notice that the section on that is already there in the pyguide

docs/styleguide/PyGuide.md
vshampor: Done

AlexanderDokuchaev (approved):

daniil-lyakhov (approved):

PR #2206, opened 2023-10-18
[Common] `port_id` and `parallel_edges_ids` are added to the graph for analysis

daniil-lyakhov:
### Changes

`port_id` and `parallel_edges_ids` are added to the graph for analysis

image

### Reason for changes

To test `parallel_edges_ids` and `port_id` in graph tests as well as the structural information.


nncf/common/graph/graph.py
```diff
@@ -607,12 +607,12 @@ def get_graph_for_structure_analysis(self, extended: bool = False) -> nx.DiGraph
                 else:
                     attrs_edge["style"] = "solid"
                 label["shape"] = edge[NNCFGraph.ACTIVATION_SHAPE_EDGE_ATTR]
+                label[
+                    "ports"
+                ] = f"{edge[NNCFGraph.OUTPUT_PORT_ID_EDGE_ATTR]}\u2192{edge[NNCFGraph.INPUT_PORT_ID_EDGE_ATTR]}"
```
vshampor, line 612: I understand that you tried to make the representation prettier, but even in 2023 there's no guarantee that every tool can handle Unicode well enough. Use `->` to represent an arrow, that would be safer.

vshampor (changes requested):
So now all of the reference structural graphs except for torch use the format with port IDs and shapes, except for torch. Either regenerate torch reference graphs to the new format, or keep the old format in the other frameworks and only extend the reference graphs with the "parallel_input_port_ids".

daniil-lyakhov:
Closed due to inactivity

PR #2208, opened 2023-10-18
Enable SmoothQuant only for OV backend

andrey-churkin:
### Changes

Run the SmoothQuant and ChannelAlignment algorithms only for the OpenVINO backend.

### Reason for changes

Only the OpenVINO backend supports SmoothQuant and ChannelAlignment algorithms.

### Related tickets
N/A

### Tests
N/A

### DIFF
--- a/nncf/quantization/algorithms/channel_alignment/algorithm.py
+++ b/nncf/quantization/algorithms/channel_alignment/algorithm.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, List, Optional, Tuple, TypeVar
+from typing import List, Optional, Tuple, TypeVar
 
 import numpy as np
 
--- a/nncf/quantization/algorithms/channel_alignment/algorithm.py
+++ b/nncf/quantization/algorithms/channel_alignment/algorithm.py
@@ -28,7 +28,6 @@ from nncf.common.tensor_statistics.statistic_point import StatisticPointsContain
 from nncf.common.utils.backend import BackendType
 from nncf.common.utils.backend import get_backend
 from nncf.quantization.algorithms.algorithm import Algorithm
-from nncf.quantization.algorithms.channel_alignment.backend import ALGO_BACKENDS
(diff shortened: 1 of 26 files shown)


nncf/quantization/algorithms/channel_alignment/algorithm.py
```diff
@@ -97,6 +97,10 @@ def apply(
         statistic_points: Optional[StatisticPointsContainer] = None,
         dataset: Optional[Dataset] = None,
     ) -> TModel:
+        backend = get_backend(model)
+        if backend != BackendType.OPENVINO:
+            return model
```
daniil-lyakhov, line 102: Should we add a warning as it was before?

alexsu52 (changes requested):
Could check it outside of the algorithm? For example you can use available_backends.

alexsu52:
> > Could check it outside of the algorithm? For example you can use available_backends.
> 
> The backend is unknown when the pipeline is created.
> 
> https://github.com/openvinotoolkit/nncf/blob/2878672e66ad2353d831975187a1a44aa35108ba/nncf/quantization/algorithms/post_training/algorithm.py#L69-L71
> 
> One possible place where it can be handled is here.
> 
> https://github.com/openvinotoolkit/nncf/blob/2878672e66ad2353d831975187a1a44aa35108ba/nncf/quantization/algorithms/pipeline.py#L112-L115
> 
> We will skip the step if the backend is not in the available backends list. Please let me know if you are okay with this, or maybe you have another suggestion.

This solution is better.

nncf/quantization/algorithms/channel_alignment/algorithm.py
andrey-churkin: Done

daniil-lyakhov (approved):

alexsu52 (approved):

andrey-churkin:
The `available_backends` property has an incorrect implementation; it does not work.
See the following example
```python
from nncf.quantization.algorithms.channel_alignment.algorithm import ChannelAlignment
ca = ChannelAlignment()
print(ca.available_backends)
# Out: {}
```
It is not possible to obtain available backends until you call the `apply(`) or `get_statistic_points()` methods.

PR #2210, opened 2023-10-19
[Quantization][Torch] Activations are not quantizing / BN operations are not quantizing in PTQ for TRANSFORMER model type

daniil-lyakhov:
### Changes

* Activation metatypies was removed from torch INPUT_QUANTIZABLE metatypes list
* BatchNorm metatype is ignored for TRANSFORMER model type in every backend

### Reason for changes

* From [the documentation](https://docs.openvino.ai/2023.1/openvino_docs_OV_UG_lpt.html):
Only activations that are supporting INT8 execution is Relu and PRelu. So, there is no benefits to quantize any stand-alone activation for OV inference right now
* From the ticket 59548:
Runtime doesn't provide performance benefits by quantizing the stand-alone RELU's
* Batchnorm operations are always decomposed to add + mul operations in OV and fused by the `SHIFT_SCALE` fusing pattern. As add metatype is ignored for the model_type.TRANSFORMER, all batchnorms operations are ignored as well

### Related tickets

110985

### Tests
Manual Conformance test: 189


MaximProshin:
We should check https://docs.openvino.ai/2023.1/openvino_docs_OV_UG_lpt.html for this first.

nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -973,6 +978,8 @@ def get_visualized_graph(self):
             is_integer_path = edge[QuantizerPropagationStateGraph.IS_INTEGER_PATH_EDGE_ATTR]
             if is_integer_path:
                 attrs = {"color": "violet", "style": "bold"}
+            u = u.replace(__RESERVED_DOT_CHARACTER, __CHARACTER_REPLACE_TO)
+            v = v.replace(__RESERVED_DOT_CHARACTER, __CHARACTER_REPLACE_TO)
```
vshampor, line 982: Really, all of this needs to be done at one place - maybe a function that takes an NX graph and then examines all of its attribute nodes or labels and does the replacement.

nncf/torch/quantization/default_quantization.py
```diff
@@ -31,20 +31,11 @@
         operator_metatypes.PTDepthwiseConv3dSubtype,
```
vshampor, line 31: What about the rest of the `default_quantization.py` files across the frameworks?

vshampor (changes requested):

nncf/torch/quantization/default_quantization.py
daniil-lyakhov: Other frameworks are outside of the scope of this PR, the only purpose of this pr is to align OV and Torch backends. I suggest to leave this for future PRs, this PR should be merged till next Monday to make progress for Torch PTQ

nncf/common/quantization/quantizer_propagation/graph.py
daniil-lyakhov: Moved to separate PR https://github.com/openvinotoolkit/nncf/pull/2231

nncf/torch/quantization/default_quantization.py
```diff
@@ -31,20 +31,11 @@
...
         operator_metatypes.PTLinearMetatype,
         operator_metatypes.PTModuleLinearMetatype,
-        operator_metatypes.PTHardTanhMetatype,
```
vshampor, line 34: You should add a comment in-code with a link to the LPT docs and a brief explanation why we decided not to quantize single activations by default.

daniil-lyakhov: Done, please check

KodiaqQ (approved):

AlexanderDokuchaev (approved):

l-bat (approved):

alexsu52 (approved):
LGTM

MaximProshin:
@vshampor , please state explicitly if there are any remaining comments from you that should be applied or you're ok to merge it.

daniil-lyakhov:
Torch 2e2_full build 479 is successfull

vshampor (approved):

PR #2214, opened 2023-10-25
[OV] Add caching of OMZ models

l-bat:
### Changes

Added cache_dir for OMZ models

### Reason for changes

* Сhanges made according to the requirements of the validation team

### DIFF
--- a/tests/openvino/conftest.py
+++ b/tests/openvino/conftest.py
@@ -22,6 +22,7 @@ from tests.shared.paths import TEST_ROOT
 
 def pytest_addoption(parser):
     parser.addoption("--data", type=str, default=None, help="Directory path to cached data.")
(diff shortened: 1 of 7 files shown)


tests/openvino/conftest.py
```diff
@@ -22,6 +22,7 @@
 
 def pytest_addoption(parser):
     parser.addoption("--data", type=str, default=None, help="Directory path to cached data.")
+    parser.addoption("--omz_cache", type=str, default=None, help="Directory path to cached OMZ models.")
```
alexsu52, line 25: Could you reuse existing option --data?

l-bat: ok

alexsu52 (approved):
LGTM, Thanks for the contribution!

PR #2216, opened 2023-10-25
QuantizationPreset.MIXED for ModelType.TRANSFORMER by default

alexsu52:
### Changes

Made the QuantizationPreset.MIXED preset for ModelType.TRANSFORMER by default.

### Reason for changes

Quantization of transformer-based models with QuantizationPreset.MIXED preset shows the best accuracy w/o performance degradation.

### Related tickets

ref: 123235

### Tests
test_create_nncf_config
test_quantization_preset


nncf/experimental/torch/quantization/quantize_model.py
```diff
@@ -87,7 +87,7 @@ def send_to_device(tensor):
 def quantize_impl(
     model: torch.nn.Module,
     calibration_dataset: Dataset,
-    preset: QuantizationPreset,
+    preset: Optional[QuantizationPreset],
```
nikita-savelyevv, line 90: Shouldn't it be like this?

```suggestion
    preset: Optional[QuantizationPreset] = None,
```

Otherwise it would fail if `preset` parameter is not given, despite it being marked as optional.

Also similar cases below.

nikita-savelyevv (changes requested):

nncf/experimental/torch/quantization/quantize_model.py
alexsu52: it does not apply to this function. Optional[QuantizationPreset] is the same Union[None, QuantizationPreset]. I can replace if you have concern about Optional.

nikita-savelyevv (approved):

KodiaqQ (approved):

PR #2230, opened 2023-10-27
Added raising an exception when empty calibration dataset is provided

nikita-savelyevv:
### Changes

Added raising an exception when an empty calibration dataset is provided.

Now if empty calibration dataset is passed to `nncf.quantize()` the following exception will be thrown:
```
Calibration dataset must not be empty. Please provide calibration dataset with at least one sample.
```

Also added a check for non-positive `subset_size` provided to `nncf.quantize()`. Otherwise, it would error with the same statistics not collected error.

### Reason for changes

Without an explicit exception it will error out later with a message like:
```
File "/home/nsavel/workspace/openvino_notebooks/nncf/nncf/quantization/algorithms/min_max/algorithm.py", line 673, in apply
    raise RuntimeError(f"Statistics were not collected for the node {target_node_name}")
RuntimeError: Statistics were not collected for the node /model.2/m.2/Add
```

This is confusing and does not clearly reflect what is actually wrong. There have been some reports, e.g. from OTX side, when an error like this was encountered due to empty dataset. But at first it wasn't clear what's the issue actually is, and a bug in NNCF was suspected.

I personally also encounter this sometimes during experimenting and this triggers me to look for issues in NNCF, however it was just an empty calibration dataset provided by mistake.

### Tests

Added a test for empty dataset to `common/test_statistics_aggregator.py`
Added a test for non-positive `subset_size` to `tests/openvino/native/quantization/test_quantize_api.py` (openvino only).

### DIFF
--- a/nncf/quantization/quantize_model.py
+++ b/nncf/quantization/quantize_model.py
@@ -65,8 +65,8 @@ def quantize(
         into account while compressing in order to obtain the best performance
         for this type of device.
     :type  target_device: nncf.TargetDevice
-    :param subset_size: Size of a subset to calculate activations
-        statistics used for quantization.
+    :param subset_size: Size of a subset to calculate activations statistics used for quantization.
+        Must be positive.
     :param fast_bias_correction: Setting this option to `False` enables a different
         bias correction method which is more accurate, in general, and takes
         more time but requires less memory.
@@ -81,6 +81,10 @@ def quantize(
     :return: The quantized model.
     :rtype: TModel
     """
+
+    if subset_size < 1:
+        raise RuntimeError("Subset size must be positive.")
+
     backend = get_backend(model)
     if backend == BackendType.OPENVINO:
         from nncf.openvino.quantization.quantize_model import quantize_impl
--- a/nncf/common/tensor_statistics/aggregator.py
+++ b/nncf/common/tensor_statistics/aggregator.py
@@ -59,6 +59,7 @@ class StatisticsAggregator(ABC):
             if self.stat_subset_size is not None
             else None
         )
+        empty_statistics = True
         for input_data in track(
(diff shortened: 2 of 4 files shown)


daniil-lyakhov (approved):

MaximProshin:
@KodiaqQ , also take a look and merge it if it's ok

nncf/quantization/quantize_model.py
```diff
@@ -81,6 +81,10 @@ def quantize(
     :return: The quantized model.
     :rtype: TModel
     """
+
+    if subset_size < 1:
```
AlexanderDokuchaev, line 85: @alexsu52 is it correct place to add checks of parameters?

Can it be done all checks here? 
```
if not self.dataset.get_length():
    raise RuntimeError("Calibration dataset must not be empty.")
```

AlexanderDokuchaev (changes requested):

nncf/quantization/quantize_model.py
nikita-savelyevv: `get_length()` returns None for datasets that don't have `__len__` implemented, so we can't check it here for sure

AlexanderDokuchaev: Sad 😞

alexsu52: This is correct place to check value of subset_size parameter.

nncf/quantization/quantize_model.py
```diff
@@ -81,6 +81,10 @@ def quantize(
...
+        raise RuntimeError("Subset size must be positive.")
```
alexsu52, line 86:
```suggestion
        raise ValueError("Subset size must be positive.")
```

AlexanderDokuchaev (approved):

KodiaqQ (approved):

PR #2237, opened 2023-10-31
Updated readme for 4-bit weight compression

ljaljushkin:


docs/compression_algorithms/CompressWeights.md
```diff
@@ -24,26 +23,201 @@ from nncf import compress_weights
...
 from nncf import compress_weights
 from nncf import CompressWeightsMode
-compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4)
+compressed_model = compress_weights(model, mode=CompressWeightsMode.INT4_SYM)
 ```
 
-- Compress weights of 90% of layers to nf4 with the group size 64, and the rest of layers to 8-bit integer data type.
+- Generally, `INT4_SYM` mode is the fastest mixed-precision mode, but it may lead to a significant accuracy degradation or perplexity increase.
+  Compressing weights asymmetrically (`INT4_ASYM` mode) is the way to increase accuracy, however in turns it slows down inference a bit.
+  If the accuracy or perplexity is still not satisfying, there are 2 more hyper-parameters to tune: `group_size` and `ratio`.
```
MaximProshin, line 36: As the user I'd like to know which option should be tried first. Is it group_size? Do we have any recommendation here?

docs/compression_algorithms/CompressWeights.md
```diff
@@ -24,26 +23,201 @@ from nncf import compress_weights
...
 ```python
 from nncf import compress_weights
 from nncf import CompressWeightsMode
-compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4, group_size=64, ratio=0.9)
+compressed_model = compress_weights(model, mode=CompressWeightsMode.INT4_ASYM, group_size=64, ratio=0.9)
 ```
 
-##### Limitations
+- `NF4` mode can be considered for improving accuracy, but currently models quantized to nf4 should not be faster models
+  quantized to 8-bit integer on client processors. Some speed up is expected on server processors in the future releases.
```
MaximProshin, line 48: I think it's confusing. Some speedup for clients can also be expected "in the future". I would simply say "but currently models quantized to nf4 should not be faster models
  quantized to 8-bit integer."

docs/compression_algorithms/CompressWeights.md
```diff
@@ -24,26 +23,201 @@ from nncf import compress_weights
...
+  Here's the example how to compress weights to nf4 data type with group size = 128, except first embedding and last linear layers - they are compressed to 8-bit integer data type.
```
MaximProshin, line 49: "except first embedding and last linear layers - they are compressed to 8-bit integer data type" - it's already explained above for all 4-bit types. Let's remove it.

docs/compression_algorithms/CompressWeights.md
```diff
@@ -24,26 +23,201 @@ from nncf import compress_weights
...
+  Different `group_size` and `ratio` are also supported.
+
+```python
+from nncf import compress_weights
+from nncf import CompressWeightsMode
+compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4)
+```
+
+#### Evaluation results
```
MaximProshin, line 58: can you also extend the table with chatglm2 results?

ljaljushkin: This table is for `lambada_openai` task, but chatglm2 is evaluated on different tasks with hieroglyphs.
Potentially, we can create a separate table with CEval task, but for consistency we would need to evaluate int8 and nf4 on it, but it would take a whole day on server. Should we really do that?

docs/compression_algorithms/CompressWeights.md
ljaljushkin: corrected

docs/compression_algorithms/CompressWeights.md
```diff
@@ -24,26 +23,200 @@ from nncf import compress_weights
...
+  <tr>
+    <td class="tg-0pky">meta-llama/Llama-2-13b-chat-hf</td>
+    <td class="tg-0pky">int8</td>
+    <td class="tg-0pky">2.91</td>
+    <td class="tg-0pky">0</td>
+    <td class="tg-0pky">12.1</td>
+  </tr>
+  <tr>
+    <td class="tg-0pky">meta-llama/Llama-2-13b-chat-hf</td>
+    <td class="tg-0pky">int4_sym_g64_r80</td>
```
ljaljushkin, line 201: @nikita-savelyevv corrected

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -334,7 +334,7 @@ def _get_bitwidth_distribution_str(all_weight_params: List[WeightNodeParams]) ->
         num_bits_vs_num_weights_map[num_bits] = (n_total, n_internal)
 
     # Table creation
-    header = ["Num bits (N)", "% all weight", "% internal weights"]
+    header = ["Num bits (N)", "% all parameters (layers)", "% internal parameters (layers)"]
```
ljaljushkin, line 337: @andreyanufr @AlexKoff88 corrected

docs/compression_algorithms/CompressWeights.md
MaximProshin: I would still add it as there is a high interest to this model. It's not urgent and can be a separate PR to not block this one.

docs/compression_algorithms/CompressWeights.md
MaximProshin: I don't see how it was corrected. The following part is confusing in my view and could be removed:
"on client processors. Some speed up is expected on server processors in the future releases."

MaximProshin:
@ljaljushkin , can you also add a link to https://docs.openvino.ai/nightly/gen_ai_guide.html at the end?

docs/compression_algorithms/CompressWeights.md
ljaljushkin: @MaximProshin probably you are watching at the outdated version

ljaljushkin:
> @ljaljushkin , can you also add a link to https://docs.openvino.ai/nightly/gen_ai_guide.html at the end?

added

docs/compression_algorithms/CompressWeights.md
ljaljushkin: will do it in a separate PR

MaximProshin (approved):

andreyanufr (approved):

AlexKoff88 (approved):

PR #2238, opened 2023-10-31
[TORCH] Increase acceptable range for mobilenet_v2 models in compression training tests

daniil-lyakhov:
### Changes

Acceptable range for mobilenet_v2 models in compression training tests is increased

### Reason for changes

After migration to CUDA-11.8 in NNCF Torch, some torch training_compression tests started to fail due to the fact that difference between validation metrics calculated after training and metrics calculated after --resume is exceeded expected boundaries (0.2 in most cases). Different cases with mobilenet_v2 model were failed due to this problem.

### Related tickets

124083


daniil-lyakhov:
Compression training build 120

KodiaqQ (approved):

daniil-lyakhov:
Compression training build 120 is successful

vshampor (approved):

PR #2240, opened 2023-11-02
[Torch] Workaroud for torch.quantile function

daniil-lyakhov:
### Changes

torch quantile is always using numpy realization

### Reason for changes

Bug https://github.com/pytorch/pytorch/issues/64947

third party test 237 has failed

```
       RuntimeError: quantile() input tensor is too large
 ```
### Tests
Build test third party 239 is finished successfully

### DIFF
--- a/nncf/torch/tensor_statistics/collectors.py
+++ b/nncf/torch/tensor_statistics/collectors.py
@@ -81,7 +81,7 @@ class PTNNCFCollectorTensorProcessor(NNCFCollectorTensorProcessor):
             device = x.tensor.device
             result = torch.tensor(np.median(x.tensor.detach().cpu().numpy(), axis=axis, keepdims=keepdims))
             return PTNNCFTensor(result.type(x.tensor.dtype).to(device))
-        return PTNNCFTensor(torch.quantile(x.tensor, q=0.5, dim=axis, keepdim=keepdims))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: nncf/torch/tensor_statistics/collectors.py [hunk 1]
```diff
@@ -173,17 +173,10 @@ def quantile(
...
-            result = torch.quantile(
-                tensor.tensor,
-                torch.tensor(quantile, dtype=tensor.tensor.dtype, device=tensor.tensor.device),
-                axis,
-                keepdims,
-            )
+        # https://github.com/pytorch/pytorch/issues/64947
+        result = torch.tensor(
+            np.quantile(tensor.tensor.detach().cpu().numpy(), q=quantile, axis=axis, keepdims=keepdims)
+        )
```
vshampor, line 179: If the shortcomings of the torch implementation are limited to the CPU version, then the code should only fall back to the numpy implementation if the original tensor device is CPU.

daniil-lyakhov: I'm now using `torch.quantize` in case it is possible

daniil-lyakhov:
@vshampor, please merge if you don't have any comments

limaodaxia:
> @vshampor, please merge if you don't have any comments

good job

PR #2241, opened 2023-11-02
Update networkx and remove pyparsing limitation

vshampor:
### Changes
Networkx was updated to allow 3.1, pyparsing limitation was removed. Will now replace the disallowed colon symbols `:` during reads and writes of .dot graphs.

### Reason for changes
OV is now at the networkx 3.1, and we should be aligned at least on the major version for better DX.

### Related tickets
69520

### Tests
Existing graph-checking tests


nncf/common/utils/dot_file_rw.py
```diff
@@ -36,4 +38,48 @@ def get_graph_without_data(G: nx.DiGraph) -> nx.DiGraph:
...
+        hits[dot_name] += 1
+        if hits[dot_name] > 1:
+            dot_name = f"{dot_name}_{hits}"
+        if original_name != dot_name:
+            mapping[original_name] = dot_name
+
+    relabeled_graph = nx.relabel_nodes(nx_graph, mapping)
+    for _, node_data in relabeled_graph.nodes(data=True):
+        if "label" in node_data:
+            node_data["label"] = node_data["label"].replace(__CHARACTER_REPLACE_FROM, __CHARACTER_REPLACE_TO)
```
daniil-lyakhov, line 80: All reference dot graphs have original node names in label, this could potentially broke a name for a name with `^` simbol

nncf/common/utils/dot_file_rw.py
```diff
@@ -36,4 +38,48 @@ def get_graph_without_data(G: nx.DiGraph) -> nx.DiGraph:
...
+
+    for _, _, edge_data in relabeled_graph.edges(data=True):
+        if "label" in edge_data:
+            edge_data["label"] = edge_data["label"].replace(__CHARACTER_REPLACE_FROM, __CHARACTER_REPLACE_TO)
```
daniil-lyakhov, line 84: This could potentially break edges label in case '^' symbol was inside

vshampor: Indeed, but this only means that the `^` symbol will now be reserved in place of the `:` in the in-memory graphs, and I don't know of a case in our graph labels where we use this symbol now.

vshampor: Maybe I should rework this, though.

vshampor: I have reworked this to keep your original intent and to remove the cosmetic labels on loading, please re-check.

nncf/common/utils/dot_file_rw.py
```diff
@@ -36,4 +40,95 @@ def get_graph_without_data(G: nx.DiGraph) -> nx.DiGraph:
...
 
 def read_dot_graph(path: pathlib.Path) -> nx.DiGraph:
-    return nx.nx_pydot.read_dot(str(path))
+    loaded = nx.DiGraph(nx.nx_pydot.read_dot(str(path)))
+    return relabel_graph_for_dot_visualization(loaded, from_reference=True)
+
+
+class _Direction(IntEnum):
+    TO_DOT = 0
+    FROM_DOT = 0
```
daniil-lyakhov, line 49: Is that ok?

daniil-lyakhov (approved):

PR #2245, opened 2023-11-03
Set codecov target, flags and reach graph visualization

vshampor:
### Changes

Adjusted codecov.yml for a project coverage target of 90%, set up paths for per-component flags (i.e. TORCH, TENSORFLOW etc.) so that the per-component test coverage could be calculated correctly. Made codecov post their trademark "Reach" graph for visual cues to coverage differences. Codecov will now wait for all CI checks to pass before posting the first version of the comment.

### Reason for changes
Should preserve the good coverage we have right now - the target checks are informational at the moment, may want to make them blocking in the future. Should evaluate the reach graph to see whether it is helpful in estimating the coverage impact of a given PR.

### Related tickets
N/A

### Tests
Existing precommit tests


vshampor:
I'm not sure whether the graph is helpful myself. Codecov docs seem to have very little mentions about it at all, maybe this feature of theirs is on its way out, but I presume that the rectangles represent individual files, with sizes corresponding to line count in each file, and the color reflects either the increase or decrease of coverage. I propose to live with the graph for a while and see if it is indeed helpful in some cases. If not, it's easy to disable it via `codecov.yml`.


`COMMON` flag coverage percentage increased because in these lines:

https://github.com/openvinotoolkit/nncf/pull/2245/files#diff-3e878d9bdc47f29a0ab909a7db939acf08d2d2f1aaba81e6f897b32361fa40dbR52-R55

I changed the base from which the coverage is calculated for the common flag. The base before the change for the `COMMON` flag seems to have been "all lines in code under `nncf` directory", and now the base is "all lines in code either under `nncf/common` or `nncf/quantization`". The new base has lower line count (denominator in the percentage calculation), with the actual lines covered by the common precommit remaining the same (numerator), so the percentage has risen as the result.

AlexanderDokuchaev:
What's the idea of splitting by components? 
If component means backend, that is not clear what means `common_ptq` that contains backend specific code. 
`codecov.yaml` supports regexp and it looks like this may correspond to backend discussion in thehttps://github.com/openvinotoolkit/nncf/pull/1799.

What differ between `Flag` and `Component` tables in  `Additional details`?

How the [graph](https://user-images.githubusercontent.com/32988081/283198800-883525c2-d3ae-4ac5-b63e-23e4d0ff708c.png) can help, links of the graph does not contain extra information, it looks like useless information in the comment.

https://app.codecov.io/gh/openvinotoolkit/nncf/pull/2245/blob/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py
image
Why blank line is marked as covered, but definition of  `EWParamsStateNames` is marked as uncovered? What is it mean?

vshampor:
> What's the idea of splitting by components? If component means backend, that is not clear what means `common_ptq` that contains backend specific code. `codecov.yaml` supports regexp and it looks like this may correspond to backend discussion in the #1799.
> 
> What differ between `Flag` and `Component` tables in `Additional details`?
> 
> How the [graph](https://user-images.githubusercontent.com/32988081/283198800-883525c2-d3ae-4ac5-b63e-23e4d0ff708c.png) can help, links of the graph does not contain extra information, it looks like useless information in the comment.
> 
> https://app.codecov.io/gh/openvinotoolkit/nncf/pull/2245/blob/nncf/experimental/torch/nas/bootstrapNAS/elasticity/elastic_width.py image Why blank line is marked as covered, but definition of `EWParamsStateNames` is marked as uncovered? What is it mean?

I have explained the idea and the difference in the component at length in the `codecov.yml` file in this PR.

We'll live with graph for a while and see if it is useful.

Codecov is not ideal - it may report measurement-error-level coverage difference of 0.07% in case of e.g. doc-only PRs (which should obviously be disregarded) or confusing per-line coverage in the file explorer as you have discovered. I suspect that the reasons would be that Codecov builds a per-line coverage display based on the coverage measured by the precommit runs that are in internal CI merged with develop automatically, while the Codecov view shows the file state in the PR without the merge - not much we can do about it.

KodiaqQ (approved):

codecov.yml
```diff
@@ -23,15 +23,64 @@ coverage:
...
+    - component_id: onnx
+      name: onnx
+      paths:
+        - nncf/onnx
+    - component_id: openvino
+      name: openvino
+      paths:
+        - nncf/openvino
+    - component_id: quantization
+      name: common_ptq
```
AlexanderDokuchaev, line 84: The current division by components is not correct.  Now `common_ptq`  collects files for all backends.
Please split all files to components by backends (conditions was discussed in https://github.com/openvinotoolkit/nncf/pull/1799)

codecov.yml
```diff
@@ -23,15 +23,64 @@ coverage:
...
         informational: true
         only_pulls: true
         paths:
-          - "nncf/onnx"
-          - "nncf/common"  # extend this once we collect coverage reports for more than just onnx and common part of precommit
+          - "nncf"
 
 comment:
-  layout: "diff, flags, files"
+  layout: "reach, diff, files, flags, components"
```
AlexanderDokuchaev, line 33: Now the comment is overloaded with useless information and can be fully views only on vertical monitor.

image

Looks like only `diff, files` are usefully. May be `components` can be interesting too.

AlexanderDokuchaev (changes requested):

codecov.yml
vshampor: "Useless" is subjective, and you have not yet had a chance to determine which of the new sections are actually useful or not. Memorize your objections for when the time comes to disable the sections that we will consider "useless" based on experience, perhaps via voting.

AlexanderDokuchaev: If you think that it can be useful, provide scenarios how it can help in development.

codecov.yml
vshampor: If we are talking about the Codecov term for a `"component"`, then it is absolutely arbitrary what subset of files we choose to be a component. I propose the folder-based components in the form they are in now. The component "common_ptq" is now reflecting the coverage for the PTQ component of the NNCF, for all backends. I will rename it to "ptq" since this is not really common, as you pointed out.

I have also just now extended the backend-specific "components" with corresponding files based on a glob/regex. There is no easy way to test whether they matched correctly, though. It doesn't seem to be documented, or possible, to specify `ignore` sections for "component" file paths, which means that the `common` "component" coverage will not reflect PTQ common code coverage, and that the `ptq` "component" coverage will now be calculated based off a set of files that overlaps with each backend-specific "component". 

By the way, all of these troubles, including the effort spent on #1799, follow from the original bad decision to break existing file structure consistency and store backend-specific PTQ code in `*_backend.py` instead of at `nncf/*backend*` like it was done before the POT started to be merged in, and I pointed out this decision as bad from the start (@alexsu52 @KodiaqQ FYI). We will have such issues with every future tool that works on our code base and where we need component-level breakdown of the results.

codecov.yml
vshampor: The component section gives a quick rundown on which parts of NNCF the change impacts the most.

Of the flags section, only the COMMON is important really, and it shows the part of `common` code that is tested by `common` tests in precommit - we should encourage PRs that raise this number up to the target level of 80-90%. Sadly, it is not possible to only show a specific flag in the comment, so it shows everything. This may be circumvented by the `flag_regexes` [option](https://docs.codecov.com/docs/components#component-options), e.g. a `common_tested_by_common_only` component may be configured to only limit itself to the `COMMON` flag and hopefully show the same numbers as the current `COMMON` flag coverage, but this is an improvement for the future.

The diff section is the one that shows the coverage percentage of the whole project. If the comment text itself showed this already, maybe we wouldn't need it.

The files section gives a per-file breakdown of the coverage impact and the links to the corresponding per-line coverage visualization.

I do have doubts about how useful the reach graph really is, but we haven't seen it in action yet on real PRs, so we cannot jump to conclusions just now. Presumably it should help visually identify the PRs that preserve global and component coverage, but significantly reduce coverage in one file and increase the coverage in another. Maybe there also features hidden in this graph that will be helpful that we cannot yet see (since no feature overview is provided by codecov).

codecov.yml
AlexanderDokuchaev: >  It doesn't seem to be documented, or possible, to specify ignore sections for "component" file paths

https://docs.codecov.com/docs/path-regexp

codecov.yml
```diff
@@ -23,15 +23,68 @@ coverage:
...
+      name: torch
+      paths:
+        - nncf/torch
+        - nncf/**/torch_*.py
+    - component_id: tensorflow
+      name: tensorflow
+      paths:
+        - nncf/tensorflow
+        - nncf/**/tensorflow_*.py
+    - component_id: onnx
```
AlexanderDokuchaev, line 77: image
image

is it expected?

codecov.yml
vshampor: Looks like I missed it, thanks. Pushed the version with ignores in the common component.

codecov.yml
vshampor: The corresponding precommit stage did not submit a coverage report, it seems; looks like a spurious CI failure.
https://app.codecov.io/gh/openvinotoolkit/nncf/commit/66ad33290d7796d40261c09b2315111d2d35dbb1

vshampor: Codecov uploader stage runs into a 502 error sporadically, it seems:
```
16:36:12  [2023-11-22T15:36:10.687Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io/: Error: There was an error fetching the storage URL during POST: 502 - 
16:36:12  <html><head>
16:36:12  <meta http-equiv="content-type" content="text/html;charset=utf-8">
16:36:12  <title>502 Server Error</title>
16:36:12  </head>
16:36:12  <body text=#000000 bgcolor=#ffffff>
16:36:12  <h1>Error: Server Error</h1>
16:36:12  <h2>The server encountered a temporary error and could not complete your request.<p>Please try again in 30 seconds.</h2>
16:36:12  <h2></h2>
16:36:12  </body></html>
```

KodiaqQ:
@AlexanderDokuchaev, please, review.

PR #2246, opened 2023-11-03
[Torch][PTQ] Examples are updated for the new PTQ TORCH backend 

daniil-lyakhov:
### Changes

- Do not filter constant nodes for torch backend in the inference graph
- Fix version in requarements.txt for examples of post_training_quantization
    - for ssd300_vgg16 is not available to use torch 2.1.0 (failed on export to onnx Unsupported: ONNX export of operator get_pool_ceil_padding, tracing is not supporting too)
    - Update metrics
- Add to PTEngine convert inputs to model's device to sync behavior with `create_compress_model` 
- Mobilenet_v2 example converting PyTorch model to IR by tracing (without onnx).
- nncf.quantize for PyTorch works with copy of the target model

### Reason for changes

To make PTQ work properly with disconnected graphs (like in  [example](https://github.com/openvinotoolkit/nncf/blob/develop/examples/post_training_quantization/torch/ssd300_vgg16/main.py))

### Related tickets
124417

### Tests

test_examples build 128


alexsu52 (commented):
What is the reason for the disconnectedness of the graph?

daniil-lyakhov:
> What is the reason for the disconnectedness of the graph?

Ticket 124397

AlexanderDokuchaev (approved):

alexsu52 (approved):
LGTM

examples/post_training_quantization/torch/ssd300_vgg16/requirements.txt
```diff
@@ -1,7 +1,7 @@
-fastdownload
+fastdownload==0.0.7
+onnx==1.13.1
 openvino-dev==2023.1
+pycocotools==2.0.7
+torch==2.0.1  # ssd300_vgg16 can not be exported with 2.1.0: Unsupported: ONNX export of operator get_pool_ceil_padding
```
nikita-savelyevv, line 5: Have you checked if this is a known issue in torch? If not, we should file an issue.

nncf/torch/quantization/quantize_model.py
```diff
@@ -108,7 +109,8 @@ def quantize_impl(
     if target_device == TargetDevice.CPU_SPR:
         raise RuntimeError("target_device == CPU_SPR is not supported")
 
-    nncf_network = create_nncf_network(model.eval(), calibration_dataset)
+    copied_model = deepcopy(model)
+    nncf_network = create_nncf_network(copied_model.eval(), calibration_dataset)
```
nikita-savelyevv, line 113: Could you please explain the reason behind this change?

nikita-savelyevv (commented):
Good job improving nncf examples :+1: Minor comments from my side

nncf/torch/quantization/quantize_model.py
daniil-lyakhov: @AlexanderDokuchaev, could you please reply?

examples/post_training_quantization/torch/ssd300_vgg16/requirements.txt
daniil-lyakhov: Agree, I've opened a [Pytorch issue](https://github.com/pytorch/pytorch/issues/113155) and referenced it in the mentioned file

nncf/torch/quantization/quantize_model.py
AlexanderDokuchaev: All implementation of `quantize` function works with copied of the model. 
https://github.com/openvinotoolkit/nncf/blob/release_v250/nncf/torch/quantization/quantize_model.py#L252 previous implementation for torch backend.
For new `quantize_impl` it was missed.

examples/post_training_quantization/torch/ssd300_vgg16/requirements.txt
nikita-savelyevv: Thanks

nikita-savelyevv (approved):

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -540,6 +540,7 @@ def _get_quantization_target_points(
             self._backend_entity.shapeof_metatypes,
             self._backend_entity.dropout_metatypes,
             self._backend_entity.read_variable_metatypes,
+            nncf_graph_contains_constants=backend != BackendType.TORCH,
```
KodiaqQ, line 543: I'd like to note that we introduce backend-specific logic into a common part of the PTQ, which does not fit with our approach. Do you agree with that @alexsu52, @AlexanderDokuchaev, @nikita-savelyevv?

alexsu52: Yes, I agree. @daniil-lyakhov Could you update `filter_constant_nodes` pass for discontinuous graphs?

KodiaqQ (approved):

nncf/quantization/algorithms/min_max/algorithm.py
daniil-lyakhov: Done, please check

daniil-lyakhov: I tried implementation we discussed db0ad8df19da7e4c796e17dbc876c4ea9c9f2e36, but found a corner case that is breaking the pass:
Before:
image
Expected after:
image
Actual after:
image
That's because convolution node should be an input node so survive filtering, but we have no prior information yet that can help us to mark this convolution as an input node

daniil-lyakhov:
Torch precommit is failing due to problems in e00f6b779428404dfcfd803a5dc3bbb9bbb1b1d9

KodiaqQ (approved):

PR #2249, opened 2023-11-08
[PTQ][OV] Update references for the version

KodiaqQ:
### Changes

- Added new files for 2023.2 scale references (only layer names were changed) instead of the symlinks;
- Changed layer names for existing 2023.2 references;

### Reason for changes

- Alignment with the newest OV version


KodiaqQ:
openvino-nightly run - 12

l-bat (approved):

PR #2251, opened 2023-11-08
Introduced nncf.torch.wrap_model(...)

alexsu52:


tests/torch/ptq/test_wrap_model.py
```diff
@@ -0,0 +1,74 @@
...
+# limitations under the License.
+
+import pytest
+import torch
+from torch import nn
+
+from nncf.torch.model_creation import wrap_model
+
+
+class TestArgumentModel(nn.Module):
```
vshampor, line 19: The classes that do not group test cases should not start with `Test`, otherwise pytest will try to execute the `test_*` methods if there are any present in such a class and will also complain that such a class should not contain an `__init__` method.

tests/torch/ptq/test_wrap_model.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class TestKeyWordArgumentsModel(nn.Module):
+    def __init__(self, example_input) -> None:
+        super().__init__()
+        self._example_input = example_input
+
+    def forward(self, *, x, y):
+        assert x == self._example_input["x"]
+        assert y == self._example_input["y"]
```
vshampor, line 47: Execute this under `nncf.torch.dynamic_graph.context.no_nncf_trace` so that the `__eq__` nodes don't end up in the graph.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -229,6 +229,20 @@ def __init__(self, forward_args: Tuple, forward_kwargs: Dict):
     def get_forward_inputs(self, device: str = None) -> Tuple[Tuple, Dict]:
         return self._forward_args, self._forward_kwargs
 
+    @classmethod
+    def from_example_input(cls, example_input: Any) -> "ExactInputsInfo":
```
vshampor, line 233: I really think we shouldn't have an `example_input` entity with a non-trivial logic of turning it into forward args and kwargs. 

In torch's jit.trace function (https://pytorch.org/docs/stable/generated/torch.jit.trace.html) there are now both `example_input` and `example_kwarg_inputs` parameters. Their handling of these parameters is "either/or", which basically maps onto your current logic, so your solution is not much worse than theirs. But I highly suspect that the only reason they had to do this is because they designed this API poorly in the first place (see [complaint](https://discuss.pytorch.org/t/torch-jit-trace-example-inputs-how-to-set-specified-parameters/175005)) and they needed to preserve backward compatibility. I propose not to repeat the torch mistakes and switch the understanding of the `example_inputs` in the Torch case from a `Union[Tuple, Dict]` to an intuitive `(args, kwargs)` tuple, so the `from_example_input` body would look like:
```python3
PTExampleInput = Tuple[Tuple[Any, ...], Dict[str, Any]]

@classmethod
def from_example_input(cls, example_input: PTExampleInput):
    args, kwargs = example_input
    return ExactInputsInfo(args, kwargs)
```
If not, then at least the `PTExampleInput` should be defined as a typedef like above with the currently applicable type, and the corresponding type used everywhere where such an object is expected, e.g. for the current state:
```python3
PTExampleInput = Union[Tuple[Any, ...], Dict[str, Any]]  # use this below, also in .wrap_model signature and in tests
@classmethod
def from_example_input(cls, example_input: PTExampleInput):
    ... # same as your current variant
```

vshampor (changes requested):

tests/torch/ptq/test_wrap_model.py
alexsu52: Thanks for the comment. I removed `Test` from class names.

tests/torch/ptq/test_wrap_model.py
alexsu52: done

nncf/torch/dynamic_graph/io_handling.py
alexsu52: The general idea to be align with openvino https://docs.openvino.ai/2023.1/openvino_docs_OV_Converter_UG_prepare_model_convert_model_Convert_Model_From_PyTorch.html. Could you make proposal to OpenVINO team?

nncf/torch/model_creation.py
```diff
@@ -309,3 +309,23 @@ def create_compression_algorithm_builder_from_algo_names(
     else:
         builder = PTCompositeCompressionAlgorithmBuilder(config, should_init=should_init)
     return builder
+
+
+def wrap_model(model: torch.nn.Module, example_input: Any) -> NNCFNetwork:
+    """
+    Creates NNCFNetwork instance for the PyTorch model where the first item of dataset
```
AlexanderDokuchaev, line 316: Looks like to be correct, description should be like
```
    """Wraps a PyTorch model to the NNCFNetwork class.

    This function dynamically extends the instance of PyTorch model with NNCF-enabling functionality. 

    :param model: PyTorch model.
...
```
And missed dot for param.

alexsu52: done

AlexanderDokuchaev (approved):

nncf/torch/dynamic_graph/io_handling.py
vshampor: If they made it their API already, there is little they can do at this point in time.

vshampor (approved):

PR #2253, opened 2023-11-08
Supports multi-device model inference and wrapped forward functions

alexsu52:
### Changes

Supports multi-device model inference and wrapped forward functions

### Reason for changes

Support tracing "bigscience/bloomz-560m" model from HF

### Related tickets

N/A

### Tests

test_no_self_forward,  test_multidevice_model

### DIFF
--- a/nncf/torch/dynamic_graph/io_handling.py
+++ b/nncf/torch/dynamic_graph/io_handling.py
@@ -13,7 +13,7 @@ from copy import deepcopy
 from functools import partial
 from inspect import Parameter
 from inspect import Signature
-from typing import Any, Dict, List, Optional, Protocol, Set, Tuple, Type
+from typing import Any, Dict, List, Optional, Protocol, Set, Tuple, Type, Union
 
 import torch
 
@@ -87,7 +87,7 @@ class ModelInputInfo(abc.ABC):
     """
 
     @abc.abstractmethod
-    def get_forward_inputs(self, device: str = None) -> Tuple[Tuple, Dict]:
+    def get_forward_inputs(self, device: Optional[Union[str, torch.device]] = None) -> Tuple[Tuple, Dict]:
         """
         Returns the tuple of (args, kwargs) for passing into the compressed model's forward method when necessary.
         The returned arguments should be such that the model's forward with these arguments executes the main
@@ -204,7 +204,9 @@ class FillerInputInfo(ModelInputInfo):
             return FillerInputInfo(elements)
         raise RuntimeError("Invalid input_infos specified in config - should be either dict or list of dicts")
 
-    def get_forward_inputs(self, device: str = None) -> Tuple[Tuple[torch.Tensor, ...], Dict[str, torch.Tensor]]:
+    def get_forward_inputs(
+        self, device: Optional[Union[str, torch.device]] = None
+    ) -> Tuple[Tuple[torch.Tensor, ...], Dict[str, torch.Tensor]]:
         args_list = []
         kwargs = {}
         for fe in self.elements:
@@ -218,7 +220,7 @@ class FillerInputInfo(ModelInputInfo):
         return tuple(args_list), kwargs
 
 
-class ExactInputsInfo(ModelInputInfo):
+class ExampleInputInfo(ModelInputInfo):
     """
     An implementation of ModelInputInfo that defines the model input in terms of exact forward args and kwargs.
     """
@@ -227,7 +229,7 @@ class ExactInputsInfo(ModelInputInfo):
         self._forward_args = forward_args
         self._forward_kwargs = forward_kwargs
 
-    def get_forward_inputs(self, device: str = None) -> Tuple[Tuple, Dict]:
+    def get_forward_inputs(self, device: Optional[Union[str, torch.device]] = None) -> Tuple[Tuple, Dict]:
         if device is None:
             return self._forward_args, self._forward_kwargs
         to_device_fn = partial(torch.Tensor.to, device=device)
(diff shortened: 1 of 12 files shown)


nncf/torch/nncf_network.py
```diff
@@ -144,9 +144,10 @@ def get_original_forward(self) -> Callable:
         Returns the forward function of the original model, unmodified by NNCF. The returned function will
         have its 0-th implicit `self` argument bound to the model object.
         """
-        if self._original_instance_forward is not None:
-            return functools.partial(self._original_instance_forward, self._model_ref)
-        return functools.partial(self._original_unbound_forward, self._model_ref)
+        forward_parameters = self._forward_signature.parameters
+        if forward_parameters and next(iter(forward_parameters)) == "self":
```
vshampor, line 148: This is not a particularly reliable way of establishing whether the `self._forward_signature` is belonged to a bound or an unbound version of the method. `self` is not a reserved name, it's just a convention-level name for the 0-th argument. You can have:
```
import torch

class MyModule(torch.nn.Module):
    def forward(self, x, y):
        return x + y

def evil_forward(self, y):
    return self - y

module = MyModule()
module.forward = evil_forward

print(module.forward(20, 21))  # prints -1
```
Was this change strictly necessary to enable multi-device model inference? Maybe the case such as above should be added to tests?

nncf/torch/nncf_module_replacement.py
```diff
@@ -186,6 +187,9 @@ def replace_modules_by_nncf_modules(
             else:
                 replaced_module = nncf_module_from(module)
 
+            device = get_model_device(module)
+            replaced_module.to(device)
+
             inter_dict[replaced_module] = scope_set
```
vshampor, line 193: Better to get the model device before the replacer function, just in case the `nncf_module_from` does not return a copy of `module` but instead does in-place modifications.

nncf/torch/nncf_network.py
alexsu52: Could you suggest an approach to handle your case. And please provide the model in which this occurs?

nncf/torch/nncf_module_replacement.py
alexsu52: done

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -218,7 +220,7 @@ def get_forward_inputs(self, device: str = None) -> Tuple[Tuple[torch.Tensor, ..
         return tuple(args_list), kwargs
 
 
-class ExactInputsInfo(ModelInputInfo):
+class ExampleInputInfo(ModelInputInfo):
     """
```
vshampor, line 224: You should still keep the `ExactInputsInfo` in the hierarchy and inherit `ExampleInputInfo` and `LoaderInputInfo` from it, since both those classes *are* providers of exact inputs to the model's forward.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -283,7 +287,7 @@ def from_nncf_config_dataloaders(cls, config: NNCFConfig) -> Optional["ExactInpu
                 wrapped_dataloader = wrap_dataloader_for_init(dataloader)
                 dataloader_output = next(iter(wrapped_dataloader))
                 args, kwargs = wrapped_dataloader.get_inputs(dataloader_output)
-                return ExactInputsInfo(args, kwargs)
+                return LoaderInputInfo(args, kwargs)
```
vshampor, line 290: `LoaderInputInfo` has no `__init__` right now. The `__init__` you want will be in the `ExactInputsInfo` class.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -227,7 +229,7 @@ def __init__(self, forward_args: Tuple, forward_kwargs: Dict):
         self._forward_args = forward_args
         self._forward_kwargs = forward_kwargs
```
vshampor, line 230: This is the `__init__` that should be on the `ExactInputsInfo` level of the hierarchy.

vshampor (changes requested):

nncf/torch/dynamic_graph/io_handling.py
alexsu52: done

nncf/torch/dynamic_graph/io_handling.py
alexsu52: done

nncf/torch/dynamic_graph/io_handling.py
alexsu52: done

tests/torch/test_graph_building.py
```diff
@@ -646,7 +647,8 @@ def test_filler_input_info_arg_generation(filler_gen_test_struct: FillerInputInf
     "input_info",
     [
         FillerInputInfo([FillerInputElement([1, 3, 3, 3])]),
-        ExactInputsInfo((torch.Tensor([1]), torch.Tensor([1])), {"a": torch.Tensor([1]), "b": torch.Tensor([1])}),
+        ExampleInputInfo((torch.Tensor([1]), torch.Tensor([1])), {"a": torch.Tensor([1]), "b": torch.Tensor([1])}),
+        LoaderInputInfo((torch.Tensor([1]), torch.Tensor([1])), {"a": torch.Tensor([1]), "b": torch.Tensor([1])}),
     ],
     ids=["filler", "exact"],
```
vshampor, line 653: Add an ID for the new case and the remaining test should be passing.

vshampor (approved):

AlexanderDokuchaev (approved):

PR #2255, opened 2023-11-08
Improve Tensor performance

AlexanderDokuchaev:
### Changes

Add custom custom implementation of functools.singledispatch with custom arguments 'wrapper_type' to select wrapper function.
Remove `np.errstate` from divide function because it degrades performance, to disable warning add `disable_error_handling` function.

Tested on run 1000 times `tune_range` function  by cProfile.
| | calls | time |
| --- | --- | --- |
| develop | 338187 | 0.167 |
| remove np.errstate from __div__ | 262113 |  0.127 | 
| tensor_dispatch | 241006 | 0.118  |


nncf/quantization/fake_quantize.py
```diff
@@ -106,8 +106,9 @@ def tune_range(
         fval = -left_border * s
         qval = fns.round(fval)
 
-    ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
-    rb = fns.where(qval > 0.0, (qval - level_high) / qval * left_border, right_border)
+    with fns.disable_error_handling(qval):
```
alexsu52, line 109: Please, do not enter such context because user usage scenarios are not clear.

alexsu52 (changes requested):

vshampor (changes requested):
One of the reasons why the "manual dispatching" solution was originally chosen over the idiomatic OOP because it was supposed to be "simpler to work with". That analysis was wrong from the start, but now you are adding yet another layer of complexity to gain unknown improvements to NNCF performance on the whole. 

The `tune_range` function cannot be treated as representative of the NNCF performance on the whole, PTQ or QAT, and it cannot be said that a 40% improvement in the runtime of a `tune_range` function will lead to a 40% reduction in the overall NNCF runtime. Providing an own singledispatch implementation is not justifiable if the overall NNCF runtime improment due to this ends up amounting to 0.1%.

You should have first switched all the algos to the new tensor approach, then provided a performance measurement framework that could provide measurements of overall NNCF performance for a given algorithm, and then estimated the improvement due to the change in this PR using the framework. It would be also good if the framework measured the percentage of tensor operation overhead that we are seemingly so keen on optimizing, so that we don't waste time optimizing the overhead if it comprises 1% of the overall algo runtime. Then, if the overall perf improvement justifies this kind of a complexity increase that you are bringing with this PR, it will also be interesting for me to take https://github.com/vshampor/nncf/tree/nncf_tensor_saved with the OOP approach and maybe implement an `abc.ABC` on my own too with better performance over the regular Python built-in, in case we are allowed to optimize Python built-ins in this part of NNCF code. Maybe I will even beat the "manual dispatch" implementation in terms of raw performance and we will switch the code to the approach in https://github.com/vshampor/nncf/tree/nncf_tensor_saved as a result.

nncf/quantization/fake_quantize.py
AlexanderDokuchaev: removed it, tune_range will be reworked in next pr.

PR #2259, opened 2023-11-10
Fix torchvision version in the examples

vshampor:
### Changes
Allow torchvision 0.16 in the examples

### Reason for changes
Otherwise the installation of the requirements for the torch examples tries to install torchvision 0.16, which pulls the torch 2.0.1 which is different from the BKC torch v2.1

### Related tickets
N/A

### Tests
torch_nightly, torch E2E


vshampor:
torch_nightly build 50, torch E2E build 491 TBA

KodiaqQ (approved):

PR #2268, opened 2023-11-15
[OV] Support for ScaledDotProductAttention operation

alexsu52:
### Changes

- Introduced `OVScaledDotProductAttentionMetatype`
- target_input_ports attribute in `OperatorMetatype`
- Implement handling OVScaledDotProductAttentionMetatype into `QuantizerPropagationSolver`

### Reason for changes

Support quantization of ScaledDotProductAttention operation from OpenVINO opset13

### Related tickets

124573

### Tests

test_scaled_dot_product_attention_placement

PR #2277, opened 2023-11-17
OVWeightUpdateCommand copies a constant

andrey-churkin:
### Changes

- Update an implementation of `get_const_value()` method. This method was initially implemented as a workaround in [PR 1654](https://github.com/openvinotoolkit/nncf/pull/1654), but it is no longer needed and can be replaced with `const_op.data`.
- Replace `const_op.get_data()` with `const_op.data` because the `const_op.get_data()` legacy method and `.data` property should be used.
- Set the `shared_memory` parameter to `True` during constant creation to avoid copying the constant.

### Reason for changes

- The `const_od.get_data()` method copies constant.
- Do not copy the constant value during constant creation in `opset.constant(...)`.

### Related tickets

Ref: 122922

### Tests

N/A

### DIFF
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -17,7 +17,6 @@ import numpy as np
 import openvino.runtime as ov
 from openvino._pyopenvino import DescriptorTensor
 from openvino.runtime import opset9 as opset
-from openvino.runtime import opset13
 
 from nncf.common.graph.model_transformer import ModelTransformer
 from nncf.common.graph.model_transformer import TModel
@@ -359,7 +358,8 @@ class OVModelTransformer(ModelTransformer):
         const_dtype = const_node.data.dtype
         const_value = np.reshape(const_value, const_shape).astype(const_dtype)
 
-        new_const_node = opset13.constant(const_value, dtype=const_dtype, shared_memory=True)
+        # TODO(andrey-churkin): Replace on opset13.constant() in a future release
+        new_const_node = ov.op.Constant(const_value, shared_memory=True)
         new_const_node.set_friendly_name(const_node.get_friendly_name())
         const_port.replace_source_output(new_const_node.output(0))
 
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -356,7 +356,7 @@ class OVModelTransformer(ModelTransformer):
 
         const_shape = const_node.data.shape
         const_value = np.reshape(const_value, const_shape)
-        new_const_node = opset.constant(const_value, dtype=const_node.get_element_type())
+        new_const_node = opset.constant(const_value, dtype=const_node.get_element_type(), shared_memory=True)
         new_const_node.set_friendly_name(const_node.get_friendly_name())
         const_port.replace_source_output(new_const_node.output(0))
 
--- a/nncf/openvino/graph/model_transformer.py
+++ b/nncf/openvino/graph/model_transformer.py
@@ -16,6 +16,7 @@ from typing import Callable, Dict, List, Tuple
 import numpy as np
 import openvino.runtime as ov
 from openvino._pyopenvino import DescriptorTensor
+from openvino.runtime import opeset13
 from openvino.runtime import opset9 as opset
 
 from nncf.common.graph.model_transformer import ModelTransformer
@@ -358,7 +359,7 @@ class OVModelTransformer(ModelTransformer):
         const_dtype = const_node.data.dtype
         const_value = np.reshape(const_value, const_shape).astype(const_dtype)
 
-        new_const_node = ov.op.Constant(const_value, shared_memory=True)
+        new_const_node = opeset13.constant(const_value, dtype=const_dtype, shared_memory=True)
         new_const_node.set_friendly_name(const_node.get_friendly_name())
         const_port.replace_source_output(new_const_node.output(0))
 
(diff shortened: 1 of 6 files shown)


alexsu52 (commented):
Could you provide performance results?

andrey-churkin:
> Could you provide performance results?

I have used the `databricks/dolly-v2-3b` model and got the following result.

Quantization Wall Time Before (NNCF develop): 401.24 sec
Quantization Wall Time After:                              279.26 sec

nncf/openvino/graph/model_transformer.py
```diff
@@ -354,9 +354,11 @@ def _set_const_value(node_with_const: ov.Node, const_port_id: int, const_value:
...
             raise RuntimeError("Constant node was expected but could not find it.")
 
-        const_shape = const_node.get_data().shape
-        const_value = np.reshape(const_value, const_shape)
-        new_const_node = opset.constant(const_value, dtype=const_node.get_element_type())
+        const_shape = const_node.data.shape
+        const_dtype = const_node.data.dtype
+        const_value = np.reshape(const_value, const_shape).astype(const_dtype)
+
+        new_const_node = ov.op.Constant(const_value, shared_memory=True)
```
KodiaqQ, line 361: What's wrong with the `opset.constant` method?

nncf/openvino/graph/node_utils.py
```diff
@@ -95,7 +85,7 @@ def get_bias_value(node_with_bias: NNCFNode, nncf_graph: NNCFGraph, model: ov.Mo
     add_node = nncf_graph.get_next_nodes(node_with_bias)[0]
     bias_constant = get_node_with_bias_value(add_node, nncf_graph)
     ov_bias_constant = ops_dict[bias_constant.node_name]
-    return get_const_value(ov_bias_constant)
+    return ov_bias_constant.data
```
KodiaqQ, line 88: Why don't keep the `get_const_value` with the `return const_node.data`?

andrey-churkin: I've just thought that the function only with `return const_op.data` is redundant, and it is simpler to just use `const_op.data`. However, if you have certain concerns about this, I may revert it and keep the `get_const_value()` function.

nncf/openvino/graph/model_transformer.py
andrey-churkin: Unfortunately, there is no `shared_memory` parameter in `opset.constant()`. As far as I understand, this parameter will be added in the future.

nncf/openvino/graph/node_utils.py
KodiaqQ: Personally, I prefer to wrap such things that are used in several places (eg. everywhere) into the method.
But it needs to have opinions from all reviewers to decide. I do not insist on reverting.

KodiaqQ (changes requested):

nncf/openvino/graph/model_transformer.py
andrey-churkin: Yes, but we use opset9 as opset (`from openvino.runtime import opset9 as opset`). Do you suggest using `opset13` only for constants or changing the opset in the import?

nncf/openvino/graph/node_utils.py
andrey-churkin: @alexsu52 Could you please provide your opinion about this?

nncf/openvino/graph/model_transformer.py
KodiaqQ: `opset9` is quite outdated as we can see, so we can update it to `opset13` and fix the problems, IMO.
However, there is already an example of how to live with the different opsets - https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/models.py#L19 (via #2268)

andrey-churkin: Thanks! I've updated the code.

nncf/openvino/graph/node_utils.py
andrey-churkin: I've updated code according to comments.

nncf/openvino/graph/model_transformer.py
```diff
@@ -16,6 +16,7 @@
 import numpy as np
 import openvino.runtime as ov
 from openvino._pyopenvino import DescriptorTensor
+from openvino.runtime import opeset13
```
KodiaqQ, line 19:
```suggestion
from openvino.runtime import opset13
```

KodiaqQ (approved):
Please, fill the ticket (or create PR) with the migrating to the latest `opset`.

andrey-churkin:
> Please, fill the ticket (or create PR) with the migrating to the latest `opset`.

I've filled the ticket. Ref: 125777

nncf/openvino/graph/model_transformer.py
```diff
@@ -354,9 +354,12 @@ def _set_const_value(node_with_const: ov.Node, const_port_id: int, const_value:
...
 
-        const_shape = const_node.get_data().shape
-        const_value = np.reshape(const_value, const_shape)
-        new_const_node = opset.constant(const_value, dtype=const_node.get_element_type())
+        const_shape = const_node.data.shape
+        const_dtype = const_node.data.dtype
+        const_value = np.reshape(const_value, const_shape).astype(const_dtype)
+
+        # TODO(andrey-churkin): Replace on opset13.constant() in a future release
+        new_const_node = ov.op.Constant(const_value, shared_memory=True)
```
andrey-churkin, line 362: @alexsu52 @KodiaqQ The `shared_memory` parameter is not accessible in `opset13.constant()` for the our CI version of OV. It is available only on the master branch. Therefore, I am using `ov.op.Constant()` here. This will be updated in a future release.

KodiaqQ: I'm just adding a note that using `ov.op.Constant` is not consistent with our current approach of using `opset` from Python API.

alexsu52 (approved):

KodiaqQ (approved):

PR #2281, opened 2023-11-21
Enabled nf4 tests after switching to openvino==2023.2

ljaljushkin:
### Changes

as stated in title

### Reason for changes

new version of OpenVINO contains nf4 type, therefore no need to xfail nf4 tests

### Tests

nf4 tests


KodiaqQ (approved):

alexsu52 (approved):
LGTM

PR #2295, opened 2023-11-28
[PTQ][OV] Update OV references for 23.3

KodiaqQ:
### Changes

- Changed references for the LSTM model because of https://github.com/openvinotoolkit/openvino/pull/20368.
- Changed references for the `yolo-v4-tiny-tf` model due to layers name changing.
- Fixed bug with the in-place layer insertion.

### Reason for changes

- Updated initialization for ReadValue layers (for now not only zeroes can be set as the initial value).
- Probably MO frontend (TF) changed that led to layer name changes in models.

### Related tickets

- 125066

### Tests

- openvino recommit with nightly - 24


alexsu52 (approved):
LGTM

daniil-lyakhov (approved):
👍

PR #2298, opened 2023-11-29
Patch missed magic functions

AlexanderDokuchaev:
### Changes

Patch missed magic function 

```
['__ixor__', '__itruediv__', '__rtruediv__', '__abs__', '__ror__', '__ior__', '__rdiv__', '__imatmul__', '__rxor__', '__iand__', '__neg__', '__rdivmod__', '__ipow__', '__rpow__', '__invert__', '__rand__']
```

### Related tickets

124852

### Tests

models_hub_test
 - Was `91 failed, 602 passed`, after `83 failed, 610 passed`


nncf/torch/dynamic_graph/patch_pytorch.py
```diff
@@ -122,35 +122,51 @@ class FunctionsToPatchWithoutTracing:
 class MagicFunctionsToPatch:
     MAGIC_FUNCTIONS_TO_PATCH = {
         NamespaceTarget.TORCH_TENSOR: [
+            "__abs__",
```
alexsu52, line 125: Metatypes should be defined for all new operations and added support in NNCF algorithms.

daniil-lyakhov:
@AlexanderDokuchaev, is it possible to add some unit tests for this change?

nncf/torch/hardware/fused_patterns.py
```diff
@@ -53,16 +58,24 @@ def create_l2_norm_operations() -> GraphPattern:
 @PT_HW_FUSED_PATTERNS.register(HWFusedPatternNames.SHIFT_SCALE)
 def create_shift_scale() -> GraphPattern:
     pattern = GraphPattern()
-    add_node = pattern.add_node(label="ADD, SUB", type=["__add__", "__sub__"])
-    truediv_node = pattern.add_node(label="MUL, DIV", type=["__mul__", "__truediv__"])
+    add_node = pattern.add_node(label="ADD, SUB", type=[PTAddMetatype, PTSubMetatype])
```
AlexanderDokuchaev, line 61: @ljaljushkin please review changes in shift_scale patters

Found that pattern does not match to test case https://github.com/openvinotoolkit/nncf/blob/242aacdcb82035d7035476812099f8d8bf3cce03/tests/torch/test_models/synthetic.py#L367
1. `div_` and `sub_` was not matched to any metatype as result FQ was not added (not by patter)
2. Pattern does not match because of the `clone` node exists in the graph
3. Pattern does not match because of the 'is_floating_point' branch exists in the graph

ljaljushkin: Agreed that FQ shouldn't be inserted between in-place add and mul operations. 
Some models may use this pattern https://github.com/PeterL1n/RobustVideoMatting/blob/master/model/mobilenetv3.py#L37

AlexanderDokuchaev: Reverted changes for inplace ops.

kshpv (approved):

alexsu52 (approved):
LGTM

PR #2299, opened 2023-12-01
Add backup precision option

andrey-churkin:
### Changes

- Add a backup precision option that allows for preserving FQs on weights during the accuracy-aware algorithm.

### Reason for changes

See description in Ref: 123999

### Related tickets

Ref: 123999

### Tests

tests/openvino/native/quantization/test_quantizer_removal.py

### DIFF
--- a/nncf/quantization/advanced_parameters.py
+++ b/nncf/quantization/advanced_parameters.py
@@ -132,8 +132,12 @@ class AdvancedSmoothQuantParameters:
 
 class BackupMode(Enum):
     """
-    :param FP32:
-    :param INT8_WEIGHTS:
+    Specifies how to revert operations to their original precision.
+
+    :param FP32: Operations will be reverted to floating-point precision.
+    :param INT8_WEIGHTS: Certain operations (such as MatMul and Embedding) will
+        be reverted to representation with int8 weights, while all other operations
+        will revert to floating-point precision.
     """
 
     FP32 = "fp32"
@@ -232,7 +236,7 @@ class AdvancedAccuracyRestorerParameters:
     :param intermediate_model_dir: Path to the folder where the model, which was fully
         quantized with initial parameters, should be saved.
     :type intermediate_model_dir: Optional[str]
-    :param backup_mode:
+    :param backup_mode: Specifies how to revert operations to their original precision.
     :type backup_mode: BackupMode
     """
 
--- a/nncf/quantization/advanced_parameters.py
+++ b/nncf/quantization/advanced_parameters.py
@@ -135,9 +135,8 @@ class RestoreMode(Enum):
     Specifies how to revert operations to their original precision.
 
     :param ACTIVATIONS_AND_WEIGHTS: Operations will be reverted to floating-point precision.
-    :param ONLY_ACTIVATIONS: Certain operations (such as MatMul and Embedding) will
-        be reverted to representation with int8 weights, while all other operations
-        will revert to floating-point precision.
+    :param ONLY_ACTIVATIONS: Operations with weights will be reverted to representation with int8 weights,
+        while all other operations will revert to floating-point precision.
     """
 
     ACTIVATIONS_AND_WEIGHTS = "activations_and_weights"
--- a/nncf/quantization/advanced_parameters.py
+++ b/nncf/quantization/advanced_parameters.py
@@ -130,6 +130,16 @@ class AdvancedSmoothQuantParameters:
     matmul: float = 0.95
 
 
+class BackupMode(Enum):
+    """
+    :param FP32:
+    :param INT8_WEIGHTS:
+    """
+
+    FP32 = "fp32"
+    INT8_WEIGHTS = "int8_weights"
+
+
 @api()
 @dataclass
 class AdvancedQuantizationParameters:
@@ -222,6 +232,8 @@ class AdvancedAccuracyRestorerParameters:
     :param intermediate_model_dir: Path to the folder where the model, which was fully
         quantized with initial parameters, should be saved.
     :type intermediate_model_dir: Optional[str]
+    :param backup_mode:
+    :type backup_mode: BackupMode
     """
 
     max_num_iterations: int = sys.maxsize
@@ -229,6 +241,7 @@ class AdvancedAccuracyRestorerParameters:
     ranking_subset_size: Optional[int] = None
     num_ranking_workers: Optional[int] = None
     intermediate_model_dir: Optional[str] = None
+    backup_mode: BackupMode = BackupMode.FP32
 
 
 def changes_asdict(params: Any) -> Dict[str, Any]:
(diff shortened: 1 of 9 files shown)


nncf/quantization/algorithms/accuracy_control/openvino_backend.py
```diff
@@ -38,6 +39,10 @@ class OVAccuracyControlAlgoBackend(AccuracyControlAlgoBackend):
 
     # Metatypes
 
+    @staticmethod
+    def get_weighted_metatypes() -> List[OVOpMetatype]:
+        return [ov_metatypes.OVMatMulMetatype, ov_metatypes.OVEmbeddingMetatype]
```
KodiaqQ, line 44: Why the list of the weighted metatypes is limited and not correlated with the `OPERATIONS_WITH_WEIGHTS`?

nncf/quantization/advanced_parameters.py
```diff
@@ -130,6 +130,20 @@ class AdvancedSmoothQuantParameters:
...
+    """
+    Specifies how to revert operations to their original precision.
+
+    :param FP32: Operations will be reverted to floating-point precision.
+    :param INT8_WEIGHTS: Certain operations (such as MatMul and Embedding) will
+        be reverted to representation with int8 weights, while all other operations
+        will revert to floating-point precision.
+    """
+
+    FP32 = "fp32"
```
KodiaqQ, line 143: What about `FP16` precision?

andrey-churkin: `FP32` here means "Revert to original precision", so for `FP16` IR operations will be reverted to `FP16` precision. I agree that this name is a little bit confusing

nncf/quantization/algorithms/accuracy_control/openvino_backend.py
andrey-churkin: I used operation metatypes to which weight compression is applied.

https://github.com/openvinotoolkit/nncf/blob/5f2c20e35cf3c1aa284275c7d4fd7ed7593a6522/nncf/quantization/algorithms/weight_compression/openvino_backend.py#L38-L39

KodiaqQ: Why do we focus on the `weight compression` and not the `quantize` algorithm?

nncf/quantization/advanced_parameters.py
KodiaqQ: We should rename this option then. To the `ORIGINAL` for example.

andrey-churkin: Yes, I agree. I'll rename it.

nncf/quantization/algorithms/accuracy_control/openvino_backend.py
andrey-churkin: As far as I understand, the main idea behind the `backup_mode` parameter is to revert operations not only to their original precision but also to a representation with INT8 compressed weights. Therefore, we need to select a subset of operations with weights for which we want to preserve INT8 compressed weights. I followed the logic of the weights compression algorithm, wherein we compress weights only for `MatMul` and `Embedding` operations.

KodiaqQ: I don't get it.
Anyway, I'll leave the final decision to @alexsu52 and @AlexKoff88

andrey-churkin:
I have added a description of the experiments to the ticket.

nncf/quantization/algorithms/accuracy_control/openvino_backend.py
alexsu52: Good catch! I hope we can apply this logic to all quantized operations not only for operation which support for weights compression, if it does not lead to performance degradation.

nncf/quantization/advanced_parameters.py
```diff
@@ -222,13 +236,16 @@ class AdvancedAccuracyRestorerParameters:
...
+    :param backup_mode: Specifies how to revert operations to their original precision.
+    :type backup_mode: BackupMode
     """
 
     max_num_iterations: int = sys.maxsize
     tune_hyperparams: bool = False
     ranking_subset_size: Optional[int] = None
     num_ranking_workers: Optional[int] = None
     intermediate_model_dir: Optional[str] = None
+    backup_mode: BackupMode = BackupMode.FP32
```
alexsu52, line 248: @andrey-churkin @KodiaqQ What do you think about this proposal?
```
mode: RestoreMode = RestoreMode.ACTIVATIONS_AND_WEIGHTS


class RestoreMode(Enum):
    ACTIVATIONS_AND_WEIGHTS = "activations_and_weights"
    ONLY_ACTIVATIONS = "only_activations"
```
Using QuantizerGroup is more flexible and is not introduced new class but give user a chance to remove FQ from weights:

```
groups: List[QuantizerGroup] = [QuantizerGroup.ACTIVATIONS, QuantizerGroup.WEIGHTS]
```


cc` @AlexKoff88

alexsu52 (changes requested):

nncf/quantization/advanced_parameters.py
KodiaqQ: LGTM.

nncf/quantization/advanced_parameters.py
andrey-churkin: Done

nncf/quantization/advanced_parameters.py
andrey-churkin: Done

nncf/quantization/algorithms/accuracy_control/openvino_backend.py
andrey-churkin: I reverted half of all convolutions in MobileNetV2 using both `RestoreMode.ACTIVATIONS_AND_WEIGHTS` and `RestoreMode.ONLY_ACTIVATIONS`. There was no difference in performance between the resulting models, so I believe we can specify `OPERATIONS_WITH_WEIGHTS` here.

KodiaqQ (approved):

nncf/quantization/advanced_parameters.py
```diff
@@ -130,6 +130,20 @@ class AdvancedSmoothQuantParameters:
     matmul: float = 0.95
 
 
+class RestoreMode(Enum):
+    """
+    Specifies how to revert operations to their original precision.
+
+    :param ACTIVATIONS_AND_WEIGHTS: Operations will be reverted to floating-point precision.
+    :param ONLY_ACTIVATIONS: Certain operations (such as MatMul and Embedding) will
```
alexsu52, line 138: It looks like you need update the docstring taking into account https://github.com/openvinotoolkit/nncf/pull/2299#discussion_r1432667080

andrey-churkin: Thanks! Fixed.

alexsu52 (approved):
LGTM

PR #2301, opened 2023-12-05
Fix dump of patterns

AlexanderDokuchaev:
### Changes

Replace node name only if node id is string (it can also be `int` like in patterns)

### Reason for changes

```python
        for original_name in nx_graph.nodes():
>           dot_name = original_name.replace(__CHARACTER_REPLACE_FROM, __CHARACTER_REPLACE_TO)
E           AttributeError: 'int' object has no attribute 'replace'

nncf/common/utils/dot_file_rw.py:104: AttributeError
```

### Tests

tests/common/graph/test_graph_pattern.py::test_dump

### DIFF
--- a/tests/common/graph/test_graph_pattern.py
+++ b/tests/common/graph/test_graph_pattern.py
@@ -219,4 +219,5 @@ def test_join_pattern_with_special_input_node():
 def test_dump(tmp_path: Path):
     path_dot = tmp_path / "pattern.dot"
     TestPattern.first_pattern.dump_graph(path_dot)
-    assert path_dot.exists()
+    assert path_dot.is_file()
+    path_dot.unlink()
--- a/tests/common/graph/test_graph_pattern.py
+++ b/tests/common/graph/test_graph_pattern.py
@@ -11,6 +11,7 @@
 
 import copy
 import itertools
+from pathlib import Path
 
 import networkx as nx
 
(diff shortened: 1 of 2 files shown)


tests/common/graph/test_graph_pattern.py
```diff
@@ -213,3 +213,7 @@ def test_join_pattern_with_special_input_node():
             ref_pattern.add_edge(node, added_node)
 
     assert pattern == ref_pattern
+
+
+def test_dump(tmp_path):
+    TestPattern.first_pattern.dump_graph(tmp_path / "pattern.dot")
```
kshpv, line 219: What does this test check?

AlexanderDokuchaev: That dump_graph function is not failed with AttributeError.
Added check that `pattern.dot` exist.

kshpv (approved):

tests/common/graph/test_graph_pattern.py
```diff
@@ -213,3 +214,9 @@ def test_join_pattern_with_special_input_node():
             ref_pattern.add_edge(node, added_node)
 
     assert pattern == ref_pattern
+
+
+def test_dump(tmp_path: Path):
+    path_dot = tmp_path / "pattern.dot"
+    TestPattern.first_pattern.dump_graph(path_dot)
+    assert path_dot.exists()
```
kshpv, line 222: Please remove file after test

kshpv (changes requested):

tests/common/graph/test_graph_pattern.py
AlexanderDokuchaev: done

kshpv (approved):

PR #2304, opened 2023-12-05
[PTSQ 1/4][PTQ][Torch][SQ] PTSharedFnInsertionCommand and PTWeightUpdateCommand are introduced

daniil-lyakhov:
Part 1/4 of SQ support #2279

### Changes

* PTSharedFnInsertionCommand is introduced
* PTWeightUpdateCommand  is intorduced

### Reason for changes

* To make it possible to insert one operation which will be executed in several places of the model
* To make it possible to update weights of an module in-place as it done for biases

### Related tickets

124563

### Tests

* tests/torch/test_model_transformer.py is updated

### DIFF
--- a/nncf/torch/external_hook.py
+++ b/nncf/torch/external_hook.py
@@ -12,11 +12,8 @@
 from typing import Any
 
 from nncf.torch.dynamic_graph.context import TracingContext
-from nncf.torch.quantization.debug_interface import QuantizationDebugInterface
 
 EXTERNAL_OP_STORAGE_NAME = "external_op"
-EXTERNAL_QUANTIZERS_STORAGE_NAME = "external_quantizers"
-EXTERNAL_QUANTIZERS_STORAGE_PREFIX = "_nncf." + EXTERNAL_QUANTIZERS_STORAGE_NAME
 
 
 class ExternalOpCallHook:
@@ -43,24 +40,3 @@ class ExternalOpCallHook:
         replica = self._compressed_context.base_module_thread_local_replica
         storage = getattr(replica.nncf, self._storage_name)
         return storage[self._storage_key](*args, **kwargs)
-
-
-class ExternalQuantizerCallHook(ExternalOpCallHook):
-    """
-    External hook which is using quantization storage name and
-    could be constructed with a debug interface.
-    """
-
-    def __init__(
-        self,
-        context: TracingContext,
-        quantizer_storage_key: str,
-        debug_interface: QuantizationDebugInterface = None,
-    ):
-        super().__init__(EXTERNAL_QUANTIZERS_STORAGE_NAME, context, quantizer_storage_key)
-        self.debug_interface = debug_interface
-
-    def __call__(self, *args, **kwargs):
-        if self.debug_interface is not None:
-            self.debug_interface.register_activation_quantize_call(str(self._storage_key))
-        return super().__call__(*args, **kwargs)
--- a/nncf/torch/checkpoint_loading.py
+++ b/nncf/torch/checkpoint_loading.py
@@ -256,7 +256,7 @@ class NormalizedKeys:
         Returns original key if there's no ';' and operation doesn't start with EXTERNAL_QUANTIZERS_STORAGE_NAME
         """
         result = [new_key]
-        from nncf.torch.external_hooks import EXTERNAL_QUANTIZERS_STORAGE_PREFIX
+        from nncf.torch.quantization.external_quantizer import EXTERNAL_QUANTIZERS_STORAGE_PREFIX
 
         if ";" in new_key and new_key.startswith(EXTERNAL_QUANTIZERS_STORAGE_PREFIX):
             group_of_keys = new_key.split(";")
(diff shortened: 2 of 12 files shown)


nncf/common/graph/transformations/commands.py
```diff
@@ -35,6 +35,7 @@ class TransformationPriority(IntEnum):
     FP32_TENSOR_STATISTICS_OBSERVATION = 1
     PRUNING_PRIORITY = 2
     SPARSIFICATION_PRIORITY = 3
+    OP_INSERTION_PRIORITY = 4
```
daniil-lyakhov, line 38: Comment from the original PR, @KodiaqQ:
Why do we need a new priority? For which operations?

daniil-lyakhov: @KodiaqQ, you are right! Let me remove it

daniil-lyakhov: Done

KodiaqQ (approved):

AlexanderDokuchaev (changes requested):

nncf/torch/external_hook.py
```diff
@@ -0,0 +1,42 @@
...
+# limitations under the License.
+
+from typing import Any
+
+from nncf.torch.dynamic_graph.context import TracingContext
+
+EXTERNAL_OP_STORAGE_NAME = "external_op"
+
+
+class ExternalOpCallHook:
```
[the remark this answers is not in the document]
daniil-lyakhov: It is complicated to use one class in both use cases as `ExternalQuantizerCallHook` is using `QuantizationDebugInterface`, and I don't think we actually need a debug interface for `ExternalOpCallHook`.

nncf/torch/graph/transformations/commands.py
```diff
@@ -158,6 +158,33 @@ def requires_graph_rebuild(self):
...
+        self.fn = fn
+        self.op_name = op_unique_name
+        self.priority = priority
+
+    def union(self, other: "PTTransformationCommand") -> "PTTransformationCommand":
+        # TODO: keep all TransformationCommands atomic, refactor TransformationLayout instead
+        raise NotImplementedError()
+
+    def requires_graph_rebuild(self):
+        """
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

nncf/torch/external_hook.py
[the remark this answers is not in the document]
daniil-lyakhov: Ok, done

AlexanderDokuchaev (approved):

alexsu52 (commented):
Could you provide scenario where you need insert one operation which will be executed in several places of the model?

daniil-lyakhov:
> Could you provide scenario where you need insert one operation which will be executed in several places of the model?

For example: in following pattern
image
Smooth Quant multiply should be placed after input and before two matmuls like this:
image
As torch backend is using pre/post hook to insert modules, it is impossible to use one post hook after the input as it affect add branch. As we discussed with you previously, one of the solutions is to make an shared node and call shared pre-hooks before each matmul node like this:
image

nncf/torch/graph/transformations/commands.py
```diff
@@ -149,15 +149,32 @@ def union(self, other: "PTTransformationCommand") -> "PTTransformationCommand":
...
-        """
-        Return boolean flag to rebuild graph of model.
-
-        :return: Boolean flag.
-        """
         # Rebuild graph when adding quantization nodes.
         return self.priority == TransformationPriority.QUANTIZATION_PRIORITY
 
 
+class PTSharedFnInsertionCommand(PTTransformationCommand):
```
vshampor, line 156: Docstrings missing. When are you planning to move unified-scopes quantization to this new mechanism?

nncf/torch/external_hook.py
```diff
@@ -0,0 +1,42 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any
+
+from nncf.torch.dynamic_graph.context import TracingContext
+
+EXTERNAL_OP_STORAGE_NAME = "external_op"
```
vshampor, line 16: In view of the more general "op" now being introduced, does it make sense still to keep the "external_quantizer" key? If not, prepare the transition of the "external_quantizer" state dict key to "external_op", with backward compatibility and associated deprecation notices.

nncf/torch/external_hook.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2023 Intel Corporation
```
vshampor, line 1: A hook is an entity that belongs to a tracing context first, perhaps this file should have been located somewhere closer to `nncf/torch/dynamic_graph`

nncf/torch/external_hook.py
vshampor: Also, the external op storage name belongs to NNCFNetwork, not to the hook itself.

nncf/torch/graph/transformations/commands.py
daniil-lyakhov: Please see ticket 104304

nncf/torch/external_hook.py
daniil-lyakhov: That's nice idea, I think the best way to progress with that right now is to add it to refactoring list. That will be relevant when QAT initialization will be migrating to PTQ

PR #2309, opened 2023-12-08
Update readme tutorials

nikita-savelyevv:
I see that our tutorials/samples sections recently has become less readable due to increase of jupyter notebooks. I propose to reformat it as a number of tables containing link to resources with some categorization, e.g. explicitly stating compression algo, backend and machine learning domain. 

Decided not to hide the jupyter demo notebooks section in order not to sacrifice availability of this section to users. I suspect many won't notice a drop down field. Also when the section is hidden, Find on a page functionality of a browser does not work. You may still see the version with the hidden section on an older commit [here](https://github.com/openvinotoolkit/nncf/blob/4ebb148e2ab8bddbc7f0df7251a15ebe4639fe04/README.md#demos-tutorials-and-samples).

### Changes

Reformatted *Model Compression Tutorials and Samples* section as a set of tables:
- jupyter tutorials
- jupyter demos
- PTQ example scripts
- training time compression scripts

Added new notebooks:
- Distil-Whisper
- LCM
- SDXL-turbo
- ControlNet QR Code Monster
- Würstchen

Added missing YOLO v8 with Quantization with Accuracy Control PTQ example reference.

Resulting page [link](https://github.com/nikita-savelyevv/nncf/tree/readme-tutorials?tab=readme-ov-file#demos-tutorials-and-samples)


README.md
```diff
@@ -266,62 +266,76 @@ compression_ctrl.export_model("compressed_model.pb", save_format="frozen_graph")
...
+| [Joint Pruning, Quantization and Distillation for BERT](https://github.com/openvinotoolkit/openvino_notebooks/blob/main/notebooks/116-sparsity-optimization)                                                                                                                                                                                                                                             |                      Joint Pruning, Quantization and Distillation                       |  OpenVINO  |                 NLP                 |
+
+Please expand the table below to see the list of notebooks demonstrating OpenVINO conversion and inference together with NNCF compression for models from various domains.
+
```
l-bat, line 289: What do you think about moving the table below to `examples/openvino/README.md` and adding link here?

nikita-savelyevv: I personally think that it's import to highlight this in the main readme, because the list contains many trend models and it would be a pity to hide this information. Maybe in the future we could move some outdated models to a separate section.

l-bat: Unfortunately, this list is not completely aligned with NNCF-enabled OV notebooks. Also the `Backend` column only contains one value. I don't think it's needed. In my opinion, a drop-down list with different domains is more convenient for easier navigation.

nikita-savelyevv: Thanks for your suggestions!

#### Comment 1
> Unfortunately, this list is not completely aligned with NNCF-enabled OV notebooks.

Have I missed some notebooks? Could you please elaborate more on this.

#### Comment 2
> Also the Backend column only contains one value. I don't think it's needed.

For now it indeed does. Here's how it looks without Backend column: [preview](https://github.com/openvinotoolkit/nncf/blob/70c82e200a0079fb503c7388afb12a3edc12a788/README.md#demos-tutorials-and-samples). I personally would like to keep it because:
- If we remove it for one table, it won't be aligned with other tables, harming the overall look
- In the future there may be other values besides OpenVINO, in this case we'll need to bring the column back

#### Comment 3
> In my opinion, a drop-down list with different domains is more convenient for easier navigation.

Here's how it looks with two drop down sections for CV and NLP: [preview](https://github.com/openvinotoolkit/nncf/blob/623e0d594216f9fe8e1644d8fc3f311740f7d0e5/README.md#demos-tutorials-and-samples). And with Domain column removed: [preview](https://github.com/openvinotoolkit/nncf/blob/808ddd32b6489ec7d2acd13490ad2b154bfe947e/README.md#demos-tutorials-and-samples). I'd keep it as is because:
- I'm not really into hiding these. As I stated before, when these are hidden I feel like people won't bother expanding it and will miss it. Also, page search doesn't work in this case. Moreover, our page may not be indexed by search engine based on these model keywords. I have not tested it though.
- I've split it into two sections: CV and NLP because a more fine grained differentiation would be too much. However, if we keep the Domain column, I don't see a point in this split into two sections as the information is still there. On the other hand, if we remove the Domain column, then it again does not look aligned with other tables, and moreover it gets hard to understand what every notebook actually does, i.e. CV and NLP labels poorly describe the contents.

All in all, I'd keep it as is. But maybe a third opinion would be valuable on this.

cc @MaximProshin

AlexanderDokuchaev:
Looks like `Compression Algorithm` column can be removed from some tables too.
And please sort rows by name/

MaximProshin (approved):

PR #2312, opened 2023-12-11
Fix pt sanity tests

AlexanderDokuchaev:
### Changes

- Update argument name in tests 
- accelerate==0.24.1, 0.25.0 raises exception 
  ```
  NotImplementedError: Using RTX 3090 or 4000 series doesn't support faster communication broadband via P2P or IB. Please set `NCCL_P2P_DISABLE="1"` and `NCCL_IB_DISABLE="1" or use `accelerate launch` which will do this automatically.
  ```
- Path type in addoption

### DIFF
--- a/tests/torch/conftest.py
+++ b/tests/torch/conftest.py
@@ -13,6 +13,10 @@ import random
 from pathlib import Path
 
 import pytest
+from pytest import Config
+from pytest import FixtureRequest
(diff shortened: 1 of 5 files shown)


AlexanderDokuchaev:
nightly/job/torch_nightly/75/

alexsu52 (approved):

PR #2313, opened 2023-12-11
[PTSQ 3/4][Torch] Removable hooks

daniil-lyakhov:
Part 3/4 of SQ support https://github.com/openvinotoolkit/nncf/pull/2279.

### Changes

* Hooks groups are introduced for the nncf_network
* `remove_hooks_group` method for the nncf_network is intorduced
* NNCFNetwork.insert_at_point now inserts only one function but could be called several times with the same target point

### Reason for changes

* To make it possible to collect statistics on a model with hooks.
* To make it easier to remove hooks from the NNCFNetwork 

### Related tickets

124563

### Tests

* test_temporary_insert_at_point from `tests/torch/test_nncf_network.py`
* `/tests/torch/test_statistics_aggregator.py` is updated to check statistics are collected correctly in case some hooks were inserted previously


tests/torch/data/reference_graphs/quantized/adjust_paddings/vpu_bd_int8.dot
```diff
@@ -1,20 +1,20 @@
...
 "3 MultiBranchesModel/NNCFConv2d[conv_a]/SymmetricQuantizer/symmetric_quantize_0" [color=green, id=3, label="AFQ_[B:8 M:S SGN:U PC:Y]_#3_G0", style=filled, type=symmetric_quantize];
 "4 MultiBranchesModel/NNCFConv2d[conv_a]/conv2d_0" [color=purple, id=4, label="DW_Conv2d_k3x3_PAD_#4", style=filled, type=conv2d];
 "5 MultiBranchesModel/MaxPool2d[max_pool_b]/AsymmetricQuantizer/asymmetric_quantize_0" [color=green, id=5, label="AFQ_[B:8 M:A SGN:S PC:N]_#5_G1", style=filled, type=asymmetric_quantize];
 "6 MultiBranchesModel/MaxPool2d[max_pool_b]/max_pool2d_0" [id=6, label="max_pool2d_#6", style=filled, type=max_pool2d];
-"7 MultiBranchesModel/NNCFConv2d[conv_b]/ModuleDict[pre_ops]/UpdateWeight[0]/SymmetricQuantizer[op]/symmetric_quantize_0" [color=red, id=7, label="WFQ_[B:4 M:S SGN:U PC:Y]_#7_G1", style=filled, type=symmetric_quantize];
+"7 MultiBranchesModel/NNCFConv2d[conv_b]/ModuleDict[pre_ops]/UpdateWeight[1]/SymmetricQuantizer[op]/symmetric_quantize_0" [color=red, id=7, label="WFQ_[B:4 M:S SGN:U PC:Y]_#7_G1", style=filled, type=symmetric_quantize];
```
daniil-lyakhov, line 9: @ljaljushkin, is it ok if references will be containing the index of op insertion? This should be reproducible for all test cases

daniil-lyakhov: As I found, this counter should be at least thread local because test `est_autoq_precision_init/test_can_broadcast_initialized_precisions_in_distributed_mode/` requires this

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,74 @@
...
+        Adds operation to registered hooks registry.
+
+        :param op: Operation to set to registered hooks registry.
+        """
+        if self._hook_id is not None:
+            raise RuntimeError("Attempt to use one HookHandle for two hooks.")
+
+        hooks_registry = self.hooks_registry_ref()
+        if hooks_registry:
+            hook_id = max(hooks_registry.keys())
```
daniil-lyakhov, line 58: Do not sort strings, sort ints only

nncf/torch/nncf_network.py
```diff
@@ -418,12 +429,22 @@ def insert_at_point(self, point: PTInsertionPoint, fn_list: List[Callable]):
...
+        self._groups_vs_hooks_handlers[group].extend(handles)
+
+    def remove_hooks_group(self, group: HookGroups) -> None:
+        """
+        Removes all hooks of given group from the nncf interface.
+
+        :param group: Target hook group to remove all hooks from.
+        """
+        for handle in self._groups_vs_hooks_handlers[group]:
+            handle.remove()
```
daniil-lyakhov, line 447: TODO: remove handle as it is useless after the `remove` call

nncf/torch/nncf_network.py
```diff
@@ -418,12 +429,22 @@ def insert_at_point(self, point: PTInsertionPoint, fn_list: List[Callable]):
...
                 for fn in fn_list:
-                    nncf_module.register_post_forward_operation(fn)
+                    handles.append(nncf_module.register_post_forward_operation(fn))
         else:
             raise RuntimeError("Unsupported insertion type: {}".format(point.insertion_type))
+        self._groups_vs_hooks_handlers[group].extend(handles)
+
+    def remove_hooks_group(self, group: HookGroups) -> None:
+        """
+        Removes all hooks of given group from the nncf interface.
```
daniil-lyakhov, line 442: Add description about lifespan of hook handlers

nncf/torch/dynamic_graph/context.py
```diff
@@ -94,8 +96,8 @@ class TracingContext:
     def __init__(self):
         self.graph = DynamicGraph()
 
-        self._post_hooks: DefaultDict[OperationAddress, List[Callable]] = defaultdict(list)
-        self._pre_hooks: DefaultDict[PreHookId, List[Callable]] = defaultdict(list)
+        self._post_hooks: DefaultDict[OperationAddress, Dict[int, Callable]] = defaultdict(OrderedDict)
+        self._pre_hooks: DefaultDict[PreHookId, Dict[int, Callable]] = defaultdict(OrderedDict)
```
AlexanderDokuchaev, line 100: `Dict[str, Callable]` ?

AlexanderDokuchaev (approved):

daniil-lyakhov:
@vshampor, do you have an estimation when you will be able to review this PR?

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,63 @@
...
+        self._hook_id = None
+        self._op_registered = False
+
+    @property
+    def hook_id(self) -> Union[int, str]:
+        if self._hook_id is None:
+            raise RuntimeError("Attempt to use HookHandle hook id before actual hook registration.")
+        return self._hook_id
+
+    def add(self, op: Any) -> None:
```
alexsu52, line 35: `HookHandle` should not has the add method, because this is unmutable object for the hook. See example: https://github.com/pytorch/pytorch/blob/7dac2f9f2d0e41c2d4b07e4a673d448e63a4a819/torch/utils/hooks.py#L9

alexsu52 (changes requested):

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+
+    @abstractmethod
+    def remove(self) -> None:
+        """
+        Removes added operation from registered hooks registry if it is possible.
+        """
+
+
+class _HookHandle(HookHandleInterface):
```
alexsu52, line 37: Why is the private class?

nncf/torch/graph/transformations/commands.py
```diff
@@ -164,6 +164,25 @@ def requires_graph_rebuild(self):
         return True
 
 
+class PTInsertionTemporaryCommand(PTTransformationCommand):
+    """
+    Insertion operation to the models. Given fn is kept in HookGroups.TEMPORARY group of hooks.
```
alexsu52, line 169: Could you briefly explain difference between `Hook Groups.PERMANENT` and `Hook Groups.TEMPORARY` and why you added two groups instead of one?

nncf/torch/nncf_network.py
```diff
@@ -422,12 +440,23 @@ def insert_at_point(self, point: PTInsertionPoint, fn_list: List[Callable]):
...
+                    handles.append(nncf_module.register_pre_forward_operation(fn))
             elif point.insertion_type == PTInsertionType.NNCF_MODULE_POST_OP:
                 for fn in fn_list:
-                    nncf_module.register_post_forward_operation(fn)
+                    handles.append(nncf_module.register_post_forward_operation(fn))
         else:
             raise RuntimeError("Unsupported insertion type: {}".format(point.insertion_type))
+        self._groups_vs_hooks_handlers[group].extend(handles)
+
+    def remove_hooks_group(self, group: HookGroups) -> None:
```
alexsu52, line 451: Can I remove only one hook?

daniil-lyakhov: Yes, in case you insert it manually by `NNCFNetwork` api

daniil-lyakhov: Please describe a scenario when this one hook removal is required

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Privat hook handle implementation which should be instanciated only by
```
ljaljushkin, line 39:
```suggestion
    Private hook handle implementation which should be instantiated only by
```

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+def add_op_to_registry(hooks_registry: Dict[Any, Any], op: Any) -> HookHandleInterface:
+    """
+    Registers op into the hooks_registry and returns HookHandler instance.
+
+    :param hooks_registry: A dictionary of hooks, indexed by hook `id`.
+    :param op: Operation to set to registered hooks registry.
+    :return: HookHandle instance binded with hook registry and given op that
+        able to remove op from the hooks_registry.
+    """
+    if hooks_registry:
```
ljaljushkin, line 76: What about encapsulating registry and requests for adding and removing op in the single `HookRegistry` class?

ljaljushkin (changes requested):

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+        hook_id = max(map(int, hooks_registry))
+        hook_id = str(hook_id + 1)
+    else:
+        hook_id = "0"
```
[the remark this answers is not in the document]
daniil-lyakhov: Because torch.modules can use only strings as a key. There was comminated solution before, but we decided to use strings in all cases

nncf/common/hook_handle.py
daniil-lyakhov: Such approach was rejected, please see https://github.com/openvinotoolkit/nncf/pull/2313#discussion_r1449966113

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def add_op_to_registry(hooks_registry: Dict[Any, Any], op: Any) -> HookHandleInterface:
+    """
+    Registers op into the hooks_registry and returns HookHandler instance.
+
+    :param hooks_registry: A dictionary of hooks, indexed by hook `id`.
+    :param op: Operation to set to registered hooks registry.
+    :return: HookHandle instance binded with hook registry and given op that
+        able to remove op from the hooks_registry.
```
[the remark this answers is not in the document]
daniil-lyakhov: Done, thank you!

nncf/common/hook_handle.py
daniil-lyakhov: Done

nncf/common/hook_handle.py
```diff
@@ -0,0 +1,82 @@
...
+    @abstractmethod
+    def hook_id(self) -> str:
+        """
+        Key to use to retrieve handle from the registry.
+        """
+
+    @abstractmethod
+    def remove(self) -> None:
+        """
+        Removes added operation from registered hooks registry if it is possible.
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

ljaljushkin (approved):

daniil-lyakhov:
@alexsu52, please take a look

nncf/common/hook_handle.py
alexsu52: I don't share your concerns. I think developers can take responsibility to feed correct parameters in the __init__ method. The main concern that you return `_HookHandle` instance to user in `add_op_to_registry` and it does not look as good story.

nncf/torch/model_transformer.py
```diff
@@ -84,9 +84,11 @@ def _apply_insertion_transformations(model: NNCFNetwork, transformations: List[P
 
         :param model: Model to apply transformations.
         :param transformations: List of the bias correction transformations.
+        :param hooks_group: Target hooks group to pass to the NNCFNetwork.
```
alexsu52, line 87: leftover

nncf/torch/nncf_network.py
alexsu52: I think, nncf.strip is such scenario and any scenarios with collection custom statistics from the model.

alexsu52:
> @alexsu52, please take a look.
> 
> Code example to insert and remove one operation:
> 
> ```
> insertion_point = PTInsertionPoint(...)
> hook = get_hook(..)
> 
> hook_handle = model.nncf.insert_at_point(insertion_point, hook)
> ...
> hook_handle.remove()
> ```

Could you detail about `hook = get_hook(..)`? It's not clear what type of hook do you consider? torch.Module or python function?

daniil-lyakhov:
> > @alexsu52, please take a look.
> > Code example to insert and remove one operation:
> > ```
> > insertion_point = PTInsertionPoint(...)
> > hook = get_hook(..)
> > 
> > hook_handle = model.nncf.insert_at_point(insertion_point, hook)
> > ...
> > hook_handle.remove()
> > ```
> 
> Could you detail about `hook = get_hook(..)`? It's not clear what type of hook do you consider? torch.Module or python function?

There was no typing before, but the restrictions are following:
1) It should be a callable object which can receive target operation input/output
2) It should be a torch.Module in case target point is WEIGHTS_OPERATION
3) In case callable has parameters and target point is not WEIGHTS_OPERATION, it should be placed in the module dict (like `PTSharedFnInsertionCommand`) to make it possible to save and to train the model in the distributed mode.

I understand that this looks like a mess, but I suggest to deal with it in follow up prs as this PR is blocking SQ and it does not introduce any of this restrictions. I would happily make it simpler, but I afraid extensive refactoring is required

nncf/common/hook_handle.py
daniil-lyakhov: Done

nncf/torch/model_transformer.py
daniil-lyakhov: Done

alexsu52 (approved):

PR #2317, opened 2023-12-13
[Torch] `num_edges_expected` filled for each Torch metatype

daniil-lyakhov:
### Changes

 `num_edges_expected` filled for each Torch metatype

### Reason for changes

To fix `filter_constant_nodes` for cases when node with not filled yet metatype should be considered as an input node

### Tests
Build test_examples 169 is finished successfully


alexsu52 (approved):
LGTM

PR #2318, opened 2023-12-14
Added option for compressing first and last layers

ljaljushkin:
### Changes

as stated in the title

### Reason for changes

some models, e.g. Qwen, may benefit from quantization of the first and last layers to 4-bit
but by default, they were always quantized to 8-bit

### Tests

test_compress_first_and_last
test_mixed_precision

### DIFF
--- a/nncf/quantization/quantize_model.py
+++ b/nncf/quantization/quantize_model.py
@@ -291,7 +291,7 @@ def compress_weights(
                 "Default values of `ratio` (1) and `group_size` (-1) parameters can not be overridden"
             )
         if all_layers is not None:
-            raise AttributeError("INT8 modes does not support `all_layers` option, set it to None.")
+            raise AttributeError("INT8 modes do not support `all_layers` option, set it to None.")
     else:
         if ratio is None:
             ratio = 1
--- a/nncf/quantization/quantize_model.py
+++ b/nncf/quantization/quantize_model.py
@@ -246,7 +246,7 @@ def compress_weights(
     ratio: Optional[float] = None,
     group_size: Optional[int] = None,
     ignored_scope: Optional[IgnoredScope] = None,
-    all_layers: bool = False,
+    all_layers: Optional[bool] = None,
 ) -> TModel:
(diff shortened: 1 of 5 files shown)


nncf/quantization/quantize_model.py
```diff
@@ -299,7 +302,7 @@ def compress_weights(
 
         return compress_weights_impl(model, mode, ratio, group_size, ignored_scope)
```
l-bat, line 303: Should we add a warning/error that option is not supported in the TORCH backend?

MaximProshin:
@ljaljushkin , @AlexKoff88 shouldn't it be everything including embeddings?

ljaljushkin:
> @ljaljushkin , @AlexKoff88 shouldn't it be everything including embeddings?

@MaximProshin this options allows compressing all layers, embedding are included

nncf/quantization/quantize_model.py
ljaljushkin: thanks, makes sense

MaximProshin:
> > @ljaljushkin , @AlexKoff88 shouldn't it be everything including embeddings?
> 
> @MaximProshin this options allows compressing all layers, embedding are included

then I suggest to rename the option

nncf/quantization/quantize_model.py
```diff
@@ -287,6 +290,8 @@ def compress_weights(
                 "INT8 mode assumes per-channel quantization of all layers in 8 bit. "
                 "Default values of `ratio` (1) and `group_size` (-1) parameters can not be overridden"
             )
+        if all_layers is not None:
+            raise AttributeError("INT8 modes does not support `all_layers` option, set it to None.")
```
l-bat, line 294:
```suggestion
            raise AttributeError("INT8 modes do not support `all_layers` option, set it to None.")
```

l-bat (approved):

AlexKoff88 (approved):

andreyanufr (approved):

PR #2320, opened 2023-12-15
Adjust OV Constant to NF4 dtype

l-bat:
### Changes

relates to https://github.com/openvinotoolkit/nncf/pull/2315
OV nightly job: 32

### Reason for changes

To fix error with INT4 weight compression

### Related tickets

125777

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/openvino_backend.py
+++ b/nncf/quantization/algorithms/weight_compression/openvino_backend.py
@@ -14,6 +14,8 @@ from typing import List, Optional, Tuple, TypeVar
 
 import numpy as np
 import openvino.runtime as ov
+import openvino.runtime.opset12 as opset12
+from openvino.helpers import pack_data
 from openvino.runtime import opset13 as opset
 
 from nncf.common.graph import NNCFNode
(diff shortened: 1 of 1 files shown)


l-bat:
run openvino pre-commit tests

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -449,3 +451,26 @@ def _set_weight_compression_config(
...
+def create_constant_node(data: np.ndarray, dtype: ov.Type, name: str) -> ov.op.Constant:
+    """
+    Creates a constant node with the given constant value, data type, and name.
+
+    :param data: The constant value.
+    :param dtype: The data type of the constant node.
+    :param name: The name to be assigned to the constant node.
+    :return: The OpenVINO constant node.
+    """
+    if dtype in [ov.Type.i4, ov.Type.u4]:
```
AlexKoff88, line 465: I would implement i4 and u4 the same way as it is done for NF4. I consider this as a problem on OpenVINO side. So we can easily switch to opset13 once it is fixed.
In a nutshell, my motivation is the following: don't do the packing on your side because it is up to OpenVINO how to pack data types internally.

l-bat:
Problem was fixed on OV side https://github.com/openvinotoolkit/openvino/pull/21674
openvino-nightly job: 33

PR #2324, opened 2023-12-19
Added non-vnni test make command

KodiaqQ:
### Changes

- Updated Makefile with the `test-openvino-avx2` command.
- Updated sanity tests with AVX2 metrics.

### Reason for changes

- More robust testing pipeline.

### Related tickets

- 123702

### Tests

 - CI should use the new Makefile command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,11 @@ install-openvino-dev: install-openvino-test install-pre-commit
 test-openvino:
 	pytest ${COVERAGE_ARGS} tests/openvino $(DATA_ARG) --junitxml ${JUNITXML_PATH}
 
+test-openvino-avx2:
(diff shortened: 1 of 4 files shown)


KodiaqQ:
> I think the idea was to not introduce a new option but rather check and, if needed, change tests so that they could be run on both vnni and non-vnni hosts in case e.g. of github actions where we can't control which host is assigned every time. @alexsu52 , am I right? Or does this solution will automatically work in both cases?

The current solution does not depend on the device. Based on the results we can conclude that the suggested pipeline with the environment variable works and we need to adapt our sensitive tests for it. That was done in this PR. If we need something different, let's discuss and I'll do it.

PR #2325, opened 2023-12-19
Update MLFlow version

KodiaqQ:
### Changes

- As stated in the title

### Reason for changes

- Regular update

### DIFF
--- a/examples/torch/requirements.txt
+++ b/examples/torch/requirements.txt
@@ -3,7 +3,7 @@ pillow>=8.0.1
 tensorboard>=2.1
 matplotlib>=3.3.3
 defusedxml>=0.7.0rc1
-mlflow>=2.9.2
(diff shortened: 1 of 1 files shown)


examples/torch/requirements.txt
```diff
@@ -3,7 +3,7 @@ pillow>=8.0.1
 tensorboard>=2.1
 matplotlib>=3.3.3
 defusedxml>=0.7.0rc1
-mlflow==2.9.1
+mlflow>=2.9.2
```
vshampor, line 6: Better to use `~2.9.2` or `~2.9`, so that the patch or minor versions are up-to-date automatically, but at the same time no API-breaking major version changes happen without us knowing

vshampor (changes requested):

vshampor (approved):

PR #2326, opened 2023-12-19
Fix JSON schema link in docs

vshampor:
### Changes
The doc build action now generates TOC links to the JSON schema page in a slightly different fashion so that the TOC link is valid at all sub-pages

### Reason for changes
The original link only worked on the root page and 404-d otherwise.

### Related tickets
126069

### Tests
Deployment at https://vshampor.github.io/nncf/index.html


MaximProshin (approved):
LGTM

PR #2330, opened 2023-12-21
Fix pruning compression stage check

daniil-lyakhov:
#1149 reopened with tests

### Changes
Filter pruning algorithm now uses proper calculation of current pruning level to determine compression stage.
Affects SC.

### Reason for changes
Compression stage of the model pruned by flops was calculated incorrectly.
Affects Adaptive Compression training as best checkpoints were not saved.

### Related tickets
103420
### Tests
test_compression_stage


alexsu52 (approved):
LGTM

PR #2336, opened 2023-12-25
Update patch for roberta-large-mnli in third_party_integration

AlexanderDokuchaev:
### Changes

- Remove `attention_mask` from `input_info` to disable tracing. The placement of the FQ has not changed.
- Bump version of transformers and optimum

### Reason for changes 

After changes in the graph trace, `attention_mask` branch was added in NNCFgraph as result added FQ for mask. 
But since the mask is used values 0 and finfo(float32).min, inf and nan appear at backpropagation, as result parameters xof FQ are set nan.

### Related tickets

127936

### Tests

test_sanity_third_party.TestTransformers.test_glue_train

### DIFF
--- a/tests/post_training/requirements.txt
+++ b/tests/post_training/requirements.txt
@@ -8,8 +8,8 @@ onnxruntime==1.14.1
 pytest
 pytest-cov
 openvino-dev==2023.2.0
-optimum[onnxruntime,openvino]==1.8.8
-optimum-intel @ git+https://github.com/huggingface/optimum-intel@47428484598986443686b7c8ba825b4a3d7b4f73
+optimum[onnxruntime,openvino]==1.16.0
+optimum-intel @ git+https://github.com/huggingface/optimum-intel@622f585962e5e0ff5f927fd1a96fbc02b60a2e65
 soundfile==0.12.1
 librosa==0.10.0
 memory-profiler==0.61.0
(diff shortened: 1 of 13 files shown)


AlexanderDokuchaev:
ubuntu20_3dparty/285/

tests/post_training/requirements.txt
```diff
@@ -2,14 +2,14 @@
 torch==2.1.0
 --extra-index-url https://download.pytorch.org/whl/cpu
 torchvision==0.16.0
-transformers==4.30.0
+transformers==4.36.0
```
MaximProshin, line 5: please use ~=4.36 instead

tests/torch/pruning/experimental/test_nodes_grouping.py
```diff
@@ -482,6 +482,11 @@ def test_groups(desc: GroupTestDesc, mocker, tmp_path):
     compare_nx_graph_with_reference(nx_graph, path_to_dot, sort_dot_graph=False)
 
     filtered_groups = select_largest_groups(not_filtered_groups)
+    print(filtered_groups)
+
+    for x in filtered_groups:
+        print(x)
+    # print(desc.ref_groups)
```
ljaljushkin, line 489: please remove debug prints

ljaljushkin (approved):
LGTM except `prints`

tests/post_training/requirements.txt
AlexanderDokuchaev: This сan lead to a hidden version update that breaks the tests, like in this pr that requited update reference.
In tests better use strict version.

MaximProshin (approved):

PR #2337, opened 2023-12-26
Fix overflow in collectors

AlexanderDokuchaev:
### Changes

Cast float16 and float32 to float64 for mean and median.

### Reason for changes

Error on calculation parameters of FQ for large scale.

### Tests

test_reducers_and_aggregators.py::test_overflow

### DIFF
--- a/nncf/onnx/statistics/collectors.py
+++ b/nncf/onnx/statistics/collectors.py
@@ -52,11 +52,18 @@ class ONNXNNCFCollectorTensorProcessor(NNCFCollectorTensorProcessor):
         return ONNXNNCFTensor(np.maximum(x1.tensor, x2.tensor))
 
     @staticmethod
-    def mean(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims=False) -> NNCFTensor:
+    def mean(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims: bool = False) -> NNCFTensor:
+        if x.tensor.dtype in [np.float32, np.float16]:
+            res = np.mean(x.tensor, axis=axis, keepdims=keepdims, dtype=np.float64).astype(dtype=x.tensor.dtype)
+            return ONNXNNCFTensor(res)
         return ONNXNNCFTensor(np.mean(x.tensor, axis=axis, keepdims=keepdims))
 
     @staticmethod
-    def median(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims=False) -> NNCFTensor:
+    def median(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims: bool = False) -> NNCFTensor:
+        if x.tensor.dtype in [np.float32, np.float16]:
+            t = x.tensor.astype(dtype=np.float64)
+            res = np.median(t, axis=axis, keepdims=keepdims).astype(dtype=x.tensor.dtype)
+            return ONNXNNCFTensor(res)
         return ONNXNNCFTensor(np.median(x.tensor, axis=axis, keepdims=keepdims))
 
     @classmethod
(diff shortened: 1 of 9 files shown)


nncf/onnx/statistics/collectors.py
```diff
@@ -70,7 +77,16 @@ def masked_mean(
         if mask is None:
             return cls.mean(x, axis=axis, keepdims=keepdims)
         masked_x = np.ma.array(x.tensor, mask=mask.tensor)
-        return ONNXNNCFTensor(np.ma.mean(masked_x, axis=axis, keepdims=False).data)
+        if x.tensor.dtype in [np.float32, np.float16]:
+            result = np.ma.mean(masked_x, axis=axis, keepdims=keepdims, dtype=np.float64)
+            if isinstance(result, np.ma.MaskedArray):
+                result = result.data
+            return ONNXNNCFTensor(result.astype(dtype=x.tensor.dtype))
+        else:
```
daniil-lyakhov, line 85: Can we use `dtype` as a parameter and reduce amount of branches here?

AlexanderDokuchaev: The problem is that cast result to type of input tensor required only for float16 and float32, for integer cast is no need, because result is float64. Еhere is no way without using conditions.

nncf/onnx/statistics/collectors.py
```diff
@@ -76,17 +74,12 @@ def masked_mean(
...
-                result = result.data
-            return ONNXNNCFTensor(result.astype(dtype=x.tensor.dtype))
-        else:
-            result = np.ma.mean(masked_x, axis=axis, keepdims=keepdims)
-            if isinstance(result, np.ma.MaskedArray):
-                result = result.data
-            return ONNXNNCFTensor(result)
+        result = np.ma.mean(masked_x, axis=axis, keepdims=keepdims, dtype=comp_dtype)
+        if isinstance(result, np.ma.MaskedArray):
+            result = result.data
```
daniil-lyakhov, line 81: Don't forget to specify return type in this case

nncf/torch/tensor_statistics/collectors.py
```diff
@@ -71,11 +71,14 @@ def max(cls, *args) -> NNCFTensor:
...
+    def mean(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims: bool = False) -> NNCFTensor:
+        comp_dtype, out_dtype = _get_computing_dtype(x.tensor.dtype)
+        return PTNNCFTensor(
+            torch.mean(x.tensor, axis=axis, keepdims=keepdims, dtype=comp_dtype).to(dtype=out_dtype, copy=False)
+        )
 
     @staticmethod
-    def median(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims=False) -> NNCFTensor:
+    def median(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims: bool = False) -> NNCFTensor:
         # See https://github.com/pytorch/pytorch/issues/61582
```
daniil-lyakhov, line 82: `_get_computing_dtype` is missed for the median function

nncf/torch/tensor_statistics/collectors.py
```diff
@@ -85,42 +88,50 @@ def median(x: NNCFTensor, axis: Union[int, Tuple[int, ...], List[int]], keepdims
...
+        # Implemented in numpy as torch.masked.median is not implemented yet
         if mask is None:
             return cls.median(x, axis=axis, keepdims=keepdims)
+
         device = x.tensor.device
-        masked_x = np.ma.array(x.tensor.detach().cpu().numpy(), mask=mask.tensor.detach().cpu().numpy())
+        dtype = x.tensor.dtype
+        np_tensor = x.tensor.detach().cpu().numpy()
+        if np_tensor.dtype in [np.float32, np.float16]:
+            np_tensor = np_tensor.astype(dtype=np.float64)
```
daniil-lyakhov, line 116: Please use `_get_computing_dtype` here

nncf/onnx/statistics/collectors.py
AlexanderDokuchaev: This is done below
`return ONNXNNCFTensor(result.astype(dtype=out_dtype, copy=False))`

nncf/torch/tensor_statistics/collectors.py
AlexanderDokuchaev: Done. For quantile it is not needed.

nncf/torch/tensor_statistics/collectors.py
AlexanderDokuchaev: done

daniil-lyakhov (approved):

PR #1958, opened 2023-07-10
[DOCS] Model Zoo refactoring

nikita-savelyevv:
### Changes

- Created a single models list at `./docs/ModelZoo.md` instead of list of models in the main `README.md` + list of models in every sample's `README.md`, resulting in duplication
- Refactored models tables to be in HTML format instead of Markdown, as it's more customizable

### Reason for changes

- Improving NNCF documentation
- Removing duplication in models tables
- Shortening main readme page

PR #2339, opened 2023-12-27
Enable SIM rules in ruff

AlexanderDokuchaev:
### Changes

- Enable checks from https://pypi.org/project/flake8_simplify/
- removed `tools/update_eval_results.py` and `.file-header` as outdated


ljaljushkin:
Cool improvement! I saw many comments in PR's to make fixes like that.

daniil-lyakhov (approved):
👍

tools/update_eval_results.py
```diff
@@ -1,574 +0,0 @@
-# Copyright (c) 2023 Intel Corporation
```
andrey-churkin, line 1: We don't need this tool anymore?

nncf/tensorflow/quantization/init_range.py
```diff
@@ -65,15 +65,15 @@ def get_init_config_for_quantization_point(self, layer: tf.keras.layers.Layer, i
...
-                        RangeInitConfig(
-                            pl_config.init_type, pl_config.num_init_samples, pl_config.init_type_specific_params
-                        )
+            )
+            if should_be_considered and (group == pl_config.target_group or pl_config.target_group is None):
+                matches.append(
+                    RangeInitConfig(
+                        pl_config.init_type, pl_config.num_init_samples, pl_config.init_type_specific_params
                     )
+                )
```
nikita-savelyevv, line 76: Was this also an automatic refactor? I'd suppose creating `should_be_considered ` is unnecessary.

nncf/torch/compression_method_api.py
```diff
@@ -195,9 +195,9 @@ def _get_transformation_layout(self, target_model: NNCFNetwork) -> PTTransformat
     def _handle_frozen_layers(self, target_model: NNCFNetwork):
         scopes_of_frozen_layers = []
         for weighted_node in target_model.nncf.get_weighted_original_graph_nodes():
-            if not weighted_node.layer_attributes.weight_requires_grad:
-                if self._should_consider_scope(weighted_node.node_name):
-                    scopes_of_frozen_layers.append(weighted_node.node_name)
+            should_be_considered = self._should_consider_scope(weighted_node.node_name)
+            if not weighted_node.layer_attributes.weight_requires_grad and should_be_considered:
+                scopes_of_frozen_layers.append(weighted_node.node_name)
```
nikita-savelyevv, line 200: The same question as above.

nncf/torch/quantization/init_range.py
```diff
@@ -71,13 +71,13 @@ def get_init_config_for_quantization_point(self, qp: QuantizationPointBase) -> R
...
-                        RangeInitConfig(
-                            pl_config.init_type, pl_config.num_init_samples, pl_config.init_type_specific_params
-                        )
+            should_be_considered = should_consider_scope(qid, pl_config.ignored_scopes, pl_config.target_scopes)
+            if should_be_considered and (group == pl_config.target_group or pl_config.target_group is None):
+                matches.append(
+                    RangeInitConfig(
+                        pl_config.init_type, pl_config.num_init_samples, pl_config.init_type_specific_params
                     )
+                )
```
nikita-savelyevv, line 80: Here too

tests/torch/pruning/filter_pruning/test_algo.py
```diff
@@ -711,8 +711,8 @@ def test_flops_calculator(model_module, all_weights, pruning_flops_target, ref_f
     config["compression"]["algorithm"] = "filter_pruning"
     config["compression"]["params"]["all_weights"] = all_weights
     config["compression"]["params"]["prune_first_conv"] = True
-    config["compression"]["pruning_init"] = 0.5 if not pruning_flops_target else pruning_flops_target
-    if pruning_flops_target:
+    config["compression"]["pruning_init"] = pruning_flops_target if pruning_flops_target is not None else 0.5
+    if pruning_flops_target is not None:
```
nikita-savelyevv, line 715: 1. Before `pruning_init` would become 0.5 if `pruning_flops_target` **was not None**. Now it'll be equal 0.5 if `pruning_flops_target` **is None**. Was there a bug before?
2. Was this check designed for `pruning_flops_target` being None, 0 or both? Now it'll only be triggered by None.
3. What's now the correct way to check for 0 and None? Are checks of the form `if a:` are not accepted now?

tests/torch/test_models/swin.py
```diff
@@ -272,7 +272,7 @@ def __init__(
     def forward(self, x):
         H, W = self.input_resolution
         B, L, C = x.shape
-        assert L == H * W, "input feature has wrong size"
+        assert H * W == L, "input feature has wrong size"
```
nikita-savelyevv, line 275: What's the rule in such comparisons?

nncf/tensorflow/quantization/init_range.py
AlexanderDokuchaev: No, but without it autoformatting looks bad 
```python
            if should_consider_scope(qid, pl_config.ignored_scopes, pl_config.target_scopes) and (
                group == pl_config.target_group or pl_config.target_group is None
            ):
```
With `should_be_considered` variable more clear.

tests/torch/test_models/swin.py
AlexanderDokuchaev: https://docs.astral.sh/ruff/rules/yoda-conditions/

tests/torch/pruning/filter_pruning/test_algo.py
AlexanderDokuchaev: Yes, looks like it was a bug, because `bool(None)` and `bool(0.0)` == False,
But this changes does not effect on test (model PruningTestBatchedLinear final final pruning level is 0 in both case).

nncf/tensorflow/quantization/init_range.py
nikita-savelyevv: Ok, I see. I'd personally would choose the autoformatted version just to be unified with the rest of the code, because otherwise it's unclear when to create a separate variable for a condition and when not to create. But this way is also fine if you like it more.

tests/torch/test_models/swin.py
nikita-savelyevv: Thanks!

tools/update_eval_results.py
AlexanderDokuchaev: The script was outdate since https://github.com/openvinotoolkit/nncf/pull/1958#issuecomment-1641775114. 
Think that half a year is enough to make decision to remove it.

andrey-churkin: @alexsu52 FYI

alexsu52 (approved):

tools/update_eval_results.py
alexsu52: + @KodiaqQ

nncf/torch/model_analyzer.py
```diff
@@ -81,7 +81,4 @@ def is_quantized_weights(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
     :param nncf_graph: The NNCF graph.
     :return bool: return `True` if the node is quantized.
     """
-    for prev_node in nncf_graph.get_previous_nodes(node):
-        if prev_node.node_type in OP_NAMES_QUANTIZE_NODE:
-            return True
-    return False
+    return any(prev_node.node_type in OP_NAMES_QUANTIZE_NODE for prev_node in nncf_graph.get_previous_nodes(node))
```
KodiaqQ, line 84: Well, the previous version looks slightly better than this, even with unapplicable rules.

tools/update_eval_results.py
KodiaqQ: We need to finalize the decision about the tool remove before removing it.
Even in the provided PR https://github.com/openvinotoolkit/nncf/pull/1958, there is a suggestion to track this in a separate ticket.
So, this PR shouldn't delete this.

AlexanderDokuchaev: Ok, create 129482

nncf/torch/model_analyzer.py
AlexanderDokuchaev: Yes, that's debatable, for me `any` looks better and shorter. But ok, disabled this rule and reverted changes.

KodiaqQ (approved):

andrey-churkin (approved):

nikita-savelyevv (approved):

PR #2321, opened 2023-12-18
Data-based mixed-precision for weight compression

ljaljushkin:
### Changes

Introduced 4 new criteria (HAWQ_IN, MEAN_VAR, MEAN_MAX, MAX_VAR) for assigning different quantization precision based on the given dataset

### Reason for changes

determine more accurate configuration by taking into account activations

### Related tickets

126378

### Tests

test_data_based_criterion
test_mixed_precision

- [x] proper docstrings and typehints
- [x] final check of accuracy on llama, qwen

the MAX_ACTIVATION_VARIANCE metric shows improvement on wikitext and who-what-benchmark for the 4 considered models.
image

Manual validation on llama-2 shows that output is without German words.
image

MEAN_VAR
image
MAX_VAR
image
HAWQ_IN
image
INT8_ERROR
image

PR #2344, opened 2023-12-27
Added note about execution time of data-based mixed precision weight compression

ljaljushkin:
Data-based mixed precision weight compression has been added in #2321 as an alternative for data-free method.
Added note to documentation about speed of the new method.
Generally, data-based approach is slower on 20-40% than data-free method.
image


MaximProshin (commented):
@ljaljushkin , I don't think we should say about exact percentage as it's HW dependent and "takes more time" should be enough here.
@AlexKoff88 , what do you think?

ljaljushkin:
> enerally, data-based approach is slower on 20-40% than data-free metho

I can paraphrase: `takes more time (~30% slower on Intel(R) Xeon(R) Gold 6430L)` to mention an approximate overhead. I don't think it will be significantly different on other CPUs.

MaximProshin (approved):

PR #2360, opened 2024-01-02
Add release notes for release v2.8.0

vshampor:


ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
 # Release Notes
 
+## New in Release 2.8.0
+
+Post-training Quantization:
+
+- Breaking changes:
+  - `nncf.quantize` signature has been changed to add `mode: Optional[nncf.QuantizationMode] = None` as its 3-rd argument, between the original `calibration_dataset` and `preset` arguments.
```
AlexKoff88, line 8: Why is the first bullet relates to breaking changes? It looks like an extension of the current API but not the change. I mean the existing users' calibration pipeline will not be broken with this change after update of the NNCF version.

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+  - (Common) `nncf.common.quantization.structs.QuantizationMode` has been renamed to `nncf.common.quantization.structs.QuantizationScheme`
+- General:
+  - (OpenVINO) Changed default OpenVINO opset from 9 to 13.
+- Features:
+  - (OpenVINO) Added support for quantizing the ScaledDotProductAttention operation from OpenVINO opset 13.
+  - (Common) Split the `nncf.CompressWeights.INT8` into `nncf.CompressWeights.INT8_SYM` and `nncf.CompressWeights.INT8_ASYM`, which specify [symmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#symmetric-quantization) and [asymmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#asymmetric-quantization) for weight compression respectively.
```
AlexKoff88, line 14:
```suggestion
  - (Common) Renamed the `nncf.CompressWeights.INT8` to `nncf.CompressWeights.INT8_ASYM` and introduce `nncf.CompressWeights.INT8_SYM` that can be efficiently used with dynamic 8-bit quantization of activations.
```

AlexKoff88: It looks like a breaking change, btw.

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+  The original `nncf.CompressWeights.INT8` enum value is now deprecated.
+  - (OpenVINO) Added FP8 quantization support via `nncf.QuantizationMode.FP8_E4M3` and `nncf.QuantizationMode.FP8_E5M2` enum values, invoked via passing one of these values as an optional `mode` argument to `nncf.quantize`.
```
AlexKoff88, line 16:
```suggestion
  - (OpenVINO) Added FP8 quantization support via `nncf.QuantizationMode.FP8_E4M3` and `nncf.QuantizationMode.FP8_E5M2` enum values, invoked via passing one of these values as an optional `mode` argument to `nncf.quantize`. Currently, OpenVINO supports inference of FP8-quantized models in reference mode with no performance benefits and can be used for accuracy projections.
```

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+  - (Common) Added an `all_layers: Optional[bool] = None` argument to `nncf.compress_weights` to indicate whether embeddings and last layers of the model should be compressed to a primary precision.
```
AlexKoff88, line 17:
```suggestion
  - (Common) Added an `all_layers: Optional[bool] = None` argument to `nncf.compress_weights` to indicate whether embeddings and last layers of the model should be compressed to a primary precision. It is relevant to 4-bit quantization only.
```

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+  - (Common) Added a `sensitivity_metric: Optional[nncf.parameters.SensitivityMetric] = None` argument to `nncf.compress_weights` for finer control over the sensitivity metric for assigning quantization precision to layers.
```
AlexKoff88, line 18: I guess `nncf.compress_weights` was also extended with `dataset` parameter. 
@ljaljushkin to help with formulating this change.

ReleaseNotes.md
MaximProshin: I agree and I would mention it in the bullet about FP8 below

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
 # Release Notes
 
+## New in Release 2.8.0
+
+Post-training Quantization:
+
+- Breaking changes:
+  - `nncf.quantize` signature has been changed to add `mode: Optional[nncf.QuantizationMode] = None` as its 3-rd argument, between the original `calibration_dataset` and `preset` arguments.
+  - (Common) `nncf.common.quantization.structs.QuantizationMode` has been renamed to `nncf.common.quantization.structs.QuantizationScheme`
```
MaximProshin, line 9: Maybe it's useful to clearly state that "It's not impacting any user API".

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+- General:
+  - (OpenVINO) Changed default OpenVINO opset from 9 to 13.
+- Features:
+  - (OpenVINO) Added support for quantizing the ScaledDotProductAttention operation from OpenVINO opset 13.
+  - (Common) Split the `nncf.CompressWeights.INT8` into `nncf.CompressWeights.INT8_SYM` and `nncf.CompressWeights.INT8_ASYM`, which specify [symmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#symmetric-quantization) and [asymmetric](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/Quantization.md#asymmetric-quantization) for weight compression respectively.
+  The original `nncf.CompressWeights.INT8` enum value is now deprecated.
```
MaximProshin, line 15: It should also be listed in the deprecation section below

ReleaseNotes.md
MaximProshin: @vshampor , you can add it like: 
Added 4-bit data-aware weights compression. For that ‘dataset’ optional parameter has been added to nncf.compress_weights() and can be used to minimize accuracy degradation of compressed models (note that this option increases the compression time).

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+## New in Release 2.8.0
+
+Post-training Quantization:
+
+- Breaking changes:
+  - `nncf.quantize` signature has been changed to add `mode: Optional[nncf.QuantizationMode] = None` as its 3-rd argument, between the original `calibration_dataset` and `preset` arguments.
+  - (Common) `nncf.common.quantization.structs.QuantizationMode` has been renamed to `nncf.common.quantization.structs.QuantizationScheme`
+- General:
+  - (OpenVINO) Changed default OpenVINO opset from 9 to 13.
+- Features:
```
MaximProshin, line 12: Missed #123999. You can add it like:
Post-training Quantization with Accuracy Control, nncf. quantize_with_accuracy_control(), has been extended by ‘restore_mode’ optional parameter to revert weights to int8 instead  of the original precision. This parameter helps to reduce the size of the quantized model and improves its performance. By default, it’s disabled and model weights are reverted to the original precision in nncf.quantize_with_accuracy_control().

MaximProshin (changes requested):

ReleaseNotes.md
```diff
@@ -1,5 +1,36 @@
...
+  - (PyTorch) `nncf.strip(..., do_copy=True)` now actually returns a deepcopy (stripped) of the model object.
+  - (PyTorch) Post-hooks can now be set up on operations that return `torch.return_type` (such as `torch.max`).
+  - (PyTorch) Improved dynamic graph tracing for various tensor operations from `torch` namespace.
+  - (PyTorch) More robust handling of models with disjoint traced graphs when applying PTQ.
+  Developers are advised to use [optimum-intel](https://github.com/huggingface/optimum-intel) instead.
+- Improvements:
+  - Reformatted the tutorials section in the top-level `README.md` for better readability.
+- Deprecations/Removals:
+  - (Common) Removed legacy external quantizer storage names.
+  - (PyTorch) The Git patch for integration with HuggingFace `transformers` repository is marked as deprecated and will be removed in a future release.
```
MaximProshin, line 32: add the reference

MaximProshin (changes requested):

ReleaseNotes.md
vshampor: The existing pipelines will be broken for the following use case:
```
quantized_model = nncf.quantize(model, dataset, QuantizationPreset.Performance)
```
and basically any usage where more than 2 of the `nncf.quantize` arguments have been used positionally.

ReleaseNotes.md
vshampor: This is not a breaking change at this point since the original `INT8` enum value is preserved.

ReleaseNotes.md
vshampor: Done

ReleaseNotes.md
vshampor: Done

ReleaseNotes.md
vshampor: Done

ReleaseNotes.md
vshampor: Looks like I added the link to a wrong place, it should be fine now.

ReleaseNotes.md
vshampor: In a previous NNCF version, the `nncf.QuantizationMode` with its original meaning had been marked as an API entity: 
https://github.com/openvinotoolkit/nncf/blob/aa4408b5470be7e168eb91d434be3b11e96f12a7/nncf/common/quantization/structs.py#L24-L34

in a transitive fashion since it had been used in another API entity, namely, `nncf.quantization.advanced_parameters.QuantizationParameters`:
https://github.com/openvinotoolkit/nncf/blob/aa4408b5470be7e168eb91d434be3b11e96f12a7/nncf/quantization/advanced_parameters.py#L90

In the current state the `QuantizationParameters` has incorrect/misleading documentation and type hint with respect to the same parameter due to:
https://github.com/openvinotoolkit/nncf/blob/634e6b1df142b9d6cb5d0f08acbfdffb2fa9e7c6/nncf/quantization/advanced_parameters.py#L22

and also all previous code that instantiated `QuantizationParameters` will be incorrect because `nncf.common.quantization.structs.QuantizationMode`, which you in general need for that instantiation, does not exist anymore.

No deprecations have been announced ahead of time for this change, and no time was given for the users to adjust their code. It is therefore a breaking change as stated originally.

MaximProshin: I guess it was missed during the review then. @AlexKoff88 , I don't think we can fix it anyhow but do you see any impact on users, optimum-intel, etc?

ReleaseNotes.md
```diff
@@ -1,5 +1,40 @@
...
+  By default, the backup precision is assigned for the embeddings and last layers.
+- Fixes:
+  - (OpenVINO) Models with embeddings (e.g. `gpt-2`, `stable-diffusion-v1-5`, `stable-diffusion-v2-1`, `opt-6.7b`, `falcon-7b`, `bloomz-7b1`) are now more accurately quantized.
+  - (PyTorch) `nncf.strip(..., do_copy=True)` now actually returns a deepcopy (stripped) of the model object.
+  - (PyTorch) Post-hooks can now be set up on operations that return `torch.return_type` (such as `torch.max`).
+  - (PyTorch) Improved dynamic graph tracing for various tensor operations from `torch` namespace.
+  - (PyTorch) More robust handling of models with disjoint traced graphs when applying PTQ.
+- Improvements:
+  - Reformatted the tutorials section in the top-level `README.md` for better readability.
+- Deprecations/Removals:
```
MaximProshin, line 33: Let's also add a deprecation about the Docker files.

ReleaseNotes.md
AlexKoff88: I agree to keep it here if it really breaks the BC. But maybe it is worth adding info on how the code should be changed, e.g.:
Use `quantized_model = nncf.quantize(model, dataset, mode=QuantizationMode.PERFORMANCE)` instead of Use `quantized_model = nncf.quantize(model, dataset, QuantizationPreset.PERFORMANCE)`.

ReleaseNotes.md
AlexKoff88: No impact as these are details of internal implementation.

ReleaseNotes.md
MaximProshin: has anyone tried to reproduce it to ensure we broke the BC here?
cc @alexsu52 @KodiaqQ

ReleaseNotes.md
vshampor: Done

ReleaseNotes.md
```diff
@@ -1,5 +1,41 @@
...
+  - (Common) Post-training Quantization with Accuracy Control - `nncf.quantize_with_accuracy_control()` has been extended by `restore_mode` optional parameter to revert weights to int8 instead of the original precision.
+  This parameter helps to reduce the size of the quantized model and improves its performance.
+  By default, it's disabled and model weights are reverted to the original precision in `nncf.quantize_with_accuracy_control()`.
+  - (Common) Added an `all_layers: Optional[bool] = None` argument to `nncf.compress_weights` to indicate whether embeddings and last layers of the model should be compressed to a primary precision. This is relevant to 4-bit quantization only.
+  - (Common) Added a `sensitivity_metric: Optional[nncf.parameters.SensitivityMetric] = None` argument to `nncf.compress_weights` for finer control over the sensitivity metric for assigning quantization precision to layers.
```
ljaljushkin, line 22:
```suggestion
  - (OpenVINO) Added an `all_layers: Optional[bool] = None` argument to `nncf.compress_weights` to indicate whether embeddings and last layers of the model should be compressed to a primary precision. This is relevant to 4-bit quantization only.
  - (OpenVINO) Added a `sensitivity_metric: Optional[nncf.parameters.SensitivityMetric] = None` argument to `nncf.compress_weights` for finer control over the sensitivity metric for assigning quantization precision to layers.
```
Currently, it's relevant to int4 compression in the OpenVINO backend only

ReleaseNotes.md
```diff
@@ -1,5 +1,41 @@
...
+
+Post-training Quantization:
+
+- Breaking changes:
+  - `nncf.quantize` signature has been changed to add `mode: Optional[nncf.QuantizationMode] = None` as its 3-rd argument, between the original `calibration_dataset` and `preset` arguments.
+  - (Common) `nncf.common.quantization.structs.QuantizationMode` has been renamed to `nncf.common.quantization.structs.QuantizationScheme`
+- General:
+  - (OpenVINO) Changed default OpenVINO opset from 9 to 13.
+- Features:
+  - (OpenVINO) Added 4-bit data-aware weights compression. For that `dataset` optional parameter has been added to `nncf.compress_weights()` and can be used to minimize accuracy degradation of compressed models (note that this option increases the compression time).
```
ljaljushkin, line 13:
```suggestion
  - (OpenVINO) Added 4-bit data-aware mixed-precision weights compression. For that `dataset` optional parameter has been added to `nncf.compress_weights()` and can be used to minimize accuracy degradation of compressed models (note that this option increases the compression time). 
```

ReleaseNotes.md
KodiaqQ: I've tested locally on the `nncf/examples/post_training_quantization/openvino/mobilenet_v2/main.py`. 
The scenario with the `nncf.quantize(model, dataset, QuantizationPreset.PERFORMANCE)` would throw:
`RuntimeError: overflow_fix option with OverflowFix.FIRST_LAYER value is not supported with the mode option!`. 
Because the mode is not `None` other options must be disabled as well.

MaximProshin:
@AlexKoff88 , @ljaljushkin , @KodiaqQ , please re-check it as all your comments have been applied.

KodiaqQ (approved):

ljaljushkin (approved):

PR #2371, opened 2024-01-05
[PTQ][MinMax][Micro refactoring] Use RangeInitCollectorParams across all backends

daniil-lyakhov:
### Changes

`RangeInitCollectorParams` is used in common code during statistic collector creation.

### Reason for changes

 To keep logic of `use_abs_max` parameter common

### Related tickets

109437


kshpv (approved):

AlexanderDokuchaev (approved):

nncf/common/quantization/initialization/range.py
```diff
@@ -160,6 +160,28 @@ def __init__(self, is_weights: bool, mode: QuantizationMode, per_channel: bool):
...
+        True for Weights, False for Activations.
+        """
+        return self._is_weights
+
+    @property
+    def mode(self) -> QuantizationMode:
+        """
+        Returns quantization mode: symmetric or asymmetric.
+        """
+        return self._mode
```
KodiaqQ, line 176: Why are we still trying to implement `mode` instead of the `scheme`? `QuantizationMode` is not `symmetric or asymmetric` anymore. I hope that @alexsu52 can clear it here.

nncf/common/quantization/initialization/range.py
```diff
@@ -160,6 +160,28 @@ def __init__(self, is_weights: bool, mode: QuantizationMode, per_channel: bool):
...
+
+    @property
+    def per_channel(self) -> bool:
+        """
+        Returns quantization granularity.
+        """
+        return self._per_channel
```
KodiaqQ, line 183: Since `_per_channel` contains only a `boolean` value, why not call it `_is_per_channel`? To align it with `_is_weights`.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -366,11 +367,16 @@ def _get_stat_collector(
         """
         range_estimator_params = self._get_range_estimator_parameters(target_point, quantizer_config)
 
+        collector_params = RangeInitCollectorParams(
+            is_weights=target_point.is_weight_target_point(),
+            mode=quantizer_config.mode,
+            per_channel=quantizer_config.per_channel,
+        )
```
KodiaqQ, line 374: As I can see, we calculate `RangeInitCollectorParams` instead of using `QuantizerConfig` here. Why do we need that? How does it help?
Based on the ticket, I understood that `RangeInitCollectorParams` calculates in the Torch backend as well and we can re-use it. But it's not, based on the PR's code.

daniil-lyakhov: The main reason behind is to [calculate parameter `use_abs_max` in one place](https://github.com/openvinotoolkit/nncf/pull/2371#issue-2067303016). And as this `RangeInitCollectorParams` are used in Torch to calculate them, I decided to align all backends and use its logic in OV and ONNX as well.

Alternative solution is to implement new way of `use_abs_max` processing and migrate legacy torch code on it, but I believe this approach will do better

nncf/common/quantization/initialization/range.py
daniil-lyakhov: Good catch, this name is from legacy quantization class, we can rename it

nncf/common/quantization/initialization/range.py
daniil-lyakhov: Done

nncf/common/quantization/initialization/range.py
KodiaqQ: When and where the renaming would be done?

daniil-lyakhov: Let's discuss it offline with the team

daniil-lyakhov: Done, please check

KodiaqQ (approved):

PR #2287, opened 2023-11-23
[PTQ][OV] Drop old, set new references

KodiaqQ:
### Changes

- Dropped 23.1 outdated references.
- Added 23.3 references (symlink-like).

PR #2377, opened 2024-01-08
[OV] Reference search instead of symlinks 

daniil-lyakhov:
### Changes

* Reference search is used instead of symlinks
* OV test transform fn skips pot in case it is not available

### Reason for changes

* To enable pre-commit tests for OV 2024.0 version

### Tests
ubuntu20_precommit_openvino/41/ is finished successfully


KodiaqQ:
@daniil-lyakhov if you add references for the new OpenVINO version, you should drop references for the older one.
Also, versions for 2024.0 should be copied from 2023.3 at least, not 2023.2 since they are not actual.
Next, which references were changed in comparison with 2023.3 and why? Have you analysed it?

tests/openvino/test_transform_fn.py
```diff
@@ -62,6 +64,12 @@ def multiple_inputs_as_dict_transform_fn(data_item):
     ],
 )
 def test_transform_fn(model, transform_fn, use_pot: bool):
+    if use_pot:
+        try:
+            importlib.import_module("openvino.tools.pot")
+        except ImportError:
+            pytest.skip("POT is not available in the current environment.")
+
```
KodiaqQ, line 72: Why do not drop POT from this test as well?

KodiaqQ (changes requested):
@daniil-lyakhov if you add references for the new OpenVINO version, you should drop references for the older one.
Also, versions for 2024.0 should be copied from 2023.3 at least, not 2023.2 since they are not actual.
Next, which references were changed in comparison with 2023.3 and why? Have you analysed it?

daniil-lyakhov:
> @daniil-lyakhov if you add references for the new OpenVINO version, you should drop references for the older one. Also, versions for 2024.0 should be copied from 2023.3 at least, not 2023.2 since they are not actual. Next, which references were changed in comparison with 2023.3 and why? Have you analysed it?

It's my mistake, I've copied 2023.3 metrics, and I didn't change any of them, it worked as is.

tests/openvino/test_transform_fn.py
daniil-lyakhov: I suggest to do it in a separate PR where all pot references will be removed

KodiaqQ (changes requested):
The 2024.0 references still contain symlinks to the 2023.2.
See https://github.com/openvinotoolkit/nncf/pull/2287/ as the reference PR.

daniil-lyakhov:
Build ubuntu20_precommit_openvino/42/

daniil-lyakhov:
openvino-nightly/job/trigger/25/

daniil-lyakhov:
openvino-nightly/job/trigger/25/

KodiaqQ (approved):

tests/openvino/native/common.py
```diff
@@ -80,3 +82,33 @@ def get_openvino_version() -> str:
...
+    Get path to actual reference file.
+
+    :param rel_path: Relative path to reference file.
+
+    :return: Path to reference file or raise RuntimeError.
+    """
+    root_dir = OPENVINO_NATIVE_TEST_ROOT / "data"
+    current_ov_version = version.parse(get_openvino_version())
+
+    def is_version(dir_path: Path) -> bool:
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

PR #2391, opened 2024-01-12
[Experimental][Collectors] Decouple inplace reducers from base reducers

daniil-lyakhov:
### Changes

InplaceReducers interface was extracted from the base implementation of reducer

### Reason for changes

To make it easier to implement reducers for backends


alexsu52 (approved):
I like it!

PR #2395, opened 2024-01-16
Removed mlflow usage

ljaljushkin:
### Changes

removed usage of mlflow

### Reason for changes

AFAIK, it's not used during the development in the team and there was no interest from customers to extend it. 
Also, it's extra dependency and often leads to errors in validation, that required several workarounds (SafeMLFLow and `mock_mlflow`):
```
E           Exception: Run with UUID 14ae21d2c1d349839bca0371583cfc78 is already active. To start a new run, first end the current run with mlflow.end_run(). To start a nested run, call start_run with nested=True
```
### Tests

- [x] nightly torch: WIP build 108
- [x] nightly torch: WIP build 110


vshampor (approved):

examples/torch/requirements.txt
```diff
@@ -3,8 +3,7 @@ pillow>=8.0.1
 tensorboard>=2.1
 matplotlib>=3.3.3
 defusedxml>=0.7.0rc1
-mlflow~=2.9
-returns>0.14
+tabulate~=0.9.0
```
ljaljushkin, line 6: `returns` is not needed anymore since it was for `SafeMLFlow` only 
`tabulate` is needed for optional dumping of debug information in the AutoQ algorithm

vshampor (approved):

PR #2396, opened 2024-01-16
[PTQ][OV] Add empty tensors handling

KodiaqQ:
### Changes

- Added `keep_empty_stats` for reducers;
- Updated `get_raw_stat_collector` to save empty data;

### Reason for changes

- More accurate statistics collection for BC case;

### Related tickets

- 129429

### Tests

 - Updated `test_reducers_and_aggregators.py`


nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -103,10 +110,11 @@ def __eq__(self, __o: object) -> bool:
             isinstance(__o, self.__class__)
             and self._reduction_axes == __o._reduction_axes
             and self._inplace == __o.inplace
+            and self._keep_empty_stats == __o._keep_empty_stats
         )
 
     def __hash__(self) -> int:
-        return hash((self.__class__.__name__, self.inplace, self._reduction_axes))
+        return hash((self.__class__.__name__, self.inplace, self._reduction_axes, self._keep_empty_stats))
```
daniil-lyakhov, line 117: As hash is always overriding please update hashes for each reducer

KodiaqQ: Outdated.

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -299,9 +302,9 @@ def register_inputs(self, inputs: Dict[int, List[NNCFTensor]]) -> None:
         for reducer in self._reducers:
             reducer_hash = hash(reducer)
             input_ = inputs[reducer_hash]
-            if any(tensor.is_empty() for tensor in input_):
-                continue
-            reduced_inputs[reducer_hash] = reducer(input_)
+            reduced_input = reducer(input_)
```
daniil-lyakhov, line 305: What if None will be valid output for the reducers?

nncf/openvino/statistics/collectors.py
```diff
@@ -195,6 +196,11 @@ def get_output_names(self, target_node_name: str, port_id: int) -> List[str]:
         return [get_result_node_name(target_node_name, port_id)]
 
 
+class OVRawReducer(RawReducer):
+    def get_output_names(self, target_node_name: str, port_id: int) -> List[str]:
+        return [get_result_node_name(target_node_name, port_id)]
```
daniil-lyakhov, line 201: `get_output_names` method does not exist anymore, there is no need to make a backend-specific realization for this reducer

KodiaqQ: Then, why wasn't it dropped for other reducers? Why do we have backend reducers now?

KodiaqQ: I don't get what should I do with this. Do I need to remove `OVRawReducer` and `OVNoopReducer` as well?

kshpv:
It looks like the fix is general for OV,ONNX and PT. The name of PR confused me

nncf/openvino/statistics/collectors.py
daniil-lyakhov: Yes, if you don't mind. I made a PR recently https://github.com/openvinotoolkit/nncf/pull/2391 and forgot to delete OVNoopReducer, PTNoopReducer and ONNXNoopReducer

tests/common/experimental/test_reducers_and_aggregators.py
```diff
@@ -170,12 +170,14 @@ def squeeze_tensor(self, ref_tensor: List[Any], axes: Optional[Tuple[int]] = Non
     def cast_tensor(self, tensor, dtype: Dtype):
         pass
 
-    def test_noop_reducer(self, reducers):
-        reducer = reducers["noop"]()
-        input_ = np.arange(24).reshape((1, 2, 3, 4))
-        reduced_input = reducer([self.get_nncf_tensor(input_)])
+    @pytest.mark.parametrize(
+        "reducer_name, input_data", [("noop", np.arange(24).reshape((1, 2, 3, 4))), ("raw", np.array([]))]
```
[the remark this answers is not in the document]
KodiaqQ: Which cases do we need to cover?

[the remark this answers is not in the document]
kshpv: Reducing non empty and empty tensors in a row.

nncf/openvino/statistics/collectors.py
KodiaqQ: `get_output_names` also exists for other Reducers and needs for `TensorCollector.get_output_info`.
Why do you suggest to remove it?

daniil-lyakhov: Please rebase your PR

KodiaqQ: Ok, then I've got it.
What should we do with the [ONNX backend](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/onnx/statistics/collectors.py#L199-L232) and [Torch backend](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/tensor_statistics/collectors.py#L226-L259) then? Why should we keep all these classes?

daniil-lyakhov: 1) OV has inplace reducers implementation and other backend does not have it
2) method `_get_processor` is still backend-specific

KodiaqQ: For some reducers, we're going to have common implementations, for some - not. That confuses me. And there is still no answer as to why we can't drop **pass** reducers for ONNX and Torch.

daniil-lyakhov: > For some reducers, we're going to have common implementations, for some - not. That confuses me. 

Reducers which do nothing except filtering can be backend-agnostic, reducers which do something with input tensor cannot be backend-agnostic yet. 

We can drop PTNoopReducer, OVNoopReducer, ONNXNoopReducer and OVRawReducer, PTRawReducer, ONNXRawReducer, no problems.  I suggest to drop them

KodiaqQ: We can also drop ONNXMinReducer, ONNXMaxReducer, ONNXAbsMaxReducer, etc. since they do nothing, can't we?

KodiaqQ: Where is the implementation [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/onnx/statistics/collectors.py#L199-L232)?

daniil-lyakhov: Please see this https://github.com/openvinotoolkit/nncf/blob/develop/nncf/onnx/statistics/collectors.py#L191-L197

KodiaqQ: What a mess with these classes.

daniil-lyakhov: Awful, I agree

tests/common/experimental/test_reducers_and_aggregators.py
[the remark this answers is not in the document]
KodiaqQ: Thank you. Added.

tests/common/experimental/test_reducers_and_aggregators.py
```diff
@@ -498,17 +505,9 @@ def test_mad_percentile_aggregators_not_implemented_aggregation_axes(
...
+        ["min", "max", "abs_max", "mean", "quantile", "abs_quantile", "batch_mean", "mean_per_ch"],
     )
     def test_reducers_name_hash_equal(self, reducer_name, reducers):
-        if reducer_name == "noop":
-            reducers_instances = [reducers[reducer_name]() for _ in range(2)]
-            assert hash(reducers_instances[0]) == hash(reducers_instances[1])
-            assert reducers_instances[0] == reducers_instances[1]
-            assert reducers_instances[0].name == reducers_instances[1].name
-            assert len(set(reducers_instances)) == 1
-            return
```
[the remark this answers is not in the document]
daniil-lyakhov: Honestly, I would keep it as it is and add RawReducer here, just to check that base class hash works as expected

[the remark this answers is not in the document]
daniil-lyakhov: Maybe it is better to do a new micro test for them with this check

daniil-lyakhov (approved):

tests/common/experimental/test_reducers_and_aggregators.py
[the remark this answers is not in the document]
KodiaqQ: Don't we have these checks for the base class? To be honest, I just don't want to add copy-pasted classes to the `reducers` pytest fixture list.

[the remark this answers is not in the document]
daniil-lyakhov:
```
  @pytest.mark.parametrize("reducer_cls", [NoopReducer, RawReducer])
    def test_other_reducers_name_hash_equal(self, reducer_cls):
         reducers_instances = [reducer_cls() for _ in range(2)]
         assert hash(reducers_instances[0]) == hash(reducers_instances[1])
         assert reducers_instances[0] == reducers_instances[1]
         assert reducers_instances[0].name == reducers_instances[1].name
         assert len(set(reducers_instances)) == 1
```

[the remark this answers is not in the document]
KodiaqQ: Looks good 👍 
Added.

alexsu52 (approved):

nikita-savelyevv (approved):

kshpv (approved):

PR #2400, opened 2024-01-17
As/release/8bit weights compression

alexsu52:
cherry pick to release #2333


KodiaqQ:
- Is that a release stopper?
- Do we need to update ReleaseNotes with these changes?
- Which validation runs do we need to have for the release branch?
- Does it change E2E PT metrics?
cc @MaximProshin, @AlexKoff88

alexsu52:
>     * Do we need to update ReleaseNotes with these changes?
#2406 
>     * Which validation runs do we need to have for the release branch?
builds: 306, 308, 309
>     * Does it change E2E PT metrics?
No

PR #2411, opened 2024-01-22
Reverted casting operations in do_integer_quantization function.

alexsu52:
### Changes

Reverted casting operations in do_integer_quantization function.

### Reason for changes

numpy operations with uint8 is very slow.

### Related tickets

ref: 130239

### Tests

N/A


MaximProshin:
I don't think we should map it to the release branch unless there are other reasons than the compression time or if we still plan to run release checks. cc @KodiaqQ

KodiaqQ:
For now, we have no plans and almost have no time for another validation.
Also, no numbers were provided to show the severity of degradation compared with the previous release.

PR #2412, opened 2024-01-22
Replaced RuntimeError

Om-Doiphode:
### Changes

Built-in exception type `RuntimeError` was replaced with appropriate custom-made exception functions.
Fixes issue https://github.com/openvinotoolkit/openvino/issues/22192

### Reason for changes

Accomplishing this would enable better user/developer experience for NNCF, and also would allow the developers to apply [fuzzing](https://github.com/google/atheris) techniques more effectively.


nncf/errors.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    pass
+
+class ModuleNotFoundError(Exception):
+    """
+    Raised when a required module is not found.
+    """
+    pass
+
+class ParameterNotFoundError(Exception):
```
vshampor, line 71: This `Parameter` is semantically `torch.nn.Parameter`, while the `Parameter` in the `ParameterNotSupportedError` below is a generic "compression algorithm API parameter", which may be confusing. Based on the usage, I propose to merge this `ParameterNotFoundError` into `nncf.InternalError`. Also, for the remaining `nncf.ParameterNotSupportedError` please add some words into the docstring explaining that the `Parameter` here actually refers to a compression algo parameter (such as a parameter passed in an argument of an API call).

nncf/errors.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    pass
+
+class ParameterNotSupportedError(Exception):
+    """
+    Raised when an unsupported parameter is encountered.
+    """
+    pass
+
+class NotSupportedError(Exception):
```
vshampor, line 83: This one currently also only applies to the compression algo parameters, propose to merge it into `nncf.ParameterNotSupportedError`

nncf/errors.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    Raised when a type or operation is not supported.
+    """
+    pass
+
+class ExcessParameterError(Exception):
```
vshampor, line 89: Now this `Parameter`, if left as-is, would have a third meaning, adding to the confusion. Based on the usages, I propose to merge these cases into `nncf.InternalError`, as these are currently raised in more or less obscure cases that are only fixable by the developers.

nncf/errors.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    pass
+
+class UnsupportedDatasetError(Exception):
+    """
+    Raised when an unsupported dataset is encountered
+    """
+    pass
+
+class InvalidFolderPathError(Exception):
```
vshampor, line 23: I would make this more general, say, `InvalidPathError`

vshampor (changes requested):
Great job - this is pretty much what I wanted to see. I have approved the CI checks, but they won't be very informative, I suppose, since your code is currently not exactly runnable - remember that you have to `import` your new error classes, e.g. `from nncf import InternalError`, before you can raise them.

I'll keep an eye on this and if your future pushes do not automatically trigger the CI checks, or if you can't see the results of the check via web, I suggest that you run the associated unit tests locally before pushing, for better chances of CI checks passing. You can do this by running `make test-*backend*`, where `*backend*` can be either `torch`, `tensorflow`, `openvino`, `onnx` or `common`, and will run the corresponding parts of the precommit checks accordingly. Before running each of the above, though, make sure that your virtual environment has the necessary test requirements - you can install those by running `make install-*backend*-test` before each `make test-*backend*` invocation. Check the respective make targets directly in the `Makefile` when in doubt, there are merely console commands there.

Some tests may fail for you locally if you don't have >1 GPU, or even any GPU at all, but these are the minority, are easy to spot and we'll rely on the CI runs for those.

Om-Doiphode:
Hi @vshampor, I have made the necessary changes, please review. Thanks!

Om-Doiphode:
Hi @vshampor, I tried running the tests on my system, but it is taking too much time to run all the tests. In 2 hours, it could complete only 78% of the tests and then it got stuck there.

vshampor:
> Hi @vshampor, I tried running the tests on my system, but it is taking too much time to run all the tests. In 2 hours, it could complete only 78% of the tests and then it got stuck there.

Understandable. I think that the test results for each Github precommit check are visible if you click on the "Details" link, I should have directed you there so you could have run only the failing set of the tests. Anyway, you've done most of the work already, I've pushed some commits of my own with minor changes like linting and `pytest.raises` adjustments. Let's see if we have green checks now.

vshampor:
The install tests are passing.

vshampor (approved):

PR #2415, opened 2024-01-22
Hide an irrelevant warning during tune_range call

nikita-savelyevv:
When `tune_range()` is called with a `left_border` being equal to 0. This leads to the following warning
```
/home/nsavel/workspace/nncf/nncf/experimental/tensor/tensor.py:84: RuntimeWarning: invalid value encountered in multiply
  return Tensor(self.data * unwrap_tensor_data(other))
```

at this line:

```
rb = fns.where(qval > 0.0, (qval - level_high) / qval * left_border, right_border)
```

When we divide by `qval` which equals 0, this results in a NaN during computation. At the same time if `qval` is zero then `where` condition does not hold and `rb==right_border`. Seems like numpy computes both results nevertheless of the condition.

### Changes

Do not show the unnecessary warning because it is confusing.

### Reason for changes

Improving UX

### Tests

Added a test for `tune_range`. Not sure about the location, feel free to suggest a better one.

### DIFF
--- a/nncf/quantization/fake_quantize.py
+++ b/nncf/quantization/fake_quantize.py
@@ -122,11 +122,8 @@ def tune_range(
         qval = fns.round(fval)
 
     ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
-    with warnings.catch_warnings():
-        # If `qval` is 0 `rb` will equal `right_border`, and we don't want to show an unnecessary division by 0 warning
-        warnings.simplefilter("ignore")
-        rb_then_result = (qval - level_high) / qval * left_border
-    rb = fns.where(qval > 0.0, rb_then_result, right_border)
+    qval_ = fns.where(qval == 0.0, fns.ones_like(qval), qval)
+    rb = fns.where(qval > 0.0, (qval - level_high) / qval_ * left_border, right_border)
 
     range_a = right_border - ra
     range_b = rb - left_border
--- a/nncf/quantization/fake_quantize.py
+++ b/nncf/quantization/fake_quantize.py
@@ -121,7 +121,7 @@ def tune_range(
         qval = fns.round(fval)
 
     ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
-    qval_ = fns.where(qval == 0.0, fns.ones_like(qval), qval)
+    qval_ = fns.where(qval == 0.0, 1.0, qval)
     rb = fns.where(qval > 0.0, (qval - level_high) / qval_ * left_border, right_border)
 
     range_a = right_border - ra
--- a/nncf/quantization/fake_quantize.py
+++ b/nncf/quantization/fake_quantize.py
@@ -9,6 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


nncf/quantization/fake_quantize.py
```diff
@@ -121,7 +122,11 @@ def tune_range(
         qval = fns.round(fval)
 
     ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
-    rb = fns.where(qval > 0.0, (qval - level_high) / qval * left_border, right_border)
+    with warnings.catch_warnings():
```
alexsu52, line 125: Could you fix it instead of using `warnings.catch_warnings()`?

nikita-savelyevv: I could, for example like this:
```
qval_ = fns.where(qval == 0.0, fns.ones_like(qval), qval)
rb = fns.where(qval > 0.0, (qval - level_high) / qval_ * left_border, right_border)
```

But I decided just to intercept the warning on purpose. I wouldn't say that something is _wrong_ here to fix it, it's just the way numpy works processing both `where` branches even though only one of them will be used. 

Do you think it's a good idea to solve "cosmetic" problems by changing the computation logic?

alexsu52: IMHO:This is more clear for me, because changes add 1 line of code and don't require additional comment in the code.

nikita-savelyevv: Ok, I'll check whether there is difference in performance and switch to the code above if it's not significant

nikita-savelyevv: With warning catching `tune_range` is about 7% slower, with a solution above it's about 20%. But still it's nothing compared to the rest of the pipeline. So I replace the warning catching with addition of one line.

alexsu52: it looks like not good.  Maybe you can reuse ideas from here https://discuss.pytorch.org/t/torch-divide-only-where-denominator-is-non-zero/133968

nikita-savelyevv: I've tried it out. For numpy this is indeed a preferred way. But for pytorch I had to do a workaround like this:
```
@numeric.divide.register(torch.Tensor)
def _(x1: torch.Tensor, x2: Union[torch.Tensor, float], where: Optional[torch.Tensor] = None) -> torch.Tensor:
    result = torch.clone(x1)
    if where is not None:
        result[where] = torch.divide(x1[where], x2[where])
    return result
```
This is far less effective, than it currently is. I propose to keep it as it is in the PR now.

nncf/quantization/fake_quantize.py
```diff
@@ -121,7 +121,8 @@ def tune_range(
         qval = fns.round(fval)
 
     ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
-    rb = fns.where(qval > 0.0, (qval - level_high) / qval * left_border, right_border)
+    qval_ = fns.where(qval == 0.0, fns.ones_like(qval), qval)
```
alexsu52, line 124:
```suggestion
    qval_ = fns.where(qval == 0.0, 1.0, qval)
```

alexsu52 (approved):
LGTM

PR #2418, opened 2024-01-23
[Experimental tensor] numeric.quantile is introduced

daniil-lyakhov:
### Changes

Numeric.quantile is introduced

### Reason for changes

To make it possible to calculate quantile on `nncf.experimental.tensor.Tensor`

### Tests

test_fn_quantile

### DIFF
--- a/nncf/experimental/tensor/functions/numpy_numeric.py
+++ b/nncf/experimental/tensor/functions/numpy_numeric.py
@@ -185,7 +185,7 @@ def _(
     q: Union[float, List[float]],
     axis: Union[int, Tuple[int]] = None,
     keepdims: Optional[bool] = None,
-) -> Union[float, Union[np.ndarray, np.generic]]:
+) -> Union[np.ndarray, np.generic]:
     if keepdims is None:
         keepdims = np._NoValue
     return np.array(np.quantile(a, q=q, axis=axis, keepdims=keepdims))
--- a/nncf/experimental/tensor/functions/numpy_numeric.py
+++ b/nncf/experimental/tensor/functions/numpy_numeric.py
@@ -188,10 +188,7 @@ def _(
 ) -> Union[float, Union[np.ndarray, np.generic]]:
     if keepdims is None:
         keepdims = np._NoValue
-    ret_val = np.quantile(a, q=q, axis=axis, keepdims=keepdims)
-    if isinstance(ret_val, np.ndarray):
-        return ret_val
-    return np.array(ret_val)
+    return np.array(np.quantile(a, q=q, axis=axis, keepdims=keepdims))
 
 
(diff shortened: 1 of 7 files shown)


nncf/experimental/tensor/functions/numpy_numeric.py
```diff
@@ -179,6 +179,21 @@ def _(a: Union[np.ndarray, np.generic], decimals: int = 0) -> np.ndarray:
...
+    q: Union[float, List[float]],
+    axis: Union[int, Tuple[int]] = None,
+    keepdims: Optional[bool] = None,
+) -> Union[float, Union[np.ndarray, np.generic]]:
+    if keepdims is None:
+        keepdims = np._NoValue
+    ret_val = np.quantile(a, q=q, axis=axis, keepdims=keepdims)
+    if isinstance(ret_val, np.ndarray):
+        return ret_val
+    return np.array(ret_val)
```
alexsu52, line 194:
```suggestion
    retrun np.array(np.quantile(a, q=q, axis=axis, keepdims=keepdims))
```

alexsu52 (approved):

nncf/experimental/tensor/functions/torch_numeric.py
```diff
@@ -191,6 +192,28 @@ def _(a: torch.Tensor, decimals=0) -> torch.Tensor:
...
+    if len(a) <= 16_000_000 and isinstance(axis, int):
+        result = torch.quantile(
+            a,
+            torch.tensor(q, dtype=a.dtype, device=a.device),
+            axis,
+            keepdims,
+        )
+    else:
+        result = torch.tensor(np.quantile(a.detach().cpu().numpy(), q=q, axis=axis, keepdims=keepdims))
+    return result.type(a.dtype).to(device)
```
AlexanderDokuchaev, line 214: In numpy quantile always return float64. But in torch it can be float32 or float64 depends from input dtype. 
In your implementations it will cast to input dtype.
Think that need to return dtype like in numpy. 

And torch implementation supports only float32 and float64, in other case need to use numpy.

```python 
if len(a) <= 16_000_000 and isinstance(axis, int) and a.dtype in [torch.float32, torch.float64]:
    return torch.quantile(...).to(torch.float64)
return torch.tensor(np.quantile(...)).to(device)
```

AlexanderDokuchaev (changes requested):

nncf/experimental/tensor/functions/torch_numeric.py
daniil-lyakhov: I hope it won't broke quantization parameters precision 🙏

daniil-lyakhov: Done

AlexanderDokuchaev (approved):

PR #2422, opened 2024-01-24
[PT] check only major and minor version

AlexanderDokuchaev:
### Changes

- Use `SpecifierSet` to check version of PT and TF
- Remove unused `BKC_TORCHVISION_VERSION`
- Fix strict check version of TF that never raised before


nncf/version.py
```diff
@@ -11,6 +11,6 @@
 
 __version__ = "2.7.0"
 
-BKC_TORCH_VERSION = "2.1.0"
+BKC_TORCH_VERSION = "2.1.2"
 BKC_TORCHVISION_VERSION = "0.16.0"
```
alexsu52, line 15:
```suggestion
BKC_TORCHVISION_VERSION = "0.16.2"
```

nncf/version.py
```diff
@@ -11,6 +11,6 @@
 
 __version__ = "2.7.0"
 
-BKC_TORCH_VERSION = "2.1.0"
+BKC_TORCH_VERSION = "2.1.2"
```
alexsu52, line 14:
```suggestion
BKC_TORCH_VERSION = "2.1.*"
```
What do you think?

AlexanderDokuchaev: `from packaging import version` does not support `*` in version.
`LegacyVersion` and `LegacySpecifier` that support it was deprecated and removed.

Think that using version without wild card it's more correct way.

alexsu52: I'm trying to avoid misunderstanding here. You should keep in mind that patch version is not checked. Could you suggest explicitly solution? For example, Using `BKC_TORCH_VERSION = "2.1"` if nncf decelerates support torch~=2.1.0?

My general concern is following. If you need to specify the torch version with exactly patch you will have to fix code again.

alexsu52 (approved):

nncf/tensorflow/tf_internals.py
```diff
@@ -18,7 +18,7 @@
 
 from nncf.tensorflow import tensorflow_version
 
-if version.parse(tensorflow_version) < version.parse("2.13"):
+if tensorflow_version < version.parse("2.13"):
```
KodiaqQ, line 21: Why can't we fix `2.13` somewhere as a variable instead of spreading hard-coded versions everywhere?

setup.py
```diff
@@ -128,7 +128,7 @@ def find_version(*file_paths):
 ]
 
 TORCH_EXTRAS = [
-    "torch>=2.0,<2.2;python_version < '3.11'",
+    "torch==2.1.2;python_version < '3.11'",
```
KodiaqQ, line 131: Why not `2.1.*`?

nncf/version.py
```diff
@@ -11,6 +11,6 @@
 
 __version__ = "2.9.0"
 
-BKC_TORCH_VERSION = "2.1.2"
-BKC_TORCHVISION_VERSION = "0.16.2"
-BKC_TF_VERSION = "2.12.*"
+BKC_TORCH_SPEC = "==2.1.*"
+BKC_TF_SPEC = "==2.12.*"
+STRICT_TF_SPEC = ">=2.8.4,<2.14.0"
```
KodiaqQ, line 16: Why the `docs/Installation.md` and `README.md` weren't updated?

nncf/tensorflow/tf_internals.py
AlexanderDokuchaev: What are you suggesting?

setup.py
AlexanderDokuchaev: To have fixed version in tests.

nncf/version.py
AlexanderDokuchaev: What was changed to update md files?

KodiaqQ (approved):

PR #2431, opened 2024-01-26
Avoiding overflow on summing 32-bit int on Windows

ljaljushkin:
### Changes

Explicitly use unsigned 64-bit integer for number of weight in weight compression.

### Reason for changes

The overflow is happening on Windows, when 32-bit integers are summed up: 
image
it leads to incorrect mixed precision selection.

### Related tickets

bug with chatglm3 on Windows - 130819

### Tests

- [x] manual check

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/config.py
+++ b/nncf/quantization/algorithms/weight_compression/config.py
@@ -11,6 +11,8 @@
 from dataclasses import dataclass
 from typing import Optional, Tuple, TypeVar
 
+import numpy as np
+
 from nncf.common.graph.graph import NNCFNode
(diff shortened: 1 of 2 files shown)


andreyanufr (approved):

alexsu52 (commented):
What do you think about adding a test?

ljaljushkin:
> What do you think about adding a test?
I found it's challenging to implement. It's reproduced in a specific Windows setup with a specific large 7B model. 
Any suggestions for the test?
The fix is relatively easy and was verified by the customer.

alexsu52 (approved):

PR #2438, opened 2024-01-31
[Fixed]: Issues resolved raised by mypy for issue Make #22313

AtharvaPore01:
### Changes

**Missing Type Annotations:**

- Many errors stem from functions or variables lacking explicit type annotations.

**Incorrect Function Signatures:**

- Some errors indicate mismatches between function arguments, return types, or overridden methods.

**List and Array Operations:**

- Errors involve incorrect usage of lists or arrays, such as incompatible types or invalid operations.

**Missing Return Statements:**

- Functions marked as returning a value must return something.

**Missing Type Annotations**

- A significant number of functions are missing type annotations. This is indicated by errors like `Function is missing a type annotation for one or more arguments [no-untyped-def]`.

**Incorrect or Missing Implementation of `__eq__` Method:**

- There were several errors related to the implementation of the `__eq__` method in different classes. The method's signature must accept an argument of type `object` and then check if the passed object is of the expected type.

**Type Argument Issues in Generic Types:**

- Errors like `Missing type parameters for generic type` suggest that you are using generic types (like `ndarray`, `deque`, `List`) without specifying their type parameters.

**Incompatible Types and Assignments:**

- There were various errors where the types of variables or return types didn't match what was expected. This can be due to incorrect assignment or returning the wrong type from a function.

**Mypy-specific Issues and Recommendations:**

- Some errors were specific to mypy, a static type checker, which includes recommendations on how to make the code compliant with type checking, such as adhering to the expected method signatures in subclasses.

Miscellaneous Errors:

- These include issues like methods missing return type annotations (even when they don't return any value, `-> None` should be used), the use of a function as a type that is invalid, and a method in a subclass that does not return a value, contradicting the superclass's method signature.

**Other Issues:**

- A few errors relate to missing attributes or incorrect usage of specific functions.

### Reason for changes

As stated in issue [#22313](https://github.com/openvinotoolkit/nncf/issues/2493), the main reason behind solving the errors above as they were raised by running the mypy.ini file.

### Ticket Number: 
[#22313](https://github.com/openvinotoolkit/nncf/issues/2493)


nncf/common/graph/transformations/commands.py
```diff
@@ -140,6 +140,7 @@ def __init__(self, target_type: TargetType):
         :param target_type: Type of the target point.
         """
         self._target_type = target_type
+        self.target_node_name: str = ""
```
daniil-lyakhov, line 143: I suggest to add the `target_node_name` to the initialization parameters list as all inherited classes requires this parameter. Don't forget to update the docstring

nncf/common/tensor_statistics/aggregator.py
```diff
@@ -50,21 +49,15 @@ def collect_statistics(self, model: TModel, graph: NNCFGraph) -> None:
 
         merged_statistics = self._get_merged_statistic_points(self.statistic_points, model, graph)
         transformation_layout = self._get_transformation_layout_extra_outputs(merged_statistics)
-        model_with_outputs = model_transformer.transform(transformation_layout)
+        model_with_outputs: ModelTransformer = model_transformer.transform(transformation_layout)
```
daniil-lyakhov, line 52:
```suggestion
        model_with_outputs: TModel = model_transformer.transform(transformation_layout)
```

nncf/common/tensor_statistics/aggregator.py
```diff
@@ -87,7 +80,10 @@ def register_statistic_points(self, statistic_points: StatisticPointsContainer)
 
         for _, _statistic_points in self.statistic_points.items():
             for _statistic_point in _statistic_points:
-                for _, tensor_collectors in _statistic_point.algorithm_to_tensor_collectors.items():
+                for (
+                    _,
+                    tensor_collectors,
+                ) in _statistic_point.algorithm_to_tensor_collectors.items():
```
daniil-lyakhov, line 86:
```suggestion
                for tensor_collectors in _statistic_point.algorithm_to_tensor_collectors.values():
```

nncf/common/tensor_statistics/aggregator.py
```diff
@@ -50,21 +49,15 @@ def collect_statistics(self, model: TModel, graph: NNCFGraph) -> None:
...
-        model_with_outputs = model_transformer.transform(transformation_layout)
+        model_with_outputs: ModelTransformer = model_transformer.transform(transformation_layout)
         engine = factory.EngineFactory.create(model_with_outputs)
 
-        dataset_length = self.dataset.get_length()
-        total = (
-            min(dataset_length or self.stat_subset_size, self.stat_subset_size)
-            if self.stat_subset_size is not None
-            else None
-        )
```
daniil-lyakhov, line 61: There was logic before: if `self.stat_subset_size` is None then subset size is equal to the length of the whole dataset. As default value is 0 for `self.stat_subset_size` and this check is gone, I don't see the place where this logic is implemented. Could you please explain where is it?

nncf/common/tensor_statistics/collectors.py
```diff
@@ -12,22 +12,23 @@
...
 
-ReductionAxes = Tuple[int]
+ReductionAxes = Tuple[int, ...]
 
 
 class TensorStatisticCollectorBase(ABC):
     """Collector estimate statistics at the quantization point based on the provided reduction shape."""
 
-    def __init__(self, reduction_shape: Optional[ReductionAxes] = None, num_samples: Optional[int] = None):
+    def __init__(self, reduction_shape: ReductionAxes = None, num_samples: Optional[int] = None):
```
daniil-lyakhov, line 31: Why `Optional` gone?

nncf/common/tensor_statistics/aggregator.py
AtharvaPore01: The logic to set `self.stat_subset_size` to the length of the whole dataset if it is None is not explicitly present. However, there are steps taken during the registration of statistic points that indirectly handle the calculation of `self.stat_subset_size`.

Screenshot 2024-02-12 at 3 44 35 PM

In this part of the code, it iterates over the registered statistic points and their associated tensor collectors. For each tensor collector, it checks if `self.stat_subset_size` is None. If it is None, it sets `self.stat_subset_size` to the num_samples attribute of the tensor collector.
If `self.stat_subset_size` is not None, it compares it with the num_samples attribute of the tensor collector and updates `self.stat_subset_size` to be the maximum of the two values.
This logic effectively sets `self.stat_subset_size` based on the maximum number of samples encountered among all tensor collectors associated with the registered statistic points. While it doesn't directly set it to the length of the whole dataset if `self.stat_subset_size` is None, it achieves a similar effect by considering the maximum number of samples encountered during registration.

nncf/common/tensor_statistics/aggregator.py
```diff
@@ -87,7 +79,7 @@ def register_statistic_points(self, statistic_points: StatisticPointsContainer)
 
         for _, _statistic_points in self.statistic_points.items():
             for _statistic_point in _statistic_points:
-                for _, tensor_collectors in _statistic_point.algorithm_to_tensor_collectors.items():
+                for tensor_collectors in _statistic_point.algorithm_to_tensor_collectors.values():
                     for tensor_collector in tensor_collectors:
```
daniil-lyakhov, line 83: Is this a real issue?

nncf/common/tensor_statistics/collectors.py
```diff
@@ -40,10 +45,10 @@ def __init__(self, reduction_shape: Optional[ReductionAxes] = None, num_samples:
         self._num_samples = num_samples
 
     @property
-    def num_samples(self) -> int:
+    def num_samples(self) -> Optional[int]:
         return self._num_samples
 
-    def register_input(self, x: TensorType) -> TensorType:
+    def register_input(self, x: torch.Tensor) -> torch.Tensor:
```
daniil-lyakhov, line 51: torch.Tensor in common code, common code could use only common types

nncf/common/tensor_statistics/collectors.py
```diff
@@ -266,7 +298,7 @@ def squeeze(x: NNCFTensor, dim: Optional[Union[int, Tuple[int, ...]]] = None) ->
 
     @staticmethod
     @abstractmethod
-    def sum(tensor: NNCFTensor) -> TensorElementsType:
+    def sum(tensor: TensorElementsType) -> TensorElementsType:
```
daniil-lyakhov, line 301:
```suggestion
    def sum(tensor: NNCFTensor) -> TensorElementsType:
```

nncf/common/tensor_statistics/collectors.py
```diff
@@ -406,20 +440,21 @@ def __init__(self, use_abs_max: bool, reduction_shape: ReductionAxes, num_sample
...
-        min_reduced = self._tensor_processor.reduce_min(x, self._reduction_shape)
-        if self._use_abs_max:
+    def _register_input_common(self, x: NNCFTensor) -> None:
+        if self._tensor_processor is not None:
+            min_reduced: None = self._tensor_processor.reduce_min(x, self._reduction_shape)
+        if self._use_abs_max and self._tensor_processor is not None:
             x = self._tensor_processor.abs(x)
-        max_reduced = self._tensor_processor.reduce_max(x, self._reduction_shape)
+        if self._tensor_processor is not None:
+            max_reduced: None = self._tensor_processor.reduce_max(x, self._reduction_shape)
```
daniil-lyakhov, line 457: Please refactor it. And please remove None typehint

nncf/common/tensor_statistics/collectors.py
```diff
@@ -455,36 +490,38 @@ def __init__(
...
+        if self._tensor_processor is not None:
+            min_reduced: int = self._tensor_processor.reduce_min(x, self._reduction_shape)
+        if self._use_abs_max and self._tensor_processor is not None:
             x = self._tensor_processor.abs(x)
-        max_reduced = self._tensor_processor.reduce_max(x, self._reduction_shape)
+        if self._tensor_processor is not None:
+            max_reduced: int = self._tensor_processor.reduce_max(x, self._reduction_shape)
 
-        if self._use_per_sample_stats:
+        if self._use_per_sample_stats and self._tensor_processor is not None:
```
daniil-lyakhov, line 509: Please refactor it. And please remove None typehint

nncf/common/tensor_statistics/collectors.py
```diff
@@ -508,39 +545,52 @@ def __init__(
...
-        stacked_max = self._tensor_processor.stack(self._all_max_values)
-        if self._use_means_of_maxs:
+    def _max_aggregate(self) -> None:
+        if self._tensor_processor is not None:
+            stacked_max = self._tensor_processor.stack(self._all_max_values)
+        if self._use_means_of_maxs and self._tensor_processor is not None:
             return self._tensor_processor.mean(stacked_max, axis=0)
-        return self._tensor_processor.reduce_max(stacked_max, axis=0)
+        if self._tensor_processor is not None:
+            return self._tensor_processor.reduce_max(stacked_max, axis=0)
```
daniil-lyakhov, line 562: Please refactor it. And please remove None typehint

nncf/common/tensor_statistics/collectors.py
```diff
@@ -508,39 +545,52 @@ def __init__(
...
     """
 
-    def _min_aggregate(self):
-        stacked_min = self._tensor_processor.stack(self._all_min_values)
-        return self._tensor_processor.mean(stacked_min, axis=0)
+    def _min_aggregate(self) -> None:
+        if self._tensor_processor is not None:
+            stacked_min = self._tensor_processor.stack(self._all_min_values)
+        if self._tensor_processor is not None:
+            return self._tensor_processor.mean(stacked_min, axis=0)
```
daniil-lyakhov, line 574: I think this `if self._tensor_processor is not None:` should be done in a different place

nncf/common/tensor_statistics/aggregator.py
daniil-lyakhov: Got you. But despite the fact that this should work in every case, I prefer explicit double check

mlukasze:
hey @AtharvaPore01 
do you have time to finish this PR? Or should we move it back to the pool?

AtharvaPore01:
@mlukasze  Hi, Sorry for the delay from my side. I appreciate your patience. I'm currently tied up with some urgent tasks but will prioritize completing the PR as soon as possible. Your understanding is greatly appreciated.

MaximProshin:
@AtharvaPore01 , do you plan to continue to work on this PR?

PR #2444, opened 2024-02-01
[PTQ][OV] Force renaming for friendly names duplicates

KodiaqQ:
### Changes

- Added the method for OpenVINO backend to rename friendly names duplicates;

### Reason for changes

- Robust friendly names usage in NNCF;

### Related tickets

- 131073

### Tests

- Updated `tests/openvino/native/test_model_utils.py` with the test;


nncf/openvino/quantization/quantize_model.py
```diff
@@ -308,6 +309,8 @@ def quantize_impl(
     """
     Implementation of the `quantize()` method for the OpenVINO backend.
     """
+    model = remove_friendly_name_duplicates(model)
```
KodiaqQ, line 312: @kshpv, could you please ask if we should also check duplicates for the **if** bodies?

nncf/openvino/graph/model_utils.py
```diff
@@ -60,3 +60,20 @@ def get_start_nodes_for_activation_path_tracing(nncf_graph: NNCFGraph) -> List[N
...
+    Removes diplicates of node names (friendly_name attribute) in the model.
+
+    :param model: ov.Model instance to update.
+    :return: Updated ov.Model without duplicated names.
+    """
+    existing_names = set()
+    for op in model.get_ops():
+        friendly_name = op.get_friendly_name()
+        if friendly_name in existing_names:
+            friendly_name = friendly_name + "0"
```
andrey-churkin, line 76: Is there such a case when we have three operations in the model with the same friendly name?

nncf/openvino/quantization/quantize_model.py
kshpv: Ask whom?

KodiaqQ: Could you please say if should we check duplicates for the **if** bodies?

nncf/openvino/quantization/quantize_model.py
```diff
@@ -362,6 +365,8 @@ def quantize_with_accuracy_control_impl(
     """
     Implementation of the `quantize_with_accuracy_control()` method for the OpenVINO backend.
     """
+    model = remove_friendly_name_duplicates(model)
```
kshpv, line 368: Do you need this call?

kshpv (changes requested):

nncf/openvino/quantization/quantize_model.py
KodiaqQ: Yes, we need it. Why not?
We have the `quantize_with_tune_hyperparams` method inside the `quantize_with_accuracy_control_impl` along with `quantize_impl` that uses `model`.

kshpv: Does it mean that for AAQ there are many calls of `remove_friendly_name_duplicates` that does nothing? As at the first step it is called inside `quantize_with_accuracy_control_impl` and then inside `quantize_impl`

kshpv:
Do you need to add `remove_friendly_name_duplicates` to compress weights algo?

nncf/openvino/quantization/quantize_model.py
KodiaqQ: Yes, it is. Do you have any other solution that would be more efficient in our pipeline?

KodiaqQ:
> `remove_friendly_name_duplicates`

Good catch! Updated.

KodiaqQ:
> Do you need to add `remove_friendly_name_duplicates` to compress weights algo?

Good catch! Updated.

nncf/openvino/quantization/quantize_model.py
kshpv: Not really. If there is no easy solution without refactoring all NNCF I would just drop a comment- something like: that there is overhead in AAQ of call `remove_friendly_name_duplicates`.

KodiaqQ:
@alexsu52, @kshpv, @andrey-churkin, see the results of the local AAQ testing in the ticket.

nncf/openvino/graph/model_utils.py
```diff
@@ -62,6 +62,29 @@ def get_start_nodes_for_activation_path_tracing(nncf_graph: NNCFGraph) -> List[N
...
 
 
+def remove_friendly_name_duplicates(model: ov.Model) -> ov.Model:
+    """
+    Removes diplicates of node names (friendly_name attribute) in the model.
+
+    :param model: ov.Model instance to update.
+    :return: Updated ov.Model without duplicated names.
+    """
+    rt_info_path = ["nncf", "friendly_names_were_updated"]
```
alexsu52, line 72: it looks like idea with using `rt_info` will not work without copying `rt_info` in the model transformer. Could you check?

In general it would be a good idea not to lose the rt_info of the original model in the quantized model. I think this statement can be discussed offline in the team.

KodiaqQ: Generally, our `ov.Model` creations from the `ModelTransformer` are isolated from the main model. But I've added `rt_info` copying as well for consistency.
`rt_info` can't be lost during the pipeline since we use the `ov.Model.clone()` method and we hope it correctly works with `rt_info`.

alexsu52 (approved):
Thanks for the contribution!

nncf/openvino/graph/model_transformer.py
```diff
@@ -206,9 +206,14 @@ def _insert_outputs(model: ov.Model, outputs: List[Tuple[ov.Output, int, Callabl
             OVModelTransformer._update_tensor_name([result.get_output_tensor(0)], result_name)
             extra_model_outputs.append(result)
 
-        return ov.Model(
+        model_with_outputs = ov.Model(
             results=results + extra_model_outputs, sinks=assign_ops, parameters=params, name=model.friendly_name
         )
+        rt_info_path = ["nncf"]
+        if model.has_rt_info(rt_info_path):
```
kshpv, line 213: I could suggest to implement the special function for that

nncf/openvino/graph/model_utils.py
```diff
@@ -62,6 +62,29 @@ def get_start_nodes_for_activation_path_tracing(nncf_graph: NNCFGraph) -> List[N
...
+    friendly_names_flag = str("True")
```
kshpv, line 73:
```suggestion
    friendly_names_flag = "True"
```

kshpv (changes requested):

kshpv (approved):

andrey-churkin (approved):

PR #2446, opened 2024-02-05
[PT][WC] Add WC tests for functional models

l-bat:
### Changes

Fixed bug with `channel_idx` for Torch Conv metatypes
Fixed bug in `get_module_by_name()` if base model has weights as a `torch.nn.Parameter`

### Related tickets

124822

### Tests

Added `test_compress_weights_functional_model`


alexsu52 (approved):
Thanks for the contribution!

PR #2447, opened 2024-02-05
Add AWQ to LLM hyperparams tuning example

l-bat:
### Changes

Add AWQ parameter to to LLM hyperparams tuning example
test_examples job: 228

### Tests

Updated `tests/cross_fw/examples/run_example.py::llm_tune_params`


AlexKoff88:
@andreyanufr, @ljaljushkin please take a look as well.

ljaljushkin:
What about considering data-aware without awq, first? 
Conformance test shows it gives a better similarity (group size and ratio are the same) and works faster. 
image

ljaljushkin:
> Conformance test shows data-aware without awq gives a better similarity (group size and ratio are the same) and works faster. I wonder why do you have the opposite results. `group_size=64` `ratio=0.8` `mode=int4_sym` `tinyllama/tinyllama-1.1b-step-50k-105b` image

@l-bat 
do you have logs with similarity per configuration? Could you please post it there?

l-bat:
TinyLLama shows better results with `awq=True` for `ratio=1.0`, so the AWQ algo is enabled for further experiments.

ljaljushkin (approved):

PR #2448, opened 2024-02-06
nncf.AdvancedAccuracyRestorerParameters

alexsu52:
### Changes

nncf.AdvancedAccuracyRestorerParameters

### Reason for changes

Make it easier to create AdvancedAccuracyRestorerParameters

### Related tickets

N/A

### Tests

N/A

### DIFF
--- a/nncf/__init__.py
+++ b/nncf/__init__.py
@@ -42,6 +42,9 @@ from nncf.quantization import QuantizationPreset as QuantizationPreset
 from nncf.quantization import compress_weights as compress_weights
 from nncf.quantization import quantize as quantize
 from nncf.quantization import quantize_with_accuracy_control as quantize_with_accuracy_control
+from nncf.quantization.advanced_parameters import (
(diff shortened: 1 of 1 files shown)


nncf/__init__.py
```diff
@@ -42,6 +42,9 @@
 from nncf.quantization import compress_weights as compress_weights
 from nncf.quantization import quantize as quantize
 from nncf.quantization import quantize_with_accuracy_control as quantize_with_accuracy_control
+from nncf.quantization.advanced_parameters import (
+    AdvancedAccuracyRestorerParameters as AdvancedAccuracyRestorerParameters,
+)
```
KodiaqQ, line 47: Shouldn't we introduce an example with these parameters then?

alexsu52: Thanks for the comment. The ticket 131817 was already filed.

KodiaqQ: Great!

KodiaqQ (approved):

PR #2420, opened 2024-01-24
Using FP32 precision to calculate compressed weights

alexsu52:
### Changes

Using FP32 precision to calculate compressed weights

### Reason for changes

Time Optimization of weight compression.
Models | Weight precision | Task | 0fdac3dc131facb7353e61a9282454253c82ae6e | PR | Speed up
-- | -- | -- | -- | -- | --
  |   |   | Time (sec) | Time (sec) |  
llama-2-7b-chat-hf | FP16 | int8_asym | 239,36 | 59,98 | 3,990713571
llama-2-7b-chat-hf | FP16 | int4_sym_g128_r80 | 633,45 | 279,41 | 2,267103064
llama-2-7b-chat-hf | FP32 | int8_asym | 74,22 | 44,81 | 1,656215131
llama-2-7b-chat-hf | FP32 | int4_sym_g128_r80 | 315,83 | 243,76 | 1,295663768

### Related tickets

130239

### Tests

N/A

PR #2458, opened 2024-02-08
More strict threshold for accuracy and minor corrections

ljaljushkin:
### Changes

Introduced a more strict criterion for passing conformance tests for weight compression: similarity should match at least to the fifth decimal place.

Also use a specific dataset revision to avoid issues with different accuracy

### Reason for changes

in order to catch regressions - e.g., https://github.com/openvinotoolkit/nncf/pull/2420

### Related tickets

n/a

### Tests

- [x] weight compression conformance test build 2

image

### DIFF
--- a/tests/post_training/test_quantize_conformance.py
+++ b/tests/post_training/test_quantize_conformance.py
@@ -85,7 +85,7 @@ def fixture_wc_reference_data():
         data = yaml.safe_load(f)
         fp32_test_cases = defaultdict(dict)
         for test_case_name in data:
-            data[test_case_name]['atol'] = 1e-5
+            data[test_case_name]["atol"] = 1e-5
             model_name = test_case_name.split("_backend_")[0]
             fp32_test_cases[f"{model_name}_backend_FP32"]["metric_value"] = 1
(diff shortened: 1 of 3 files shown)


tests/post_training/test_quantize_conformance.py
```diff
@@ -85,8 +85,10 @@ def fixture_wc_reference_data():
         data = yaml.safe_load(f)
         fp32_test_cases = defaultdict(dict)
         for test_case_name in data:
+            data[test_case_name]["atol"] = 1e-5
```
AlexanderDokuchaev, line 88:
```suggestion
            data[test_case_name]["atol"] = data[test_case_name].get("atol", 1e-5)
```
To have option to change from json
and do same for fp32 in 90 and 91 lines.

AlexanderDokuchaev (changes requested):

tests/post_training/test_quantize_conformance.py
ljaljushkin: thanks, corrected

AlexanderDokuchaev (approved):

PR #2467, opened 2024-02-14
[PT] Propagate is_shared attribute from constant node

AlexanderDokuchaev:
### Changes

- If model traced with trace_parameters=True, detect shared constant nodes by number of output nodes and propagate this attribute to next nodes.

- Rename OP_NAMES_QUANTIZE_NODE to QUANTIZE_NODE_TYPES
### Reason for changes

Incorrect set `is_shared` attributes. 

### Tests

test_shared_nodes_in_wrapped_model_with_trace_parameters

### DIFF
--- a/nncf/torch/graph/graph_builder.py
+++ b/nncf/torch/graph/graph_builder.py
@@ -23,8 +23,8 @@ from nncf.torch.dynamic_graph.graph_tracer import GraphTracer
 from nncf.torch.dynamic_graph.layer_attributes_handlers import set_nodes_attributes_in_nncf_graph
 from nncf.torch.dynamic_graph.scope import Scope
 from nncf.torch.graph.graph import PTNNCFGraph
-from nncf.torch.graph.operator_metatypes import OP_NAMES_QUANTIZE_NODE
 from nncf.torch.graph.operator_metatypes import PT_OPERATOR_METATYPES
+from nncf.torch.graph.operator_metatypes import QUANTIZE_NODE_TYPES
 
 
 class GraphBuilder:
@@ -152,7 +152,7 @@ def _propagate_true_for_is_shared_attribute(node: NNCFNode, graph: NNCFGraph):
     :param graph: The NNCFGraph instance.
     """
     node.attributes[NNCFNode.IS_SHARED_ATTR] = True
-    if node.metatype in CONST_NOOP_METATYPES or node.node_type in OP_NAMES_QUANTIZE_NODE:
+    if node.metatype in CONST_NOOP_METATYPES or node.node_type in QUANTIZE_NODE_TYPES:
         for next_node in graph.get_next_nodes(node):
             _propagate_true_for_is_shared_attribute(next_node, graph)
 
--- a/nncf/torch/graph/graph_builder.py
+++ b/nncf/torch/graph/graph_builder.py
@@ -9,10 +9,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from collections import defaultdict
-from typing import Any, Callable, Dict, Optional, Set
+from typing import Any, Callable, Dict, List, Optional, Set
 
 import torch
 
+from nncf.common.graph import NNCFGraph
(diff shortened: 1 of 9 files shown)


AlexanderDokuchaev:
For example class with shared nodes

```python 
class SharedConv(nn.Module):
    INPUT_SIZE = [1, 1, 4, 4]
 
    def __init__(self):
        super().__init__()
        self.conv = create_conv(1, 1, 2, 2)
 
    def forward(self, x):
        a = self.conv(x)
        b = self.conv(x + 1)
        return a + b
 ```

Current: 
| is_shared | node_name | 
| --- | --- |
| True | /nncf_model_input_0 | 
| True | conv.weight | 
| True | conv.bias | 
| True | SharedConv/Conv2d[conv]/conv2d_0 | 
| True | SharedConv/__add___0 | 
| True | SharedConv/Conv2d[conv]/conv2d_1 | 
| True | SharedConv/__add___1 | 
| True | /nncf_model_output_0 | 

PR: 
| is_shared | node_name | 
| --- | --- |
| False | /nncf_model_input_0 | 
| True | conv.weight | 
| True | conv.bias | 
| True | SharedConv/Conv2d[conv]/conv2d_0 | 
| False | SharedConv/__add___0 | 
| True | SharedConv/Conv2d[conv]/conv2d_1 | 
| False | SharedConv/__add___1 | 
| False | /nncf_model_output_0 |

File: nncf/torch/graph/graph_builder.py [hunk 1]
```diff
@@ -134,4 +139,34 @@ def convert(dynamic_graph: DynamicGraph) -> PTNNCFGraph:
...
+
+
+def _propagate_true_for_is_shared_attribute(node: NNCFNode, graph: NNCFGraph):
+    """
+    Propagates the is_shared attribute through specific nodes in an NNCFGraph.
+
+    :param node: The start NNCFNode to process.
+    :param graph: The NNCFGraph instance.
+    """
+    node.attributes[NNCFNode.IS_SHARED_ATTR] = True
```
daniil-lyakhov, line 154: What do you think about making `NNCFNode.IS_SHARED_ATTR` an a parameter of the function?

nncf/torch/graph/graph_builder.py
```diff
@@ -134,4 +139,34 @@ def convert(dynamic_graph: DynamicGraph) -> PTNNCFGraph:
             )
 
         set_nodes_attributes_in_nncf_graph(nncf_graph)
+        # if traced_parameters:
+        #     propagate_is_shared_attribute_from_constant_nodes(nncf_graph)
         return nncf_graph
+
+
+def _propagate_true_for_is_shared_attribute(node: NNCFNode, graph: NNCFGraph):
```
daniil-lyakhov, line 147:
```suggestion
def _propagate_true_for_is_shared_attribute(node: NNCFNode, graph: NNCFGraph) -> None:
```

nncf/torch/graph/graph_builder.py
AlexanderDokuchaev: done

AlexanderDokuchaev: added `val` args

nncf/torch/graph/graph_builder.py
AlexanderDokuchaev: Added, but it's not necessary, None is default for return type.

nncf/torch/graph/graph_builder.py
daniil-lyakhov: I mean making the node attribute a parameter :) Like _Propagate_attribute(node, graph, True, NNCFNode.IS_SHAED_ATTR)

AlexanderDokuchaev: Because for another attributes condition to propagation can be different.
This function can correctly propagate only is_shared attribute

daniil-lyakhov (approved):

alexsu52 (approved):
LGTM

PR #2472, opened 2024-02-14
Remove friendly names duplicates after model weight compression

nikita-savelyevv:
### Changes

It was found that model weight compression may sometimes (randomly) result in a model with duplicate friendly names. 

With hybrid quantization pipeline we quantize a model that was compressed beforehand. In such case the compressed model may contain duplicate friendly names, but they will not be removed at quantization start because `nncf.friendly_names_were_updated` flag was set at the start of compression. 

Changes:
- Remove duplicates after model weight compression
- Don't skip duplicates removal if it was applied before (the model may have changed since the last removal)

### Related tickets
132595

### DIFF
--- a/nncf/openvino/graph/model_utils.py
+++ b/nncf/openvino/graph/model_utils.py
@@ -64,7 +64,7 @@ def get_start_nodes_for_activation_path_tracing(nncf_graph: NNCFGraph) -> List[N
 
 def remove_friendly_name_duplicates(model: ov.Model) -> ov.Model:
     """
-    Removes diplicates of node names (friendly_name attribute) in the model.
+    Removes duplicates of node names (friendly_name attribute) in the model.
 
     :param model: ov.Model instance to update.
     :return: Updated ov.Model without duplicated names.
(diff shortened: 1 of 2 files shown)


nncf/openvino/graph/model_utils.py
```diff
@@ -64,24 +64,18 @@ def get_start_nodes_for_activation_path_tracing(nncf_graph: NNCFGraph) -> List[N
...
+    Removes duplicates of node names (friendly_name attribute) in the model.
 
     :param model: ov.Model instance to update.
     :return: Updated ov.Model without duplicated names.
     """
-    rt_info_path = ["nncf", "friendly_names_were_updated"]
-    friendly_names_flag = "True"
-    if model.has_rt_info(rt_info_path) and model.get_rt_info(rt_info_path).value == friendly_names_flag:
-        return model
-
```
KodiaqQ, line 76: We should not remove this due to performance issues.
We should fix the original reason for the duplicate's existence.
cc @alexsu52

KodiaqQ (changes requested):

nncf/openvino/graph/model_utils.py
nikita-savelyevv: > We should not remove this due to performance issues.
I don't see any performance issues. According to my tests the extra for loops takes about 1 ms per 1000 model nodes. 

> We should fix the original reason for the duplicate's existence.
I agree, but this is a temporary workaround while the investigation is in progress.

nikita-savelyevv:
A proper solution was found https://github.com/openvinotoolkit/nncf/pull/2475. This WA is no longer needed.

PR #2473, opened 2024-02-14
[PT] Do not trace parameters for modules inserted by NNCF

AlexanderDokuchaev:
### Changes

Do not trace parameters for modules inserted by NNCF

### Reason for changes

Reduce number of nodes in NNCFGraph, this parameters will not used to anilazine graph or insert points. 

### Tests

test_not_trace_parameters_in_nncf_modules

### DIFF
--- a/nncf/torch/dynamic_graph/wrappers.py
+++ b/nncf/torch/dynamic_graph/wrappers.py
@@ -278,15 +278,10 @@ def wrap_parameters(model: torch.nn.Module):
 
     :param model: A model.
     """
-    from nncf.torch.external_hook import EXTERNAL_OP_STORAGE_PREFIX
-    from nncf.torch.quantization.external_quantizer import EXTERNAL_QUANTIZERS_STORAGE_PREFIX
-
-    # Ignore modules that will be added by NNCF
-    ignored_prefixes = [EXTERNAL_QUANTIZERS_STORAGE_PREFIX, EXTERNAL_OP_STORAGE_PREFIX]
-
     ctx = get_current_context()
     for name, param in model.named_parameters():
-        if any(name.startswith(ignore_prefix) for ignore_prefix in ignored_prefixes):
+        if name.startswith("_nncf"):
+            # Exclude parameters in modules which added by NNCF.
             continue
         is_reused = name in ctx.reused_parameters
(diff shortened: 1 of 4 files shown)


daniil-lyakhov (approved):

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -278,8 +278,16 @@ def wrap_parameters(model: torch.nn.Module):
 
     :param model: A model.
     """
+    from nncf.torch.external_hook import EXTERNAL_OP_STORAGE_PREFIX
```
alexsu52, line 281: Сan you move these imports to the beginning of the file?

AlexanderDokuchaev: no due to cyclical imports

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -278,8 +278,16 @@ def wrap_parameters(model: torch.nn.Module):
...
+    from nncf.torch.quantization.external_quantizer import EXTERNAL_QUANTIZERS_STORAGE_PREFIX
+
+    # Ignore modules that will be added by NNCF
+    ignored_prefixes = [EXTERNAL_QUANTIZERS_STORAGE_PREFIX, EXTERNAL_OP_STORAGE_PREFIX]
+
     ctx = get_current_context()
     for name, param in model.named_parameters():
+        if any(name.startswith(ignore_prefix) for ignore_prefix in ignored_prefixes):
```
alexsu52, line 289: Could you think of a way to get this information from the model?

AlexanderDokuchaev: Looks like no.
It can be done by excluding all parameters in model.nncf, but in this case better to exclude by `name.startswith("_nncf")` condition. but in this case

alexsu52: It make sense to exclude all parameters in model.nncf. Could you implement it?

AlexanderDokuchaev: done

alexsu52 (approved):
LGTM

PR #2481, opened 2024-02-16
[PT] Insert point for not replaced modules

AlexanderDokuchaev:
### Changes

Add option to insert point for models that wrapped with replace_modules=False


alexsu52 (approved):
LGTM

daniil-lyakhov (approved):

PR #2490, opened 2024-02-20
Added deprecation warning

KodiaqQ:
### Changes

- Added the simple deprecation warning for the **binarization** algorithm;

### Reason for changes

- Binarization deprecation;

### Related tickets

- 133129

### Tests

- N/A


alexsu52 (approved):
LGTM

MaximProshin:
@alexsu52 @KodiaqQ I think we don't need it actually because we deprecated it in 2.9 and plan to remove in 2.10

KodiaqQ:
> @alexsu52 @KodiaqQ I think we don't need it actually because we deprecated it in 2.9 and plan to remove in 2.10

Got it. After the 2.9.0 release, I'll drop this from the develop.

PR #2506, opened 2024-02-21
[PTQ] Metatypes with bias list alignment

KodiaqQ:
### Changes

- Reduced metatypes in `OPERATIONS_WITH_BIAS` list for ONNX;
- Reduced metatypes in `OPERATIONS_WITH_BIAS` list for PyTorch;
- Separated GroupConvolution from the Convolutions on ONNX;

### Reason for changes

- Alignment for algorithms between backends;

### Related tickets

- 133198
- 104166
- 106469

### Tests

- post_training_quantization/323 - failed with regressions for ONNX (as expected)
- post_training_quantization/324 - passed (updated references)

### DIFF
--- a/nncf/onnx/graph/metatypes/onnx_metatypes.py
+++ b/nncf/onnx/graph/metatypes/onnx_metatypes.py
@@ -9,6 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import deque
 from typing import Dict, List, Optional, Type
 
 import onnx
@@ -44,14 +45,15 @@ class ONNXOpMetatype(OperatorMetatype):
     @classmethod
     def determine_subtype(cls, model: onnx.ModelProto, node: onnx.NodeProto) -> Optional[Type[OperatorMetatype]]:
         matches = []
-        for subtype in cls.get_subtypes():
+        subtypes_list = deque(cls.get_subtypes())
+        while subtypes_list:
+            subtype = subtypes_list.popleft()
             if subtype.matches(model, node):
+                subtypes_list.extend(subtype.get_subtypes())
                 matches.append(subtype)
-        if len(matches) > 1:
-            raise nncf.InternalError("Multiple subtypes match operator call - cannot determine single subtype.")
         if not matches:
             return None
-        return matches[0]
+        return matches[-1]
 
 
 class ONNXOpWithWeightsMetatype(ONNXOpMetatype):
@@ -86,6 +88,22 @@ class ONNXDepthwiseConvolutionMetatype(ONNXOpWithWeightsMetatype):
 
 
 @ONNX_OPERATION_METATYPES.register()
+class ONNXGroupConvolutionMetatype(ONNXOpWithWeightsMetatype):
+    name = "GroupConvOp"
+    op_names = ["Conv"]
+    hw_config_names = [HWConfigOpName.CONVOLUTION]
+    weight_channel_axis = 0
+    weight_port_ids = [1]
+    bias_port_id = 2
+    output_channel_axis = 1
+    subtypes = [ONNXDepthwiseConvolutionMetatype]
+
+    @classmethod
+    def matches(cls, model: onnx.ModelProto, node: onnx.NodeProto) -> bool:
+        return _is_group_conv(node)
+
+
+@ONNX_OPERATION_METATYPES.register()
 class ONNXConvolutionMetatype(ONNXOpWithWeightsMetatype):
     name = "ConvOp"
     op_names = ["Conv"]
@@ -94,7 +112,7 @@ class ONNXConvolutionMetatype(ONNXOpWithWeightsMetatype):
     weight_port_ids = [1]
     bias_port_id = 2
     output_channel_axis = 1
-    subtypes = [ONNXDepthwiseConvolutionMetatype]
+    subtypes = [ONNXGroupConvolutionMetatype]
 
 
 @ONNX_OPERATION_METATYPES.register()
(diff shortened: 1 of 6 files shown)


nncf/torch/graph/operator_metatypes.py
```diff
@@ -1038,12 +1038,7 @@ def get_operator_metatypes() -> List[Type[OperatorMetatype]]:
     PTModuleConv1dMetatype,
     PTModuleConv2dMetatype,
     PTModuleConv3dMetatype,
-    PTDepthwiseConv1dSubtype,
-    PTDepthwiseConv2dSubtype,
-    PTDepthwiseConv3dSubtype,
-    PTModuleConvTranspose1dMetatype,
```
alexsu52, line 1044: It looks like `ConvTranspose` operations have similar semantic as `Convolution`. From BC perspective, what is strategy for it? Do you have a list of models where `ConvTranspose` operations are used?

Useful link: https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html

KodiaqQ: Since we have only conformance models scope for the PyTorch backend, the removal of the **ConvTranspose** doesn't affect the accuracy for any model here. So, we can decide that the models from the conformance scope don't contain **ConvTranspose** or this layer didn't work anyway, and was added with no reason and testing.
@AlexanderDokuchaev, could you please clarify?

AlexanderDokuchaev: What i should clarify? I have no list of used modules in test.
`ConvTranspose` was added as operation with bias...
But as i remember transpose convolution in OV has type `ConvolutionBackpropData`, that does not exist in OPERATIONS_WITH_BIAS for OV backend.

So, if transposed layers should not be supported it `ConvTranspose` can be removed.  In otherwise need to check that some model using `ConvTranspose` and add this kind of model to test scope if need.

And looks like `OPERATIONS_WITH_BIAS` is not correct name to current use, may be it should be `metatypes_for_fast_bias_correction`.

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -711,12 +745,9 @@ def _is_depthwise_conv(model: onnx.ModelProto, node: onnx.NodeProto) -> bool:
     :param node: Convolution node to check whether it is depthwise.
     :return: True if the convolution is depthwise, False - otherwise.
     """
-    conv_group = None
```
AlexanderDokuchaev, line 714: After changes `conv_group` used while definition is under if condition, as result possible that `conv_group` uses without definition and raise error.

nncf/openvino/graph/metatypes/groups.py
```diff
@@ -185,7 +185,6 @@
 # Contains the operation metatypes for which bias can be applied.
 OPERATIONS_WITH_BIAS = [
     ov_metatypes.OVConvolutionMetatype,
-    # TODO: add all metatypes with bias
```
AlexanderDokuchaev, line 188: Was added all metatypes with bias?

nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -85,6 +86,22 @@ def matches(cls, model: onnx.ModelProto, node: onnx.NodeProto) -> bool:
         return _is_depthwise_conv(model, node)
 
 
+@ONNX_OPERATION_METATYPES.register()
+class ONNXGroupConvolutionMetatype(ONNXOpWithWeightsMetatype):
```
AlexanderDokuchaev, line 90: Should be added groupconv metatypes in PT backend?

nncf/torch/graph/operator_metatypes.py
KodiaqQ: > What i should clarify?

Need to clarify why you've added these layers to the **OPERATORS_WITH_BIAS_METATYPES** list. **ConvolutionBackpropData** wasn't added to the OV backend for a reason: it possibly causes accuracy degradation.

nncf/openvino/graph/metatypes/groups.py
KodiaqQ: No. And they will not.

nncf/onnx/graph/metatypes/onnx_metatypes.py
KodiaqQ: If it is possible, we may extend metatypes in the PT backend. But not in this PR, I hope.

nncf/onnx/graph/metatypes/onnx_metatypes.py
KodiaqQ: Since **ONNXDepthwiseConvolutionMetatype** became the subtype of the **ONNXGroupConvolutionMetatype** that must have a _group_ attribute, it's not possible that _conv_group_ wouldn't be found.

nncf/onnx/graph/metatypes/onnx_metatypes.py
AlexanderDokuchaev: 135885

nncf/torch/graph/operator_metatypes.py
alexsu52: I hope, the comment will be resolved in 106469.

alexsu52 (approved):
LGTM

KodiaqQ:
@AlexanderDokuchaev, @kshpv, please, review.

kshpv (approved):

AlexanderDokuchaev (approved):

PR #2507, opened 2024-02-21
[PT] Return bias value after fusing layer for FBC

AlexanderDokuchaev:
### Changes

Return bias value after fusing layer, only for conv+bn (other layers does not supported by PT FBC). 
It's temporal fix, working with bias and extracted model will be reworked in 129581.

### Reason for changes

Incorrect calculation bias shift magnitude


KodiaqQ:
Do not forget to provide conformance run results.

KodiaqQ (approved):
Are there any conformance run results?

AlexanderDokuchaev:
manual/job/post_training_quantization/303
@KodiaqQ

PR #2518, opened 2024-02-26
[v.2.9.0] Added AC as dependency

KodiaqQ:
### Changes

- Added AC installation as a dependency for tests;

### Reason for changes

- AC was removed from the OV package.

### Related tickets

- 132634

### Tests

- N/A


alexsu52 (approved):
LGTM

PR #2520, opened 2024-02-26
Remove pytest dependency from run_example.py

daniil-lyakhov:
### Changes

pytest dependency is removed from run_example.py

### Reason for changes

Test example fails due to missing pytest dependency 

Found in test_examples/261/

### Tests
test_examples/262/


alexsu52 (approved):
LGTM

l-bat (approved):

PR #2522, opened 2024-02-27
PTQ quantization aware training documentation

daniil-lyakhov:
On top of #2515
### Changes

PTQ quantization aware training documentation

### Reason for changes

To improve docs coverage

### Related tickets

129584

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,8 @@ quantized_model = nncf.quantize(model, calibration_dataset)
 
 ```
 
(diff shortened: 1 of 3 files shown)


MaximProshin (commented):
Discussed offline: 
- add note to the Usage section with the ref to example
- update note in PTQ to clearly state PyTorch and remove typos
- remove note from QAT - this doc should be refactored and the note makes it even more confusing

MaximProshin (approved):

PR #2536, opened 2024-02-29
Fix remove_shapeof_subgraphs

andrey-churkin:
### Changes

img

### Reason for changes

The `remove_shapeof_subgraphs()` function removes some operations whose weights are quantized.

### Related tickets

Ref: 131843

### Tests

Current scope

### DIFF
--- a/nncf/quantization/passes.py
+++ b/nncf/quantization/passes.py
@@ -64,26 +64,26 @@ def remove_shapeof_subgraphs(
         if node.metatype in shapeof_metatypes:
             shape_of_nodes.append(node)
             continue
-        if node in infer_nodes:
+        if node.node_name in infer_nodes:
             continue
-        infer_nodes.append(node)
+        infer_nodes.append(node.node_name)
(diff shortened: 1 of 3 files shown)


nncf/quantization/passes.py
```diff
@@ -64,26 +64,26 @@ def remove_shapeof_subgraphs(
         if node.metatype in shapeof_metatypes:
             shape_of_nodes.append(node)
             continue
-        if node in infer_nodes:
+        if node.node_name in infer_nodes:
```
andrey-churkin, line 67: @KodiaqQ These changes are necessary because a node from quantized_model_graph contains NumPy arrays in its attributes. Therefore, it is not possible to perform the `NNCFNode.__eq__()` method due to this comparison
```python
self.attributes == other.attributes
```
I have not changed `NNCFNode.__eq__()` method because these attributes are custom (they were added by algorithm).

KodiaqQ (approved):

PR #2537, opened 2024-02-29
[WC] Align compression subgraphs for both weight input data types

nikita-savelyevv:
### Changes

Precision configuration of input OV model for weight compression can be one of the following three:
1. Weights and activations are in FP32 precision (model is saved with `compress_to_fp16=False`)
2. Weights are in FP16 and activations are in FP32 (model is saved with `compress_to_fp16=True`)
3. Weight and activations are in FP16 (e.g., PT model is first halfed and then converted to OV)

This PR make compression subgraphs equal for all these three cases. Compression activations are always executed in FP16. So for the first case an additional f16 -> f32 Convert node is added.

image


nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -153,19 +153,22 @@ def transform_model(
             compressed_const = opset.constant(
                 compressed_weight.tensor.data, dtype=compression_dtype, name=const_node_name
             )
-            converted_const = opset.convert(compressed_const, const_dtype)
+            converted_const = opset.convert(compressed_const, ov.Type.f32)
```
alexsu52, line 156: please, check your solution on the model with FP16 weights and activations.

alexsu52 (changes requested):
Testing models with different weight/inference precision are necessary.

nncf/quantization/algorithms/weight_compression/openvino_backend.py
alexsu52: Could you share compression subgraph for model with FP16 weights and activations?

nikita-savelyevv: Thanks, I've added the logic to correctly process the case of FP16 activations.

nikita-savelyevv:
WC manual test fails until #2569 is not merged.

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -153,51 +153,37 @@ def transform_model(
...
-            scale_const = opset.convert(scale_const, ov.Type.f32, name=f"{const_node_name}/scale_convert")
             mul = opset.multiply(
                 converted_const,
                 scale_const,
                 name=f"{const_node_name}/fq_weights_{wc_params.weight_port_id}",
             )
+            if const_dtype == ov.Type.f32:
+                mul = opset.convert(
+                    mul, ov.Type.f32, name=f"{mul.get_friendly_name()}/convert"
+                )
```
alexsu52: It looks like this code supports only fp32 and fp16. What will happen to other types?
```suggestion
            if const_dtype != ov.Type.f16:
                mul = opset.convert(
                    mul, const_dtype, name=f"{mul.get_friendly_name()}/convert"
                )
```

alexsu52 (changes requested):
Please check that you get the same graphs for PyTorch backend.

nncf/quantization/algorithms/weight_compression/openvino_backend.py
nikita-savelyevv: Thanks, changed.

nikita-savelyevv:
post training weight compression test build 34 is green

alexsu52 (approved):
LGTM

ljaljushkin:
@alexsu52 @nikita-savelyevv 
I've measured time for compression and total time for different weight compression cases:
develop
imageq
current PR
image
Seems like model inference takes almost twice longer on the validation dataset. 
Does it mean that compressed model should be saved differently in tests and on customer side?

nikita-savelyevv:
@ljaljushkin Thanks for highlighting this! 

The reason behind this is that during compression with group size, there is an additional Reshape node. In this PR, a Convert f16>f32 node is added after scale Multiply node. If Convert is added before Reshape node, then the performance drops. To fix this, I moved Convert node after Reshape node.

| Before                             | After |
|------------------------------------|----------------|
| image | image |

With this, performance is maintained after changes in the PR:
| Test case                         | Total time<br>develop branch | Total time<br>PR branch |
|-----------------------------------|------------------------------|-------------------------|
| tinyllama_data_free               | 04:18                        | 04:21                   |
| tinyllama_data_aware              | 04:06                        | 04:07                   |
| tinyllama_data_aware_awq          | 03:33                        | 03:39                   |
| tinyllama_data_aware_awq_stateful | 03:03                        | 03:03                   |

nikita-savelyevv:
post_training_weight_compression test build 42 is green. Waiting for results of OV validation across different hardware.

PR #2558, opened 2024-03-07
Add check on None

kshpv:
### Changes

Add a check on None for OverflowFix parameter.

### Reason for changes

Fix bug

### Related tickets

N/A

### Tests

Add test on a failed before scenario

### DIFF
--- a/nncf/quantization/advanced_parameters.py
+++ b/nncf/quantization/advanced_parameters.py
@@ -396,7 +396,7 @@ def apply_advanced_parameters_to_config(
     :param params: Advanced quantization parameters
     :return: advanced quantization parameters as dict in the legacy format
     """
-    config["overflow_fix"] = params.overflow_fix.value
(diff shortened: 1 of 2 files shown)


nncf/quantization/advanced_parameters.py
```diff
@@ -396,7 +396,7 @@ def apply_advanced_parameters_to_config(
     :param params: Advanced quantization parameters
     :return: advanced quantization parameters as dict in the legacy format
     """
-    config["overflow_fix"] = params.overflow_fix.value
+    config["overflow_fix"] = params.overflow_fix if params.overflow_fix is None else params.overflow_fix.value
```
KodiaqQ, line 399: We should have a test for that to prevent further pipeline issues.

kshpv: added

KodiaqQ (approved):

Issue #2579, opened 2024-03-15
[Good First Issue][NNCF]: Add tests for torch device utils


daniil-lyakhov:
Dear good first issue solvers, greetings!🐱
Don't miss the opportunity to contribute to our beloved project!

### Context

Functions `get_model_device`, `is_multidevice` and `get_model_dtype` from the file https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/utils.py are not covered by tests and don't have proper docstrings. 

### What needs to be done?

The tasks are
*) To extend file https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/test_utils.py with tests for functions mentioned above. Please check all 3 possible scenarios:
1) Model has no parameters
2) Model has parameters and all parameters placed on the same device
3) Model has parameters and parameters placed on different devices

*) To add proper docstrings for functions mentioned above

### Example Pull Requests

https://github.com/openvinotoolkit/nncf/pull/2526

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers
- [How to link your Pull Request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue-using-the-pull-request-sidebar)

### Contact points

@daniil-lyakhov

DaniAffCH:
.take

DaniAffCH:
Hi, @daniil-lyakhov thank you for having posted a GFI, I was looking for that :)

I was wondering whether it's safe to consider only the first network layer in the function `get_model_dtype`.

As far as I know, there are some edge cases in which the torch allows you to use different `dtype` within the same models as long as you keep the two workflows separate.

Consider this example:
```python3
import torch
import torch.nn as nn

class MixedTypeNet(nn.Module):
    def __init__(self):
        super(MixedTypeNet, self).__init__()
        self.fc_f16 = nn.Linear(10, 1).to(torch.float16)
        self.fc_f32 = nn.Linear(10, 1).to(torch.float32)

    def forward(self, f16, f32):
        x_f16 = torch.relu(self.fc_f16(f16))
        x_f32 = torch.relu(self.fc_f32(f32))
        return x_f16, x_f32

[...]

PR #2580, opened 2024-03-16
Tests for torch device and dtype utils

DaniAffCH:
### Changes

This PR addresses #2579.
- Add tests for torch device utils. The tests consider the case in which the model has no parameters, has all parameters on CPU, has all parameters on CUDA, and has parameters placed on different devices. In the latter, the parameters are moved on different devices randomly. 
- Add tests for torch `dtype` utils. The case in which the model has no parameters is also considered.
- Created a test torch helper class `EmptyModel` to take into account the case in which the model has no parameters at all.
- Add docstrings in utils.

### Tests

I compared all the results manually checking for their correctness. The code is also compliant with the coding style having been verified with `pre-commit run`

### DIFF
--- a/tests/torch/test_utils.py
+++ b/tests/torch/test_utils.py
@@ -36,15 +36,12 @@ def compare_saved_model_state_and_current_model_state(model: nn.Module, model_st
         assert param.requires_grad == model_state.requires_grad_state[name]
 
 
-def change_model_state(module: nn.Module, trainable_paramaters: bool = True):
+def change_model_state(module: nn.Module):
     for i, ch in enumerate(module.modules()):
         ch.training = i % 2 == 0
 
     for i, p in enumerate(module.parameters()):
-        if trainable_paramaters:
-            p.requires_grad = i % 2 == 0
-        else:
-            p.requires_grad = False
+        p.requires_grad = i % 2 == 0
 
 
 @pytest.mark.parametrize(
@@ -72,7 +69,9 @@ def test_bn_training_state_switcher(model: nn.Module):
 
     runner = DataLoaderBNAdaptationRunner(model, "cuda")
 
-    change_model_state(model, trainable_paramaters=False)
+    for p in model.parameters():
+        p.requires_grad = False
+
     saved_state = save_module_state(model)
 
     with runner._bn_training_state_switcher():
--- a/tests/torch/test_utils.py
+++ b/tests/torch/test_utils.py
@@ -55,35 +55,6 @@ def randomly_change_model_state(module: nn.Module, compression_params_only: bool
             p.requires_grad = True
 
 
-def randomly_change_model_device(module: nn.Module):
-    import random
-
-    if not torch.cuda.is_available():
-        return False
-
-    cuda = torch.device("cuda")
-
-    num_params = 0
-    num_cuda = 0
-
-    for p in module.parameters():
-        num_params += 1
-        if random.uniform(0, 1) > 0.5:
-            p.data = p.data.to(cuda)
-            num_cuda += 1
-
-    is_multidevice = 0 < num_cuda < num_params
-    return is_multidevice
-
-
-def is_model_empty(module: nn.Module):
-    try:
-        next(module.parameters())
-        return False
-    except StopIteration:
-        return True
-
-
 @pytest.mark.parametrize(
     "model", [BasicConvTestModel(), TwoConvTestModel(), MockModel(), DepthWiseConvTestModel(), EightConvTestModel()]
 )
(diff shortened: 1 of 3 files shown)


tests/torch/test_utils.py
```diff
@@ -83,3 +116,52 @@ def check_were_only_bn_training_state_changed(model: nn.Module, saved_state: _Mo
...
+    multidevice = randomly_change_model_device(model)
+    assert is_multidevice(model) == multidevice
+
+    is_empty = is_model_empty(model)
+
+    if torch.cuda.is_available() and not is_empty:
+        cuda = torch.device("cuda")
+        model = model.to(cuda)
+        assert not is_multidevice(model)
+        assert get_model_device(model).type == "cuda"
```
daniil-lyakhov, line 145: Please simplify the test:
1) Do not use `randomly_change_model_device`, test should be 100% deterministic
2) Please test only two models: empty model and Two conv test model as they cover 100% of cases
3) Please split this test on two tests: `test_empty_model_device` and `test_model_device` and remove this if section.

I expect following tests here:
`test_empty_model_device`: test creates empty model and check device
`test_model_device` skipped if cuda is not available, creates model on cpu, checks device and multidevice, move one param to cude, check devce and multidevce, move all params to cude, check device and multidevice

tests/torch/test_utils.py
```diff
@@ -83,3 +116,52 @@ def check_were_only_bn_training_state_changed(model: nn.Module, saved_state: _Mo
...
+)
+def test_model_dtype(model: nn.Module):
+    is_empty = is_model_empty(model)
+
+    model.to(torch.float16)
+    assert is_empty or get_model_dtype(model) == torch.float16
+    model.to(torch.float32)
+    assert get_model_dtype(model) == torch.float32
+    model.to(torch.float64)
+    assert is_empty or get_model_dtype(model) == torch.float64
```
daniil-lyakhov, line 167: Please split this test on two:
`test_empty_model_dtype` and `test_model_dtype`

daniil-lyakhov (changes requested):
@DaniAffCH, thank you for the contribution! Please address comments below

DaniAffCH:
Hi, @daniil-lyakhov thank you for the review, I just addressed your comments.

While refactoring I noticed that there are some tests that are not deterministic, previously developed. 
For instance:  https://github.com/openvinotoolkit/nncf/blob/e76bba2f8039266a1af42a01b1bbbcdb79ebc1d5/tests/torch/test_utils.py#L40

Should I modify them to ensure a deterministic outcome? To not modify it too much I could set a fixed random seed

tests/torch/test_utils.py
```diff
@@ -83,3 +87,46 @@ def check_were_only_bn_training_state_changed(model: nn.Module, saved_state: _Mo
...
 
     compare_saved_model_state_and_current_model_state(model, saved_state)
+
+
+def test_model_device():
+    if not torch.cuda.is_available():
+        return
+
+    model = TwoConvTestModel()
+    cuda = torch.device("cuda")
```
daniil-lyakhov, line 97:
```suggestion
@pytest.mark.skipif(not torch.cuda.is_available(), reason="Cuda is not available in current environment")
def test_model_device():
    model = TwoConvTestModel()
    cuda = torch.device("cuda")
```

daniil-lyakhov (commented):
Minor

daniil-lyakhov:
> some other tests that

Nice catch! looks like we missed this part during code review previously. This test could cause problems in future because of randomness, it would be great if you remove randomness from them

DaniAffCH:
> > some other tests that
> 
> Nice catch! looks like we missed this part during code review previously. This test could cause problems in future because of randomness, it would be great if you remove randomness from them

Perfect, done! :)

tests/torch/test_utils.py
```diff
@@ -76,10 +72,51 @@ def check_were_only_bn_training_state_changed(model: nn.Module, saved_state: _Mo
 
     runner = DataLoaderBNAdaptationRunner(model, "cuda")
 
-    randomly_change_model_state(model)
+    change_model_state(model)
```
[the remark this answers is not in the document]
DaniAffCH: Do you mean just inside the test `test_bn_training_state_switcher` or also in `test_training_mode_switcher`? In case both has to be modified I can simply change the `change_model_state` function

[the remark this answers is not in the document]
daniil-lyakhov: Please do it only in the `test_bn_training_state_switcher`, `test_training_mode_switcher` is ok

tests/torch/test_utils.py
```diff
@@ -76,10 +72,51 @@ def check_were_only_bn_training_state_changed(model: nn.Module, saved_state: _Mo
 
     runner = DataLoaderBNAdaptationRunner(model, "cuda")
 
-    randomly_change_model_state(model)
+    change_model_state(model, trainable_paramaters=False)
```
daniil-lyakhov, line 75:
```suggestion
    for p in module.parameters():
        p.requires_grad = False
```

tests/torch/test_utils.py
```diff
@@ -33,29 +36,22 @@ def compare_saved_model_state_and_current_model_state(model: nn.Module, model_st
         assert param.requires_grad == model_state.requires_grad_state[name]
 
 
-def randomly_change_model_state(module: nn.Module, compression_params_only: bool = False):
-    import random
+def change_model_state(module: nn.Module, trainable_paramaters: bool = True):
```
daniil-lyakhov, line 39: Please remove `trainable_paramaters` param
```suggestion
def change_model_state(module: nn.Module):
```

daniil-lyakhov (changes requested):

daniil-lyakhov (approved):
LGTM 👍

daniil-lyakhov:
@AlexanderDokuchaev, please review and merge if you don't have comments

AlexanderDokuchaev (approved):

Issue #2570, opened 2024-03-11
[Good First Issue][NNCF]: fix invalid error reporting in JSON schema


ljaljushkin:
### Context

Configuration for NNCF algorithms can be defined in JSON format. For instance, [config.json].(https://github.com/openvinotoolkit/nncf/blob/develop/examples/torch/classification/configs/sparsity_quantization/inception_v3_imagenet_rb_sparsity_int8.json#L17-L38)

Compression section in these configuration files are validated using [jsonschema.validate](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/config/schema.py#L163C9-L163C28)

But there's some bug with reporting error in JSON schema when a field is not defined in schema.

Steps to reproduce:

1) checkout https://github.com/ljaljushkin/nncf_pytorch/tree/nl/bug_in_schema_report

2) There's unit test checking 2 config files:
- [good config](https://github.com/ljaljushkin/nncf_pytorch/blob/nl/bug_in_schema_report/tests/torch/data/schema_validation_good_configs/bug_in_error_report.json)
- [config with some not defined field](https://github.com/ljaljushkin/nncf_pytorch/blob/nl/bug_in_schema_report/tests/torch/data/schema_validation_good_configs/bug_in_error_report_2.json#L8)

To run tests:
```
test $pytest tests/torch/test_config_schema.py -sv -k 'bug_in_error_report'
```

The expected behavior, that the first test should pass, the second — fail. 
But the reported error in the second test should be the following:
```
error = <ValidationError: "Additional properties are not allowed ('unknown_field' was unexpected)">
```
However, the second test reports this, which is not correct:
```
jsonschema.exceptions.ValidationError: 'polynomial' is not one of ['exponential', 'exponential_with_bias', 'baseline'] 
```

The message is misleading, because it's saying that the supported schedule type `polynomial` for rb_sparsity doesn't correspond to the list of supported schedule types for the filter pruning algorithm. 

https://openvinotoolkit.github.io/nncf/schema/index.html#compression_oneOf_i0_oneOf_i1_params_schedule
https://openvinotoolkit.github.io/nncf/schema/index.html#tab-pane_compression_oneOf_i0_oneOf_i9

filter pruning's definition: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/config/schemata/algo/filter_pruning.py#L88
rb sparsity's definition: 
[...]

PR #2597, opened 2024-03-23
#2570 Issue: Update config.py

LalitSP:
### Changes
Updated the validation logic in the config.py file to handle errors related to additional properties not defined in the schema. Added checks to raise a more informative ValidationError when unexpected properties are encountered during validation.

### Reason for changes
The current error reporting mechanism in the JSON schema validation is misleading, especially when encountering unexpected properties in the configuration.

### Related tickets
Ticket ID: 94933

### Tests
test $pytest tests/torch/test_config_schema.py -sv -k 'bug_in_error_report'


nncf/config/config.py
```diff
@@ -127,13 +128,17 @@ def _is_path_to_algorithm_name(path_parts: List[str]) -> bool:
     def validate(loaded_json):
         try:
             jsonschema.validate(loaded_json, NNCFConfig.schema())
+
+        except JsonSchemaValidationError as e:
+            logger.error('Invalid NNCF config supplied!')
```
ljaljushkin, line 133: Why do you need this error handler? Isn't it the same as the next one? Shouldn't the next one executed?

ljaljushkin (changes requested):
Thanks for the contribution! Please address the comment and fix 2 precommit jobs: common unit tests and type checks.

ljaljushkin:
#2570

ljaljushkin:
Due to the lack of activity @LalitSP was unassigned from the issue and PR is closed

PR #2602, opened 2024-03-26
Support string dtype for OV backend

andrey-churkin:
### Changes

- Support string dtype for OV backend

### Reason for changes

NotImplementedError: NNCF is not yet supported OpenVINO data type: string.

### Related tickets

136751

### Tests

-  test_convert_to_nncf_dtype_supported_types()
- test_convert_to_nncf_dtype_unsupported_types()


nncf/common/graph/layer_attributes.py
```diff
@@ -19,6 +19,7 @@
 class Dtype(Enum):
     FLOAT = "float"
     INTEGER = "int"
+    STRING = "string"
```
rkazants, line 22: may you have test for it?

DaniAffCH:
Hi @andrey-churkin could you please add the inverse mapping (from NNCF dtype to OV dtype) in the function `convert_nncf_dtype_to_ov_dtype` within `nncf/experimental/common/graph/netron.py`.
Otherwise, the Netron export won't work for networks containing `string` as `dtype`

nncf/common/graph/layer_attributes.py
andrey-churkin: I have added tests to be sure that we cover all available OpenVINO data types.

andrey-churkin:
@DaniAffCH Thanks! I have updated the code recently. So everything is fine now.

alexsu52 (approved):
LGTM

PR #2605, opened 2024-03-27
Added torch cuda test skipping in case it is run on a CPU-only machine

nikita-savelyevv:
### Reason for changes

Regression after #2596 .


nikita-savelyevv:
windows torch precommit build 296 has passed

alexsu52 (approved):
LGTM

PR #2611, opened 2024-03-28
Add is_subtype argument for metatype register 

AlexanderDokuchaev:
### Changes

Add `is_subtype` argument for meta type register 

### Reason for changes

Registering a nonlinear hierarchy of sub meta types.

### DIFF
--- a/nncf/common/graph/operator_metatypes.py
+++ b/nncf/common/graph/operator_metatypes.py
@@ -79,11 +79,12 @@ class OperatorMetatypeRegistry(Registry):
         super().__init__(name)
         self._op_name_to_op_meta_dict: Dict[str, Type[OperatorMetatype]] = {}
 
-    def register(self, name: Optional[str] = None) -> Callable[..., Type[OperatorMetatype]]:
+    def register(self, name: Optional[str] = None, is_subtype: bool = False) -> Callable[..., Type[OperatorMetatype]]:
         """
         Decorator for registering operator metatypes.
 
         :param name: The registration name.
+        :param is_subtype: Whether the decorated metatype is a subtype of another registered operator.
         :return: The inner function for registering operator metatypes.
         """
         name_ = name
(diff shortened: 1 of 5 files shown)


nncf/common/graph/operator_metatypes.py
```diff
@@ -100,15 +101,15 @@ def wrap(obj: Type[OperatorMetatype]) -> Type[OperatorMetatype]:
...
-            op_names = obj.get_all_aliases()
-            for name in op_names:
-                if name in self._op_name_to_op_meta_dict and not obj.subtype_check(self._op_name_to_op_meta_dict[name]):
-                    raise nncf.InternalError(
-                        "Inconsistent operator metatype registry - single patched "
-                        "op name maps to multiple metatypes!"
-                    )
-
-                self._op_name_to_op_meta_dict[name] = obj
+            if not is_subtype:
```
daniil-lyakhov, line 104:
```suggestion
            if is_subtype:
                return obj
            ....
```

daniil-lyakhov (changes requested):

nncf/common/graph/operator_metatypes.py
AlexanderDokuchaev: I'm not sure that code like this

```python
if is_subtype:
    return obj
...
return obj
```

is better that 

```python
if not is_subtype:
   ...
return obj
```

daniil-lyakhov: It is believed that less indentation increase code readability, for example to understand how this code is working for `is_type == True` one should skip if and check nothing is called after it. But do as you wish

daniil-lyakhov (approved):

AlexanderDokuchaev:
Needs to register hierarchy of subtypes like
```
        Conv2d  
     /         \ 
ModuleCond2   DepthwiseCond2d
     |    
ModuleDepthwiseCond2d
```

KodiaqQ (approved):

PR #2616, opened 2024-04-03
Deprecate install extra packages

AlexanderDokuchaev:
### Changes

- Depreciated message about installation extra requirements by `pip install nncf[torch]`
- Fix system requirements of python version (actual python>=3.8)
- Update corresponding onnx version

### Related tickets

134503


AlexanderDokuchaev:
> I would clearly list all options, i.e. that [torch], [tf], [onnx] and [openvino] are deprecated. Now it's confusing.

Deprecated all, torch as examples, it's writing after list of options. 

> Installation guide should also be updated

Updated

> Are you sure it's not possible to raise a warning when pip install nncf[] is used?

Yes

> Can you carefully check if we have it in notebooks, nncf examples, ov documentation, optimum-intel or otx somewhere?

I haven't found installation nncf with extra option.

alexsu52 (approved):
LGTM

MaximProshin (approved):

Issue #2565, opened 2024-03-08
[Good First Issue][NNCF]: check number of u8, u4 constants in weight compression tests


ljaljushkin:
### Context

The issue proposes adding additional check to [post-training compression conformance suite](https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/README.md), particularly for 
[weight compression part](blob/develop/tests/post_training/test_quantize_conformance.py#L248). 
Currently, it counts number of `FakeQuantize` ops only: [code](https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/pipelines/base.py#L381)
But it's only applicable for [quantization algorithms](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/post_training/Quantization.md#post-training-quantization), [weight compression algorithm](https://github.com/openvinotoolkit/nncf/blob/develop/docs/compression_algorithms/CompressWeights.md) creates u4 or u8 constants instead. 

### What needs to be done?

1) Add a function for counting number of `u4` and `u8` ops similar to how it's done in the Optimum tests.
[get_num_quantized_nodes](https://github.com/huggingface/optimum-intel/blob/main/tests/openvino/utils_tests.py#L122)
2) Integrate this function to [conformance test's hierarchy](https://github.com/openvinotoolkit/nncf/tree/develop/tests/post_training/pipelines)
Refer to the [get_num_compressed](https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/pipelines/base.py#L381) for quantization algorithm.
3) Add a check that actual number of u4 and u8 ops matches the reference values.
Refer to the check for the accuracy: https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/pipelines/base.py#L285
4) add number of u4 and u8 to `results.csv`
https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/test_quantize_conformance.py#L123
Example of current `results.csv`:
```
Model,Backend,Metric name,Metric value,Metric diff,RAM MiB,Compr. time,Stat. collection time,Mixed-Precision search time,AWQ time,Apply Compression time,Total time,Status
tinyllama_data_aware_awq_stateful,OV,Similarity,0.81229,-0.18771,31028,0:02:32,0:00:55,0:00:19,0:00:48,0:00:23,0:04:48,
```
More human-readable visualization:
[...]

PR #2617, opened 2024-04-05
add int4 and int8 checks

DaniAffCH:
### Changes
This PR addresses #2565

- Added a function to retrieve the number of int4 and int8 ops within the Weight Compression tests.
- Added check to ensure that the number of int4 and int8 ops match the reference values
- Added reference values for int4 and int8 ops

### Related tickets

Ticket: 132562

### Tests

Tested running:
```
pytest test_quantize_conformance.py::test_weight_compression[tinyllama_data_aware_awq_stateful_backend_OV] --data=
```


tests/post_training/data/wc_reference_data.yaml
```diff
@@ -1,5 +1,7 @@
 tinyllama_data_free_backend_OV:
   metric_value: 0.72057
+  num_int4: 228
+  num_int8: 84
 tinyllama_data_aware_backend_OV:
   metric_value: 0.83084
```
ljaljushkin, line 6: I can't commit to your branch, here are the numbers. The same for the rest of configurations.
```suggestion
  metric_value: 0.83084
  num_int4: 184
  num_int8: 128
```

ljaljushkin (changes requested):
I noticed some bug 
references numbers are different for data-free, but test is passing and results file has  different numbers
image

when run separately, results file has corrent numbers
image

tests/post_training/data/wc_reference_data.yaml
ljaljushkin: BTW, in your development, you could try to comment some parts of validate function to not evaluate models for accuracy (similarity). Potentially, it could save some memory and make it possible to check all configurations.

DaniAffCH:
> I noticed some bug references numbers are different for data-free, but test is passing and results file has different numbers image
> 
> when run separately, results file has corrent numbers image

I attempted to comment on some sections in the evaluation script. Unfortunately, I'm still unable to execute the other models as they saturate the RAM during quantization. Apparently, 16 GB is not sufficient. Since I can only execute one model, I'm unable to reproduce the bug you spotted. I need some time to find a solution for testing and debugging.

DaniAffCH:
Nice catch @ljaljushkin! I think the issue now is fixed

andrey-churkin:
@DaniAffCH Could you please rebase your branch to run CI one more time? Thanks.

ljaljushkin (approved):

ljaljushkin:
image
44 build

ljaljushkin:
Thanks, @DaniAffCH! 🎉

PR #2641, opened 2024-04-18
Fix quantile parameter not being used for MEAN_NO_OUTLIERS aggregator

nikita-savelyevv:
### Changes

When `MEAN_NO_OUTLIERS` aggregator is used, `quantile` parameter was not provided to the aggregator constructor.

### DIFF
--- a/nncf/quantization/algorithms/min_max/torch_backend.py
+++ b/nncf/quantization/algorithms/min_max/torch_backend.py
@@ -33,7 +33,8 @@ from nncf.quantization.advanced_parameters import StatisticsType
 from nncf.quantization.algorithms.min_max.backend import MinMaxAlgoBackend
 from nncf.quantization.fake_quantize import FakeConvertParameters
 from nncf.quantization.fake_quantize import FakeQuantizeParameters
-from nncf.quantization.range_estimator import RangeEstimatorParameters, AggregatorType
+from nncf.quantization.range_estimator import AggregatorType
+from nncf.quantization.range_estimator import RangeEstimatorParameters
 from nncf.torch.graph.graph import PTNNCFGraph
 from nncf.torch.graph.graph import PTTargetPoint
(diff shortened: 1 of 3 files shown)


daniil-lyakhov:
Interestingly, there were only reducers with quantile, now both reducers and aggregators are using quantile. Previously parameter `params.quantile_outlier_prob` was using only for reducers. Now in case of quantile reducer and no outliers aggregator there will be no options to select different parameter for reducer and aggregator accordingly. Let's discuss this with the team

nikita-savelyevv:
> Interestingly, there were only reducers with quantile, now both reducers and aggregators are using quantile. Previously parameter `params.quantile_outlier_prob` was using only for reducers. Now in case of quantile reducer and no outliers aggregator there will be no options to select different parameter for reducer and aggregator accordingly. Let's discuss this with the team

In my opinion using both reducer and outlier with quantile is not a very likely use case. I'd keep it as is and just add a note for quantile parameter that says that this parameter may be used for both reducers or aggregators, but not together.

nncf/quantization/algorithms/min_max/torch_backend.py
```diff
@@ -195,11 +196,14 @@ def get_statistic_collector(
...
-                aggregation_axes=aggregation_axes,
-                num_samples=num_samples,
-                tensor_processor=PTNNCFCollectorTensorProcessor,
-            )
+            kwargs = {
+                "num_samples": num_samples,
+                "aggregation_axes": aggregation_axes,
+                "tensor_processor": PTNNCFCollectorTensorProcessor,
+            }
+            if params.aggregator_type == AggregatorType.MEAN_NO_OUTLIERS:
```
daniil-lyakhov, line 204: Please specify `MedianNoOutliersAggregator` here as well

alexsu52 (approved):
LGTM

daniil-lyakhov (approved):
👍

PR #2647, opened 2024-04-22
Update ReleaseNotes.md

daniil-lyakhov:

### DIFF
--- a/ReleaseNotes.md
+++ b/ReleaseNotes.md
@@ -1,5 +1,57 @@
 # Release Notes
 
+## New in Release 2.10.0
+
+Post-training Quantization:
+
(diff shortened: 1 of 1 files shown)


MaximProshin:
I didn't get the problem with the bs example. Why should it be excluded? Regardless of whether there is a description or not, it's still available. 
IMO, there are no valuable changes here, just aligned with OV release notes which is not required. Please make sure such comments are received at the related NNCF release notes PR rather than after when it's already merged.

KodiaqQ:
@daniil-lyakhov, why didn't you open this request immediately in the release branch? Are these changes necessary for 2.10?

PR #2648, opened 2024-04-22
Fix for pre-compressed MatMuls in AWQ.

andreyanufr:
### Changes

1) Avoiding int8/int4 MatMuls processing in AWQ.
2) Tests for case of compression pre-compressed models.
3) Test for group_size == -1 and AWQ/scale estimation.

### Reason for changes

Compression incorrectly falls with AWQ and pre-compressed models with int8/int4 MatMuls.

### Related tickets

CVS-137988

### Tests

Unit tests.


ljaljushkin:
Some simple unit test for this particular problem would be useful

ljaljushkin:
> Some simple unit test for this particular problem would be useful

Here's the good example of simple unit test to consider
https://github.com/openvinotoolkit/nncf/pull/2653/files#diff-223ea638f7751f7c0c3e8f867ec9c8c132a3ccd62a9dcea2a5d158836c71c222R458-R470

nikita-savelyevv (approved):

ljaljushkin (approved):

PR #2649, opened 2024-04-23
[Torch] Do not wrap not tensor inputs

daniil-lyakhov:
### Changes

* NNCF wrapping for Tensor input parameters is skipped when actual parameter is not a tensor

### Reason for changes

* To support models which could accept different types as an input
(example: [YolovV8](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/nn/tasks.py#L77-L89) 
```
def forward(self, x, *args, **kwargs):
    if isinstance(x, dict):  # for cases of training and validating while training.
        return self.loss(x, *args, **kwargs)
    return self.predict(x, *args, **kwargs)
```

### Related tickets

138682

### Tests

* tests/torch/test_input_management.py is extended

### DIFF
--- a/nncf/torch/dynamic_graph/io_handling.py
+++ b/nncf/torch/dynamic_graph/io_handling.py
@@ -308,6 +308,9 @@ class InputInfoWrapManager:
     def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dict]:
         bound_model_params = self._fwd_signature.bind(*model_args, **model_kwargs)
         for param_name, dummy_input in self._fwd_param_names_to_dummy_inputs_odict.items():
+            if not isinstance(dummy_input, torch.Tensor):
+                continue  # Did not expect a tensor at this parameter during graph building, shouldn't wrap now too
+
             param_kind = self._fwd_signature.parameters[param_name].kind
             if param_kind is Parameter.VAR_POSITIONAL or param_kind is Parameter.VAR_KEYWORD:
                 nncf_logger.warning(
@@ -327,14 +330,6 @@ class InputInfoWrapManager:
                 bound_model_params.apply_defaults()
 
             potential_tensor = bound_model_params.arguments[param_name]
-            if not isinstance(dummy_input, torch.Tensor):
-                if isinstance(potential_tensor, torch.Tensor):
-                    raise RuntimeError(
-                        f'Original model forward parameter "{param_name}" expected to not be a tensor,'
-                        " but the tensor type recieved. Please use a tensor type as an example input"
-                        f' for the "{param_name}" parameter_name'
-                    )
-                continue  # Did not expect a tensor at this parameter during graph building, shouldn't wrap now too
 
             if potential_tensor is None:
                 # Default was None - cannot wrap as-is. Will wrap a dummy tensor as specified in
--- a/nncf/torch/dynamic_graph/io_handling.py
+++ b/nncf/torch/dynamic_graph/io_handling.py
@@ -308,8 +308,6 @@ class InputInfoWrapManager:
     def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dict]:
         bound_model_params = self._fwd_signature.bind(*model_args, **model_kwargs)
         for param_name, dummy_input in self._fwd_param_names_to_dummy_inputs_odict.items():
-            if not isinstance(dummy_input, torch.Tensor):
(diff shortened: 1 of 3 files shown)


nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -329,7 +327,22 @@ def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dic
                 bound_model_params.apply_defaults()
 
             potential_tensor = bound_model_params.arguments[param_name]
+            if not isinstance(dummy_input, torch.Tensor):
+                if isinstance(potential_tensor, torch.Tensor):
+                    raise RuntimeError(
```
alexsu52, line 332: It looks like, this change is not backwards compatible.

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -329,7 +327,22 @@ def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dic
...
+                        f'Original model forward parameter "{param_name}" expected to not be a tensor,'
+                        " but the tensor type recieved. Please use a tensor type as an example input"
+                        f' for the "{param_name}" parameter_name'
+                    )
+                continue  # Did not expect a tensor at this parameter during graph building, shouldn't wrap now too
+
             if potential_tensor is not None:
+                if not isinstance(potential_tensor, torch.Tensor):
```
alexsu52, line 340:
```suggestion
            if isinstance(potential_tensor, torch.Tensor):
```

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -329,7 +327,22 @@ def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dic
...
+                    nncf_logger.warning(
+                        f'Original model forward parameter "{param_name}" expected to be a tensor,'
+                        f" but {type(potential_tensor)} recieved."
+                    )
+                    continue
```
alexsu52, line 345: From the user stand point, an user will see warning every model inference, right? I think this will piss off any user:)

I suggest to remove this warning.

alexsu52 (changes requested):

nncf/torch/dynamic_graph/io_handling.py
daniil-lyakhov: Do you suggest to remove runtime error then?

daniil-lyakhov: Such combination alters NNCFNetwork building and the result of additional unexpected nncf_input could cause an error in an compression algo. In case we have such combination (which is unlikely) in some model, it worth investigation on how it actually works

nncf/torch/dynamic_graph/io_handling.py
daniil-lyakhov: Done

nncf/torch/dynamic_graph/io_handling.py
daniil-lyakhov: Refactored

nncf/torch/dynamic_graph/io_handling.py
daniil-lyakhov: Ok, done

nncf/torch/dynamic_graph/io_handling.py
```diff
@@ -341,6 +341,9 @@ def wrap_inputs(self, model_args: Tuple, model_kwargs: Dict) -> Tuple[Tuple, Dic
                     device = get_model_device(self._module_ref_for_device)
                 dummy_input_copy = dummy_input.clone().to(device)
                 _ = nncf_model_input(dummy_input_copy)
+            elif isinstance(potential_tensor, torch.Tensor):
+                # Skip wrapping by nncf_model_input in case potential tensor is not a torch.Tensor.
```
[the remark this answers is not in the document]
AlexanderDokuchaev: Should not we support this case, and wrap all tensors in inputs? For dict as input wrap all values, for list all elements.
For torch get input as dict or list of tensor looks like legal operation, but unusual.

alexsu52 (approved):
LGTM

nncf/torch/dynamic_graph/io_handling.py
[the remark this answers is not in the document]
daniil-lyakhov: Please provide an example when we actually need to wrap dictionary or other object

AlexanderDokuchaev (approved):

PR #2636, opened 2024-04-16
[NNCF]: Add INT8 weight compression conformance test for Tinyllama-1.1b PyTorch model

AdiKsOnDev:
### Changes

- Added the compression for `INT8` weight compression (**PyTorch** backend support) in `LMWeightCompression` class
- Added the test scenario to the `model_scope`

### Reason for changes

Requested to Benchmark changes via `whowhatbench` in issue #2527 

### Related tickets

ref: 130788
Closes issue #2527

### Tests
Following code was added in the `LMWeightCompression` class

```python
def compress(self) -> None:
    if self.backend == BackendType.FP32:
        return
    elif self.backend == BackendType.TORCH:
        start_time = time.perf_counter()
                                                                                                            
        tokenizer = transformers.AutoTokenizer.from_pretrained(self.model_id)
        self.model = transformers.AutoModelForCausalLM.from_pretrained(
            self.model_id, torch_dtype=torch.float16, device_map="cpu"
        )
                                                                                                            
        text = "The TinyLlama project aims to pretrain a 1.1B Llama model on 3 trillion tokens."
        token = tokenizer(text, max_length=500, return_tensors="pt", truncation=True)
        inputs = {"input_ids": token["input_ids"], "attention_mask": token["attention_mask"]}
                                                                                                            
        self.run_info.compression_memory_usage = memory_usage(self._compress_torch(inputs), max_usage=True)
        self.run_info.time_compression = time.perf_counter() - start_time
                                                                                                            
        return

      print("Weight compression...")
      start_time = time.perf_counter()
      self.run_info.compression_memory_usage = memory_usage(self._compress, max_usage=True)
      self.run_info.time_compression = time.perf_counter() - start_time
```

```python
    def _compress_torch(self, inputs):
        self.compressed_model = nncf.compress_weights(self.model, dataset=nncf.Dataset([inputs]))
```

PR #2651, opened 2024-04-24
Checking model type in nncf.torch.wrap_model

alexsu52:
### Changes

Added model type check in nncf.torch.wrap_model

### Reason for changes

Improve UX

### Related tickets

https://github.com/openvinotoolkit/nncf/pull/2636#discussion_r1576234452

### Tests

N/A

### DIFF
--- a/nncf/torch/model_creation.py
+++ b/nncf/torch/model_creation.py
@@ -326,6 +326,10 @@ def wrap_model(
     :param trace_parameters: Whether to trace model parameters. Default is False.
     :return: A model wrapped by NNCFNetwork.
     """
+    if not isinstance(model, torch.nn.Module):
+        raise TypeError(
+            f"The provided model type {type(model)} is incompatible. Only models inheriting from torch.nn.Module are supported."
+        )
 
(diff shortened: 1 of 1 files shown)


nncf/torch/model_creation.py
```diff
@@ -326,6 +326,8 @@ def wrap_model(
     :param trace_parameters: Whether to trace model parameters. Default is False.
     :return: A model wrapped by NNCFNetwork.
     """
+    if not isinstance(model, torch.nn.Module):
+        raise ValueError("Invalid model. Only PyTorch models inherited from torch.nn.Module are supported")
```
AlexanderDokuchaev, line 330:
```suggestion
        raise TypeError(f"The type if provided model {type(model)} is not compatible. Only models that inherit from torch.nn.Module are supported.)
```
What think about it?
ValueError is about incorrect value of argument, like -1 when expected value >0. TypeError is more correct in this case.

AlexanderDokuchaev (changes requested):

PR #2652, opened 2024-04-24
Conformance test add validation with batch_size

kshpv:
### Changes

Add arbitrary batch_size for validation dataloader.

### Reason for changes

Speed up validation

### Related tickets

138556

### Tests

Jobs


kshpv:
jobs 370 and 371 with fixed host are running

kshpv:
Based on results - median speed up per model is ~ 40%

tests/post_training/model_scope.py
```diff
@@ -250,6 +250,7 @@
         },
         "backends": [BackendType.OV],
         "batch_size": 32,
+        "validation_batch_size": 1,  # Validation is slower with batch_size > 1
```
KodiaqQ, line 253: Why do we have separate batch sizes for calibration & validation? If some models should be validated with the batch_size=1, they should be calibrated with the same batch.

tests/post_training/pipelines/image_classification_timm.py
```diff
@@ -122,13 +122,14 @@ def prepare_calibration_dataset(self):
 
     def _validate(self):
         val_dataset = datasets.ImageFolder(root=self.data_dir / "imagenet" / "val", transform=self.transform)
-        val_loader = torch.utils.data.DataLoader(val_dataset, batch_size=1, num_workers=2, shuffle=False)
+        val_loader = torch.utils.data.DataLoader(
+            val_dataset, batch_size=self.validation_batch_size, num_workers=2, shuffle=False
```
KodiaqQ, line 126: Are the 2 workers enough for the data loading with a batch more than 1?

tests/post_training/model_scope.py
kshpv: Make sense

kshpv: set bs=1 for swin

kshpv:
job with updates for swin - 375

kshpv:
Closing the PR since after obtaining the correct results on validation performance it was observed that there is no significant impact on performance with these changes

PR #2653, opened 2024-04-24
[WC] Fix ratio_defining_params

l-bat:
### Reason for changes

Fix bug in `_get_ratio_defining_params` method

### Tests

test_shared_gather_all_layers

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -182,7 +182,7 @@ class WeightCompression(Algorithm):
             )
         )
 
-        # The last MatMul layer is quantized to 4-bits if all_layers=True or if the layer is shared
+        # The last MatMul layer is quantized to 4-bits if all_layers=True
         if not self._all_layers and not is_last_layer_shared:
             ratio_defining_params = ratio_defining_params[:-1]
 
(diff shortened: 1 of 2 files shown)


tests/openvino/native/quantization/test_weights_compression.py
```diff
@@ -455,6 +455,21 @@ def test_shared_gather(mode):
             assert op.get_element_type() == weight_name_vs_type[op_name]
 
 
+@pytest.mark.parametrize("all_layers", (True, False))
+def test_shared_gather_all_layers(all_layers):
```
ljaljushkin, line 459: Did you manage to catch the incorrect behavior with the old code? Seems like this test passed before

ljaljushkin (changes requested):

tests/openvino/native/quantization/test_weights_compression.py
l-bat: Looks like the old behavior was also correct. I couldn't find incorrect configuration.

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -165,28 +165,28 @@ def _get_ratio_defining_params(
...
-            return list(filter(lambda wp: len(wp.reduction_axes) == 1, all_weight_params))
-
         ratio_defining_params = list(
             filter(
                 lambda wp: wp.node_with_weight.metatype in self._backend_entity.matmul_metatypes,
                 all_weight_params,
             )
         )
 
+        # The last MatMul layer is quantized to 4-bits if all_layers=True or if the layer is shared
```
ljaljushkin, line 175: This comment looks wrong.
`if the layer is shared` means that it has the same precision as embedding, and it's 4bit only with `all_layers`

ljaljushkin (approved):
LGTM, would be nice to correct comment, but it's minor

ljaljushkin (approved):

PR #2656, opened 2024-04-25
[TF] GA precommit

AlexanderDokuchaev:
### Changes

- Add GA for TF precommit
- Move TF sanity tests to nightly scope  
- Enable pip cache for precommit jobs
- Parallel onnx and common precommit tests
- Set num workers by env variable, instead of hardcode in makefile


alexsu52 (approved):
Great job!

PR #2672, opened 2024-05-08
Fix SQ algorithm ports usage

KodiaqQ:
### Changes

- Fixed SmoothQuant algorithm to work with Split ports correctly;

### Reason for changes

- Bugfix

### Related tickets

- 140351

### DIFF
--- a/nncf/quantization/algorithms/smooth_quant/algorithm.py
+++ b/nncf/quantization/algorithms/smooth_quant/algorithm.py
@@ -161,7 +161,8 @@ class SmoothQuant(Algorithm):
                 weight_update_command = self._backend_entity.weight_update_command(node_to_smooth, scaled_weight.data)
                 transformation_layout.register(weight_update_command)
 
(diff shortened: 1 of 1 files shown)


nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -161,7 +161,8 @@ def apply(
                 weight_update_command = self._backend_entity.weight_update_command(node_to_smooth, scaled_weight.data)
                 transformation_layout.register(weight_update_command)
 
-            activations_shape = graph.get_output_edges(source_node)[source_output_port_id].tensor_shape
```
KodiaqQ, line 164: This behaviour that utilizes the `get_output_edges` order may be the reason for the other issues in different places in NNCF. As the common approach, only identification by the needed `input-`/`output_port_id` field should be used.
@alexsu52

KodiaqQ:
@alexsu52, please, review.

alexsu52 (approved):
LGTM

PR #2676, opened 2024-05-14
Andreyan/awq extention

andreyanufr:
### Changes

Extended AWQ algorithms for patterns Act->MatMul and Act->Multiply->MatMul with insertion for extra scales after activation.

### Reason for changes

Support AWQ for wider family of LLMs

### Related tickets

CVS-141131

### Tests

Added unit tests


nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -150,19 +145,30 @@ def apply(
...
-            assert len(target_node_names) == len(merge_node_names)
             weight_params = self._all_weight_params[name_mapping[target_node_names[-1]]]
+
             if weight_params.compression_config.num_bits != 4:
                 continue
             target_node = self._nodes_to_compress[name_mapping[target_node_names[-1]]]
-            merge_node = self._nodes_to_compress[name_mapping[merge_node_names[-1]]]
+
+            # avoid matching different patterns for the same node
+            if target_node.node_name in awq_data:
```
ljaljushkin, line 159: When can it happen? Which patterns can be applied for the same node? Aren't they mutually exclusive?

ljaljushkin: This code fixes the bug in AWQ with matching multiple patterns for the same node. 
Small unit test would be helpful
image

nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -150,19 +145,30 @@ def apply(
...
+                continue
+
+            nncf_node = graph.get_node_by_key(match[0])
+
+            if self._backend_entity.is_node_with_weights(nncf_node, graph):  # pattern MatMul->Multiply->MatMul
+                merge_node_names = []
+                for weight_op_friendly_name, _ in self._backend_entity.get_weight_names_and_port_ids(nncf_node, graph):
+                    merge_node_names.append(weight_op_friendly_name)
+                assert len(target_node_names) == len(merge_node_names)
```
ljaljushkin, line 168: per offline discussion, should be deleted, since the matmul can't have more than one weight on input

ljaljushkin (changes requested):

nncf/quantization/algorithms/weight_compression/awq.py
andreyanufr: Done

ljaljushkin (approved):

PR #2679, opened 2024-05-14
Add loop body quantization for OpenVINO

andrey-churkin:
### Changes

- Add loop body quantization

### Reason for changes

Support loop body quantization for OpenVINO backend

### Related tickets

Ref: 135480


MaximProshin:
It was decided to not introduce this change in NNCF as OV decided to convert such models differently. We can re-open it if later on we get such models.

PR #2681, opened 2024-05-15
[quantization] Fix config compatible check

UsingtcNower:

### DIFF
--- a/nncf/common/quantization/structs.py
+++ b/nncf/common/quantization/structs.py
@@ -207,13 +207,12 @@ class QuantizationConstraints:
         return qconfig
 
     def is_config_compatible(self, qconfig: QuantizerConfig) -> bool:
-        is_compatible = True
         for attr_name, constraint in self.qconf_attr_vs_constraint_dict.items():
             if constraint is not None:
                 qconf_attr_value = getattr(qconfig, attr_name)
                 if qconf_attr_value != constraint:
-                    is_compatible = False
-        return is_compatible
+                    return False
+        return True
 
     def get_updated_constraints(self, overriding_constraints: "QuantizationConstraints") -> "QuantizationConstraints":
         new_dict = deepcopy(self.qconf_attr_vs_constraint_dict)


nncf/common/quantization/structs.py
```diff
@@ -213,6 +213,7 @@ def is_config_compatible(self, qconfig: QuantizerConfig) -> bool:
                 qconf_attr_value = getattr(qconfig, attr_name)
                 if qconf_attr_value != constraint:
                     is_compatible = False
```
alexsu52, line 215: What do you think about `return False` here?

```
def is_config_compatible(self, qconfig: QuantizerConfig) -> bool:
        for attr_name, constraint in self.qconf_attr_vs_constraint_dict.items():
            if constraint is not None:
                qconf_attr_value = getattr(qconfig, attr_name)
                if qconf_attr_value != constraint:
                    return False
        return True
```

UsingtcNower: > What do you think about `return False` here?
> 
> ```
> def is_config_compatible(self, qconfig: QuantizerConfig) -> bool:
>         for attr_name, constraint in self.qconf_attr_vs_constraint_dict.items():
>             if constraint is not None:
>                 qconf_attr_value = getattr(qconfig, attr_name)
>                 if qconf_attr_value != constraint:
>                     return False
>         return True
> ```

I think it should be avoided, and the reason is the developer might accidentally forget to return a value from a function. In this case, Python will return None. This can cause subtle bugs that can be difficult to debug.

UsingtcNower: > What do you think about `return False` here?
> 
> ```
> def is_config_compatible(self, qconfig: QuantizerConfig) -> bool:
>         for attr_name, constraint in self.qconf_attr_vs_constraint_dict.items():
>             if constraint is not None:
>                 qconf_attr_value = getattr(qconfig, attr_name)
>                 if qconf_attr_value != constraint:
>                     return False
>         return True
> ```

Hi @alexsu52 thanks for your review. This looks good to me.

alexsu52 (approved):
Thanks for the contribution!

PR #2682, opened 2024-05-15
Remove compress_to_fp16=False from examples

truhinnm:
### Changes

Removed `compress_to_fp16=False` from `save_model` functions in examples.

### Reason for changes

Using this parameter when saving OpenVINO model is no longer required

### Tests

Changes were tested via pytest, using example tests in project. While testing, some errors occured, but all of them were about wrong development environment settings. Setting up the invironment took most of the time.

### DIFF
--- a/examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
+++ b/examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
@@ -168,7 +168,7 @@ def run_example():
     # Benchmark performance, calculate compression rate and validate accuracy
 
     fp32_ir_path = f"{ROOT}/stfpm_fp32.xml"
-    ov.save_model(ov_model, fp32_ir_path)
+    ov.save_model(ov_model, fp32_ir_path, compress_to_fp16=False)
     print(f"[1/7] Save FP32 model: {fp32_ir_path}")
     fp32_size = get_model_size(fp32_ir_path, verbose=True)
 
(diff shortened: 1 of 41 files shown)


examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
```diff
@@ -168,7 +168,7 @@ def transform_fn(data_item):
     # Benchmark performance, calculate compression rate and validate accuracy
 
     fp32_ir_path = f"{ROOT}/stfpm_fp32.xml"
-    ov.save_model(ov_model, fp32_ir_path, compress_to_fp16=False)
+    ov.save_model(ov_model, fp32_ir_path)
```
alexsu52, line 171: OpenVINO saves model weights in FP16 precision by default. If you remove `compress_to_fp16=False` here, the model weights will be saved in FP16. This leads at a minimum to a change in the size of the model and a change in the output metrics of the example. I would suggest not to remove `compress_to_fp16=False` for FP32 models here in another examples, so as not to change the behavior of the example.

alexsu52 (changes requested):
Thanks for your contribution!

Please check that output metrics of examples don't change after your changes.

examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py
truhinnm: Sorry, I've been busy all week, finally done it. Can you check, please?

alexsu52:
run example tests: 391

alexsu52 (approved):
Thanks for the contribution!

PR #2692, opened 2024-05-22
[OpenVINO] Fix bug when Add node in not 0-index child

kshpv:
### Changes

Make searching for an Add node among all children to determine whether the node has a bias.

### Reason for changes

Add node could be not only on 0-index of node children list.

### Related tickets

141885

### Tests

TBD

### DIFF
--- a/nncf/quantization/algorithms/channel_alignment/openvino_backend.py
+++ b/nncf/quantization/algorithms/channel_alignment/openvino_backend.py
@@ -34,8 +34,8 @@ from nncf.openvino.graph.metatypes.openvino_metatypes import OVMatMulMetatype
 from nncf.openvino.graph.metatypes.openvino_metatypes import OVSubtractMetatype
 from nncf.openvino.graph.node_utils import create_bias_tensor
 from nncf.openvino.graph.node_utils import get_bias_value
-from nncf.openvino.graph.node_utils import get_node_with_bias_value
 from nncf.openvino.graph.node_utils import get_weight_value
+from nncf.openvino.graph.node_utils import is_node_with_bias
 from nncf.openvino.graph.transformations.commands import OVTargetPoint
 from nncf.openvino.statistics.collectors import OVNNCFCollectorTensorProcessor
 from nncf.openvino.statistics.collectors import OVQuantileReducer
@@ -93,16 +93,7 @@ class OVChannelAlignmentAlgoBackend(ChannelAlignmentAlgoBackend):
 
     @staticmethod
     def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
-        next_nodes = nncf_graph.get_next_nodes(node)
-        if not next_nodes:
-            return False
-
-        add_node = next_nodes[0]
-        if add_node.metatype != OVAddMetatype:
-            return False
-
-        bias_constant = get_node_with_bias_value(add_node, nncf_graph)
-        return bias_constant is not None
+        is_node_with_bias(node, nncf_graph)
 
     @staticmethod
     def get_dims_descriptor(node: NNCFNode) -> LayoutDescriptor:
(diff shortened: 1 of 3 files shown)


MaximProshin:
@kshpv , I suppose it impacts the quantization time, right? Is the new position of Add expected/correct and will not impact the inference? 
cc @alexsu52 @AlexKoff88

nncf/quantization/algorithms/channel_alignment/openvino_backend.py
```diff
@@ -93,16 +93,7 @@ def get_statistic_collector(
...
-        if not next_nodes:
-            return False
-
-        add_node = next_nodes[0]
-        if add_node.metatype != OVAddMetatype:
-            return False
-
-        bias_constant = get_node_with_bias_value(add_node, nncf_graph)
-        return bias_constant is not None
+        return is_node_with_bias(node, nncf_graph)
```
alexsu52, line 96: It is bad practice to import a function into a module where a function with the same name exists. You can update the import as follows:
"""
from nncf.openvino.graph import node_utils
node_utils.is_node_with_bias
"""

alexsu52 (commented):
Could you provide graph example where node has the bias not on the 0 port?

kshpv:
> Could you provide graph example where node has the bias not on the 0 port?

The issue is not laying on the fact that _Add_ node is not connected to 0-index output port of a parent node. The problem is that the order of node's children at NNCFGraph does not guaranteed to be the same due the incorrect usage of OpenVINO API. During the NNCFGraph creation the order of children is dependent on the iteration order here - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/openvino/graph/nncf_graph_builder.py#L97
However, the return type of `get_target_inputs()` is `set` meaning that iteration order is not guaranteed to be the same. Because of that children nodes order is not the same from run to run.

kshpv:
> @kshpv , I suppose it impacts the quantization time, right? Is the new position of Add expected/correct and will not impact the inference? cc @alexsu52 @AlexKoff88

The impact on quantization time should be tiny. The difference in Add position in node's children list at NNCFGraph does not impact the inference time of a model.

kshpv:
> > Could you provide graph example where node has the bias not on the 0 port?
> 
> The issue is not laying on the fact that _Add_ node is not connected to 0-index output port of a parent node. The problem is that the order of node's children at NNCFGraph does not guaranteed to be the same due the incorrect usage of OpenVINO API. During the NNCFGraph creation the order of children is dependent on the iteration order here - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/openvino/graph/nncf_graph_builder.py#L97 However, the return type of `get_target_inputs()` is `set` meaning that iteration order is not guaranteed to be the same. Because of that children nodes order is not the same from run to run.

The last commit introduces sorting of return nodes from `get_target_inputs()`

nncf/quantization/algorithms/channel_alignment/openvino_backend.py
kshpv: done

nncf/openvino/graph/nncf_graph_builder.py
```diff
@@ -93,10 +93,13 @@ def _add_edges_to_nncf_graph(model: ov.Model, graph: NNCFGraph) -> None:
         for op in model.get_ops():
             in_node_id = graph.get_node_by_name(op.get_friendly_name()).node_id
             for output_port_id, out in enumerate(op.outputs()):
-                node_vs_target_inputs = defaultdict(list)
-                for inp in out.get_target_inputs():
-                    node_vs_target_inputs[inp.get_node()].append(inp)
-
+                node_vs_target_inputs = OrderedDict()
+                for inp in sorted(out.get_target_inputs(), key=lambda x: x.get_node().get_friendly_name()):
```
alexsu52, line 97: why do you sort inputs by name here?

kshpv: reverted

alexsu52 (approved):
LGTM, It looks like a hot fix.

PR #2696, opened 2024-05-23
Remove dependency from typing_extensions 

AlexanderDokuchaev:
### Changes

Remove dependency from typing_extensions 

### Reason for changes

Fail with `ImportError` on python>=3.9.

nncf.common depends from typing_extensions packages
https://github.com/openvinotoolkit/nncf/blob/47187e5bbaef678508fdfdc980b258fc4e32d3db/nncf/common/accuracy_aware_training/training_loop.py#L22-L23

With python3.8 typing_extensions installs as depends of rich module, but it works only on `python<3.9`
```
Collecting typing-extensions<5.0,>=4.0.0 (from rich>=13.5.2->nncf==2.11.0.dev0+1406e7f)
```
https://github.com/Textualize/rich/blob/349042fd8912ab5f0714ff9a46a70ef8a4be4700/pyproject.toml#L31

### Tests
tests/cross_fw/install (python 3.10)

### DIFF
--- a/nncf/common/accuracy_aware_training/training_loop.py
+++ b/nncf/common/accuracy_aware_training/training_loop.py
@@ -15,12 +15,10 @@ import pathlib
 from abc import ABC
 from abc import abstractmethod
 from functools import partial
-from typing import Callable, Dict, List, Optional, Tuple, TypedDict, TypeVar, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, TypeVar, Union, cast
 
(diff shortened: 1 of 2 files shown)


alexsu52:
Is it possible to remove this dependency? Could you try to replace type hints from `typing_extensions` to built-in type hints.

alexsu52 (commented):
LGTM

alexsu52 (approved):

PR #2717, opened 2024-06-04
Removal install package with extra from tests

AlexanderDokuchaev:
### Changes

Removal install package with extra from tests

### Reason

```
WARNING: nncf 2.11.0.dev0+unknown.version does not provide the extra 'torch'
```


alexsu52 (approved):
LGTM

PR #2719, opened 2024-06-04
Fix torch operator unpatching

nikita-savelyevv:
### Changes

- Add condition to skip operator wrapping if torch is in unpatched state
- Fix test from a previous PR #2665 (remove dependency of the test on `nncf.torch`)

### Reason for changes

Fix inference of compiled torch models when `nncf.torch` is imported

### Related tickets

140265

### Tests

Added `test_operator_unpatching`

### DIFF
--- a/nncf/torch/dynamic_graph/wrappers.py
+++ b/nncf/torch/dynamic_graph/wrappers.py
@@ -26,6 +26,7 @@ from nncf.torch.dynamic_graph.context import get_current_context
 from nncf.torch.dynamic_graph.layer_attributes_handlers import get_layer_attributes_from_args_and_kwargs
 from nncf.torch.dynamic_graph.op_input_processing import OperatorInput
 from nncf.torch.dynamic_graph.operation_address import OperationAddress
+from nncf.torch.dynamic_graph.patch_pytorch_state import PATCHING_STATE
 from nncf.torch.dynamic_graph.structs import NamespaceTarget
 from nncf.torch.dynamic_graph.structs import PatchedOperatorInfo
 from nncf.torch.dynamic_graph.trace_functions import forward_trace_only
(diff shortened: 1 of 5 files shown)


tests/torch/pytorch_patch_isolated.py
```diff
@@ -80,14 +80,24 @@ def test_jit_script_exception_preserves_patching_isolated():
 
 
 def compile_and_run_test_model() -> torch.Tensor:
-    from tests.torch.helpers import BasicConvTestModel
```
nikita-savelyevv, line 83: Had to drop `BasicConvTestModel` because during its import `nncf.torch` is imported which is not acceptable for this specific test

nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -76,6 +76,12 @@ def wrap_operator(operator, operator_info: PatchedOperatorInfo):
 
     @functools.wraps(operator)
     def wrapped(*args, **kwargs):
+        from nncf.torch.dynamic_graph.patch_pytorch import operators_are_wrapped
```
alexsu52, line 79: Please think about how to implement it without conditional import.

nikita-savelyevv: It will be possible if we create a separate file `patch_pytorch_utils.py` and put there some constants from `patch_pytorch.py` such as `_OPERATORS_ALREADY_WRAPPED`. Then both `patch_pytorch.py` and `wrappers.py` will import from `patch_pytorch_utils.py` as usual. Is that better?

alexsu52 (approved):
LGTM

nikita-savelyevv:
torch nightly build 256 is green

AlexanderDokuchaev (approved):

PR #2720, opened 2024-06-05
[CI] Set actions references to specific commit hash, provide explicit top level token permissions for jobs

AlexanderDokuchaev:
### Changes

Token permissions docs: https://github.com/ossf/scorecard/blob/98ec491a888a8a0db9d83a3c7d379ae1f46321de/docs/checks.md#token-permissions
Actions by hash docs: https://github.com/ossf/scorecard/blob/98ec491a888a8a0db9d83a3c7d379ae1f46321de/docs/checks.md#pinned-dependencies

### Related tickets

143130

### DIFF
--- a/.github/workflows/build_and_publish_doc.yml
+++ b/.github/workflows/build_and_publish_doc.yml
@@ -1,4 +1,6 @@
 name: Build complete docs and publish to GH Pages
+permissions: read-all
+
 on:
   push:
     branches:
@@ -11,8 +13,6 @@ concurrency:
   group: ci-${{ github.head_ref }}-${{ github.workflow }}
   cancel-in-progress: true
 
-permissions:
-  contents: write
 jobs:
   call-build-html-doc:
     uses: ./.github/workflows/build_html_doc.yml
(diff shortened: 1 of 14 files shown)


AlexanderDokuchaev:
```
| 10 / 10 | Token-Permissions 
| 6 / 10  | Pinned-Dependencies | Warn: pipCommand not pinned by hash:
```
To get 10/10 for Pinned-Dependencies. need to add hashes for all packages

.github/workflows/build_and_publish_doc.yml
```diff
@@ -23,20 +23,22 @@ jobs:
   publish:
     needs: [call-build-html-doc, call-build-schema-page]
     runs-on: ubuntu-20.04
+    permissions:
+      attestations: write
```
KodiaqQ, line 27: Do we need `write` for all steps below? Is it possible to grant permissions for each step?

AlexanderDokuchaev: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

I did not find in documentation anything about permission to specific step, only for a job.

PR #2722, opened 2024-06-06
bandit happy

alexsu52:
### Changes

`# nosec` is added because `urlretrieve` function is used with the hard code url

### Reason for changes

Suppress bandit issue

### Related tickets

ref: 143381

### Tests

build 862

### DIFF
--- a/examples/quantization_aware_training/torch/anomalib/main.py
+++ b/examples/quantization_aware_training/torch/anomalib/main.py
@@ -44,11 +44,7 @@ def download_and_extract(root: Path, info: download.DownloadInfo) -> None:
     downloaded_file_path = root / info.url.split("/")[-1]
     print(f"Downloading the {info.name} dataset.")
     with download.DownloadProgressBar(unit="B", unit_scale=True, miniters=1, desc=info.name) as progress_bar:
-        urlretrieve(
-            url=f"{info.url}",
-            filename=downloaded_file_path,
-            reporthook=progress_bar.update_to,
-        )
+        urlretrieve(url=f"{info.url}", filename=downloaded_file_path, reporthook=progress_bar.update_to)  # nosec
     print("Checking the hash of the downloaded file.")
     download.check_hash(downloaded_file_path, info.hashsum)
     print(f"Extracting the {info.name} dataset.")


examples/quantization_aware_training/torch/anomalib/main.py
```diff
@@ -44,11 +44,7 @@ def download_and_extract(root: Path, info: download.DownloadInfo) -> None:
     downloaded_file_path = root / info.url.split("/")[-1]
     print(f"Downloading the {info.name} dataset.")
     with download.DownloadProgressBar(unit="B", unit_scale=True, miniters=1, desc=info.name) as progress_bar:
-        urlretrieve(
-            url=f"{info.url}",
-            filename=downloaded_file_path,
-            reporthook=progress_bar.update_to,
-        )
+        urlretrieve(url=f"{info.url}", filename=downloaded_file_path, reporthook=progress_bar.update_to)  # nosec
```
KodiaqQ, line 47: Why can't we use `download_and_extract` as in [OpenVINO example](https://github.com/openvinotoolkit/nncf/blob/develop/examples/post_training_quantization/openvino/anomaly_stfpm_quantize_with_accuracy_control/main.py?)? Why should we use `urlretrieve`?

KodiaqQ:
Any successful runs?

examples/quantization_aware_training/torch/anomalib/main.py
KodiaqQ: Are there any other options instead of just putting `nosec`?

alexsu52:
> Any successful runs?

build 862

examples/quantization_aware_training/torch/anomalib/main.py
alexsu52: `urlretrieve` is in the black list of bandit https://bandit.readthedocs.io/en/1.7.2/blacklists/blacklist_calls.html, as far I understand, the risk of using `urlretrieve` is in allowing use of ‘file:’’ or custom schemes, but the example uses the hard code url and cann't be specified by user.

KodiaqQ (approved):

PR #2737, opened 2024-06-14
Excluded comprasion operations from quantizable operations

alexsu52:
### Changes

Excluded comparison operations from quantizable operations for transformer models

### Reason for changes

Quantization of comparison operations is lead to the runtime error:
```
RuntimeError: Exception from src/inference/src/cpp/infer_request.cpp:223:
Exception from src/plugins/intel_cpu/src/node.cpp:559:
Multiply node with name 'self_attn/aten::mul/Multiply' Exception from src/plugins/intel_cpu/src/shape_inference/custom/eltwise.cpp:45:
Eltwise shape infer input shapes dim index: 1 mismatch
```

### Related tickets

ref 143448

### Tests

N/A

### DIFF
--- a/nncf/quantization/algorithms/min_max/onnx_backend.py
+++ b/nncf/quantization/algorithms/min_max/onnx_backend.py
@@ -233,6 +233,10 @@ class ONNXMinMaxAlgoBackend(MinMaxAlgoBackend):
                 om.ONNXSqrtMetatype,
                 om.ONNXReciprocalMetatype,
                 om.ONNXBatchNormMetatype,
+                # Сomparison operations
+                om.ONNXGreaterMetatype,
+                om.ONNXLessMetatype,
+                om.ONNXEqualMetatype,
             ]
             if device != TargetDevice.CPU_SPR:
                 types.append(om.ONNXMulLayerMetatype)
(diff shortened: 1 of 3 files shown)


nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -233,6 +233,10 @@ def get_ignored_metatypes(model_type: ModelType, device: TargetDevice) -> List[O
                 om.ONNXSqrtMetatype,
                 om.ONNXReciprocalMetatype,
                 om.ONNXBatchNormMetatype,
+                # Сomparison operations
+                om.ONNXGreaterMetatype,
+                om.ONNXLessMetatype,
+                om.ONNXEqualMetatype,
```
KodiaqQ, line 239: Why only for `Transformer` models? Is there any reason we can't remove them from the configuration?

alexsu52: It's required wide validation. I filed a ticket to do it 144218.

KodiaqQ: Was there any validation for `Transformer` models at least?

alexsu52: run post_training_quantization: 402

PR #2746, opened 2024-06-18
[PTQ] MinMax make empty cache != initialized cache

kshpv:
### Changes

Make noninitialied cache a `None
`
### Reason for changes

To differentiate empty cache from noninitialized cache.

### Related tickets

143659

### Tests

N/A

### DIFF
--- a/tests/common/quantization/test_minmax.py
+++ b/tests/common/quantization/test_minmax.py
@@ -9,6 +9,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import collections
+import types
+
 import pytest
 
 import nncf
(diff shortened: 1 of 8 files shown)


tests/post_training/test_templates/test_min_max.py
```diff
@@ -166,3 +167,42 @@ def test_get_channel_axes_matmul_torch(self, weight_shape, ref_axes):
...
+    @abstractmethod
+    def get_backend():
+        "Returns backend."
+
+    @staticmethod
+    @abstractmethod
+    def get_no_quantized_ops_graph():
+        "Returns a NNCFGraph with no operations for quantization."
+
+    @pytest.mark.parametrize("run_nums", (2,))
```
AlexanderDokuchaev, line 183: Looks like it can be hardcoded in test body, no need to add extra parameter for one value

tests/onnx/quantization/test_min_max.py
```diff
@@ -79,3 +82,13 @@ def test_get_channel_axes_deptwiseconv_node_ov(self):
...
+
+
+class TestONNXCommonMinMax(TemplateTestCommonMinMax, TestONNXMinMaxAlgorithm):
+    @staticmethod
+    def get_backend():
+        return BackendType.ONNX
+
+    @staticmethod
+    def get_no_quantized_ops_graph():
+        return NNCFGraphToTest(ONNXIdentityMetatype).nncf_graph
```
AlexanderDokuchaev, line 94: Why used different metatypes as conv_metatype in NNCFGraphToTest?

And this tests looks like not backend specific that should check only cache. Could you patch functions test it without run backend specific code?

tests/post_training/test_templates/test_min_max.py
kshpv: done

tests/onnx/quantization/test_min_max.py
kshpv: made a test backend agnotic

tests/common/quantization/test_minmax.py
```diff
@@ -206,3 +209,33 @@ def test_mode_with_quantization_params(mode, activations_quantization_params, we
...
+        nonlocal called
+        called += 1
+        # Set up cache
+        self._quantization_target_points_to_qconfig = collections.OrderedDict()
+        self._unified_scale_groups = []
+        return self._quantization_target_points_to_qconfig, self._unified_scale_groups
+
+    run_nums = 2
+    algo = MinMaxQuantization()
+    MinMaxQuantization._find_quantization_target_points = types.MethodType(foo, algo)
```
AlexanderDokuchaev, line 234:
```suggestion
    algo._find_quantization_target_points = types.MethodType(foo, algo)
```

AlexanderDokuchaev (changes requested):

tests/common/quantization/test_minmax.py
kshpv: changed

AlexanderDokuchaev (approved):

PR #2747, opened 2024-06-18
Update Fast-/BC algorithms

KodiaqQ:
### Changes

- Updated FastBiasCorrection algorithm to correct work with transposed MatMul layers.
- Updated BiasCorrection algorithm to avoid recursion depth error.

### Reason for changes

- FastBiasCorrection algorithm adaptation.
- BiasCorrection improvement for many correctable layers.

### Related tickets

- 144240


nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -229,60 +229,62 @@ def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) ->
...
+                        continue
 
-            for input_edge in nncf_graph.get_input_edges(node):
-                fill_subgraph_input_ids(input_edge)
+                if node in visited_nodes:
+                    continue
+                visited_nodes.append(node)
+                edges_queue.extend(nncf_graph.get_input_edges(node))
 
+        nncf_logger.info(f"Processing subgraph data for {node.node_name} node")
```
alexsu52, line 276: Could you share the output of the BC before and after this change, to look a difference?

KodiaqQ: That was added for debugging purposes. Removed.

nncf/openvino/graph/node_utils.py
```diff
@@ -468,3 +468,39 @@ def get_weighted_layer_attributes(
...
+    if (
+        node.metatype == OVMatMulMetatype
+        and node.layer_attributes is not None
+        and node.layer_attributes.input_attributes is not None
+        and "transpose" in node.layer_attributes.input_attributes
+    ):
+        transpose = node.layer_attributes.input_attributes["transpose"]
+        channel_axis = calculate_port_based_channel_axis(port_id, transpose)
+
+    return channel_axis
```
kshpv, line 495: As I can see the main logic of this function is only used for `OVMatMulMetatype`. Also, the `port_id` is used only for `OVMatMulMetatype`. I suggest to reimplement this in a following way:
```suggestion
def get_activation_channel_axis(node: NNCFNode, port_id: Optional[int] = None) -> int:
    """
    Returns axis number of the activation tensor which correspond to it channel.

    :param node: NNCFNode instance.
    :param port_id: Specified input port id if needed.
    :return: Channel axis number.
    """
    channel_axis = 1

    if (
        node.metatype == OVMatMulMetatype
        and node.layer_attributes is not None
        and node.layer_attributes.input_attributes is not None
        and "transpose" in node.layer_attributes.input_attributes
    ):
        if port_id is None:
            raise nncf.InternalError("port_id is required for OVMatMulMetatype with transpose.")
        if port_id > 1:
            raise nncf.InternalError(f"{node.metatype.name} can not take more than 2 input tensors.")

        transpose = node.layer_attributes.input_attributes["transpose"]
        channel_axis = calculate_port_based_channel_axis(port_id, transpose)

    return channel_axis
```

Also, can we avoid these checks?
```
and node.layer_attributes is not None
and node.layer_attributes.input_attributes is not None
and "transpose" in node.layer_attributes.input_attributes
```

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -229,60 +229,61 @@ def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) ->
...
-
-            # Since we need to find the inputs for the subgraph,
-            # we can take only those layers for which we have already collected statistics.
-            if input_id in self._collected_stat_inputs_map:
-                activation_id = self._collected_stat_inputs_map[input_id]
-                if activation_id not in subgraph_output_ids:
-                    subgraph_input_ids.append(input_id)
-                    return
+        def fill_subgraph_output_ids(main_node: NNCFNode):
+            edges_queue = deque(nncf_graph.get_output_edges(main_node))
```
kshpv, line 233: It looks like you don't really use `deque` specific functions, so it makes sense to utilize `list` instead.

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -229,60 +229,61 @@ def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) ->
...
+            while edges_queue:
+                edge = edges_queue.pop()
+                node = edge.to_node
+                if node in visited_nodes:
+                    continue
+                visited_nodes.append(node)
```
kshpv, line 239: It makes sense to use a `set` for `visited_nodes`

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -229,60 +229,61 @@ def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) ->
...
-            input_id = (edge.to_node.node_name, edge.input_port_id)
-
-            # Since we need to find the inputs for the subgraph,
-            # we can take only those layers for which we have already collected statistics.
-            if input_id in self._collected_stat_inputs_map:
-                activation_id = self._collected_stat_inputs_map[input_id]
-                if activation_id not in subgraph_output_ids:
-                    subgraph_input_ids.append(input_id)
-                    return
+        def fill_subgraph_output_ids(main_node: NNCFNode):
```
kshpv, line 232: Could you add small descriptions of these functions?

kshpv (changes requested):

nncf/openvino/graph/node_utils.py
KodiaqQ: This is not correct.

> is only used for OVMatMulMetatype

This is also used for the `Convolution` metatype, for example.

> Also, the port_id is used only for OVMatMulMetatype

This check also verifies `Convolution`, since it may take only 2 input ports.

> Also, can we avoid these checks?

We can't avoid it since it is also applicable for `Convolution`.

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: These methods aren't new. There are comments that exists for `_get_subgraph_data_for_node` at all.

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Sure.

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Sure.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -95,3 +95,7 @@ def is_node_with_bias(node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
     @staticmethod
     def get_node_names_for_input_output_statistics(node: NNCFNode, nncf_graph: NNCFGraph) -> Tuple[str, str]:
         return node.node_name, node.node_name
+
+    @staticmethod
+    def get_activation_channel_axis(node: NNCFNode, port_id: int) -> int:
+        return node.metatype.output_channel_axis
```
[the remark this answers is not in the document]
KodiaqQ: There are no examples of such models from the ONNX backend. If you have any, share it, please.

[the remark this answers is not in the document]
kshpv: It looks like you leave a technical debt for you or your colleagues.
From the provided ticket you can try to convert the model into ONNX and try to reproduce. Otherwise, you can create a synthetic model.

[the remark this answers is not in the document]
KodiaqQ: Do you know any options to convert IR model into ONNX?

nncf/quantization/algorithms/bias_correction/algorithm.py
kshpv: It will make the understanding of these functions easier. Also, it will help to understand whether these two functions can be simplified and merged into one.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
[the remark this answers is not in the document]
kshpv: The original model came from PyTorch, am I right?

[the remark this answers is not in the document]
KodiaqQ: I see what you mean. Give me some time to completely dive into the way of getting the model and then convert it.
I hope that @alexsu52 and the customers are OK with that postponement.
ONNX and PyTorch model's support with these options is crucial for us, as @kshpv states.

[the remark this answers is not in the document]
kshpv: Another option could be to separate work into two parts: initially to add support for OpenVINO and then to add the same for ONNX and PyTorch, but the final decision is up to you

nncf/openvino/graph/node_utils.py
```diff
@@ -468,3 +468,36 @@ def get_weighted_layer_attributes(
...
+    :param port_id: Specified input port id if needed.
+    :return: Channel axis number.
+    """
+    channel_axis = 1
+
+    if node.metatype == OVMatMulMetatype:
+        if port_id is None:
+            raise nncf.InternalError("port_id is required for OVMatMulMetatype with transpose.")
+        if port_id > 1:
+            raise nncf.InternalError(f"{node.metatype.name} can not take more than 2 input tensors.")
```
daniil-lyakhov, line 487: I suggest to make a `get_linear_weights_layout_from_node` function in the same fashion as the `get_linear_weights_layout_from_node` function.

This should help to write tests quickly as  tests for `get_linear_weights_layout_from_node` should be similar to this one

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -229,60 +228,61 @@ def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) ->
...
-                return
-
-            for output_edge in nncf_graph.get_output_edges(node):
-                fill_subgraph_output_ids(output_edge)
+        def fill_subgraph_output_ids(main_node: NNCFNode):
+            edges_queue = nncf_graph.get_output_edges(main_node)
+            while edges_queue:
+                edge = edges_queue.pop()
+                node = edge.to_node
+                if node in visited_nodes:
```
daniil-lyakhov, line 247: Too many indents IMHO, I would put while body here and in the `fill_subgraph_input_ids` in a separate function

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
```diff
@@ -107,3 +107,7 @@ def get_node_names_for_input_output_statistics(node: NNCFNode, nncf_graph: NNCFG
         next_norm_node = get_potential_fused_node(input_node_name, nncf_graph)
         output_node_name = next_norm_node.node_name if next_norm_node else input_node_name
         return input_node_name, output_node_name
+
+    @staticmethod
+    def get_activation_channel_axis(node: NNCFNode, port_id: int) -> int:
+        return node.metatype.output_channel_axis
```
daniil-lyakhov, line 113: Please don't forget to test this

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Done.

nncf/quantization/algorithms/fast_bias_correction/torch_backend.py
KodiaqQ: There was nothing new introduced.

nncf/openvino/graph/node_utils.py
KodiaqQ: As you wish. Done.

nncf/quantization/algorithms/bias_correction/algorithm.py
KodiaqQ: Done.

nncf/openvino/graph/node_utils.py
daniil-lyakhov: Nice! Thanks a lot!

daniil-lyakhov (approved):

nncf/quantization/algorithms/bias_correction/algorithm.py
daniil-lyakhov: Nice!

KodiaqQ:
@alexsu52, @daniil-lyakhov, @kshpv, review, please.

nncf/onnx/graph/node_utils.py
```diff
@@ -209,3 +213,20 @@ def get_quantized_tensor_shape(
...
+    :param node: NNCFNode instance.
+    :param port_id: Port ID.
+    :return: Channel axis number.
+    """
+    # Gemm metatype supports only 2D inputs according to the documentation -
+    # https://onnx.ai/onnx/operators/onnx__Gemm.html
+    gemm_shape = [0, 1]
+    trans_attr = "transA" if not port_id else "transB"
+    transpose = node.layer_attributes.node_attrs[trans_attr]
+    channel_axis = -1 - port_id if transpose else -2 + port_id
```
kshpv, line 231: I agree that this approach is now aligned with OpenVINO, but why did you choose the logic of axis processing of OpenVINO instead of ONNX? For me, it seems that ONNX supports all input shapes, while OV doesn't. But maybe I am wrong...

nncf/onnx/graph/nncf_graph_builder.py
```diff
@@ -176,10 +176,13 @@ def _get_node_attrs(node: onnx.NodeProto, model: onnx.ModelProto) -> Dict[str, A
     :param model: ONNX model.
     :return : Node attributes.
     """
+    # We need to collect list of the inputs.
```
kshpv, line 179: I think you can safely remove these comments or put them in the docstring in a way that describes the content of attribute dict.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
```diff
@@ -78,7 +80,19 @@ def get_bias_value(node: NNCFNode, nncf_graph: NNCFGraph, model: onnx.ModelProto
 
     @staticmethod
     def get_activation_port_ids_for_bias_node(node: NNCFNode) -> Tuple[int, int]:
-        return 0, 0
+        layer_attrs = node.layer_attributes
+        inputs = deepcopy(layer_attrs.node_attrs["inputs"])
```
kshpv, line 84: Why do you utilize input names instead of metatypes attributes such as `bias_port_id` and `weight_port_ids`? It will free storing 'inputs' for each node in its attribute.

nncf/quantization/algorithms/bias_correction/algorithm.py
```diff
@@ -216,24 +215,28 @@ def _is_node_correctable(self, node: NNCFNode, nncf_graph: NNCFGraph) -> bool:
             node, nncf_graph
         )
 
-    def _get_subgraph_data_for_node(self, node: NNCFNode, nncf_graph: NNCFGraph) -> Dict[str, List[str]]:
+    def _find_subgraph_output_ids(self, nncf_graph: NNCFGraph, main_node: NNCFNode) -> List[Tuple[str, int]]:
```
kshpv, line 218: Can be these two traversal functions merged into one single traverse from a node with a bias up to collected statistics?

kshpv (changes requested):

nncf/onnx/graph/node_utils.py
```diff
@@ -209,3 +213,20 @@ def get_quantized_tensor_shape(
...
+
+def calculate_gemm_channel_axis(node: NNCFNode, port_id: int) -> int:
+    """
+    Calculates Gemm channel axis based on the port and node attributes.
+
+    :param node: NNCFNode instance.
+    :param port_id: Port ID.
+    :return: Channel axis number.
+    """
+    # Gemm metatype supports only 2D inputs according to the documentation -
```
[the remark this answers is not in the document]
alexsu52: @kshpv, please, file a ticket for this comment.

nncf/quantization/algorithms/bias_correction/algorithm.py
alexsu52: It was implemented in `LayerWiseEngine` and can be used in BC.

nncf/quantization/algorithms/fast_bias_correction/onnx_backend.py
alexsu52: @kshpv could you clarify your idea? The code snippet will be good.

kshpv: I see that for some metatypes such as Gemm the activation and weight ports are not determined. So this approach is applicable then. 
However, the new proposed layer attribute is not necessary for this function. Because in `layer_attrs` there is a `weight_attrs` mapping from `port_id` to a weight info. 
So, the information of the weight port (from `weight_attrs`) and bias port (from metatype) is enough for detecting the activation port, am I right?
If this is correct my suggestion is to remove adding new layer_attrs in these changes for ONNX
@alexsu52

nncf/onnx/graph/node_utils.py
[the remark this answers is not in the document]
kshpv: One more corner case is to add support of FBC or BC algorithm for MatMul -> Add subgraph for ONNX. From a specification MatMul could have arbitrary dimensions https://onnx.ai/onnx/operators/onnx__MatMul.html

kshpv (approved):

PR #2759, opened 2024-06-21
Fix default value for ignored scope in calibrate.py

kshpv:
### Changes

Change None to []

### Reason for changes

Incorrect API usage

### Related tickets

144809

### Tests

Manually tested on failed scenario

### DIFF
--- a/tests/openvino/tools/calibrate.py
+++ b/tests/openvino/tools/calibrate.py
@@ -349,7 +349,7 @@ def map_ignored_scope(ignored):
             if op.get("attributes") is not None:
                 raise ValueError('"attributes" in the ignored operations ' "are not supported")
             ignored_operations.append(op["type"])
-    return {"ignored_scope": IgnoredScope(names=ignored.get("scope"), types=ignored_operations)}
+    return {"ignored_scope": IgnoredScope(names=ignored.get("scope", []), types=ignored_operations)}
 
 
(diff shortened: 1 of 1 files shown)


tests/openvino/tools/calibrate.py
```diff
@@ -349,7 +349,7 @@ def map_ignored_scope(ignored):
             if op.get("attributes") is not None:
                 raise ValueError('"attributes" in the ignored operations ' "are not supported")
             ignored_operations.append(op["type"])
-    return {"ignored_scope": IgnoredScope(names=ignored.get("scope"), types=ignored_operations)}
+    return {"ignored_scope": IgnoredScope(names=ignored.get("scope", []), types=ignored_operations)}
```
alexsu52, line 352: Why is this a mistake? It looks like NNCF should support empty list in `IgnoredScope`.

alexsu52 (changes requested):

tests/openvino/tools/calibrate.py
kshpv: > It looks like NNCF should support empty list in IgnoredScope.

I agree, so the default value was changed accordingly from None to an empty list.

alexsu52 (approved):
LGTM

KodiaqQ (approved):

Issue #2562, opened 2024-03-08
[Good First Issue][NNCF]: Dump actual_subset_size to ov.Model


l-bat:
### Context

After applying quantization to the ov.Model in Neural Network Compression Framework ([NNCF](https://github.com/openvinotoolkit/nncf/tree/develop)), the quantization parameters, including `subset_size`, are dumped to the meta section of the OpenVINO IR. `subset_size` represents the size of the dataset used for calibration.
https://github.com/openvinotoolkit/nncf/blob/09960b9b71a58f3277d0964531f8ee08365d7c72/nncf/openvino/quantization/quantize_model.py#L102
 
However, inconsistencies arise when the dataset size is less than the provided or default 'subset_size'. To address this confusion, it is proposed to also dump the `actual_subset_size`, which denotes the number of data samples used to calculate activation statistics. This addition will improve clarity and accuracy in managing quantization parameters and assist in reproducing quantization results.

### What needs to be done?

* Dump `actual_subset_size` parameter to ov.Model meta section.
* Add tests

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)

### Contact points

@l-bat

AiGaf1:
.take

andrey-churkin:
@AiGaf1 Hi, are you still working on this task? Do you need any help? Please inform us if you do not plan to continue working on this task. Thanks!

RitikaxShakya:
Hello! Is this task completed? If not i wish to work on this issue.

p-wysocki:
I reopened the issue due to current assignee's inactivity.

@RitikaxShakya I assigned you.

p-wysocki:
Hello @RitikaxShakya, are you still working on that issue? Do you need any help?

awayzjj:
.take

awayzjj:
Hi @l-bat I created a PR after testing locally, and the XML output is as expected:
<img width="705" alt="截屏2024-06-29 11 46 58" src="https://github.com/openvinotoolkit/nncf/assets/38181615/4509f628-042f-4720-b288-b7e0b5516d93">

I ran the pytest in tests/openvino, and the original tests did not break. But I have 2 questions
1. which file I should edit to add my unit test. 
2. how to implement the unit test, should I check the output XML to verify whether the `actual_subset_size` property exists?

Thank you very much!

l-bat:
Hi @awayzjj! 
[...]

PR #2769, opened 2024-06-29
dump actual_subset_size

awayzjj:
### Changes

Dump actual_subset_size to ov.Model

### Reason for changes

Inconsistencies arise when the dataset size is less than the provided or default 'subset_size'. 

### Related tickets
Closes: https://github.com/openvinotoolkit/nncf/issues/2562
### Tests
WIP


awayzjj:
@l-bat HI, could you please help me review the PR?

nncf/openvino/quantization/quantize_model.py
```diff
@@ -57,6 +57,13 @@
 TTensor = TypeVar("TTensor")
 
 
+def get_actual_subset_size(dataset, subset_size):
+    dataset_length = dataset.get_length()
+    if dataset_length:
+        return min(dataset_length, subset_size)
+    return subset_size
```
l-bat, line 64: if `__len()__` is not implemented, the actual subset size should be obtained as:
```python
actual_subset_size = 0
for data in islice(calibration_dataset.get_inference_data(), subset_size):
    actual_subset_size += 1
```

tests/openvino/native/quantization/test_quantization_pipeline.py
```diff
@@ -197,3 +197,20 @@ def test_ignored_scope_dump(ignored_options, expected_dump, tmp_path):
             assert dumped_model.get_rt_info(rt_path) == value
         else:
             assert dumped_model.has_rt_info(rt_path) is False
+
+
+@pytest.mark.parametrize("subset_size, expected_actual_subset_size", [[1, 1], [2, 1]])
+def test_dump(subset_size, expected_actual_subset_size, tmp_path):
```
l-bat, line 203: Please add 2 tests:
1. `__len()__` implemented
2.  `__len()__` not implemented

mlukasze:
hey @awayzjj will you continue work on this?

awayzjj:
@mlukasze Yes, sorry for the delay, I will finish it this weekend :)

nncf/openvino/quantization/quantize_model.py
awayzjj: Sorry for the late reply. I found that the suggested changes do not work because the `calibration_dataset.get_inference_data()` iterator is consumed before dump_parameters. As a result, `actual_subset_size` ends up being zero. Could you please provide some suggestions on how to handle this issue?
Thank you very much! @l-bat

l-bat: @awayzjj after discussing within the team, we decided **not** to add `actual_subset_size` to `rt_info` because `rt_info` is intended to contain quantization parameters provided to algorithms. If the provided dataset is shorter than `subset_size`, we should display a warning during the statistics collection process: https://github.com/openvinotoolkit/nncf/blob/408c67dd67fadd20b4d76ce4bc20c2b215258720/nncf/common/tensor_statistics/aggregator.py#L76

l-bat:
The contributor has decided to unassign themselves from this task

PR #2663, opened 2024-05-03
[OpenVINO] Update models for tests 

kshpv:
### Changes

Replace model at tests from OMZ models to analogs at `tests/torch/test_models`.
All graph, scales references were updated accordingly.
New dependency for OV tests were added: `timm`, `addict`, `efficientnet_pytorch`

### Reason for changes

To vanish places of OMZ usage.

### Related tickets

136314

### Tests

All references were updated

PR #2779, opened 2024-07-02
[OV] Clean test environment

AlexanderDokuchaev:
### Changes

Removed `tensorflow` from test environment.

### Reason for changes

After https://github.com/openvinotoolkit/nncf/pull/2663 module is not required in tests. It was need for yolov4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,7 +64,6 @@ install-openvino-test:
 	pip install -U pip
 	pip install -e .
 	pip install "git+https://github.com/openvinotoolkit/open_model_zoo.git@37f60eb#egg=accuracy_checker&subdirectory=tools/accuracy_checker"
-	pip install tensorflow==2.12.0 # Install tensorflow before to avoid conflict on install for typing-extensions
 	pip install -r tests/openvino/requirements.txt
 	pip install -r tests/cross_fw/install/requirements.txt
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -64,7 +64,6 @@ install-openvino-test:
 	pip install -U pip
 	pip install -e .
 	pip install "git+https://github.com/openvinotoolkit/open_model_zoo.git@37f60eb#egg=accuracy_checker&subdirectory=tools/accuracy_checker"
-	pip install tensorflow==2.12.0 # Install tensorflow before to avoid conflict on install for typing-extensions
```
KodiaqQ, line 67: Have you checked that `install-openvino-test` make command uses only for pre-commit tests and does not affect other internal jobs?

KodiaqQ: Thank you for the explanation. 
My question was about the consequences of this change. Are there any jobs (including internal runs) that utilize `install-openvino-test` make command that may fail without TF?

AlexanderDokuchaev: As i have wrote i dont found any code that required TF. As owner of OV part may be you know more that me.
Please provide list of jobs that may fail.

alexsu52 (approved):

KodiaqQ (approved):

PR #2784, opened 2024-07-04
Support the scale estimation inside GPTQ

alexsu52:
### Changes
- The scale estimation algorithm is used to calculate quantization parameters inside GPTQ.
- Combination GPTQ and AWQ algorithm is supported.
- The layer-wise scheduler is deterministic.

### Reason for changes

Improve model compression accuracy for low bits.
image

### Related tickets

ref: 143637

### Tests

job/manual/job/post_training_weight_compression/152/

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -124,7 +124,12 @@ class WeightCompression(Algorithm):
 
         if self._gptq:
             gptq_params = self._advanced_parameters.gptq_params
-            self._gptq_algo = GPTQ(gptq_params.damp_percent, gptq_params.block_size, gptq_params.subset_size)
+            self._gptq_algo = GPTQ(
+                damp_percent=gptq_params.damp_percent,
+                block_size=gptq_params.block_size,
+                subset_size=gptq_params.subset_size,
+                scale_estimation=self._scale_estimation,
+            )
             self._gptq_statistics = None
 
     @property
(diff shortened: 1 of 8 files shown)


nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -542,7 +547,7 @@ def _get_activations(
         statistics_aggregator = StatisticsAggregatorFactory.create(model, dataset)
         statistics_aggregator.register_statistic_points(statistic_container)
 
-        if self._gptq:
+        if self._gptq and not self._awq:
             self._gptq_statistics = self._gptq_algo.get_statistic_points(
                 model, graph, nodes_to_compress, self._backend_entity
```
ljaljushkin, line 552: Are statistics collected twice for AWQ, Scale Estimation and GPTQ? can it be reused?

alexsu52: GPTQ statistics include only statistics for the first step. The layer-wise scheduler specify statistic points taking into account parameters of scheduler. In common case, statistics points can not be matched to AWQ and Scale Estimation statistic points. I mean GPTQ adds additional statistics points. 

If we talk about the same statistics, in practice they will be merged because `StatisticsAggregatorFactory` has functionality of merging statistics. I guess, that this doesn't result in additional memory usage.

Let's discuss offline, if my answer is not clear.

ljaljushkin (approved):

andreyanufr (approved):
Dev SE and PR SE has the same ppl numbers.

PR #2787, opened 2024-07-04
[Torch] Use metatypes instead of types in pattern matching

daniil-lyakhov:
### Changes

* ~~GraphPattern type/metatype attrs are renamed~~
* `Metatypes` are used to identify type of a node instead of `type`
* Elementwise metatypes updated with inplace types

### Reason for changes

As number of types is going to grow (due to new TorchFX backend), it is easier to specify metatype one time instead of specifying types, which should be updated each time new type is arrived. 

### Related tickets

145981


nncf/common/graph/patterns/patterns.py
```diff
@@ -78,8 +78,8 @@ class GraphPattern:
     """
 
     LABEL_ATTR = "label"
-    METATYPE_ATTR = "type"
-    NODE_TYPE_ATTR = "metatype"
+    TYPE_ATTR = "type"
+    METATYPE_ATTR = "metatype"
     ANY_PATTERN_NODE_TYPE = "ANY_PATTERN_NODE"
```
daniil-lyakhov, line 83: That's confusing 🤔

daniil-lyakhov:
@kshpv please advice what should we do with this patterns: `nncf/experimental/onnx/hardware/fused_patterns.py`

kshpv (commented):
I would ask to run PTQ validation for these changes to be sure that there is no regression with patterns

daniil-lyakhov:
> I would ask to run PTQ validation for these changes to be sure that there is no regression with patterns

Build post_training_quantization/422/ is OK

kshpv (approved):

nncf/torch/graph/pattern_operations.py
```diff
@@ -10,54 +10,82 @@
...
+        om.PTDepthwiseConv1dSubtype,
+        om.PTModuleConv1dMetatype,
+        om.PTModuleDepthwiseConv1dSubtype,
+        # Conv2D
+        om.PTConv2dMetatype,
+        om.PTDepthwiseConv2dSubtype,
+        om.PTModuleConv2dMetatype,
+        om.PTModuleDepthwiseConv2dSubtype,
+        # Conv3D
+        om.PTConv2dMetatype,
```
AlexanderDokuchaev, line 31:
```suggestion
```

dublicate

nncf/torch/graph/pattern_operations.py
```diff
@@ -10,54 +10,82 @@
...
-    GraphPattern.METATYPE_ATTR: ["layer_norm"],
+    GraphPattern.METATYPE_ATTR: [om.PTLayerNormMetatype, om.PTModuleLayerNormMetatype],
     GraphPattern.LABEL_ATTR: "LAYER_NORMALIZATION",
 }
 
-RELU_OPERATIONS = {GraphPattern.METATYPE_ATTR: ["relu", "relu_", "hardtanh"], GraphPattern.LABEL_ATTR: "RELU"}
+RELU_OPERATIONS = {
+    GraphPattern.METATYPE_ATTR: [
+        om.PTRELUMetatype,
+        om.PTRELU6Metatype,
```
AlexanderDokuchaev, line 72: Adding new operation `relu6` to patterns, is it correct changes?

AlexanderDokuchaev (changes requested):

nncf/torch/graph/pattern_operations.py
daniil-lyakhov: That is a mistake, thanks for the catch!

daniil-lyakhov:
@AlexanderDokuchaev, I've checked metatypes one more time and have applied your comments, please check

AlexanderDokuchaev (approved):

PR #2790, opened 2024-07-05
Scale for covering full range of compressed weights.

andreyanufr:
### Changes

Implemented negative scale for symmetric compression to cover cases when positive weight has more wide range of values and now we can provide 8 bits for this case instead of 7.

### Reason for changes

Improvement in data-free accuracy.

### Related tickets

CVS-146491

### Tests

See changed files


nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -155,6 +155,31 @@ def calculate_e2m1_scale(weight: Tensor, reduction_axes: ReductionAxes, max_val=
     return scale
 
 
+def calculate_signed_scale(weight: Tensor, reduction_axes: ReductionAxes, num_bits=4) -> Tensor:
+    """
+    Calculates the scale for e2m1 quantization.
```
AlexKoff88, line 160: Is this message for "e2m1" relevant here?

AlexKoff88:
@xiao1228, @ZoranZomborat, can you please confirm if negative quantization scales are suitable for NPU? This is relevant only for weight quantization and asymmetric scheme. Asymmetric quantization will remain the same.

ZoranZomborat:
> @xiao1228, @ZoranZomborat, can you please confirm if negative quantization scales are suitable for NPU? This is relevant only for weight quantization and asymmetric scheme. Asymmetric quantization will remain the same.

for all NPU HW revisions we do support negative quantization scales;
we're more limited to items like:

- only per tensor zero point;
- zero point to be positive (need for the activation range to have zero included)
- have symmetric quantization (zp=0)  for types like i8/i4

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

ljaljushkin (changes requested):
Does it change metrics in conformance tests? please update them if yes

andreyanufr:
> Does it change metrics in conformance tests? please update them if yes

yes, it changes: post_training_weight_compression 125
I will update references.

ljaljushkin (approved):

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -155,6 +155,31 @@ def calculate_e2m1_scale(weight: Tensor, reduction_axes: ReductionAxes, max_val=
...
+    :param weight: Weight array to compress.
+    :param reduction_axes: Axes along which to reduce (collect) different statistics (e.g., min, max).
+    :param num_bits: number of bits in compression.
+    :return: Scale tensor.
+    """
+    level_high = 2 ** (num_bits - 1)
+    scale = fns.max(fns.abs(weight), axis=reduction_axes, keepdims=True)  # [a1, r//gs, 1, a2]
+
+    w_min = fns.abs(fns.min(weight, axis=reduction_axes, keepdims=True))
+    w_max = fns.abs(fns.max(weight, axis=reduction_axes, keepdims=True))
```
alexsu52, line 171: Could you check this implementation by performance, memory usage and accuracy?

```suggestion
    w_min = fns.min(weight, axis=reduction_axes, keepdims=True)
    w_max = fns.max(weight, axis=reduction_axes, keepdims=True)
    
    scale = fns.where(-w_min > w_max, w_min, w_max)
    scale /= level_high
    
    eps = fns.finfo(scale).eps
    scale = fns.where(fns.abs(scale) < eps, eps, scale)
```

tests/openvino/native/quantization/test_gptq.py
```diff
@@ -95,7 +95,15 @@ def find_params(self, x, weight=False):
             self.zero = xmin
         else:
             if self.sym:
-                self.scale = (xmax - xmin) / (self.maxq - 1)
+                maxq2 = self.maxq // 2 + 1
```
alexsu52, line 98: Please, simplify this code.

alexsu52:
@andreyanufr, please add PR description as well.

AlexKoff88 (approved):

AlexKoff88:
I discussed the possibility of using negative scales and all CPU, GPU, NPU architects agreed that this is possible. I suggest merging this PR as it is in high readiness. The main reason is that it fixes an incorrect quantization scheme for symmetric 4-bit quantization. The rest of the comments can be resolved post merge.

@alexsu52, please merge.

alexsu52 (approved):
Comments will be implemented in the follow-up PR.

PR #2802, opened 2024-07-10
Use a graph to find the output port id for the activation quantization target point

andrey-churkin:
### Changes

Before the changes in the PR, it was possible to insert the quantize operation after some layer only on its first output (`output port_id = 0`). The PR adds changes that allow the insertion of the quantize operation for operations with multiple outputs (they have different output port IDs), but only when one is used.

### Reason for changes

The selected FQ operations are not inserted unless changes are made.
MatMul_problem

### Related tickets

- 146088

### Tests

pre-commit scope

### DIFF
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -348,7 +348,7 @@ class NNCFGraph:
             edges.extend(self._get_edges(from_node, node))
         return sorted(edges, key=lambda x: x.input_port_id)
 
-    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> Optional[NNCFGraphEdge]:
+    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> NNCFGraphEdge:
         """
         Returns the input edge for a given node, where edge.input_port_id == port_id is True.
 
@@ -358,12 +358,16 @@ class NNCFGraph:
             given node.
         """
         edges = [e for e in self.get_input_edges(node) if e.input_port_id == port_id]
+        if len(edges) == 0:
+            raise nncf.ValidationError(
+                f"Node {node.node_name} does not contain input edge connected to {port_id} port ID."
+            )
+
         if len(edges) > 1:
             raise nncf.InternalError(
                 "Unsupported graph. More than one edge was found for a given node by the specified input port ID."
             )
-        edge = None if len(edges) == 0 else edges[0]
-        return edge
+        return edges[0]
 
     def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
         """
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -348,7 +348,7 @@ class NNCFGraph:
             edges.extend(self._get_edges(from_node, node))
         return sorted(edges, key=lambda x: x.input_port_id)
 
-    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> NNCFGraphEdge:
+    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> Optional[NNCFGraphEdge]:
         """
         Returns the input edge for a given node, where edge.input_port_id == port_id is True.
 
@@ -360,9 +360,10 @@ class NNCFGraph:
         edges = [e for e in self.get_input_edges(node) if e.input_port_id == port_id]
         if len(edges) > 1:
             raise nncf.InternalError(
-                "Unsupported graph. More than one edge was found for a given " "node by the specified input port ID."
+                "Unsupported graph. More than one edge was found for a given node by the specified input port ID."
             )
-        return edges[0]
+        edge = None if len(edges) == 0 else edges[0]
+        return edge
 
     def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
         """
--- a/nncf/common/graph/graph.py
+++ b/nncf/common/graph/graph.py
@@ -348,19 +348,46 @@ class NNCFGraph:
             edges.extend(self._get_edges(from_node, node))
         return sorted(edges, key=lambda x: x.input_port_id)
 
+    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> NNCFGraphEdge:
+        """
+        Returns the input edge for a given node, where edge.input_port_id == port_id is True.
+
+        :param node: The node for which to retrieve the input edge.
+        :param port_id: The ID of the input port to filter the edges.
+        :return: An input edge connected to the specified input port ID of the
+            given node.
+        """
+        edges = [e for e in self.get_input_edges(node) if e.input_port_id == port_id]
+        if len(edges) > 1:
+            raise nncf.InternalError(
+                "Unsupported graph. More than one edge was found for a given " "node by the specified input port ID."
+            )
+        return edges[0]
+
     def get_output_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
         """
         Returns edges of output tensors sorted by output port ID.
 
         :param node: Producer node.
-        :return:  List of output edges for the node sorted by output port ID.
+        :return: List of output edges for the node sorted by output port ID.
         """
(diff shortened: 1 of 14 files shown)


nncf/quantization/algorithms/min_max/openvino_backend.py
```diff
@@ -138,10 +138,17 @@ def create_convert_insertion_command(
...
+
         if target_point.type == TargetType.PRE_LAYER_OPERATION:
-            return nncf_graph.get_input_edges(node)[target_point.port_id].tensor_shape
+            edges = [e for e in nncf_graph.get_input_edges(node) if e.input_port_id == target_point.port_id]
+            return edges[0].tensor_shape
         elif target_point.type == TargetType.POST_LAYER_OPERATION:
-            return nncf_graph.get_output_edges(node)[target_point.port_id].tensor_shape
+            edges = [e for e in nncf_graph.get_output_edges(node) if e.output_port_id == target_point.port_id]
+            return edges[0].tensor_shape
+
```
andrey-churkin, line 151: The following statement

```python
nncf_graph.get_output_edges(node)[target_point.port_id]
```
it is an incorrect usage of the API because the index in the `nncf_graph.get_output_edges(node)` list is not the same as `target_point.port_id`. For example, the `nncf_graph.get_output_edges(node)` list may contain only one edge that connects `output_port_id = 1` of a `node` with some `input_port_id` of node `U` (LSTMSequence case). In that case, `target_point.port_id = 1`, but the list contains only one element with index 0.

nncf/onnx/graph/node_utils.py
```diff
@@ -171,10 +171,15 @@ def _get_activation_tensor_shape(
     :param target_point: Determines from input or ouput of a node take a shape info.
     :return: None, if there is no shape info, otherwise - tensor shape.
     """
+    # NOTE: Assumes that all output/input edges for the `node` with `output_port_id`/`input_port_id`
+    # equal to `target_point.port_id` should have the same `tensor_shape` value.
+
     if target_point.type == TargetType.PRE_LAYER_OPERATION:
-        shape = nncf_graph.get_input_edges(node)[target_point.port_id].tensor_shape
```
KodiaqQ, line 175: Just to note. There are several places in the NNCF using this approach.

nncf/onnx/graph/node_utils.py
```diff
@@ -171,10 +171,15 @@ def _get_activation_tensor_shape(
...
+        edges = [e for e in nncf_graph.get_input_edges(node) if e.input_port_id == target_point.port_id]
+        shape = edges[0].tensor_shape
```
KodiaqQ, line 179: Are we able to wrap this and introduce the new method in the NNCFGraph to get a needed edge by port ID? Minor, just about better code re-usability.

andrey-churkin: Good suggestion! I've added two methods `get_input_edge_by_port_id()` and ` get_output_edges_by_port_id()`.

nncf/onnx/graph/node_utils.py
andrey-churkin: Yes, I've changed probably all of them now.

nncf/common/graph/graph.py
```diff
@@ -348,20 +348,48 @@ def get_input_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
...
+        :param port_id: The ID of the input port to filter the edges.
+        :return: An input edge connected to the specified input port ID of the
+            given node.
+        """
+        edges = [e for e in self.get_input_edges(node) if e.input_port_id == port_id]
+        if len(edges) > 1:
+            raise nncf.InternalError(
+                "Unsupported graph. More than one edge was found for a given node by the specified input port ID."
+            )
+        edge = None if len(edges) == 0 else edges[0]
```
andrey-churkin, line 365: `None` is required here to handle case when a `node` does not have any input edges

KodiaqQ: IMHO, this would lead to issues due to the requirement to check the return value every time.
Can we handle it somehow otherwise? For example, filtering the nodes that may miss input edge (like Constants, I think)?

nncf/onnx/graph/node_utils.py
```diff
@@ -172,9 +171,13 @@ def _get_activation_tensor_shape(
     :return: None, if there is no shape info, otherwise - tensor shape.
     """
     if target_point.type == TargetType.PRE_LAYER_OPERATION:
-        shape = nncf_graph.get_input_edges(node)[target_point.port_id].tensor_shape
+        edge = nncf_graph.get_input_edge_by_port_id(node, target_point.port_id)
```
KodiaqQ, line 174: `if edge` condition was missed? Since `nncf_graph.get_input_edge_by_port_id` may return `None`.

nncf/openvino/graph/node_utils.py
```diff
@@ -167,10 +167,10 @@ def get_node_with_bias_value(add_node: NNCFNode, nncf_graph: NNCFGraph) -> Optio
     const_port_ids = add_node.layer_attributes.get_const_port_ids()
     assert len(const_port_ids) == 1
     bias_port_id = const_port_ids[0]
-    bias_constant = nncf_graph.get_input_edges(add_node)[bias_port_id].from_node
+    bias_constant = nncf_graph.get_input_edge_by_port_id(add_node, bias_port_id).from_node
```
KodiaqQ, line 170: Same here and below: `nncf_graph.get_input_edge_by_port_id` result verification was missed?

nncf/onnx/graph/node_utils.py
KodiaqQ: Well, maybe the introduction of the common method `nncf_graph.get_input_edge_by_port_id` was not a good suggestion from my side.
But, in the previous version, there was no input edge existing verification for the ONNX backend. Why do we need it now in new method?

nncf/common/graph/graph.py
andrey-churkin: As discussed, I added an exception for cases where a node does not have input edges.

KodiaqQ (approved):

nncf/common/graph/graph.py
```diff
@@ -348,20 +348,52 @@ def get_input_edges(self, node: NNCFNode) -> List[NNCFGraphEdge]:
             edges.extend(self._get_edges(from_node, node))
         return sorted(edges, key=lambda x: x.input_port_id)
 
+    def get_input_edge_by_port_id(self, node: NNCFNode, port_id: int) -> NNCFGraphEdge:
```
alexsu52, line 351: Could briefly explain why you added this function instead of using get_input_edges(node)[port_id]?

andrey-churkin: We apply passes, such as `remove_shapeof_subgraphs()` and `filter_constant_nodes()`, to the graph that can invalidate the order of input edges. For this reason, it is safer to use the `get_input_edge_by_port_id()` method.

alexsu52 (approved):

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -701,7 +705,18 @@ def _get_activation_quantization_target_point(
...
+            # we haven't encountered a model with operations that have multiple output edges with different
+            # output port IDs. Currently, such models are not supported. Usually, `output_port_id = 0` is used.
+            # However, there are operations, such as LSTMSequence, where the `output_port_id` changes from case
+            # to case. Therefore, the code below is required to dynamically determine the `output_port_id` where
+            # the quantize operation should be inserted."
+            node = nncf_graph.get_node_by_name(node_name)
+            unique_output_port_ids = set(e.output_port_id for e in nncf_graph.get_output_edges(node))
+            if len(unique_output_port_ids) > 1:
+                raise nncf.InternalError(f"Cannot determine the output_port_id for the operation: {node_name}")
+            output_port_id = next(iter(unique_output_port_ids))
```
kshpv, line 718: Minor
```suggestion
            output_port_id = unique_output_port_ids.pop()
```

kshpv (approved):

PR #2807, opened 2024-07-12
Revert round bahaviour as in u4 for i4.

andreyanufr:
### Changes

Revert zero point adding to round to align rounding error between u4 and i4 compression.

### Reason for changes

Different results for round(+-3.4555559) + 8 and round(+-3.4555559 + 8) which lead to different results in int4_sym compression for some models.


andreyanufr:
post_training_weight_compression/106

PR #2810, opened 2024-07-15
added references for 2024.4 nightly

ljaljushkin:
### Changes

as stated in the title

### Reason for changes

build 17 has another metrics for 2024.4

### Related tickets

n/a

### Tests

build 20

### DIFF
--- a/tests/post_training/data/wc_reference_data_2024.4.yaml
+++ b/tests/post_training/data/wc_reference_data_2024.4.yaml
@@ -1,0 +1,6 @@
+tinyllama_data_aware_awq_scale_estimation_backend_OV:
+  metric_value: 0.84038
+tinyllama_data_aware_awq_scale_estimation_stateful_backend_OV:
+  metric_value: 0.84038
+tinyllama_data_aware_gptq_backend_OV:
+  metric_value: 0.81936


tests/post_training/data/wc_reference_data_2024.4.yaml
KodiaqQ, line 1: Can we introduce 2024.2 references instead of the `wc_reference_data.yaml` (just renaming) and create a symlink from this to 2024.2 file?

ljaljushkin: Symlink limits the validation for Windows only. 
`wc_reference_data.yaml` name is also used for html generation in CI, it would be not easy to change every release @AlexanderDokuchaev

KodiaqQ: Is there any weight compression validation on Windows? How then the graph tests works?
I mean, are there any strong limits to using the existing approach here as it was done for the graph tests?

ljaljushkin: > Can we introduce 2024.2 references instead of the `wc_reference_data.yaml` (just renaming) and create a symlink from this to 2024.2 file? As it was done in https://github.com/openvinotoolkit/nncf/tree/develop/tests/openvino/native/data.

BTW, sym links were intentionally removed in that approach: https://github.com/openvinotoolkit/nncf/commit/810e6d21678d726ba6220e20f0a4cc3735e3b347

ljaljushkin: Note, after switching from 2024.2 to 2024.3 as a "target" package of OV for validation, the tests will still work.
Optionally, references from 2024.3 file can migrate to a "target" reference file `wc_reference_data.yaml`

KodiaqQ: > sym links were intentionally removed in that approach: [810e6d2](https://github.com/openvinotoolkit/nncf/commit/810e6d21678d726ba6220e20f0a4cc3735e3b347)

So, there exists an option to keep only two references instead of the three with two duplicates. Is that correct?
My point is that we should keep three reference files for now: `wc_reference_data.yaml`, `wc_reference_data_2024.3.yaml`, `wc_reference_data_2024.4.yaml`.
Reasons to not keep `wc_reference_data.yaml` and switch to versions approach:
1. It contains unnamed references in comparison to `wc_reference_data_20xx.x.yaml`.
2. It contains the duplicate of the `wc_reference_data_2024.4.yaml` now.
3. It is not obvious how (when and who would update this) `wc_reference_data.yaml` should be supported in the future.

ljaljushkin: 1. the current version of openvino is in the setup.py
2. No, reference are not duplicated. It contains only the ones that different with the current target version.
3. we can agree, that once the target version of OV is changed in setup.py, in the same PR wc_references_<target>.yaml is merged to wc_references.yaml 

The current expectation that we don't need to update this accuracy references often. It only happens once.
Once we move to 2024.3, the references with versions are not needed anymore until we see the accuracy discrepancy next time (supposedly not soon).

I agree that the approach with keeping `wc_reference_data_20xx.x.yaml` is better, but it will require some effort with CI scripts that run conformance tests in a parallel, and we are not sure that it will often be needed. 
Let's see whether this effort is required first.

AlexanderDokuchaev: @KodiaqQ it's fast fix according limitation of CI and current state of test.
More general change for tests i will made on this week.

AlexanderDokuchaev (approved):

PR #2815, opened 2024-07-16
[PTQ] Remove logs from autogenerated ignored node names

kshpv:
### Changes

From:

The first log is printed from the user Ignored Scope.
The second and the third ones are from the inner algorithms source and should not be printed.

image

To:

image

### Reason for changes

Improve UX by removing logs from autogenerated ignored node names.

### Related tickets

N/A

### Tests

Manually on a provided root cause


alexsu52 (approved):
LGTM

PR #2817, opened 2024-07-16
[Torch FX] qdq ops extension

anzr299:
### Changes

get the tensor shape for quantize and dequantize operations in torch fx nncf_graph_builder.py and add it to the graph edge.

(Cleanest way I could find to achieve this, I swear)

### Reason for changes

quantize and dequantize operations for torch fx did not support tensor shape to add to the graph edge.

### DIFF
--- a/nncf/experimental/torch/fx/nncf_graph_builder.py
+++ b/nncf/experimental/torch/fx/nncf_graph_builder.py
@@ -122,7 +122,8 @@ class GraphConverter:
         """
         output_port_id = 0
         if source_node.op in ("get_attr",):
-            tensor_shape = tuple(getattr(model, source_node.target).shape)
+            tensor = getattr(model, source_node.target)
+            tensor_shape = tuple(tensor.shape)
         elif "val" in source_node.meta:
             if source_nncf_node.metatype is om.PTBatchNormMetatype:
                 tensor = source_node.meta["val"][0]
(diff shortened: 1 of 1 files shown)


nncf/experimental/torch/fx/nncf_graph_builder.py
```diff
@@ -0,0 +1,143 @@
...
+                tensor = source_node.meta["val"][output_idx]
+                # Assume every split outputs corresponds to an unique output_port_id
+                output_port_id = output_idx
+            else:
+                tensor = source_node.meta["val"]
+            tensor_shape = tuple(tensor.shape)
+        else:
+            nncf_logger.info(f"Edge shape between {source_node.name} and {dist_node.name} is unknown.")
+            tensor_shape = None
+        if "quantize" in dist_node.name:
```
anzr299, line 140: This is the addition for adding shape to quantize, dequantize node edges. It essentially checks if the next node is a quantize or dequantize node and adds the value of current node to that node so the edge is added when traversing the qdq node.

daniil-lyakhov:
@anzr299, please rebase your changes

anzr299:
> @anzr299, please rebase your changes

Done

nncf/experimental/torch/fx/nncf_graph_builder.py
daniil-lyakhov: I don't think we should set any values inside the function which is reading values, I believe we need to find a way this meta is being build and rebuild it automatically using the torch fx functionality

anzr299: yeah, that sounds right. I was finding a way to add it into the qdq function for torch fx.

PR #2820, opened 2024-07-17
Fix pass to remove ShapeOf subgraph

andrey-churkin:
### Changes

It is possible to remove a node in the Shape Of subgraph only if all output edges have an integer dtype. An additional condition has been added to check for this.

### Reason for changes

In the picture below, integer edges are colored blue, and float edges are colored black. Nodes with a red border are removed during the ShapeOf removal pass, but they shouldn't be.

shapeof_pass

### Related tickets

Ref: 146088

### Tests

TODO


andrey-churkin:
Implemented here https://github.com/openvinotoolkit/nncf/pull/2902

PR #2822, opened 2024-07-18
[OpenVINO] Add HW patterns

kshpv:
### Changes

Add HW patterns covering mask_rcnn_resnet50_atrous_coco model.
From an execution graph perspective, the subgraph of a model with a new quantization scheme executes up to 3x faster.
Before | After
image

### Reason for changes
Improve performance. 

Local measurements:
Before: 1.8666 FPS | After: 2.0266 FPS 
Speed up ~ 8-9%
### Related tickets

144218

### Tests

Manually tested on the model.


alexsu52 (approved):
LGTM

PR #2814, opened 2024-07-15
Set upper limit for pydot version 3.0.0

AlexanderDokuchaev:
### Changes

Set upper limit for pydot version 3.0.0

### Reason for changes

Pydot 3.0.0 broke saving and loading with special characters

PR #2833, opened 2024-07-22
Set upper limit for pydot version 3.0.0 (#2814)

AlexanderDokuchaev:
### Changes

Set upper limit for pydot version 3.0.0

### Reason for changes

Pydot 3.0.0 broke saving and loading with special characters


alexsu52 (approved):
LGTM

PR #2840, opened 2024-07-24
Remove gather metatype from input quantizable

daniil-lyakhov:
### Changes

Gather metatype is removed from input quantizable metatypes

### Reason for changes

To prevent FQ propagation through gathers (relevant for the ScadeDotProductAttention quantization):
Before:
image

After:
image


daniil-lyakhov:
Quantizers are being inserted for gathers which have only one output:
image

daniil-lyakhov:
Fix is not valid

Issue #2766, opened 2024-06-27
[TorchFX] Torch FX/PyTorch 2 Export Quantization


alexsu52:
### 🚀 Feature request

Quantization is a widely used technique to accelerate models, particularly when using the [torch.compile](https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html). For detailed tutorials and demonstrations on model quantization using PyTorch 2 Export Quantization, please refer to the following resources:
-  [Support for PyTorch 2 export quantization in OpenVINO backend](https://docs.openvino.ai/2024/openvino-workflow/torch-compile.html#support-for-pytorch-2-export-quantization-preview)
- [PyTorch 2 Export Post Training Quantization with X86 Backend through Inductor](https://pytorch.org/tutorials/prototype/pt2e_quant_ptq_x86_inductor.html) 

These guides show how to obtain a quantized model via the [PyTorch 2 Export Quantization API](https://pytorch.org/docs/2.3/quantization.html) and run it using `torch.compile`. However OpenVINO provide backend for `torch.compile`, but NNCF does not support quantization PyTorch 2 Export (`torch.fx.GraphModule`) models and users have to use `X86InductorQuantizer` to quantize models. Comparisons between PyTorch 2 Export INT8 models quantized by `X86InductorQuantizer` and OpenVINO INT8 models quantized by `NNCF` show that `NNCF` produces more accurate and efficient INT8 models.

Feature request is to support for `torch.fx.GraphModule` models in `nncf.quantize` to enable the creation of accurate and highly efficient models using `torch.compile` with the OpenVINO backend.

### Feature Use Case

```
import torch
import nncf

# initialize a floating point model​
float_model = M().eval()​

# program capture​
# NOTE: this API will be updated to torch.export API in the future,​ but the captured result should mostly stay the same​
model = capture_pre_autograd_graph(float_model, *example_inputs)

# quantization​
quantized_model = nncf.quantize(model, calibration_dataset)

# compile quantized model with OpenVINO bac​kend
compiled_model = torch.compile(quantized_model, backend='openvino')
```

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

alexsu52:
@daniil-lyakhov, please, analyze this feature request and open issues as sub-tasks of this feature request.

alexsu52:
I suugest to introduce the following API in NNCF, to support third-party quantizers and better alignment with PyTorch 2 Export Quantization API:
```
[...]

PR #2841, opened 2024-07-24
[TorchFX] Conformance test init

daniil-lyakhov:
### Changes

Conformance test for resnet18

### Reason for changes

To extend testing scope for the TorchFX backend

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/2766

### Tests

post_training_quantization/442 is successfull

### DIFF
--- a/tests/post_training/data/ptq_reference_data.yaml
+++ b/tests/post_training/data/ptq_reference_data.yaml
@@ -22,6 +22,16 @@ hf/hf-internal-testing/tiny-random-gpt2_backend_OV:
   metric_value: null
 hf/hf-internal-testing/tiny-random-gpt2_backend_TORCH:
   metric_value: null
+torchvision/resnet18_backend_FP32:
+  metric_value: 0.6978
+torchvision/resnet18_backend_OV:
+  metric_value: 0.6948
+torchvision/resnet18_backend_TORCH:
+  metric_value: 0.69152
+torchvision/resnet18_backend_CUDA_TORCH:
+  metric_value: 0.69152
+torchvision/resnet18_backend_FX_TORCH:
+  metric_value: 0.6946
 timm/crossvit_9_240_backend_CUDA_TORCH:
   metric_value: 0.689
 timm/crossvit_9_240_backend_FP32:
--- a/tests/post_training/data/ptq_reference_data.yaml
+++ b/tests/post_training/data/ptq_reference_data.yaml
@@ -26,6 +26,8 @@ torchvision/resnet18_backend_FP32:
   metric_value: 0.6978
 torchvision/resnet18_backend_OV:
   metric_value: 0.6948
+torchvision/resnet18_backend_ONNX:
+  metric_value: 0.6948
 torchvision/resnet18_backend_TORCH:
   metric_value: 0.69152
 torchvision/resnet18_backend_CUDA_TORCH:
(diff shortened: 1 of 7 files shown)


daniil-lyakhov:
Build post_training_quantization/430 is successfull

tests/post_training/test_quantize_conformance.py
```diff
@@ -247,7 +247,7 @@ def test_ptq_quantization(
     err_msg = None
     test_model_param = None
     start_time = time.perf_counter()
-    try:
+    if True:
```
AlexanderDokuchaev, line 250: Looks like used for debug

tests/post_training/data/ptq_reference_data.yaml
```diff
@@ -22,6 +22,16 @@ hf/hf-internal-testing/tiny-random-gpt2_backend_OV:
   metric_value: null
 hf/hf-internal-testing/tiny-random-gpt2_backend_TORCH:
   metric_value: null
+torchvision/resnet18_backend_FP32:
```
AlexanderDokuchaev, line 25: What differ between torchvision and timm models?
If it's same but timm can not be exported to FX, timm/resnet18 can be removed

tests/post_training/pipelines/base.py
```diff
@@ -367,6 +368,11 @@ def save_compressed_model(self) -> None:
             )
             self.path_compressed_ir = self.output_model_dir / "model.xml"
             ov.serialize(ov_model, self.path_compressed_ir)
+        elif self.backend == BackendType.FX_TORCH:
+            exported_model = torch.export.export(self.compressed_model, (self.dummy_tensor,))
```
AlexanderDokuchaev, line 372: Compressed FX model can not be converted to OV model directly?

tests/post_training/pipelines/image_classification_torchvision.py
```diff
@@ -0,0 +1,169 @@
...
+        return transform_fn
+
+    def prepare_calibration_dataset(self):
+        dataset = datasets.ImageFolder(root=self.data_dir / "imagenet" / "val", transform=self.transform)
+        loader = torch.utils.data.DataLoader(dataset, batch_size=self.batch_size, num_workers=2, shuffle=False)
+
+        self.calibration_dataset = nncf.Dataset(loader, self.get_transform_calibration_fn())
+
+    def _validate(self):
+        val_dataset = datasets.ImageFolder(root=self.data_dir / "imagenet" / "val", transform=self.transform)
```
AlexanderDokuchaev, line 126: Looks like it can be reused from image_classification_timm.py

AlexanderDokuchaev (changes requested):

tests/post_training/pipelines/base.py
daniil-lyakhov: Yes, the error is
```
 Failed: Couldn't get TorchScript module by tracing. With exception:
E           
E                   Calling train() or eval() is not supported for exported models.
E                   Please call `torch.ao.quantization.move_exported_model_to_train(model)` (or eval) instead.
E           
E                   If you cannot replace the calls to `model.train()` and `model.eval()`, you may override
E                   the behavior for these methods by calling `torch.ao.quantization.allow_exported_model_train_eval(model)`,
E                   which does the above automatically for you. Note that this has limited effect on switching
E                   behavior between train and eval modes, and should be used only for special ops such as dropout
E                   and batchnorm.

```

tests/post_training/pipelines/image_classification_torchvision.py
daniil-lyakhov: Done

tests/post_training/data/ptq_reference_data.yaml
daniil-lyakhov: I've removed timm/resnet18 and replaced with torchvision/resnet18

tests/post_training/test_quantize_conformance.py
daniil-lyakhov: My bad, done!

daniil-lyakhov:
Build post_training_quantization/442 is green

AlexanderDokuchaev (approved):

PR #2845, opened 2024-07-26
[Torch FX] Add test for stable diffusion in model tests

anzr299:
### Changes

Extended the model tests for torch fx to include stable diffusion

### Reason for changes

To accomodate stable diffusion in model test


anzr299:
@daniil-lyakhov @alexsu52

PR #1071, opened 2022-01-17
Fix quantizer propagation algo

asenina:
### Related tickets
#60615

### Tests
Tests were added in `test_quantizer_propagation_solver.py`

PR #2854, opened 2024-07-30
[Solver] Check all output branches are quantized before merging quantizers

daniil-lyakhov:
### Changes

Quantizer merge logic updated to check that all output branches are quantized before quantizers merging and propagating up.

### Reason for changes

To prevent merging of quantizers in case of ScaledDotProductAttention op, which should have quantizers on [0, 1] input ports and shouldn't have a quantizer on the 3 input port.

### Related tickets

148211
#2766 

### Tests

* Common solver test for ScaleDotProductAttention branch merging and quantization initialization
* Graph tests for torch/ov backends


nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -769,6 +769,38 @@ def recursive_helper(curr_node_key: str, target_node_list: List[str]):
...
+        """
+        Returns True if all pathes from the given node to the first
+        input quantable nodes have an activation quantizer, False otherwise.
+
+        :param node_key: Given node key.
+        :return: True if all pathes from the given node to the first
+        input quantable nodes have an activation quantizer, False otherwise.
+        """
+
+        def recursive_helper(curr_node_key: str):
```
KodiaqQ, line 782: From the cases based on the `BiasCorrection` algorithm, the recursion is not the better option since it has limited depth and the output case may not be reached in optimal time. 
Is it possible to use the `while` option here?
Major.

nncf/common/quantization/quantizer_propagation/graph.py
```diff
@@ -769,6 +769,38 @@ def recursive_helper(curr_node_key: str, target_node_list: List[str]):
...
+                elif successor_node_type in [
+                    QuantizerPropagationStateGraphNodeType.PRE_HOOK,
+                    QuantizerPropagationStateGraphNodeType.POST_HOOK,
+                ]:
+                    successor_quantizer = successor[QuantizerPropagationStateGraph.PROPAGATING_QUANTIZER_NODE_ATTR]
+                    if successor_quantizer:
+                        continue
+                if not recursive_helper(successor_key):
+                    return False
+            return True
```
KodiaqQ, line 800: The output conditions are not very clear to me. Maybe method renaming (`recursive_helper` to something more obvious) would also help.
Minor.

nncf/common/quantization/quantizer_propagation/solver.py
```diff
@@ -1233,6 +1233,13 @@ def check_branching_transition(
         dom_op_quantizers = set()
         for op_node_key in dom_op_node_keys:
             op_node = quant_prop_graph.nodes[op_node_key]
+
+            # Check all branches have a quantizer on it before the merge
+            if op_node["op_meta"].target_input_ports is not None:
+                all_branches_are_quantized = quant_prop_graph.all_outputs_are_quantized(branching_node_key)
+                if not all_branches_are_quantized:
+                    return TransitionStatus.SHOULD_NOT_TRANSITION
```
KodiaqQ, line 1241: Are you sure that we should in `check_brancking_transition` deny the merging? Also, there exists more appropriate status like `TransitionStatus.SHOULD_MERGE`.
Major.

daniil-lyakhov: My changes are based on https://github.com/openvinotoolkit/nncf/pull/1071, which is quite similar to my target case: [This ](https://github.com/openvinotoolkit/nncf/pull/1071/files#diff-39f33773245f182909c8b9227aa638aac9e4bb39837264477d9a99cf92c11b31R1219-R1222)check was added to prevent outputs quantization (please see comment https://github.com/openvinotoolkit/nncf/pull/1071#issuecomment-1403654945). As you can see, `TransitionStatus.SHOULD_NOT_TRANSITION` was used here. Status `TransitionStatus.SHOULD_MERGE` will merge quantizer and will put them on the post hook of the branching node. In the target case of this pr that means that third branch of the `ScaledDotProductAttention` is being wrongly quantized

daniil-lyakhov:
bloomz-560m quantization with not fused ScaledDotProductAttention op checked locally

KodiaqQ (approved):

PR #2861, opened 2024-08-01
[TorchFX] Pre-hook insertion support

daniil-lyakhov:
### Changes

Torch FX pre-hook insertion support

### Reason for changes

To enable vit_b_16 quantization

### Related tickets

#2766 

### Tests

test_quantized_models is updated by vit_b_16 and swin_v2_s


nncf/experimental/torch/fx/transformations.py
```diff
@@ -170,51 +169,63 @@ def insert_one_qdq_after_node(model: torch.fx.GraphModule, target_node: torch.fx
...
+
+    # use the same qparams from quantize op
     dq_inputs = [quantized_node] + quantize_op_inputs[1:]
-    user_dq_nodes = []
-    with graph.inserting_after(quantized_node):
-        for user in target_node.users:
-            if user is quantized_node:
-                continue
-            user_dq_nodes.append((user, graph.call_function(dequantize_op, tuple(dq_inputs), {})))
+    if target_point.target_type == TargetType.OPERATOR_POST_HOOK:
```
AlexanderDokuchaev, line 180: TargetType contains 8 values, if-else is not correct way to check type.
Add check that target_type contains only expected values, or use if-elif-else-error.

And looks that  reusing PTTargetPoint and PTTargetType is bad idea, contain too many unrelated code.

nncf/experimental/torch/fx/transformations.py
```diff
@@ -170,51 +169,63 @@ def insert_one_qdq_after_node(model: torch.fx.GraphModule, target_node: torch.fx
...
-def _get_target_node(graph: torch.fx.Graph, target_point: PTTargetPoint) -> torch.fx.Node:
-    """
-    Returns TorchFX graph node correspondent to the target point.
-
-    :param graph: Target torch.fx.Graph.
-    :param target_point: A target point to find the target node.
-    :return: TorchFX graph node correspondent to the target point.
-    """
-    # TODO(dlyakhov): Support node insertion on a specific input port id.
+def _get_data_for_insertion(
```
AlexanderDokuchaev, line 218: Bad function , too general code. That should be spited to several short function this clean description.
Example:
- get_graph_node_by_name (already exists)
- get_fead_node
- get_insertion_point_ctx

AlexanderDokuchaev (changes requested):

nncf/experimental/torch/fx/transformations.py
daniil-lyakhov: Done

nncf/experimental/torch/fx/transformations.py
daniil-lyakhov: Done

AlexanderDokuchaev (approved):

PR #2862, opened 2024-08-01
[PTQ][Common] Use inference_nncf_graph instead nncf_graph in get_weights

daniil-lyakhov:
### Changes

Use inference_nncf_graph instead nncf_graph in get_weights

### Reason for changes

To avoid quantizing weight nodes that are present in `nncf_graph` and absent in `inference_nncf_graph`

### Related tickets

#2766 

### Tests

test_ptq_params: test_constant_branch_weighted_node_quantization is introduced


daniil-lyakhov:
Onnx tests diff:
image

daniil-lyakhov:
Turns out some operations on constant paths should be quantized (like positional embedding in GPT-2 model), so this fix is invalid

PR #2622, opened 2024-04-10
[PTQ][MinMax][Torch] One shared quantizer is used for all unified scale quantization points

daniil-lyakhov:
### Changes

* MinMax: new backend method `create_unified_scales_quantizers_insertion_commands` is introduced: it receives several target points and one quantization parameter. Depending on implementation, one or several insertion commands are generated and returned back to the common algorithm. 

### Reason for changes

* Torch backend requires one `PTSharedFNInsertionCommand`  to make quantizers aligned during QAT in comparison with OV/ONNX backend, which can use separate commands/quantizers for each insertion point without any restrictions

### Related tickets

104304

### Tests

[Template test] test_ptq_params: test_unified_scales_command_creation
test_create_shared_quantizer_insertion_command

### Jobs
manual/job/post_training_quantization/350/: passed

PR #2877, opened 2024-08-08
Fixed unified FakeConverts duplication

KodiaqQ:
### Changes

- Fixed the `MinMax` algorithm for the `FakeConvert` insertion case.

### Reason for changes

- Incorrect `MinMax` behaviour introduced in https://github.com/openvinotoolkit/nncf/pull/2622.

### Related tickets

- 149128

### Tests

- Manual

Incorrect behaviour - two identical layers were inserted:
image
Corrected behaviour - only one layer for each branch was inserted:
image


daniil-lyakhov:
@KodiaqQ, looks legit, thank you for the fix! Is it possible to make a minimal synthetic reproducer for this model?

KodiaqQ:
> @KodiaqQ, looks legit, thank you for the fix! Is it possible to make a minimal synthetic reproducer for this model?

Added.

daniil-lyakhov (approved):

PR #2892, opened 2024-08-19
Add weighted progress tracking for weight compression

nikita-savelyevv:
### Changes

During nncf weight compression, `rich` progress bar is used to display the progress. In this PR, progress bar is changed to be weighted according to model weights. With these changes, each weight contributes proportional amount of percent to the progress bar.

Iteration number was removed from weight compression progress bar to avoid confusion between different speeds in percent and iteration coordinates. For example now a single weight might contribute 5-10% to the whole progress.

### Reason for changes

The time it takes to compress a weight is roughly proportional to its size, so incrementing the progress by 1 for each weight is not ideal.

Especially after #2803 when weight sorting was added. Now, the largest weights come first and the smallest ones are at the end of the compression. This leads to misleading time estimation when progress contribution from every weight is equal.

Weights sizes for tinyllama-1.1b for reference:
weight_size_hist

weight_size_cumsum_hist

### DIFF
--- a/nncf/common/logging/track_progress.py
+++ b/nncf/common/logging/track_progress.py
@@ -97,7 +97,7 @@ class WeightedProgress(Progress):
             task.fields["completed_steps"] = 0
 
     @staticmethod
-    def weighted_advance(task: Task, advance: float):
+    def weighted_advance(task: Task, advance: float) -> float:
         """
         Perform weighted advancement based on an integer step value.
         """
@@ -110,7 +110,7 @@ class WeightedProgress(Progress):
         return weighted_advance
 
     @staticmethod
-    def get_weighted_completed(task: Task, completed: float):
+    def get_weighted_completed(task: Task, completed: float) -> float:
         """
         Get weighted `completed` corresponding to an integer `completed` field.
         """
@@ -137,7 +137,7 @@ class track:
         update_period: float = 0.1,
         disable: bool = False,
         show_speed: bool = True,
-        weights: List[float] = None,
+        weights: Optional[List[float]] = None,
     ):
         """
         Track progress by iterating over a sequence.
--- a/nncf/common/logging/track_progress.py
+++ b/nncf/common/logging/track_progress.py
@@ -65,29 +65,6 @@ class WeightedProgress(Progress):
     A class to perform a weighted progress tracking.
     """
 
-    COMPLETED_STEPS_ATTR = "_completed_steps"
-
-    def weighted_advance(self, task: Task, advance: float):
-        """
-        Perform weighted advancement based on an integer step value.
-        """
-        if advance % 1 != 0:
-            raise Exception(f"Unexpected advance value: {advance}.")
-        advance = int(advance)
-        current_step = self.get_task_completed_steps(task)
-        weighted_advance = sum(task.fields["weights"][current_step : current_step + advance])
-        self.set_task_completed_steps(task, current_step + advance)
-        return weighted_advance
-
-    @staticmethod
-    def get_weighted_completed(task: Task, completed: float):
-        """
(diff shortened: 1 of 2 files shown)


ljaljushkin (approved):

nncf/common/logging/track_progress.py
```diff
@@ -59,6 +60,65 @@ def render(self, task: "Task") -> Text:
...
+        if completed is not None:
+            kwargs["completed"] = self.get_weighted_completed(task, completed)
+
+        super().reset(task_id, **kwargs)
+
+        if completed == 0:
+            task.fields["completed_steps"] = 0
+
+    @staticmethod
+    def weighted_advance(task: Task, advance: float):
```
AlexanderDokuchaev, line 100:
```suggestion
    def weighted_advance(task: Task, advance: float) -> float:
```

nncf/common/logging/track_progress.py
```diff
@@ -59,6 +60,65 @@ def render(self, task: "Task") -> Text:
...
+        """
+        Perform weighted advancement based on an integer step value.
+        """
+        if advance % 1 != 0:
```
AlexanderDokuchaev, line 104: In which case is it possible?
May be better return 0 in this case, to avoid exception by tracker?

nncf/common/logging/track_progress.py
```diff
@@ -59,6 +60,65 @@ def render(self, task: "Task") -> Text:
...
+            raise Exception(f"Unexpected `advance` value: {advance}.")
+        advance = int(advance)
+        current_step = task.fields["completed_steps"]
+        weighted_advance = sum(task.fields["weights"][current_step : current_step + advance])
+        task.fields["completed_steps"] = current_step + advance
+        return weighted_advance
+
+    @staticmethod
+    def get_weighted_completed(task: Task, completed: float):
```
AlexanderDokuchaev, line 113:
```suggestion
    def get_weighted_completed(task: Task, completed: float) -> float:
```

nncf/common/logging/track_progress.py
```diff
@@ -77,6 +137,7 @@ def __init__(
         update_period: float = 0.1,
         disable: bool = False,
         show_speed: bool = True,
+        weights: List[float] = None,
```
AlexanderDokuchaev, line 140:
```suggestion
        weights: Optional[List[float]] = None,
```

AlexanderDokuchaev (changes requested):

nncf/common/logging/track_progress.py
nikita-savelyevv: > In which case is it possible?

The `update` method is exposed and can be called by a user if tracker is used as a context manager. For example here https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/pipelines/image_classification_base.py#L65

> May be better return 0 in this case, to avoid exception by tracker?

I believe it is fine to keep it as exception because the wrong value can only arise in case of wrong usage by a user. Because as I understand, when using it as `for in track(iterable)`, the `advance` argument will always be integer. Also, if something changes inside `rich` and this will no longer be the case, we'll be notified by failed tests.

nncf/common/logging/track_progress.py
nikita-savelyevv: Done

nncf/common/logging/track_progress.py
nikita-savelyevv: Done

nncf/common/logging/track_progress.py
nikita-savelyevv: Done

AlexanderDokuchaev (approved):

PR #2897, opened 2024-08-21
Add early stopping for memory monitoring

nikita-savelyevv:
### Changes

Stop addition of new memory values as soon as possible in order to correctly represent the tracked task.

### Reason for changes

When running memory monitor as a CLI tool, sometimes it may happen that the last value is recorded in the main process, i.e. when the child process has already died. This is represented as a drop in memory, because the child process memory is deallocated. This PR tries to avoid such behavior.

### DIFF
--- a/tools/memory_monitor.py
+++ b/tools/memory_monitor.py
@@ -23,6 +23,7 @@ from typing import Callable, List, Optional, Tuple
 
 import matplotlib.pyplot as plt
 import psutil
+from tabulate import tabulate
 
 logger = logging.getLogger("memory_monitor")
 
@@ -366,6 +367,7 @@ if __name__ == "__main__":
         for mm in memory_monitors:
             mm._monitoring_thread_should_stop = True
 
+    summary_data = []
     for mm in memory_monitors:
         mm.stop()
         for fz in (True, False):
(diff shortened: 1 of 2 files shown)


tools/memory_monitor.py
```diff
@@ -348,14 +354,24 @@ def _subtract_first_element(data):
...
+        for mm in memory_monitors:
+            mm._monitoring_thread_should_stop = True
+
+    for mm in memory_monitors:
+        mm.stop()
+        for fz in (True, False):
+            time_values, memory_values = mm.get_data(memory_from_zero=fz)
+            # Most probably the last value is recorded once the child process has already died
+            time_values, memory_values = time_values[:-1], memory_values[:-1]
+            mm.save_memory_logs(
```
AlexanderDokuchaev, line 375: What you think about add summary to stdout, somthing like like:

```python
from tabulate import tabulate
...
    summary_data = []
    for mm in memory_monitors:
    ...
            summary_data.append([mm.memory_type.value, fz, max(memory_values)])
    print(tabulate(summary_data, headers=["Memory type", "From zero", "Pick"]))
```
```
Memory type    From zero          Pick
-------------  -----------  ----------
rss            True            41.7188
rss            False          103.848
system         True            41.6367
system         False        71775.9
```

AlexanderDokuchaev (approved):

tools/memory_monitor.py
nikita-savelyevv: Added

PR #2899, opened 2024-08-21
[TorchFX] Performance check

daniil-lyakhov:
### Changes

* Performance code is introduced to confirm supported models are performant

### Related tickets

#2766


alexsu52 (commented):
Why did you create test instead of a tool?

PR #2902, opened 2024-08-22
Preserve the constant subgraph

andrey-churkin:
### Reason for changes

Removing the ShapeOf subgraph may lead to the removal of some input edges of operations, and QPS doesn't place FQ on these edges. In this case, the operations are quantized incorrectly because not all required inputs are quantized. These operations don't change their runtime precision in the execution graph, which means we do not have a fully quantized model.

### Changes

To resolve the described problem, this PR introduces the following algorithm:

- **Step 1:** Find all nodes belonging to the ShapeOf subgraphs in the graph. Let's denote the found nodes as $S$. At this step, we don't remove the found subgraphs; we only store the nodes they consist of. See `find_shapeof_subgraphs()` method.

- **Step 2:** Check the producers for the specified nodes (currently, we are considering only nodes of the LSTMSequence and Convolution types). If any of these producers are in set $S$, it means that we will lose this input edge when the ShapeOf subgraph is removed. Therefore, we try to exclude float subgraphs from this ShapeOf subgraph. Let's denote the nodes that are part of such float subgraphs as $P$. We can see that $P \subset S$. See `find_preserved_nodes()` method.

- **Step 3:** Find all nodes belonging to the Constant subgraphs in the graph. Let's denote the found nodes as $C$. At this step, we don't remove the found subgraphs; we only store the nodes they consist of. See `find_constant_subgraphs()` method.

- **Step 4:** Remove $S \cup C - P$ nodes from the graph.

### Related tickets

- 146088

### DIFF
--- a/nncf/quantization/passes.py
+++ b/nncf/quantization/passes.py
@@ -152,7 +152,7 @@ def find_constant_subgraphs(
     :return: A list of nodes belonging to constant subgraphs.
     """
     if not input_nodes:
-        return nncf_graph
+        return []
 
     visited_nodes = set()
     nodes_queue = collections.deque(input_nodes)
--- a/nncf/quantization/passes.py
+++ b/nncf/quantization/passes.py
@@ -36,7 +36,8 @@ def transform_to_inference_graph(
     """
     shapeof_subgraphs = find_shapeof_subgraphs(nncf_graph, shapeof_metatypes, input_nodes)
     nncf_graph.remove_nodes_from(shapeof_subgraphs)
-    filter_constant_nodes(nncf_graph, input_nodes)
+    constant_subgraphs = find_constant_subgraphs(nncf_graph, input_nodes)
+    nncf_graph.remove_nodes_from(constant_subgraphs)
     remove_nodes_and_reconnect_graph(nncf_graph, dropout_metatypes)
     return nncf_graph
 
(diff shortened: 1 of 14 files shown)


andrey-churkin:
post_training_quantization Build # 456

KodiaqQ (approved):

nncf/quantization/passes.py
```diff
@@ -34,27 +36,33 @@ def transform_to_inference_graph(
...
+
+    nodes_to_drop = set([*shapeof_subgraphs, *constant_subgraphs]).difference(preserved_nodes)
+    nncf_graph.remove_nodes_from(nodes_to_drop)
+
     remove_nodes_and_reconnect_graph(nncf_graph, dropout_metatypes)
     return nncf_graph
 
 
-def remove_shapeof_subgraphs(
+def find_shapeof_subgraphs(
```
alexsu52, line 50: It looks like this function does not find shape of subgraph, but find all nodes after shapeof operation until nodes which have path to an input node.

Let discuss algorithm offline.

alexsu52: The goal of follow-up PR generalize algorithm for all quantizable operations.

alexsu52 (approved):
LGTM

nncf/quantization/passes.py
andrey-churkin: ref: 150942

PR #2912, opened 2024-08-23
[TorchFX] Fix depthwise + transpose conv weights quantization

daniil-lyakhov:
### Changes

* Nodes with weights are updated with depthwise and transpose convs

### Reason for changes

* To correctly quantize depthwise and transpose convolutions as they are missing weights quantizers right now 

### Related tickets

#2766 

### Tests

* mobilenet_v3_small reference quantized graph is updated


AlexanderDokuchaev (approved):

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -348,6 +349,6 @@ def get_ignored_names_by_layer_attributes(nncf_graph: NNCFGraph) -> Set[str]:
     def get_weight_nodes(nncf_graph: NNCFGraph) -> List[NNCFNode]:
         retval = set()
         for node in nncf_graph.get_all_nodes():
-            if node.metatype in [om.PTConv1dMetatype, om.PTConv2dMetatype, om.PTConv3dMetatype, om.PTLinearMetatype]:
```
AlexanderDokuchaev, line 351: If you using PTOperationType please use it as in PT backend, no need add extra variables

AlexanderDokuchaev (changes requested):

nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: I thought the same at first, but PT backend is using PTModule... metatypes only, and FX backend requires PTConv... metatypes

daniil-lyakhov: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/torch/graph/operator_metatypes.py#L1113-L1129

AlexanderDokuchaev: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/torch_backend.py#L365
Use PT parts only like they used in PT backend or dont use it, or add comment why is it impossible and how it will be changed after.

daniil-lyakhov: Done

AlexanderDokuchaev (approved):

PR #2915, opened 2024-08-26
Fix track context manager

nikita-savelyevv:
### Changes

Fix `track` usage as a context manager. This is a regression after #2892.


AlexanderDokuchaev:
@nikita-savelyevv What regression? And why is it not cached by tests?

nikita-savelyevv:
> What regression?

There will be an error if `track` is used as a context manager:
```
Failed: 'Progress' object has no attribute 'progress'
```

> And why is it not cached by tests?

There are no tests for the progress tracker. I've added them now.

AlexanderDokuchaev (approved):

PR #2924, opened 2024-08-28
Activation Sparsity OV backend

nikita-savelyevv:


nikita-savelyevv:
How to run: https://github.com/nikita-savelyevv/24h1-nncf-sparsify-activations-example/blob/main/README.md

PR #2932, opened 2024-08-29
Fix torch compile applied to model forward

nikita-savelyevv:
### Changes

Check whether model is compiled based on `"_torchdynamo_orig_callable"` property of the model forward.

### Reason for changes

`torch.compile` can be applied not only to the model itself, but to `forward()` method only. For example:
```
model.forward = torch.compile(model.forward)
```
In this case the model itself doesn't change and it won't be an instance of `torch._dynamo.OptimizedModule`.

### Related tickets

143796

### Tests

Added test when `torch.compile` is applied this way. It does not fail without the fix though, because the issue is sporadic.

### Relates to
#2665, #2719

### DIFF
--- a/nncf/torch/dynamic_graph/wrappers.py
+++ b/nncf/torch/dynamic_graph/wrappers.py
@@ -13,7 +13,6 @@ from copy import deepcopy
 from typing import Callable, List, Tuple
 
 import torch
-from torch._dynamo import OptimizedModule
 from torch.nn import DataParallel
 
 from nncf.common.graph.definitions import MODEL_CONST_OP_NAME
@@ -136,7 +135,7 @@ def wrap_module_call(module_call):
         from nncf.torch.dynamic_graph.patch_pytorch import unpatching_module_call
 
         # If called on a model compiled by torch dynamo, we unpatch torch operators and invoke original module call
-        if isinstance(self, OptimizedModule):
+        if "_torchdynamo_orig_callable" in self.forward.__dict__:
             return unpatching_module_call(self, *args, **kwargs)
 
         ctx = get_current_context()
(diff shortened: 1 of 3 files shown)


nncf/torch/dynamic_graph/wrappers.py
```diff
@@ -136,7 +135,7 @@ def wrapped(self, *args, **kwargs):
         from nncf.torch.dynamic_graph.patch_pytorch import unpatching_module_call
 
         # If called on a model compiled by torch dynamo, we unpatch torch operators and invoke original module call
-        if isinstance(self, OptimizedModule):
+        if "_torchdynamo_orig_callable" in self.forward.__dict__:
```
nikita-savelyevv, line 138: It seems that in theory any model method can be compiled in isolation. The fix presented here won't work for this case. To deal with such problem, we could do something like this:
```
for attr in self.__dict__.values():
    if hasattr(attr, "__dict__") and "_torchdynamo_orig_callable" in attr.__dict__:
        return unpatching_module_call(self, *args, **kwargs)
```
Such logic should protect the inference from being patched in case any of the model's methods is compiled. This however will introduce some overhead during every patched module call. I propose to keep it as is for now until we stumble across an issue where this can actually help.

daniil-lyakhov (approved):
LGTM, the `OptimizedModel` always has the `_torchdynamo_orig_callable` https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/eval_frame.py#L148 method + appears in the compiled model https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/convert_frame.py#L413 . Hope it will stay as it is

nikita-savelyevv:
> LGTM, the `OptimizedModel` always has the `_torchdynamo_orig_callable` method + appears in the compiled model `https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/convert_frame.py#L413` . Hope it will stay as it is

I've added a check so tests will fail if this changes https://github.com/nikita-savelyevv/nncf/blob/fix-torch-compile-forward/tests/torch/pytorch_patch_isolated.py#L104

AlexanderDokuchaev (approved):

PR #2935, opened 2024-08-30
[TorchFX] Shared Constant Support

anzr299:
### Changes

Added a transformation to check and replace repeated shared constant nodes. 
1. New transformation builder was added to traverse through the graph and check nodes with the same target. 
2. Update the `constant_update_fn` to first check all consumer nodes of the constant node followed by updating the input node to updated constant for all consumer nodes
3. Included this transformation in quantize model implementation before applying other transformations.

### Reason for changes

Torch Fx graph is created with distinct nodes for shared constants. This causes redundancy in the graph and later in the nncf graph for operations applied for constant nodes. 

### Tests

TODO


nncf/experimental/torch/fx/transformations.py
```diff
@@ -166,8 +197,11 @@ def constant_update_fn(model: torch.fx.GraphModule, node: torch.fx.Node, value:
             f"Constant on input port {input_port_id} for {node} is expected,"
             f" but node {args[input_port_id]} is present."
         )
+    weight_node = args[input_port_id]
+    consumer_nodes = list(weight_node.users.keys())
     args[input_port_id] = new_constant
-    node.args = tuple(args)
+    for node in consumer_nodes:
+        node.replace_input_with(weight_node, new_constant)
```
daniil-lyakhov, line 204: What if I have two nodes which share one constant, and I want to update constant only for the first node? This code will replace constant for both operations

nncf/experimental/torch/fx/quantization/quantize_model.py
```diff
@@ -68,7 +69,7 @@ def quantize_impl(
     original_graph_meta = model.meta
 
     copied_model = deepcopy(model)
-
+    shared_constant_create_transformation_builder()(copied_model)
```
daniil-lyakhov, line 72: Please consider finding a way to cope with shared nodes without a model modification, model modification won't be approved

daniil-lyakhov (changes requested):

nncf/experimental/torch/fx/transformations.py
anzr299: ah, good point. Maybe I can add a flag and create another builder which updates constant for all the nodes and calls this update function with the flag set to true.
something like this:
```
def update_const_builder():
    update_const_fn()

def update_const_fn(flag=False):
    #Update constant function

def update_shared_const():
    update_const_fn(flag=True)
```

daniil-lyakhov: We can try to receive `List[PTTargetPoint]` in the transformation instead of `PTTargetPoint`

nncf/experimental/torch/fx/quantization/quantize_model.py
anzr299: I assumed it would be good to have a central change so that the graph is updated in one place and can be used everywhere. Otherwise in every use case we would have to search for nodes with the same target.

daniil-lyakhov: We have `is_shared` method in `NNCFNode`, and we even have a ticked to support that https://github.com/openvinotoolkit/nncf/issues/2938 (duplicate from jira). Maybe it is possible to use this attribute to help with your issue?

anzr299: Oh, I will check it out. Thank you very much!

nncf/experimental/torch/fx/transformations.py
anzr299: I'm sorry, I didn't understand. Where do you suggest I change this?

nncf/experimental/torch/fx/quantization/quantize_model.py
anzr299: So would you suggest I check and update this attribute for the NNCFnode in the nncf graph builder for torch fx?

PR #2949, opened 2024-09-04
Improve weight compression time and memory for BF16 OV models

nikita-savelyevv:
### Changes

Avoid retrieving weight data if not necessary.

For llama-3-8b with BF16 weights there is about 1.5x time and 1.26x peak memory improvement. Please see the figure below.
| Before         | After   |
|-------------------|-------|
| system_memory_usage_from-zero |  system_memory_usage_from-zero |

### Reason for changes

For BF16 dtype when constant data is fetched, it is loaded into memory resulting in additional time and memory overhead.

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/openvino_backend.py
+++ b/nncf/quantization/algorithms/weight_compression/openvino_backend.py
@@ -106,6 +106,24 @@ class OVWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
         weight_tensor = get_const_value(weight_node)
         return Tensor(weight_tensor)
 
+    def get_weight_dtype(
+        self, node_with_weight: NNCFNode, weight_port_id: int, model: ov.Model, graph: NNCFGraph
+    ) -> TensorDataType:
+        weight_name = node_with_weight.layer_attributes.constant_attributes[weight_port_id]["name"]
+        weight_node = self.name_to_node_mapping[weight_name]
+        ov_dtype = weight_node.output(0).get_element_type()
+        dtype_map = {
+            ov.Type.f16: TensorDataType.float16,
+            ov.Type.bf16: TensorDataType.bfloat16,
+            ov.Type.f32: TensorDataType.float32,
+            ov.Type.f64: TensorDataType.float64,
+            ov.Type.i8: TensorDataType.int8,
+            ov.Type.i32: TensorDataType.int32,
+            ov.Type.i64: TensorDataType.int64,
+            ov.Type.u8: TensorDataType.uint8,
+        }
+        return dtype_map.get(ov_dtype)
+
     def set_weight(
         self, node_with_weight: NNCFNode, weight_port_id: int, model: ov.Model, graph: NNCFGraph, weight: Tensor
     ):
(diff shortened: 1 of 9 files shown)


nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -106,6 +106,28 @@ def get_weight(self, node_with_weight: NNCFNode, weight_port_id: int, model: ov.
         weight_tensor = get_const_value(weight_node)
         return Tensor(weight_tensor)
 
+    def get_weight_dtype(
+        self, node_with_weight: NNCFNode, weight_port_id: int, model: ov.Model, graph: NNCFGraph
+    ) -> TensorDataType:
+        weight_name = node_with_weight.layer_attributes.constant_attributes[weight_port_id]["name"]
```
alexsu52, line 112: I think suboptimal constant type definition may be a common problem for NNCF. What do you think about adding dtype to constant_attributes and reusing it in the algorithm?

nncf/quantization/algorithms/weight_compression/torch_backend.py
```diff
@@ -170,6 +170,16 @@ def get_weight(
...
         return Tensor(weight)
 
+    def get_weight_dtype(
+        self, node_with_weight: NNCFNode, weight_port_id: int, model: torch.nn.Module, graph: NNCFGraph
+    ) -> TensorDataType:
+        return self.get_weight(node_with_weight, weight_port_id, model, graph).dtype
+
+    @staticmethod
+    def get_weight_shape(node_with_weight: NNCFNode, weight_port_id: int, graph: NNCFGraph) -> Tuple:
+        weight_node = graph.get_previous_nodes(node_with_weight)[weight_port_id]
```
alexsu52, line 180:
```suggestion
        weight_node = get_const_node(node_with_weight, weight_port_id, graph)
```

alexsu52 (commented):
Good stuff!

nncf/quantization/algorithms/weight_compression/openvino_backend.py
nikita-savelyevv: Added dtype to OV node constant_attributes

nncf/quantization/algorithms/weight_compression/torch_backend.py
nikita-savelyevv: Thanks, changed

alexsu52 (approved):
LGTM

nikita-savelyevv:
> Did you run `post_training_weight_compression_performance`? It would be useful to see how we compare on this scope with previous behavior.

Thanks for the suggestion. Will run. Do you know, are there models with bf16 weights?

ljaljushkin:
> > Did you run `post_training_weight_compression_performance`? It would be useful to see how we compare on this scope with previous behavior.
> 
> Thanks for the suggestion. Will run. Do you know, are there models with bf16 weights?

Don't think we can run on SPR via CI. maybe run locally on it?

nikita-savelyevv:
I did not notice any changes in time or memory of the `post_training_weight_compression_performance` build (id 14) against the results for develop (id 15). As I understand this is because there `compress_weights()` is called on a model with weights in FP32 precision. At least I see that an original tiny-llama model contains only fp32 constants. It is also saved into `fp32_models` directory.

I also ran `post_training_weight_compression` job (id 158). There are no regressions.

ljaljushkin (approved):

PR #2979, opened 2024-09-20
NNCF helpers

KodiaqQ:
### Changes

- Added helpers module to NNCF.
- Added dataset helper.

### Reason for changes

- Extended data-free usage.

### Related tickets

- 152550

### Tests

- Added `tests/common/test_helpers.py`


AlexKoff88:
I am ok with the code but two comments from my side:
- Why is it not a part of `nncf.data` namespace?
- Why do we need requirements file inside, especially with "-c constraints.txt"?

KodiaqQ:
> * Why is it not a part of `nncf.data` namespace?

Since NNCF does not have `helpers` namespace, it was introduced. Should I introduce `nncf.data.helpers` module for a corresponded namespace?

> * Why do we need requirements file inside, especially with "-c constraints.txt"?

The provided code utilizes `torch` and `transformers` modules. They are not part of the main NNCF requirements, so I provided an external `requirements.txt` list for the helpers module.

AlexKoff88:
> > * Why is it not a part of `nncf.data` namespace?
> 
> Since NNCF does not have `helpers` namespace, it was introduced. Should I introduce `nncf.data.helpers` module for a corresponded namespace?

I missed that part during the last review but I would like to name the function `generate_text_data(model: PreTrainedModel, tokenizer: PreTrainedTokenizer, max_length: int = 32,  num_samples: int = 128) -> List[str]`
It is up to you how to arrange it internally. I would like to see the import like this: `from nncf.data import generate_text_data`.

> 
> > * Why do we need requirements file inside, especially with "-c constraints.txt"?
> 
> The provided code utilizes `torch` and `transformers` modules. They are not part of the main NNCF requirements, so I provided an external `requirements.txt` list for the helpers module.

Ok, thanks.

AlexanderDokuchaev:
@KodiaqQ @AlexKoff88 @alexsu52 
Is planning to add other helper function to nncf.helpers?
May be will be better to locate the function in `nncf.data` or `nncf/quantization/algorithms` ?

KodiaqQ:
> function in `nncf.data` or `nncf/quantization/algorithms` ?

Please suggest the appropriate place for this function considering that this method requires `transformers` & `torch` Python packages and shouldn't be mixed with the `nncf.Dataset`.

nncf/helpers/dataset.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) 2024 Intel Corporation
```
alexsu52, line 1: I would suggest to move this file to `nncf/data/generators.py`

nncf/helpers/requirements.txt
```diff
@@ -0,0 +1,3 @@
+-c ../../constraints.txt
```
alexsu52, line 1: I would suggest to remove requirements.txt and implement lazy checking of installed packages. Many repositories use this approach for additional requirements, such as `transformers`, `optimum` and etc.

alexsu52 (changes requested):

nncf/helpers/dataset.py
KodiaqQ: Why `generators`? How it would correlate with the requirement to have `from nncf.data import generate_text_data` import?

nncf/helpers/requirements.txt
KodiaqQ: Could you please provide the proper example? As I understand, the current approach is not valid for you.

andreyanufr (approved):

nncf/helpers/requirements.txt
alexsu52: Example: 
https://github.com/huggingface/optimum/blob/2fb5ea5ca7ca8ea887af2851cce80ab2545d3f4f/optimum/gptq/quantizer.py#L742

nncf/helpers/dataset.py
alexsu52: It does not influence to the requirement to have from `nncf.data import generate_text_data` import. Could you make a proposal taking account removing `helpers` folder?

KodiaqQ: No proposals anymore.

nncf/helpers/requirements.txt
KodiaqQ: Done.

tests/common/test_dataset_generators.py
```diff
@@ -0,0 +1,74 @@
...
+from torch.random import manual_seed
+from transformers import AutoModelForCausalLM
+from transformers import AutoTokenizer
+
+import nncf
+from nncf.data import generate_text_data
+from tests.cross_fw.shared.helpers import load_json
+from tests.cross_fw.shared.paths import TEST_ROOT
+
+manual_seed(0)
```
AlexanderDokuchaev, line 22: Please use fixture to set seed like https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/conftest.py#L256
to guaranty that it will not affect to another tests and seed will be set individually for each test

tests/common/requirements.txt
```diff
@@ -1,4 +1,6 @@
 -c ../../constraints.txt
+torch
```
AlexanderDokuchaev, line 2: As it used torch and transformers, i suggest to add this test to tests/torch

nncf/data/generators.py
```diff
@@ -0,0 +1,85 @@
...
+
+TModel = TypeVar("TModel")
+TTokenizer = TypeVar("TTokenizer")
+
+
+def generate_text_data(
+    model: TModel, tokenizer: TTokenizer, seq_len: int = 32, dataset_size: int = 128, unique_tokens_lower_limit: int = 5
+) -> List[str]:
+    """
+    Generates text dataset based on the model output.
```
AlexanderDokuchaev, line 26: It will be good to add to docstring  limitations for model and that function requires torch and transformers modules.

AlexanderDokuchaev (changes requested):

nncf/data/generators.py
```diff
@@ -0,0 +1,85 @@
...
+    for vocab_size_name in vocab_size_names:
+        if hasattr(model.config, vocab_size_name):
+            vocab_size = getattr(model.config, vocab_size_name)
+
+    step_num = max(1, vocab_size // dataset_size)
+    ids_counter = 0
+
+    while len(generated_data) < dataset_size:
+        # Creating the input for pre-generate step
+        input_ids = torch.tensor([[ids_counter % vocab_size]])
```
AlexanderDokuchaev, line 70: This function will be used only for torch modules or for OV too?
May be will useful to add option to return numpy.array to avoid installation torch? (add argument pt_tensor)

AlexanderDokuchaev (changes requested):

nncf/data/generators.py
KodiaqQ: OV can pass torch tensor as the input too.
It is not possible to avoid torch installation due to `transformers.PreTrainedModel` being the input. 
See the source for this class - https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/modeling_utils.py#L1297, which contains `(nn.Module, ModuleUtilsMixin, GenerationMixin, PushToHubMixin, PeftAdapterMixin)` as the base classes.

nncf/data/generators.py
KodiaqQ: Added.

tests/common/test_dataset_generators.py
KodiaqQ: Done.

tests/common/requirements.txt
KodiaqQ: Done. From now, we'll test the common code in `tests/torch`.

AlexanderDokuchaev (approved):

alexsu52 (commented):
I would suggest to add an example of usage `generate_text_data` in `https://github.com/openvinotoolkit/nncf/tree/develop/examples/llm_compression`. It could be done in the follow-up PR, but you need to think about convenient use of the function in this PR.

nncf/data/generators.py
```diff
@@ -0,0 +1,89 @@
...
+
+BASE_VOCAB_SIZE = 12000
+
+TModel = TypeVar("TModel")
+TTokenizer = TypeVar("TTokenizer")
+
+
+def generate_text_data(
+    model: TModel, tokenizer: TTokenizer, seq_len: int = 32, dataset_size: int = 128, unique_tokens_lower_limit: int = 5
+) -> List[str]:
```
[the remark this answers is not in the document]
KodiaqQ: @AlexKoff88 asked me to modify this in a few comments below:

> I missed that part during the last review but I would like to name the function generate_text_data(model: PreTrainedModel, tokenizer: PreTrainedTokenizer, max_length: int = 32,  num_samples: int = 128) -> List[str]

I changed as was asked. For now, you're requesting to roll back this change. Please, discuss this with @AlexKoff88 and provide your final decision on that.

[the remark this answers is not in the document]
MaximProshin: @AlexKoff88 , can you please describe the use case with some example so that it's clear how the new API will be used? I see that there is some misunderstanding.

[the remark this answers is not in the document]
KodiaqQ: @alexsu52, are there any updates on that? What is the final decision?

KodiaqQ:
> I would suggest to add an example of usage `generate_text_data` in `https://github.com/openvinotoolkit/nncf/tree/develop/examples/llm_compression`. It could be done in the follow-up PR, but you need to think about convenient use of the function in this PR.

An example of the non-API usage method was added.

alexsu52:
> > I would suggest to add an example of usage `generate_text_data` in `https://github.com/openvinotoolkit/nncf/tree/develop/examples/llm_compression`. It could be done in the follow-up PR, but you need to think about convenient use of the function in this PR.
> 
> An example of the non-API method usage was added.

Great example! Please add test for the example.

KodiaqQ:
@AlexKoff88, re-review an example, please.

KodiaqQ:
> Great example! Please add test for the example.

Done.

alexsu52 (approved):
LGTM

MaximProshin (approved):

AlexKoff88 (approved):

KodiaqQ:
Since there are some issues with the `test_examples`, I propose to disable this example in tests until the issue is investigated and merge this functionality.
@alexsu52, @MaximProshin, what do you think?

MaximProshin:
@KodiaqQ 

> Since there are some issues with the `test_examples`, I propose to disable this example in tests until the issue is investigated and merge this functionality. @alexsu52, @MaximProshin, what do you think?

I agree

alexsu52:
> Since there are some issues with the `test_examples`, I propose to disable this example in tests until the issue is investigated and merge this functionality. @alexsu52, @MaximProshin, what do you think?

Good idea!

PR #2986, opened 2024-09-24
Add deprecation message for nncf.torch.create_compressed_model

AlexanderDokuchaev:
### Changes

Add deprecation massage for `nncf.torch.create_compressed_model`

### Related tickets

152996

### DIFF
--- a/nncf/torch/model_creation.py
+++ b/nncf/torch/model_creation.py
@@ -103,7 +103,14 @@ def create_compressed_model(
     """
 
     warning_deprecated(
-        "The 'nncf.torch.create_compressed_model' function is deprecated and will be removed in a future release."
+        "The 'nncf.torch.create_compressed_model' function is deprecated and will be removed in a future release.\n"
+        "To perform post training quantization (PTQ) or quantization aware training (QAT),"
+        " use the new nncf.quantize() API:\n"
+        " - https://github.com/openvinotoolkit/nncf?tab=readme-ov-file#post-training-quantization\n"
+        " - https://github.com/openvinotoolkit/nncf?tab=readme-ov-file#training-time-quantization\n"
+        "Examples:\n"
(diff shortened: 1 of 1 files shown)


nncf/torch/model_creation.py
```diff
@@ -100,6 +101,11 @@ def create_compressed_model(
         is an instance of CompositeCompressionController) and the model ready for compression parameter training wrapped
         as an object of NNCFNetwork.
     """
+
+    warning_deprecated(
+        "The 'nncf.torch.create_compressed_model' function is deprecated and will be removed in a future release."
```
MaximProshin, line 106: The deprecation message includes a recommendation of what to use instead. Let's add it.

MaximProshin (changes requested):

nncf/torch/model_creation.py
AlexanderDokuchaev:
```
The 'nncf.torch.create_compressed_model' function is deprecated and will be removed in a future release.
For quantization use 'nncf.quantize', another algorithms are deprecated.
```
ok?

MaximProshin: @AlexKoff88 , we deprecate API here but we have to state if there is an alternative recommended way. For the quantization, it's nncf.quantize() but there is no one for other algorithms. Are you ok if we state them as deprecated?

AlexKoff88: It is a good idea and we can say something like "For QAT, use the new nncf.quantize() API from the example ...". I would omit to mention other algorithms as there is no clarity of what we will keep eventually.

MaximProshin: @AlexanderDokuchaev , please update the warning and then we will merge the PR.

AlexanderDokuchaev: Updated

MaximProshin (approved):

AlexKoff88 (approved):

PR #2988, opened 2024-09-25
[OV] Add adaptive poolings metatypes

kshpv:
### Changes

Add https://docs.openvino.ai/2024/documentation/openvino-ir-format/operation-sets/operation-specs/pooling/adaptive-max-pool-8.html and https://docs.openvino.ai/2024/documentation/openvino-ir-format/operation-sets/operation-specs/pooling/adaptive-avg-pool-8.html

Based on the description of the operations I make their treatments for quantization the same as for AvgPool and MaxPool. I checked the correct quantization scheme with this PR for litehernet from ticket 150502


alexsu52 (approved):
LGTM

PR #2984, opened 2024-09-24
[Torch FX] Post Quantize Weights Compression

anzr299:
### Changes

Transformation for removing fake quantize nodes and saving all weights to disk in int8 format after quantization. It works as follows:
1. Reshape the scale if qdq operation is per-channel.
2. Pattern match the quantize-dequantize nodes.
3. Filter the matches to only include quantize-dequantize ops with constant input.
4. Replace with the multiplication of the scale and input.

### Reason for changes

To compress the model after quantization

### Tests

Add `test_post_quantization_compression()` in `tests/torch/fx/test_model_transformer.py` which checks the data type of all weights in the model after applying quantization and also checks the value after the decompression step (element-wise multiplication operation).

### Tickets
#2766

PR #2996, opened 2024-09-30
[Conformance] TorchFX/OV backends Alignment

daniil-lyakhov:
### Changes

* ~~Constant folding is applied to all TorchFX models before the quantization~~
* Some torchvision models (swin_v2_s, vit_16_b) are exported by `torch.export.export` before ov conversation
* Moc transformations are applied to openvino compressed models after the compression

After the #2984 
* Fixed `_compress_qdq_constant_transformation` for per tensor case

### Reason for changes

* To align TorchFX/OV quantized models

### Related tickets

#2766

### Tests

post_training_quantization/504/ is finished successfully


nncf/experimental/torch/fx/constant_folding.py
```diff
@@ -0,0 +1,266 @@
+# Copyright (c) 2024 Intel Corporation
```
alexsu52, line 1: I don't find where you use constant folding transformation.

tests/post_training/pipelines/base.py
```diff
@@ -418,6 +418,9 @@ def save_compressed_model(self) -> None:
             ov.serialize(ov_model, self.path_compressed_ir)
         elif self.backend in OV_BACKENDS:
             self.path_compressed_ir = self.output_model_dir / "model.xml"
+            from openvino._offline_transformations import apply_moc_transformations
+
+            apply_moc_transformations(self.compressed_model, cf=True)
```
alexsu52, line 423: Could you please explain the reason for using this transformation here?

nncf/experimental/torch/fx/constant_folding.py
daniil-lyakhov: https://github.com/openvinotoolkit/nncf/pull/2996/files#diff-76750172cf1be3020f4d3a5a71609a952f1c009e24960424d4d02175613317a5R82-R83

tests/post_training/pipelines/base.py
daniil-lyakhov: This is done to align amount of FQ between OV and FX backends for the VIT model 
image

nncf/experimental/torch/fx/constant_folding.py
alexsu52: Could you explain why you use the constant folding transformation only in tests. How will user apply the constant folding transformation to the model?

daniil-lyakhov: Fixed, please check

alexsu52:
Please rerun post_training_quantization build.

daniil-lyakhov:
> Please rerun post_training_quantization build.

/post_training_quantization/504/ is finished successfully

alexsu52 (approved):

AlexanderDokuchaev:
Please add unit tests for constant folding and docstrings.

daniil-lyakhov:
> Please add unit tests for constant folding and docstrings.

tested by `test_models` by `swin_v2_s` and `vit_b_16` models. The same test approach is done for other transformation.
If you think we definitely need a unit test - let's create an issue for that

AlexanderDokuchaev:
@alexsu52  i will not review PR that adding transformation that will applied for any model in `nncf.quantize` without unit tests and any documentation.

daniil-lyakhov:
@AlexanderDokuchaev, please take a look

anzr299:
In the case where the user gives a model with already inserted Quantize-Dequantize or Quantize-random_nodes-Dequantize, shouldn't it be ignored by constant folding?
cc @alexsu52

nncf/experimental/torch/fx/transformations.py
```diff
@@ -669,13 +669,17 @@ def _compress_qdq_constant_transformation(model: torch.fx.GraphModule, matches)
...
+        def get_const(arg: Union[torch.fx.Node, float, int]):
+            if isinstance(arg, torch.fx.Node):
+                return get_tensor_constant_from_node(arg, model)
+            return arg
+
         weight_node = get_const(nodes_map["weight"])
         scale_node = get_const(nodes_map["scale"])
         zp_node = get_const(nodes_map["zero_point"])
-        axis = nodes_map["axis"]
+        axis = nodes_map.get("axis")
```
anzr299, line 682: I think we can use the same function for axis too.
```suggestion
        axis = get_const(nodes_map["axis"])
```

anzr299:
> @anzr299, I fixed `_compress_qdq_constant_transformation` function in the per tensor case and tested it by `test_compress_post_quantize_transformation`, take a look

Oh, yes that implementation did not consider `per_tensor` quantization. Thank you for pointing it out!

nncf/experimental/torch/fx/transformations.py
daniil-lyakhov: Nope, per tensor case has no "axis" key, `[]` raises key error. Intended axis value for per tensor case - None. `.get` return None https://docs.python.org/3/library/stdtypes.html#dict.get

anzr299: Sorry about that, I have updated the suggestion. I meant to say that we can pass this also to the `get_const` funciton to keep it the same as others

daniil-lyakhov: But it will always be an int or None, I don't think we should check is it `torch.fx.Node` or not because it is never a `torch.fx.Node`

daniil-lyakhov: Done

anzr299 (approved):

AlexanderDokuchaev (changes requested):

tests/torch/fx/test_model_transformer.py
```diff
@@ -406,56 +410,90 @@ def get_shared_constant_nodes(nncf_graph: NNCFGraph):
...
+    conv_node = get_graph_node_by_name(model.graph, node_name)
+    if per_channel:
+        with model.graph.inserting_before(conv_node):
+            scale_node = create_getattr_from_value(
+                model,
+                model.graph,
+                "scale_node",
+                torch.ones(
+                    [
+                        3,
```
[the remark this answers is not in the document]
daniil-lyakhov: Done

tests/torch/fx/test_model_transformer.py
```diff
@@ -406,56 +410,90 @@ def get_shared_constant_nodes(nncf_graph: NNCFGraph):
...
-    q_after, dq_after = count_q_dq(quantized_after_insertion)
-    assert q_before == 5
-    assert dq_before == 6
-    assert q_after == 6
-    assert dq_after == 7
+
+    model_with_correct_pattern = _capture_model(model, ex_input)
+    insert_qdq_nodes(model_with_correct_pattern, correct_pattern=True, per_channel=is_per_channel)
+    compress_post_quantize_transformation(model_with_correct_pattern)
+    graph_name = f"compress_post_quantize_{'ph' if is_per_channel else 'pt'}_valid.dot"
```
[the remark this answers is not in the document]
daniil-lyakhov: Renamed, done

nncf/experimental/torch/fx/transformations.py
```diff
@@ -669,13 +669,17 @@ def _compress_qdq_constant_transformation(model: torch.fx.GraphModule, matches)
     for match in matches:
         mul_node = match.replacements[0]
         sub_node = match.replacements[1]
-        weight_node, scale_node, zp_node, axis = None, None, None, None
         nodes_map = {node.name: match.nodes_map[node] for node in match.nodes_map}
-        get_const = partial(get_tensor_constant_from_node, model=model)
+
+        def get_const(arg: Optional[Union[torch.fx.Node, float, int]]):
```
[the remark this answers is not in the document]
daniil-lyakhov: Done, please check

AlexanderDokuchaev (approved):

PR #3011, opened 2024-10-14
compress_weights telemetry update

KodiaqQ:
### Changes

- Added telemetry for the `nncf.compress_weights` backend-specific `impl`;
- Added missed wrapper for the `nncf.qantize` PyTorch `impl`;
- Fixed reported event for `nncf.quantize_with_accuracy_control` OpenVINO `impl`.
- Added missed `app_name`, `app_version` fields for each event.

### Reason for changes

- Telemetry update.

### Related tickets

- 154833

### Tests

- TBD

### DIFF
--- a/nncf/openvino/quantization/quantize_model.py
+++ b/nncf/openvino/quantization/quantize_model.py
@@ -51,6 +51,7 @@ from nncf.quantization.quantize_model import quantize_with_tune_hyperparams
 from nncf.quantization.quantize_model import warning_model_no_batchwise_support
 from nncf.quantization.telemetry_extractors import CompressionStartedWithCompressWeightsApi
 from nncf.quantization.telemetry_extractors import CompressionStartedWithQuantizeApi
+from nncf.quantization.telemetry_extractors import CompressionStartedWithQuantizeWithAccuracyControlApi
 from nncf.scopes import IgnoredScope
 from nncf.scopes import validate_ignored_scope
 from nncf.telemetry.decorator import tracked_function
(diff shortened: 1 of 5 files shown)


AlexKoff88 (approved):

nncf/openvino/quantization/quantize_model.py
```diff
@@ -366,6 +369,18 @@ def quantize_impl(
     )
 
 
+@tracked_function(
+    NNCF_OV_CATEGORY,
+    [
+        CompressionStartedWithCompressWeightsApi(),
```
MaximProshin, line 375: I would extend it by 'all_layers'. cc @AlexKoff88

ljaljushkin: What about `ratio` and `group_size`?

alexsu52 (approved):

nncf/openvino/quantization/quantize_model.py
MaximProshin: > What about `ratio` and `group_size`?

How do you plan to use that information?

MaximProshin:
Every new data in telemetry requires a legal approval and we can't merge the PR until then. I will try to force it so it doesn't take a lot of time. 
@alexsu52 @AlexKoff88 fyi

nncf/openvino/quantization/quantize_model.py
ljaljushkin: It might show how much the ratio and group size have been reduced in order to obtain an accurate model, as an example.

MaximProshin: Though for gs it's still ok as I think most of users will have 64 or 128

ljaljushkin: Ratio would show how often mixed precision is used, and how often customers choose full int4 model.
Often usage of mixed precision with small ratio and low group size might hint the need for more advanced technics to recover accuracy, as vector quantization, fine-tuning ...

MaximProshin: Agreed to keep only mode and algo.

ljaljushkin (approved):

MaximProshin (approved):
Feel free to merge it

PR #2995, opened 2024-09-27
Update aggregator.py

zina-cs:
### Changes

Added an error message

### Reason for changes

send warning message to avoid Inconsistencies arise when the dataset size is less than the provided or default 'subset_size'.

### Related tickets

Closes: https://github.com/openvinotoolkit/nncf/issues/2562

I had an inquiry:
I noticed that subset_size is sometimes put as 100, or 300, or specified in the advanced parameters. Should a default be used here, or could you point me to where I can find the correct subset_size to be imported?

PR #3013, opened 2024-10-14
Update main.py

zina-cs:
### Changes

adding subset_size as per @l-bat :
https://github.com/openvinotoolkit/nncf/tree/develop/examples/post_training_quantization/onnx/yolov8_quantize_with_accuracy_control - subset_size=128

### Related tickets

#2995 

After the confirmation that I understood the requirement correctly, I'll add subset_size to the rest


l-bat:
@zina-cs, please avoid opening a new PR for each file. Instead, apply all comments to the initial PR:https://github.com/openvinotoolkit/nncf/pull/2995

PR #3014, opened 2024-10-15
[Torch] INT4 weight compression

alexsu52:


nncf/quantization/algorithms/weight_compression/torch_backend.py
```diff
@@ -235,17 +236,35 @@ def transform_model(
...
+                    result_shape=weight.shape,
+                    result_dtype=weight.dtype,
+                )
+
+            # pack tensor
+            packed_tensor = decompressor.pack_weight(compressed_weight.tensor.data)
 
             # sets compressed tensor
-            compressed_parameter = torch.nn.Parameter(packed_tensor.data, requires_grad=False)
+            compressed_parameter = torch.nn.Parameter(packed_tensor, requires_grad=False)
```
ljaljushkin, line 267: Why not via `register_buffer`? Will the packed data be fine-tuned?

nncf/quantization/quantize_model.py
```diff
@@ -506,20 +510,25 @@ def compress_weights(
...
+        if unsupported_options:
             raise AttributeError(
-                "TorchFX backend supports only INT8_ASYM, INT8_SYM modes for weight compression, "
-                f"but given {mode.value} mode."
+                f"TorchFX backend does not support {', '.join(unsupported_options)} option(s). Set them to None."
             )
 
-        if backup_mode is not None:
-            raise AttributeError("TorchFX backend does not support backup_mode option.")
+        if ratio is not None and ratio != 1:
```
ljaljushkin, line 529: As far as I can see, Torch and Torch FX have the same processing of parameters. Does it make sense to combine them in a single function?

nncf/torch/quantization/quantize_functions.py
```diff
@@ -275,3 +276,62 @@ def decompress_symmetric(input: torch.Tensor, scale: torch.Tensor) -> torch.Tens
...
+    :param tensor: A tensor of dtype `torch.uint8` where each element represents a uint4 value.
+        The tensor should contain values in the range [0, 15].
+    :return: A packed tensor of dtype `torch.uint8` where each element packs two uint4 values.
+    :raises ValueError: If the input tensor is not of type `torch.uint8`.
+    """
+    if tensor.dtype != torch.uint8:
+        raise ValueError(f"Invalid tensor dtype {tensor.type}. torch.uint8 type is supported.")
+    packed_tensor = tensor.contiguous()
+    packed_tensor = packed_tensor.reshape(-1, 2)
+    packed_tensor = torch.bitwise_and(packed_tensor[..., ::2], 15) | packed_tensor[..., 1::2] << 4
```
ljaljushkin, line 295: Some unit tests for 4bit signed/unsigned packing/unpacking would be useful.

nncf/quantization/algorithms/weight_compression/torch_backend.py
alexsu52: To be honest, I took into account the fact that NNCF replaces torch.nn.Parameter in the module. I did not check, using `regester_buffer` is possible or not, also I didn't try to train compressed data.

nncf/quantization/quantize_model.py
alexsu52: Thanks for this comment. Thanks for this comment. We are going to refactor it when the torch.fx backend will be ready to move out from experiments module.

cc' @daniil-lyakhov, @anzr299

nncf/torch/quantization/quantize_functions.py
alexsu52: added test_pack_uint4 and test_pack_int4

nncf/torch/quantization/quantize_functions.py
```diff
@@ -275,3 +276,62 @@ def decompress_symmetric(input: torch.Tensor, scale: torch.Tensor) -> torch.Tens
...
+    return packed_tensor
+
+
+@register_operator()
+def unpack_uint4(packed_tensor: torch.Tensor) -> torch.Tensor:
+    """
+    Unpacks a tensor where each uint8 element stores two uint4 values back into a tensor with
```
ljaljushkin, line 302:
```suggestion
    Unpacks a tensor, where each uint8 element stores two uint4 values, back into a tensor with
```

nncf/torch/quantization/quantize_functions.py
```diff
@@ -275,3 +276,62 @@ def decompress_symmetric(input: torch.Tensor, scale: torch.Tensor) -> torch.Tens
...
+    Packs a tensor containing uint4 values (in the range [0, 15]) into a tensor with uint8 values,
+    where each element stores two uint4 values.
+
+    :param tensor: A tensor of dtype `torch.uint8` where each element represents a uint4 value.
+        The tensor should contain values in the range [0, 15].
+    :return: A packed tensor of dtype `torch.uint8` where each element packs two uint4 values.
+    :raises ValueError: If the input tensor is not of type `torch.uint8`.
+    """
+    if tensor.dtype != torch.uint8:
+        raise ValueError(f"Invalid tensor dtype {tensor.type}. torch.uint8 type is supported.")
```
ljaljushkin, line 292: What about raising nncf-specific error here?

ljaljushkin (changes requested):

nncf/torch/quantization/quantize_functions.py
alexsu52: done

ljaljushkin (approved):

andreyanufr (approved):

PR #3023, opened 2024-10-17
Remove warning from examples

l-bat:
### Changes

Add `subset_size` to PTQ samples after merging https://github.com/openvinotoolkit/nncf/pull/2995 to remove the warning message
CI: https://github.com/openvinotoolkit/nncf/actions/runs/11436915496


zina-cs:
.take

alexsu52 (changes requested):
Could you take subset size from datasets instead of using magic constants if `len` is implemented for datasets?

alexsu52 (approved):

PR #3027, opened 2024-10-18
[TorchFX] Bias fusing is removed from default transformations

daniil-lyakhov:
### Changes

* Bias fusing is removed from default transformations
* `constant_folding` is updated to remove inplace operations without users
* `extract_model` is updated to support original model output as a subgraph output

### Reason for changes

To make it possible to apply quantization the same way it done by X86Quantizer

### Related tickets

#2766
110985

### Tests
* All int8 references are updated and checked manually
* `test_constant_folding` and `test_constant_folding_with_constraints` are updated with a constant subgraph which contains an inplace op (`relu_`)
* `test_model_extraction_with_original_output` is introduced
* conformance test post_training_quantization/557 have finished successfully


daniil-lyakhov:
@anzr299, I believe this should fix the https://github.com/openvinotoolkit/nncf/pull/3077, please check

nncf/experimental/torch/fx/constant_folding.py
```diff
@@ -252,6 +265,8 @@ def constant_fold(
         for node in erased_params:
             gm.graph.erase_node(node)
 
-        gm.graph.eliminate_dead_code()
+        # Custom _is_impure function allows to eliminate all layers with zero
+        # users including inplace ops like relu_ besides output and placeholders.
+        gm.graph.eliminate_dead_code(_is_impure)
```
anzr299, line 270: It looks like this is the default behavior where `node.is_impure()` is checked which checks if node.op is placeholder or output https://github.com/pytorch/pytorch/blob/c418a9ac7501184d4a8cb4443f49e727d43f6be0/torch/fx/graph.py#L1815

anzr299: Oh, alright. I went over the documentation (https://pytorch.org/docs/stable/fx.html#torch.fx.Node.is_impure) instead of the code.

anzr299 (approved):

alexsu52 (approved):

PR #3042, opened 2024-10-28
Data-free mixed precision algorithm for Torch and Torch FX backends

alexsu52:
### Changes

Data-free mixed precision algorithm for Torch and Torch FX backends

### Reason for changes

For more accurate weight compression

### Related tickets

ref: 153918

### Tests

NNCF/job/manual/job/post_training_weight_compression/228

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/mixed_precision.py
+++ b/nncf/quantization/algorithms/weight_compression/mixed_precision.py
@@ -102,20 +102,13 @@ class MixedPrecisionCriterion(Algorithm):
             weight_param.compression_config = self._primary_config
             num_weights_in_4bit += weight_param.num_weights
 
-    @property
-    def available_backends(self) -> List[BackendType]:
-        return [BackendType.OPENVINO]
-
+    @abstractmethod
     def _set_backend_entity(self, model: TModel) -> None:
-        model_backend = get_backend(model)
-        if model_backend == BackendType.OPENVINO:
-            from nncf.quantization.algorithms.weight_compression.openvino_backend import OVMixedPrecisionAlgoBackend
+        """
+        Creates a helper class with a backed-specific logic of the algorithm.
 
-            self._backend_entity = OVMixedPrecisionAlgoBackend(model)
-        else:
-            raise nncf.UnsupportedBackendError(
-                "Cannot return backend-specific entity because {} is not supported!".format(model_backend.value)
-            )
+        :param model: Backend-specific input model.
+        """
 
     @abstractmethod
     def get_statistic_points(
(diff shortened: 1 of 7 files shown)


nncf/quantization/algorithms/weight_compression/mixed_precision.py
```diff
@@ -102,20 +102,13 @@ def apply(
...
 
-    @property
-    def available_backends(self) -> List[BackendType]:
-        return [BackendType.OPENVINO]
-
+    @abstractmethod
     def _set_backend_entity(self, model: TModel) -> None:
-        model_backend = get_backend(model)
-        if model_backend == BackendType.OPENVINO:
-            from nncf.quantization.algorithms.weight_compression.openvino_backend import OVMixedPrecisionAlgoBackend
```
ljaljushkin, line 112: is `OVMixedPrecisionAlgoBackend` still needed?

alexsu52: Thanks, typo.

nncf/quantization/quantize_model.py
```diff
@@ -511,7 +511,6 @@ def compress_weights(
             )
 
         options = {
-            "sensitivity_metric": sensitivity_metric,
             "awq": awq,
```
andreyanufr, line 514: What happens if someone pass awq=False, scale_estimation=False, gptq=False ?

nncf/quantization/quantize_model.py
```diff
@@ -511,20 +511,22 @@ def compress_weights(
...
             "scale_estimation": scale_estimation,
             "gptq": gptq,
             "lora_correction": lora_correction,
         }
-        unsupported_options = [name for name, value in options.items() if value is not None]
+        unsupported_options = [name for name, value in options.items() if value]
         if unsupported_options:
             raise nncf.ParameterNotSupportedError(
-                f"Torch backend does not support {', '.join(unsupported_options)} option(s). Set them to None."
+                f"Torch backend does not support {', '.join(unsupported_options)} option(s). Set them to False."
```
ljaljushkin, line 522: The behavior is changed but not for all backends, I'd align in all places then.

ljaljushkin: From the user standpoint, it's a bit confusing that explicit disabling of not supported algorithms is not the same as having them disabled by default. 

On the other hand, default values for some unsupported options might be even more confusing.
For example, for `all_layers` and `backup_mode`. 

So I'd leave the current behavior on develop - no explicit values for unsupported options.

andreyanufr: I agree

alexsu52: The last commit was reverted.

ljaljushkin (approved):

andreyanufr (approved):

PR #2942, opened 2024-09-03
GPTQ + Scale Eestimation conformance test

alexsu52:
### Changes

replaced tinyllama_data_aware_gptq test by tinyllama_data_aware_gptq_scale_estimation_stateful test in the weight compression conformance test

### Reason for changes

Test regression in the GPTQ + Scale Estimation pipeline

### Related tickets

ref 143637

### Tests

tinyllama_data_aware_gptq_scale_estimation_stateful
NNCF/job/manual/job/post_training_weight_compression/226

PR #3044, opened 2024-10-28
[AWQ] Cast `fns.quantile()` result to float32

nikita-savelyevv:
### Changes

Cast `fns.quantile()` result to float32 inside AWQ algorithm.

### Reason for changes

`fns.quantile()` for numpy backend returns `np.float64` value. In AWQ it is used as a clip lower bound, resulting in float64 result. Then via chain reaction it leads to weights and activations being converted to float64. 

As I understand, processing in float64 is not necessary. At the same time it leads to increased running time. Below are measurements for compression time with AWQ enabled before and after the changes.

| Model           | develop (sec.) | branch (sec.) |
|-----------------|----------------|---------------|
| tiny-llama-1.1b | 123            | 109 (-11%)    |
| phi3_mini-3.7b  | 487            | 419 (-14%)    |
| llama3-8b       | 1091           | 912 (-16%)    |

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/awq.py
+++ b/nncf/quantization/algorithms/weight_compression/awq.py
@@ -36,6 +36,7 @@ from nncf.quantization.algorithms.weight_compression.weight_lowering import do_i
 from nncf.quantization.algorithms.weight_compression.weight_lowering import do_nf4_dequantization
 from nncf.quantization.algorithms.weight_compression.weight_lowering import do_nf4_quantization
 from nncf.quantization.passes import transform_to_inference_graph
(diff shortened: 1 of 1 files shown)


nikita-savelyevv:
Post-training weight compression conformance test shows no accuracy degradations (`NNCF/job/manual/job/post_training_weight_compression/229`). I've compared against build `NNCF/job/manual/job/post_training_weight_compression/226`, which was run for a PR #2942 that added only a test and no functionality changes.

andreyanufr (approved):

ljaljushkin (approved):

PR #3054, opened 2024-11-01
Make SE algorithm also return zero points

nikita-savelyevv:
### Changes

1. Made Scale Estimation return zero points besides scales.
2. Minor fixes to `do_int_quantization` function, made its signature more strict.

### Reason for changes

1. Currently zero points need to be re-computed for the weights for which SE computed scales.
2. Avoid unnecessary computations inside `do_int_quantization`. Avoid wrong usage of the function.

### Tests

Added a test to verify correctness of feeding precomputed quantization parameters to `do_int_quantization`.

### Tickets
Prerequisite to 139047.

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/weight_lowering.py
+++ b/nncf/quantization/algorithms/weight_compression/weight_lowering.py
@@ -358,6 +358,12 @@ def do_int_quantization(
     """
     assert config.is_integer(), "The function supports integer quantization only"
     group_size = config.group_size
+    is_asym = config.mode in [CompressWeightsMode.INT8_ASYM, CompressWeightsMode.INT4_ASYM]
+    if is_asym and (precomputed_scale is None) != (precomputed_zero_point is None):
+        raise ValueError(
+            "If precomputed quantization parameters are provided, both scale and zero point are required "
+            "for asymmetric quantization."
+        )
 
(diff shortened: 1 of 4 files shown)


nikita-savelyevv:
No regressions on NNCF/job/manual/job/post_training_weight_compression/240

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -358,6 +358,12 @@ def do_int_quantization(
     """
     assert config.is_integer(), "The function supports integer quantization only"
     group_size = config.group_size
+    is_asym = config.mode in [CompressWeightsMode.INT8_ASYM, CompressWeightsMode.INT4_ASYM]
+    if is_asym and (precomputed_scale is None) != (precomputed_zero_point is None):
+        raise ValueError(
```
andreyanufr, line 363: Is it working with data free approach for INT4_ASYM?

nikita-savelyevv: During data-free asymmetric compression, both `precomputed_scale` and `precomputed_zero_point` are None, so the if condition does not hold and the exception is not raised.

andreyanufr (approved):

ljaljushkin (approved):

PR #3055, opened 2024-11-01
Divide as `a * (1.0 / b)` during weight compression

nikita-savelyevv:
### Changes

- Introduce `fns.reciprocal(x) = 1.0 / x` operation
- Replace regular division `a / b` with `a * fns.reciprocal(b)` during weight compression -related computations.

| Model                         | Develop PPL | Branch PPL | Develop Similarity | Branch Similarity |
|-------------------------------|-------------|------------|---|---|
| phi3-mini int4_asym           | 10.96023    | 10.95891 (-0.012%)   |   0.89861   |  0.89635 (-0.25%)  |
| phi3-mini int4_asym SE |  10.59984   | 10.60182 (+0.019%)  |   0.88692   |  0.88924  (+0.25%) |
| phi3-mini int4_asym GPTQ | 10.73838    | 10.72351 (-0.139%)  |   0.88555   |  0.86824 (-1.96%)  |
| phi3-mini int4_asym GPTQ + SE | 10.51594    | 10.50970 (-0.059%)  |  0.871897    |  0.89654 (+2.82%)  |

### Reason for changes

Align with how division is performed in OpenVINO. Prerequisite to #2727. This should be merged first in order to later track only the differences introduced by the OV computations.

### Related tickets

139047

### Tests

Added a test comparing inverted division result to a reference value.
NNCF/job/manual/job/post_training_weight_compression/271/
NNCF/job/manual/job/post_training_quantization/576/

Also weekly validation is requested in 158012


nikita-savelyevv:
`NNCF/job/manual/job/post_training_weight_compression/256/` has passed.

There is ~0.06 (0.86697 -> 0.80292) drop in similarity for `tinyllama_data_aware_gptq_scale_estimation_stateful` task in the CI. But when running locally I observe ~0.015 (0.83399 -> 0.85908) improvement.

nncf/tensor/functions/numeric.py
```diff
@@ -905,3 +905,42 @@ def ceil(a: Tensor) -> Tensor:
     :return: An array of the same type as a, containing the ceiling values.
     """
     return Tensor(ceil(a.data))
+
+
+@functools.singledispatch
+@tensor_guard
+def inverted_divide(a: Union[Tensor, float], b: Union[Tensor, float], invert: Optional[bool] = True) -> Tensor:
```
alexsu52, line 912: Taking into account UX, I would suggest that NNCF should be aligned with numpy and torch. The `reciprocal` function is used to solve your task `a*(1.0 / b)` as follow: `a * reciprocal(b)`. I would suggest to introduce `reciprocal` function instead of `inverted_divide` and `inplace_inverted_divide`.

numpy: https://numpy.org/doc/2.1/reference/generated/numpy.reciprocal.html
torch: https://pytorch.org/docs/stable/generated/torch.reciprocal.html

alexsu52 (approved):
LGTM

alexsu52 (changes requested):

nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -418,7 +418,7 @@ def estimate_scales(weight: Tensor, target: Tensor, zero_mask: Tensor, importanc
     :param importance: The importance values tensor.
     :return: The estimated scales
     """
-    ideal_scale = fns.abs(weight) / (fns.abs(target) + zero_mask)
+    ideal_scale = fns.abs(weight) * fns.reciprocal(fns.abs(target) + zero_mask)
```
alexsu52, line 421: How did you prove that this approximation is better for `cw= w*(1/scale) + zp`?

alexsu52 (changes requested):

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -154,7 +154,7 @@ def calculate_signed_scale(weight: Tensor, reduction_axes: ReductionAxes, num_bi
     w_max = fns.max(weight, axis=reduction_axes, keepdims=True)
 
     scale = fns.where(w_abs_min >= w_max, w_abs_min, -w_max)
-    scale /= level_high
+    scale *= 1.0 / level_high
```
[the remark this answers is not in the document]
nikita-savelyevv: I have changed every division operation in the compression pipeline because this is how it will work inside an openvino model during scale computation. This way we'll have a complete alignment. 

As I understand every division operation inside an OV graph is performed in the form of `a * (1.0 / b)`. This also applies the the operation you mention.

nncf/quantization/fake_quantize.py
```diff
@@ -355,11 +355,11 @@ def calculate_scale_zero_point(
     :return: Scale and Zero point values.
     """
     levels = level_high - level_low if narrow_range else level_high - level_low + 1
-    scale = ((input_high - input_low) / (levels - 1)).astype(TensorDataType.float32)
+    scale = ((input_high - input_low) * (1.0 / (levels - 1))).astype(TensorDataType.float32)
```
alexsu52, line 358: I would suggest to run the conformance test for the post training quantization.

nikita-savelyevv: Do you think this is neccessary? `calculate_scale_zero_point` is only used from `weight_lowering.py`.

alexsu52: https://github.com/openvinotoolkit/nncf/blob/8d501c7951b6a6407775576b8e78959414223755/nncf/onnx/quantization/quantizer_parameters.py#L18

nncf/quantization/algorithms/weight_compression/scale_estimation.py
nikita-savelyevv: With this change, I get `0.84391` similarity for `tinyllama_data_aware_gptq_scale_estimation_stateful` conformance test. Without it I got `0.8188` similarity.

However, PPL on phi3 model shows contradicting results, i.e. the metric is better without this change. Please see the updated PR description.

Because of the facts this change is not required for the alignment of compression methods for #2727, and division via reciprocal is in general less accurate, I propose to not to change it here.

nikita-savelyevv:
PTQ conformance test run (NNCF/job/manual/job/post_training_quantization/554) has no additional degradation compared to the run on develop branch (NNCF/job/manual/job/post_training_quantization/556)

nikita-savelyevv:
The problem with division is resolved by aligning OV division with NP division by disabling `ConvertDivide` and `ConvertDivideWithConstant` transformations. These transformation can be disabled for target division nodes by setting `"nonconvertable_divide_0"` rt_info flag.

PR #3061, opened 2024-11-05
Collect statistics from subset in weight compression

ljaljushkin:
### Changes

Set a subset size for collecting statistics

### Reason for changes

`subset size` option is ignored and the whole dataset is used

### Related tickets

n/a

### Tests

- [x] unit test for setting subset size
- [x] manual/job/post_training_weight_compression/245

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -498,9 +498,7 @@ class WeightCompression(Algorithm):
                 matmul_nodes_to_compress, graph
             )
             if statistic_points is None:
-                statistic_points = self.get_statistic_points(
-                    model, graph, matmul_input_to_output_nodes_map.keys(), self._subset_size
-                )
+                statistic_points = self.get_statistic_points(model, graph, matmul_input_to_output_nodes_map.keys())
                 statistic_points = self._collect_statistics(dataset, graph, model, statistic_points)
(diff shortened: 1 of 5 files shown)


andreyanufr (approved):

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -759,15 +759,13 @@ def get_statistic_points(
         model: TModel,
         graph: NNCFGraph,
         nodes_and_port_ids: Iterable[Tuple[NNCFNode, int]],
-        subset_size: Optional[int] = None,
     ) -> StatisticPointsContainer:
```
nikita-savelyevv, line 762: What do you think about also removing `subset_size` parameter from `MixedPrecisionCriterion.get_statistic_points()` method for two algorithms to be more aligned? I believe `subset_size` can be provided to criterion constructor similarly as it is done for `WeightCompression` algorithm.

nikita-savelyevv (commented):
A minor comment on my side

nncf/quantization/algorithms/weight_compression/algorithm.py
ljaljushkin: Had to make it optional for data-free mixed precision, but probably it's better since the parameter is passed only once.

nikita-savelyevv (approved):

kshpv (approved):

PR #3063, opened 2024-11-06
Update PTQ reference for 24.5

KodiaqQ:
### Changes

- As stated in the title.

### Reason for changes

- OV started to produce more correct SE graphs.
- NNCF SE pattern started to match with the model. The number of FQ layers was reduced from 79 to 61.

### Related tickets

- N/A

### Tests

- openvino-nightly post_training_quantization/68/ - success

Pattern wasn't applied for a model with OV 24.4 because of the Reshape layer:

There is no Reshape with OV 24.5:


KodiaqQ:
@openvinotoolkit/nncf-maintainers, can someone review this, please. Thank you.

alexsu52 (approved):
LGTM

PR #3064, opened 2024-11-06
[PT] handle inner functions

AlexanderDokuchaev:
### Changes

Add handle inner function for  `relu`,  `multi_head_attention_forward`  and `batch_norm`.
Analog of https://github.com/openvinotoolkit/nncf/blob/5c1a029104bd89ba4e554f8db82af87f0aa6ec35/nncf/torch/dynamic_graph/patch_pytorch.py#L390-L399

### Related tickets

152996

### Tests

test_compare_torch_function_with_handle_inner_function - comparing code of function with torch implementation to detect changes in torch
test_inner_functions - test traced graph of model


tests/torch2/function_hook/test_handle_inner_functions.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+@pytest.mark.parametrize("model_cls", (ReluModel, BatchNormModel, MultiHeadAttention))
+def test_inner_functions(model_cls: nn.Module, regen_ref_data: bool):
+    model = wrap_model(model_cls()).eval()
+    graph = build_graph(model, *model_cls.example_input())
+
+    dot_graph = to_pydot(graph)
+    ref_file = REF_DIR / f"inner_functions_{model_cls.__name__}.dot"
+    compare_with_reference_file(str(dot_graph), ref_file, regen_ref_data)
```
alexsu52, line 110: I would suggest to compare with torch implementation instead of the gold values.

AlexanderDokuchaev: This test check traced graph for this functions, that should contain all inner functions instead of one function from torch.nn.functions
What and how should be compared with torch implementation?

PR #2954, opened 2024-09-06
Added metrics for 2024.5 nightly

ljaljushkin:
### Changes

as stated in the title
This PR introduced changes in accuracy: https://github.com/openvinotoolkit/openvino/pull/26147
but it's coming from float error:
image

### Reason for changes

As result validation with openvino nightly failed:
image

### Related tickets

150613
151260

PR #3065, opened 2024-11-07
Returned OV 2024.3 references for OV 2024.5 in Weight Compression tests

ljaljushkin:
### Changes

Revert of https://github.com/openvinotoolkit/nncf/pull/2954
Use [references for the OV 2024.3](https://github.com/ljaljushkin/nncf_pytorch/blob/873fbb80ab07bd41a3b1a2e21f68bc1a7a394810/tests/post_training/data/wc_reference_data.yaml) when run conformance tests with OV 2024.5. 

### Reason for changes

According to the CPU team: 
> 1) In 24.4 time frame, RMS impl has been enabled (https://github.com/openvinotoolkit/openvino/pull/26147), which introduce some accuracy change due to vrsqrtss which is done by approximation. 
> 2) In 24.5 time frame, fix has been applied (https://github.com/openvinotoolkit/openvino/pull/26817) to use vsqrtss instead. With that we can see the result can be covered to that prior to https://github.com/openvinotoolkit/openvino/pull/26147
> With that (in 24.5), the accuracy result is recovered to that prior to RMS impl https://github.com/openvinotoolkit/openvino/pull/26147. Let's say that 24.4 may has some accuracy issue in some case. 24.5 fix that issue.

### Related tickets

156605
156776
156237
151260

### Tests

- [x] 47 build of weekly/job/openvino-nightly/job/post_training_weight_compression
image
- [x] 48 build for xfail tests
image

### DIFF
--- a/tests/post_training/data/wc_reference_data_2024.5.yaml
+++ b/tests/post_training/data/wc_reference_data_2024.5.yaml
@@ -6,3 +6,23 @@ tinyllama_int4_data_free_backend_TORCH:
   metric_value: 0.73873
   num_int4: 114
   num_int8: 84
+tinyllama_awq_backup_mode_none_backend_OV:
+  metric_value: 0.84783
+  num_int4: 208
+  num_int8: 0
+tinyllama_data_aware_awq_scale_estimation_backend_OV:
+  metric_value: 0.85502
+  num_int4: 94
+  num_int8: 124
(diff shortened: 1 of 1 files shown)


MaximProshin (approved):

alexsu52 (commented):
Please update metrics for xfail tests as well to keep history of changes.

PR #3080, opened 2024-11-12
[GHA] test install

AlexanderDokuchaev:

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -1,0 +1,78 @@
+name: Test install
+permissions: read-all
+
+on:
+  pull_request:  ## TODO: remove
+
+jobs:
+  install-cpu:
+    name: Test install [${{ matrix.backend }}]
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        backend:  ["torch", "tf", "onnx", "openvino"]
+    defaults:
+      run:
+        shell: bash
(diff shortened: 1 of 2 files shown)


alexsu52 (commented):
I would suggest to apply nightly schedule to the install tests, taking into account time of tests 
{3460857A-51BA-4F49-A227-28999AC5C4E7}

The install tests should be able to run manually.

MaximProshin:
> @alexsu52 @MaximProshin what you think about 2 triggers for test install:
> 
> 1. nightly:  full test cpu, gpu, all type of installation, linux and windows
> 2. post_merge: only cpu, only one build_s, only linux
> 
> Or second trigger can be used as precommit (will be fast test)

@AlexanderDokuchaev , why do you want to have some checks post merge? Do we often break smth here?

AlexanderDokuchaev:
@MaximProshin not often, but it's cheap, simple, it save time of developer who will investigate falls

alexsu52:
@AlexanderDokuchaev I agree with adding the fast test to the precommit. This solution is better than introducing post-merge phase. Could you share test time of only cpu, only one build_s, only linux?

AlexanderDokuchaev:
https://github.com/openvinotoolkit/nncf/actions/runs/12001473810

alexsu52 (approved):
LGTM

MaximProshin (approved):

PR #3081, opened 2024-11-13
[ONNX] Fix sporadic results in BC

kshpv:
### Changes

1. This PR addresses an issue using `ONNXRuntime==1.19.2` where a tensor used as both an input and output in a model shares the same memory. This causes unexpected behavior: updating the input tensor inadvertently modifies the statistics data due to memory overlap.
The issue was confirmed by calling `np.shares_memory(input_data['image'], outputs['image'])`, which returned `True`, indicating that the input and output tensors share memory. After applying the proposed changes, the same check now returns `False`, confirming that memory sharing is resolved.
To fix this, the `ONNXEngine` logic has been updated to create a copy of any output tensor that is also used as a model input. This ensures that the input tensor and statistics data remain independent, avoiding unintended side effects.

2. Merge RawReducer and NoopReducer
3. Minor fixes (remove warnings + fix bug in BC)

### Reason for changes

Regression

### Related tickets

156025

### Tests

PTQ run 549


kshpv:
545 job

kshpv:
549 ptq - passed

nncf/quantization/fake_quantize.py
```diff
@@ -122,11 +122,13 @@ def tune_range(
...
 
-    ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
     with warnings.catch_warnings():
         # If `qval` is 0 `rb` will equal `right_border`, and we don't want to show an unnecessary division by 0 warning
+        # The same for (qval - level_high)
         warnings.simplefilter("ignore")
+        ra_then_result = qval / (qval - level_high) * right_border
         rb_then_result = (qval - level_high) / qval * left_border
+    ra = fns.where(qval < level_high, ra_then_result, left_border)
     rb = fns.where(qval > 0.0, rb_then_result, right_border)
```
nikita-savelyevv, line 132: Could you please expand https://github.com/openvinotoolkit/nncf/blob/f0873ca9bc229fee8b5de101877229b8ca590a01/tests/common/quantization/test_tune_range.py#L19 to cover the new case?

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -412,23 +412,18 @@ def __init__(self, tensor_collectors: List[TensorCollector]) -> None:
...
-class RawReducer(NoopReducer):
-    def __init__(self):
-        super().__init__()
-
-    def __call__(self, x: List[Tensor]):
-        return self._reduce_out_of_place(x)
+    def _reduce_out_of_place(self, x: List[Tensor]) -> List[Tensor]:
+        output = []
+        for tensor in x:
+            output.append(tensor.clone())
```
daniil-lyakhov, line 425: Could you please expand the test to check that tensors were actually cloned?

kshpv: done. test passes on PR and do not on develop

nikita-savelyevv (approved):

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -412,23 +412,15 @@ def __init__(self, tensor_collectors: List[TensorCollector]) -> None:
...
-
-
-class RawReducer(NoopReducer):
-    def __init__(self):
-        super().__init__()
-
-    def __call__(self, x: List[Tensor]):
-        return self._reduce_out_of_place(x)
+    def _reduce_out_of_place(self, x: List[Tensor]) -> List[Tensor]:
+        return [tensor.clone() for tensor in x]
```
KodiaqQ, line 423: What is the reason for `clone`? What about increased memory consumption here?
Also, the return format was changed (at least in terms of the final dimension). Was it tested for every use case in NNCF?

nncf/quantization/fake_quantize.py
```diff
@@ -122,11 +122,13 @@ def tune_range(
         fval = -left_border * s
         qval = fns.round(fval)
 
-    ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
     with warnings.catch_warnings():
         # If `qval` is 0 `rb` will equal `right_border`, and we don't want to show an unnecessary division by 0 warning
+        # The same for (qval - level_high)
         warnings.simplefilter("ignore")
+        ra_then_result = qval / (qval - level_high) * right_border
```
KodiaqQ, line 129: Are we sure that we hide only an `irrelevant` warning here? 
What was the reason for the initial warnings? 
What is the reason for hiding them?

nncf/experimental/common/tensor_statistics/collectors.py
kshpv: during the work of reducer the tensor is reduced and used but when it comes to BC the value of the tensor from time to time looks like a garbage. After adding clone the value is stable so I do not know the exact reason but it looks like the without clone the initial tensor can be screwed up.
I tested on performance jobs and it passes but it could potentially lead to increase of memory consumption 
Why does return format is changed? I did not get it

KodiaqQ: From a high perspective, it looks like we are hiding the bug since we do not know the exact reason for the `garbaged tensor` but are trying to avoid it by cloning. Is it eligible for ONNX only or also for OV/PT backends?
Also, unnecessary memory consumption may lead to huge regression in other backends and huge models in every scenario, since RawReducer is used in different algorithms.

The format wasn't changed, you're correct.

nncf/quantization/fake_quantize.py
KodiaqQ: Thus, we're hiding the occurrence of errors because of `inf/nan` in the future, instead of reviewing the reasons for these errors. Am I right?

kshpv: no, because the ra and rb tensor will not have inf/nan because of fns.where condition, so the hiding these messages will not lead to any nan/inf values

nncf/experimental/common/tensor_statistics/collectors.py
kshpv: My assumption that this tensor is cleaned by garbage collector as noone changes nncf.Tensor inside this reducer (this is proved because the issue is fixed by clone). 
I think clone is the right behavior here as we can not rely on the fact that the tensor came will not be modified/removed

KodiaqQ: Any observations, on why the `numpy` version downgrade also resolves the issue?
From my side, these changes introduce regressions and look unnecessary because the initial reason for these is not clear.

nncf/quantization/fake_quantize.py
KodiaqQ: Thus, I see no reason to hide these calculations under the `warnings.catch_warnings()`.
For me, the existence of the `inf/nan` as the `ra_then_result` is the issue as in any other part.

kshpv: If you think so could you provide a solution to this better than I proposed and @nikita-savelyevv proposed in initial PR?

nncf/experimental/common/tensor_statistics/collectors.py
kshpv: I searched for the differences in memory management for numpy between versions and didn't find any info.
What are exact regressions? I do not see any regression for now (all tests passed, jobs passed)

nncf/quantization/fake_quantize.py
KodiaqQ: The only better solution is to analyse the root cause of the `inf/nan` instead of hiding them.

nncf/experimental/common/tensor_statistics/collectors.py
KodiaqQ: Memory consumption changes weren't analysed as I understood. This is a potential regression.

nncf/quantization/fake_quantize.py
kshpv: So no proposal, no solution from your side?

nncf/experimental/common/tensor_statistics/collectors.py
kshpv: Why do you think so? I run 2 performance jobs with memory consumption analyze....

nncf/quantization/fake_quantize.py
KodiaqQ: Again and again. The proposal is to analyse the issues with `inf/nan` instead of hiding them. I don't know the root causes of the issue so I can't propose anything to fix it.
But I do not agree with just dropping them from the view.

nncf/experimental/common/tensor_statistics/collectors.py
KodiaqQ: Provide the memory analysis then.

kshpv: https://github.com/openvinotoolkit/nncf/pull/3081#issuecomment-2476131286

nncf/quantization/fake_quantize.py
kshpv: The root cause is that there are some zeros in denominator and these values are valid, because of that nan/inf gotten during the calculation but during filtration (fns.where) these values are filtered. 
So, there is nothing bad with these values and because of this they were hidden to not confuse anyone
What exactly do you think is bad in this pipeline?

nncf/experimental/common/tensor_statistics/collectors.py
KodiaqQ: Ok. I'll do it for you.
Reference (post_training_quantization_performance - 58):
image
Your run:
image
This is only one example.

nncf/quantization/fake_quantize.py
KodiaqQ: > some zeros in denominator and these values are valid

I don't get how it can be valid. Any examples?

nncf/experimental/common/tensor_statistics/collectors.py
kshpv: good job, did you find anything suspicious, because I did not

nncf/quantization/fake_quantize.py
kshpv: https://github.com/openvinotoolkit/nncf/pull/3081/files#diff-83c7c13863051f18f7e62d4d596695e2abb1d084ed6449c3fb78b42a86024d1fR27

nncf/experimental/common/tensor_statistics/collectors.py
KodiaqQ: Memory consumption increased as we can see. Need to verify other models and experiment with some LLMs, since their consumer memory a way higher.
It also was decreased for some reason. This may be because of the test instability as well.

nncf/quantization/fake_quantize.py
KodiaqQ: This test is irrelevant to me also. I wouldn't approve this either.

kshpv: @alexsu52 could you add your thoughts on this because you approved the initial PR - https://github.com/openvinotoolkit/nncf/pull/2415 ?

nncf/experimental/common/tensor_statistics/collectors.py
KodiaqQ: I couldn't find any memory threshold for the tests in the code. Can you help here? What is the threshold?

kshpv: There are examples that are influenced by these changes have the same numbers:
develop:
image
PR:
image

develop:
image
PR:
image

Moreover, I looked at some nightly build and saw around 3% deviations there. 
Also I looked at LLM memory consumption tests as the most memory intensive and didn't see any difference. (manual job vs nightly)

Based on this data I do not see any regression here

KodiaqQ: Please provide the LLM memory consumption graphs also.
As I found, here is your PR:
image
Latest develop run:
image

kshpv: You looked at data-free method which should not be influenced by the proposed changes

KodiaqQ: As you wish.
I will not approve this change, since the issue analysis was not done properly.

KodiaqQ:
@kshpv, feel free to merge this PR as you need.

nncf/quantization/fake_quantize.py
```diff
@@ -122,11 +122,13 @@ def tune_range(
         fval = -left_border * s
         qval = fns.round(fval)
 
-    ra = fns.where(qval < level_high, qval / (qval - level_high) * right_border, left_border)
```
alexsu52, line 125: @kshpv, could you briefly explain the reason of these changes?

kshpv: During the run of ptq for `torchvision/mobilenet_v3_small_BC` the runtimewarning appears:
```
RuntimeWarning: invalid value encountered in multiply
    return Tensor(self.data * unwrap_tensor_data(other))
```

kshpv:
Locally memory consumption with the changes is similar to `ONNXRuntime==1.17.3`

nncf/quantization/fake_quantize.py
```diff
@@ -122,11 +121,14 @@ def tune_range(
...
-    with warnings.catch_warnings():
-        # If `qval` is 0 `rb` will equal `right_border`, and we don't want to show an unnecessary division by 0 warning
-        warnings.simplefilter("ignore")
-        rb_then_result = (qval - level_high) / qval * left_border
+    # Avoid division by zero
+    qval_nonzero = fns.where(qval == 0, fns.ones_like(qval), qval)
```
alexsu52, line 125: Please, check the performance of the function after your changes. @nikita-savelyevv, it looks like we discussed it already. Could you remember us the solution?

nikita-savelyevv: As I remember ignoring the warning was the solution with the least impact.

kshpv: Perf measurment:
with warnings context manager: ~17.648 sec
without warning context manager but with: ~20.809 sec
```
qval_nonzero = fns.where(qval == 0, fns.ones_like(qval), qval)
qval_not_high = fns.where(qval - level_high == 0, fns.ones_like(qval), qval - level_high)
```

kshpv: > @kshpv you can use just `1.0` instead of `fns.ones_like(qval)`
> 
> ```
> qval_nonzero = fns.where(qval == 0, 1.0 , qval)
> qval_not_high = fns.where(qval - level_high == 0, 1.0 , qval - level_high)
> ```

I checked your proposed version and it has the same performance as with `fns.ones_like(qval)` :(

kshpv:
Found the open issue with the same problem on ONNXRuntime - https://github.com/microsoft/onnxruntime/issues/21922

nncf/quantization/fake_quantize.py
alexsu52: I suggest to use implementation with the best performance.

alexsu52 (approved):
LGTM

AlexanderDokuchaev (approved):

PR #3083, opened 2024-11-13
Set statistics path to example directory in tiny_llama_find_hyperparams

AlexanderDokuchaev:
### Changes

Save statistics in example directory
Add statistics path to .gitignore

### Tests
https://github.com/openvinotoolkit/nncf/actions/runs/11822254518


kshpv (approved):
Thanks!

PR #3017, opened 2024-10-15
[WeightCompression] Statistics caching

kshpv:
### Changes

Add statistics saving and loading for the `WeightCompression `algorithm:

1. Statistics are cached for all configurations such as:
`awq = True, scale_estimation=True with all type of sensitivities.`
3. Then the statistics are dumped in a directory which can be reused for any `weights_compression()` configuration.

The example for tinyllama was updated with this functionality.

More changes:
1. Make all statistics used in `WeightCompression` aligned with `TesnorStatistics` from `nncf/experimental/common/tensor_statistics/statistics.py`
2. Extend `StatisticsAggregator` by the logic of loading and saving statistics.
3. Dumping statistics is done using pickle and gzip. Serialization methods were added for `Tensor`.
4. Introduced `statistics_serializer `and `statistics_validator` to handle the statistics loading/dumping to the file.

Statistics sizes

Model  | subset size | statistics directory size | statistics collection time |
-- |  -- | -- | -- |
tinyllama | 128 | 100 MB | 61 sec |
Phi-3-mini-4k-instruct | 128 | 258 MB | 51 sec | 
ruDialoGPT-medium | 128 | 80 MB | 9 sec | 
 llama-3.1-8b | 128 | 393 MB | 95 sec | 

### Reason for changes

Speed up compression configuration finding.

### Related tickets

153129

### Tests

Test coverage were extended by tests on `statistics_serializer`, `statistics_validator`, `StatisticsAggregator` and on `WeightCompression` algorithm with the proposed functional.

PR #3097, opened 2024-11-19
[Release] Add symlink verification before opening file

KodiaqQ:
### Changes

- As stated in the title.

### Reason for changes

- Fix after #3017.

### Related tickets

- 157068

### Tests

- Manual

### DIFF
--- a/nncf/common/tensor_statistics/statistics_serializer.py
+++ b/nncf/common/tensor_statistics/statistics_serializer.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Tuple, cast
 
 import nncf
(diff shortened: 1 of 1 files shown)


nncf/common/tensor_statistics/statistics_serializer.py
```diff
@@ -71,6 +72,7 @@ def load_from_dir(dir_path: str) -> Tuple[Dict[str, Any], Dict[str, str]]:
             continue  # Skip the metadata file
 
         try:
+            fail_if_symlink(statistics_file)
```
kshpv, line 75: Could you add a specific function, such as `safe_open,` and use it here and below?

PR #3105, opened 2024-11-21
[TorchFX] YoloV11 support

daniil-lyakhov:
### Changes

* Torch SDPA pattern is updated
* As the concat node has his input nodes in format `args=([inp_1, ..., inp_n], dim)`, thus it should be treated differently. Retrieving concat inputs by input port id was supported in each TorchFX transformation

### Reason for changes

* To support quantization of ultralytics/yolo11n in TorchFX backend

### Related tickets

#2766 
157032

### Tests

* `tests/torch/fx/test_model_transformer.py` and `tests/torch/fx/test_compress_weights.py` are updated to check all cases with the concat node. All .`dot` / `.json`  were checked manually.
* `tests/torch/fx/test_models.py` is updated with `YOLO11N_SDPABlock` synthetic model to check the correctness of SDPA pattern matching


alexsu52 (approved):
LGTM

anzr299 (approved):

PR #3112, opened 2024-11-25
[GHA] examples on windows

AlexanderDokuchaev:
### Changes

Add windows for test examples

### Related tickets

153650

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -112,15 +112,13 @@ jobs:
           export LIB="${LIB};$(python -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))")"
           export LIB="${LIB};$(python -c "import sys; print(sys.prefix + '/libs')")"
           export INCLUDE="${INCLUDE};$(python -c "import sysconfig; print(sysconfig.get_path('include'))")"
-          echo $LIB
-          echo $INCLUDE
           python -m pytest -s -ra tests/cross_fw/examples \
             --junit-xml=pytest-results.xml \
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev:
https://github.com/openvinotoolkit/nncf/actions/runs/12033549644?pr=3112

.github/workflows/examples.yml
```diff
@@ -12,6 +12,10 @@ on:
       pytest_args:
         description: 'Pytest arguments'
         default: ''
+      skip_windows:
+        description: 'Skip tests on Windows'
+        type: boolean
+        default: true
```
alexsu52, line 18: Could you briefly explain how to use this option and why this default value is true?

MaximProshin (approved):

.github/workflows/examples.yml
AlexanderDokuchaev: To dont run windows job by default for manual run

alexsu52: I would suggest to set false by default.

AlexanderDokuchaev: I made it as true, to avoid unnecessary launch on windows:
- it's longer
- linux tests covered 99% falls
- 99% changes is no OS specific

In my mind, windows checks should be run only if:
- add new examples
- changes os specific part

alexsu52: Let's discuss it offline.

AlexanderDokuchaev: Changed default value for skip_windows to false

PR #3117, opened 2024-11-27
Fix imports in FP8 example

kshpv:
Workaround for issue https://github.com/huggingface/optimum-intel/issues/993

### DIFF
--- a/examples/llm_compression/openvino/smollm2_360m_fp8/requirements.txt
+++ b/examples/llm_compression/openvino/smollm2_360m_fp8/requirements.txt
@@ -2,4 +2,4 @@ datasets
 openvino==2024.5
 optimum-intel[openvino]
 transformers
-onnx<1.16.2
+onnx==1.17.0
--- a/examples/llm_compression/openvino/smollm2_360m_fp8/main.py
+++ b/examples/llm_compression/openvino/smollm2_360m_fp8/main.py
@@ -10,9 +10,9 @@
 # limitations under the License.
 from functools import partial
 
-import datasets
 import numpy as np
 import openvino as ov
+from datasets import load_dataset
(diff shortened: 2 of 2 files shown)


kshpv:
Windows job 292 - success

examples/llm_compression/openvino/smollm2_360m_fp8/requirements.txt
```diff
@@ -2,4 +2,4 @@ datasets
 openvino==2024.5
 optimum-intel[openvino]
 transformers
-onnx<1.16.2
+onnx==1.17.0
```
MaximProshin, line 5: I think it can be removed now, right?

kshpv: Every example has this requirement. If we remove it here, do not we need to remove it elsewhere? @AlexanderDokuchaev

AlexanderDokuchaev: @MaximProshin 
It's part of our styleguide https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#65-bkc-management
Need to avoid problem this hidden updating version of packages in tests, and know what we are testing.

MaximProshin: Then I need to understand why we have onnx requirement in LLM examples specifically. Can you explain please?

kshpv: @MaximProshin let's keep it as is

MaximProshin (approved):

AlexanderDokuchaev (approved):

Issue #3120, opened 2024-11-28
[Good First Issue][NNCF][MinMax]: move `get_statistic_collector` to the common part


daniil-lyakhov:
### Context

MinMax algorithm backend has the  [`get_statistic_collector`](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/backend.py#L232-L252) method. Previously, backends have different implementations and require different reducers/aggregators maps, thus this method was justified. Now all reducers and aggregators were unified, and we just have 4 identical method copies in each backend: [ONNX](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/onnx_backend.py#L177-L220), [OpenVINO](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/openvino_backend.py#L167-L210), [Torch](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/torch_backend.py#L161-L206), [TorchFX](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/torch_fx_backend.py#L153-L198). (Openvino is actually using `inplace` attribute besides other backend, but this attribute is common as well and could be passed to each backend)

### What needs to be done?

The task is to move `get_statistic_collector` implementation to the [common algorithm file](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/algorithm.py) and to adjust all the min max tests: 
* https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_min_max.py
* https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_ptq_params.py
* https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_quantizer_config.py

### Example Pull Requests

https://github.com/openvinotoolkit/nncf/pull/2856

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers
[...]

PR #3132, opened 2024-12-06
[MinMax] Move get_statistic_collector from backend to common 

rk119:
### Changes

- Added a common `REDUCERS_MAP` in `nncf/experimental/common/tensor_statistics/collectors.py` for backends PT, FX and ONNX.
- Removed redundant reducers maps in backend specific collectors and updated relevant tests that were previously using them.
- Added the `get_statistic_collector` method in the common `nncf/quantization/algorithms/min_max/algorithm.py` file and defined a `reducers_map` property to each backend file.
- Changed the default value of `inplace_statistics` to False in `nncf/quantization/advanced_parameters.py` since the attribute `inplace` in `get_statistic_collector` in `nncf/quantization/algorithms/min_max/algorithm.py` is now passed for all backends, it needs to be False by default and True only for OpenVino.
- Removed the `get_statistic_collector` method definition from PT, FX, ONNX and OpenVino `min_max` backend specific files.
- Removed the abstract `get_statistic_collector` method from `min_max/backend.py` as it is no longer needed.

### Reason for changes

To remove redundant method definitions in all backends.

### Related tickets

#3120


rk119:
Hi @daniil-lyakhov,

Can you review the changes made so far in this PR? The min_max tests do not require any updates so far from what I have observed.

nncf/quantization/advanced_parameters.py
```diff
@@ -234,7 +234,7 @@ class AdvancedQuantizationParameters:
     # General parameters
     overflow_fix: OverflowFix = None
     quantize_outputs: bool = False
-    inplace_statistics: bool = True
+    inplace_statistics: bool = False
```
daniil-lyakhov, line 237: Please keep default `inplace_statistics` value false. All backends except OpenVINO ignore this parameter, it is safe to have such a default value

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -204,6 +209,7 @@ def __init__(
         self._preset = preset
         self._ignored_scope = IgnoredScope() if ignored_scope is None else ignored_scope
         self.quantizer_propagation_rule = quantizer_propagation_rule
+        self.reducer_map = REDUCERS_MAP
```
daniil-lyakhov, line 212: Please introduce a `reducers_map` property in the `MinMaxAlgoBackend` class and do not introduce any attributes in the  `MinMaxQuantization` class

daniil-lyakhov (changes requested):
Thank you for the contribution! Couple comments from my side:

nncf/quantization/algorithms/min_max/algorithm.py
rk119: Yes, my bad. On it.

nncf/quantization/advanced_parameters.py
rk119: I'm sorry if I misunderstood this, since I already changed it to False, do you mean True?

nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -111,6 +108,10 @@ def hw_config(self) -> HWConfig:
     def quant_trait_op_dict(self) -> Dict[int, OperatorMetatype]:
         return DEFAULT_ONNX_QUANT_TRAIT_TO_OP_DICT
 
+    @property
+    def reducer_map(self) -> Dict[StatisticsType, TensorReducerBase]:
+        return REDUCERS_MAP
+
```
daniil-lyakhov, line 114: Please don't forget to introduce the abstract `reducer_map` property in the `MinMaxAlgoBackend`

nncf/quantization/advanced_parameters.py
rk119: Since the param `inplace` is passed now for all backends in `_get_statistic_collector`. Keeping the default value to True causes an error in some tests for backends other than Openvino.

For example FX `test_bias_correction.py`:

image

When I investigated, I realized that the value of `inplace` was True when passed [here](https://github.com/openvinotoolkit/nncf/pull/3132/files#diff-439982cdce21bd7deb841ee26d4601401f3ca51bd4af3fbe13b18a4d7407da2bR495) for all backends causing the error since some of the reducers [here](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/common/tensor_statistics/collectors.py#L410-L541) do not implement `get_inplace_fn`. 

I tried changing [this](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/algorithm.py#L146) to False as default but since the `inplace_statistics` value in [advanced_parameters.py](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/advanced_parameters.py#L237) is passed to MinMaxQuantization in [create_ptq_pipeline](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/post_training/pipeline.py#L118) it updates [this value]((https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/algorithm.py#L146)) to True as well, being the source of the error.

nncf/quantization/algorithms/min_max/onnx_backend.py
rk119: Yep!

nncf/quantization/advanced_parameters.py
daniil-lyakhov: All redusers have a default implementation of `get_inplace_fn`, please check this https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/common/tensor_statistics/collectors.py#L73-L79

rk119: Yea, it returns `None` though, unlike OV.

daniil-lyakhov: Could you please send a trace back of the errors you have? From my perspective, `get_inplace_fn` could not cause this errors because it is being called only in openvino code (only one real use here https://github.com/openvinotoolkit/nncf/blob/develop/nncf/openvino/statistics/aggregator.py#L70)

rk119: Sure.

image
image

daniil-lyakhov: The problem appears here: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/common/tensor_statistics/collectors.py#L85-L88 reducer just returns given input value all the time. Let's mimic previous logic and introduce `supports_inplace_statistics` method to `MinMaxAlgoBackend`

rk119: > The problem appears here: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/experimental/common/tensor_statistics/collectors.py#L85-L88 reducer just returns given input value all the time. 

Oh yes, I remember coming across this. I guess I did not give a clear response earlier on the cause of the error because I traced back and investigated way before and forgot a few things. Should've noted down :) Thank you for clearing up.

> Let's mimic previous logic and introduce `supports_inplace_statistics` method to `MinMaxAlgoBackend`

Alright 👍.

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -474,15 +477,74 @@ def _get_stat_collector(
...
-        return self._backend_entity.get_statistic_collector(
+        return self._get_statistic_collector(
             range_estimator_params,
             collector_params.use_abs_max,
             reduction_axes,
             aggregation_axes,
-            self._inplace_statistics,
-            num_samples=num_samples,
+            self._backend_entity.supports_inplace_statistics,
+            num_samples,
```
daniil-lyakhov, line 486:
```suggestion
            self._inplace_statistics,
            num_samples=num_samples,
```

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -474,15 +477,74 @@ def _get_stat_collector(
...
+
+        :param range_estimator_params: Parameters that specify estimators types.
+        :param use_abs_max: Wheather reduce absolute values of input tensors or not.
+        :param reduction_axes: Axes for reducer.
+        :param aggregation_axes: Axes for aggregator.
+        :param inplace: Whether to calculate statistic inplace or not.
+        :param num_samples: Maximum number of samples to collect.
+        :return: TensorCollector for the statistics calculation.
+        """
+        collector = TensorCollector(MinMaxTensorStatistic)
```
daniil-lyakhov, line 509:
```suggestion
        if not self._backend_entity.supports_inplace_statistics:
            inplace = False
        collector = TensorCollector(MinMaxTensorStatistic)
```

daniil-lyakhov (changes requested):

nncf/quantization/algorithms/min_max/openvino_backend.py
```diff
@@ -35,15 +32,14 @@
...
-from nncf.quantization.advanced_parameters import StatisticsType
 from nncf.quantization.algorithms.min_max.backend import MinMaxAlgoBackend
 from nncf.quantization.fake_quantize import FakeConvertParameters
 from nncf.quantization.fake_quantize import FakeQuantizeParameters
-from nncf.quantization.range_estimator import AggregatorType
+from nncf.quantization.range_estimator import StatisticsType
 
 
 class OVMinMaxAlgoBackend(MinMaxAlgoBackend):
+
```
daniil-lyakhov, line 42: Please remove this empty line from each MinMaxAlgoBackend and we good to go

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -483,6 +486,67 @@ def _get_stat_collector(
...
+        :param use_abs_max: Wheather reduce absolute values of input tensors or not.
+        :param reduction_axes: Axes for reducer.
+        :param aggregation_axes: Axes for aggregator.
+        :param inplace: Whether to calculate statistic inplace or not.
+        :param num_samples: Maximum number of samples to collect.
+        :return: TensorCollector for the statistics calculation.
+        """
+        if not self._backend_entity.supports_inplace_statistics:
+            inplace = False
+        collector = TensorCollector(MinMaxTensorStatistic)
```
daniil-lyakhov, line 511:
```suggestion
            inplace = False

        collector = TensorCollector(MinMaxTensorStatistic)
```

daniil-lyakhov (approved):
LGTM, thanks!

daniil-lyakhov:
@AlexanderDokuchaev, please review and merge if you have no comments

AlexanderDokuchaev (approved):

AlexanderDokuchaev:
@rk119 Well done

PR #3137, opened 2024-12-11
Dynamic OV model builder

KodiaqQ:
### Changes

- Added `ModelBuilder` class.
- Updated FastBC algorithm to utilize new approach.

### Reason for changes

- Algorithm speed up.

### Related tickets

- 122317

### Tests

- Added tests/openvino/native/test_model_builder.py.
- Conversion jobs were run using DLB scope.

Results (develop run - manual/post_training_quantization_performance/91, 92; branch - manual/post_training_quantization_performance/90)

Model | Backend | FBC time (develop, OV) | FBC time (develop, PT) | FBC time (branch, OV) | Diff (develop - branch, OV)
-- | -- | -- | -- | -- | --
hf/bert-base-uncased | OV | 00:00:02 | - | 00:00:01 | **00:00:01**
torchvision/resnet18 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
torchvision/mobilenet_v3_small_BC | OV | 00:00:01 | - | 00:00:01 | 00:00:00
torchvision/vit_b_16 | OV | 00:00:02 | - | 00:00:01 | **00:00:01**
torchvision/swin_v2_s | OV | 00:00:05 | - | 00:00:01 | **00:00:04**
timm/crossvit_9_240 | OV | 00:00:01 | 00:00:00 | **00:00:01** | 00:00:00
timm/darknet53 | OV | 00:00:01 | 00:00:00 | **00:00:01** | 00:00:00
timm/deit3_small_patch16_224 | OV | 00:00:01 | 00:00:00 | 00:00:00 | **00:00:01**
timm/dla34 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/dpn68 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/efficientnet_b0 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/efficientnet_b0_BC | OV | 00:00:04 | - | 00:00:04 | 00:00:00
timm/efficientnet_lite0 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/hrnet_w18 | OV | 00:00:13 | 00:00:03 | **00:00:04** | **00:00:09**
timm/inception_resnet_v2 | OV | 00:00:08 | 00:00:03 | **00:00:04** | **00:00:04**
timm/levit_128 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/mobilenetv2_050 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/mobilenetv2_050_BC | OV | 00:00:03 | - | 00:00:03 | 00:00:00
timm/mobilenetv3_small_050 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/mobilenetv3_small_050_BC | OV | 00:00:01 | - | 00:00:01 | 00:00:00
timm/regnetx_002 | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/resnest14d | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/swin_base_patch4_window7_224 | OV | 00:00:03 | 00:00:00 | **00:00:02** | **00:00:01**
timm/tf_inception_v3 | OV | 00:00:01 | 00:00:01 | 00:00:01 | 00:00:00
timm/vgg11 | OV | 00:00:01 | 00:00:00 | **00:00:01** | 00:00:00
timm/visformer_small | OV | 00:00:00 | 00:00:00 | 00:00:00 | 00:00:00
timm/wide_resnet50_2 | OV | 00:00:01 | 00:00:00 | **00:00:01** | 00:00:00


nncf/openvino/graph/model_builder.py
```diff
@@ -0,0 +1,222 @@
...
+                )
+            elif node_type == "Parameter":
+                # We've created Parameter nodes on the previous step.
+                clone_node = graph_node
+                parameters.append(clone_node)
+            else:
+                # We have to have args as the inputs since all of them are nodes and are required to be as input.
+                args = [clone_nodes.popleft() for _ in graph_node.inputs()]
+
+                clone_node = self._node_factory.create(node_type, args, attrs)
```
KodiaqQ, line 212: Currently, `NodeFactory` will create nodes without names since naming is done using a decorator (see https://github.com/openvinotoolkit/openvino/blob/master/src/bindings/python/src/openvino/utils/decorators.py#L25-L34).
It shouldn't be a problem, since the FastBC algorithm utilizes `Parameter` and `Result` nodes for statistics. But for BiasCorrection algorithm utilizes the names.
The current approach may be changed by adding the corresponding constructor calls into _metatypes,_ for example, for each layer.

nncf/openvino/graph/node_utils.py
```diff
@@ -121,18 +121,22 @@ def get_const_value(const_node: ov.Node) -> np.ndarray:
...
 
     :param node_with_bias: The node that corresponds to the operation with bias.
     :param nncf_graph: NNCFGraph instance.
     :param model: The model that contains this operation.
+    :param node_mapping: Original nodes mapping cache.
     :return: The bias value that is applied to the output tensor of the node's operation.
     """
-    ops_dict = {op.get_friendly_name(): op for op in model.get_ops()}
+    if node_mapping is None:
+        node_mapping = {op.get_friendly_name(): op for op in model.get_ops()}
```
KodiaqQ, line 137: This needs to avoid failures for the algorithms that utilize `get_bias_value`, but without mapping for now.

nncf/quantization/algorithms/fast_bias_correction/openvino_backend.py
```diff
@@ -113,3 +119,11 @@ def get_node_names_for_input_output_statistics(node: NNCFNode, nncf_graph: NNCFG
...
     def get_activation_channel_axis(node: NNCFNode, port_id: int, input_shape: Tuple[int]) -> int:
         return get_activation_channel_axis(node, port_id, input_shape)
+
+    def extract_submodel(self, model_transformer, input_id, output_id):
+
+        return self._model_builder.build(
+            input_ids=[input_id],
+            output_ids=[output_id],
+            node_mapping=self._node_mapping,
+        )
```
KodiaqQ, line 129: To keep the original behaviour of the other backends and reduce code changing, the overloading approach was used here.

KodiaqQ:
@alexsu52, @andrey-churkin, could you please review not PR (need to discuss a number of questions), but the proposal of the new graph-building approach (for FastBiasCorrection and as the possible extension of the BiasCorrection)? Thanks.
All details were stored in the corresponding ticket. See the latest comments.
cc @MaximProshin

nncf/quantization/algorithms/fast_bias_correction/openvino_backend.py
```diff
@@ -33,6 +34,12 @@
 
 
 class OVFastBiasCorrectionAlgoBackend(FastBiasCorrectionAlgoBackend):
+
+    def __init__(self, model):
+        # Node mapping caching to reduce time for calculations
+        self._node_mapping = {op.get_friendly_name(): op for op in model.get_ops()}
```
andrey-churkin, line 40: Just in case. Is that situation possible?
```python
backend.extract_submodel(...)

# some calls that invalidate `node_mapping`

backend.extract_submodel(...)
```
It is some theoretical situation but we need to know about that when we use OVFastBiasCorrectionAlgoBackend.

nncf/openvino/graph/model_builder.py
```diff
@@ -0,0 +1,222 @@
...
+        (w)
+         |
+        (c) (il) (ih) (ol) (oh)
+          \   |   |   /   /
+           (fake quantize) (parameter)
+                  \           /
+                  (convolution)
+                        |
+                    (result)
+        Based on the above graph, the return value would look like this:
```
andrey-churkin, line 79: What are `input_ids` and `output_ids` in that example? Or you just want to demonstrate how nodes are organized inside returned list?

nncf/openvino/graph/model_builder.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+class ModelBuilder:
+    """
+    The purpose of the ModelBuilder is to build a new OpenVINO model from input and output points.
+    This Builder was created to reduce the number of model cloning that is required for ModelTransformer to work.
+    """
+
+    def __init__(self):
+        self._node_factory = NodeFactory()
```
andrey-churkin, line 30: Maybe it makes sense to move that inside `build()`. In that case
```python
ModelBuilder.build(...)

# instead of

model_builder = ModelBuilder()
model_builder.build()

```

andrey-churkin (commented):
LGTM
Probably some validation required here

nncf/openvino/graph/model_builder.py
KodiaqQ: Basically, `input_ids` and `output_ids` are the lists of the points in the special format - `(node_name, port_id)`. This helps to point to the precise part of the model that may be used to define the subgraph.
I'll add this to the description.

nncf/quantization/algorithms/fast_bias_correction/openvino_backend.py
KodiaqQ: Theoretically - yes, it is possible. Following the existing code, we build the `AlgoBackend` once for the statistics point collection and once per apply. So, in practice, there are no situations that may break the mapping.
Also, there is no solution on how to avoid this break.

nncf/openvino/graph/model_builder.py
KodiaqQ: Make sense. I'll update it.

KodiaqQ: Hm. In the case of the `ModelBuilder.build(...)` we well need to have created NodeFactory() somewhere or create it for each build. That looks not convenient in comparison with the current approach. What do you think?

andrey-churkin: I think it's okay to leave it as is.

andrey-churkin (approved):
LGTM
It can be merged after testing.

KodiaqQ:
> Good job! I see that you added performance numbers in the related ticket, but the value of the PR description will be more if you add summary speed up numbers in the PR as well.

Updated.

PR #3138, opened 2024-12-11
Added try for backends

KodiaqQ:
### Changes

- Added `try-catch` blocks for `isinstance` during backend calculation.

### Reason for changes

- Corrupted frameworks may lead to exceptions even if these frameworks were not intended to be used.

### Related tickets

- 158806

### Tests

- Manual

### DIFF
--- a/nncf/common/utils/backend.py
+++ b/nncf/common/utils/backend.py
@@ -11,7 +11,7 @@
 import importlib
 from copy import deepcopy
 from enum import Enum
-from typing import List, TypeVar
+from typing import Callable, List, TypeVar
 
 import nncf
 
@@ -24,6 +24,16 @@ class BackendType(Enum):
     TENSORFLOW = "Tensorflow"
     ONNX = "ONNX"
     OPENVINO = "OpenVINO"
+
+
+def result_verifier(func: Callable) -> Callable:
+    def verify_result(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except AttributeError:
+            return False
+
+    return verify_result
 
 
 def get_available_backends() -> List[BackendType]:
@@ -51,6 +61,7 @@ def get_available_backends() -> List[BackendType]:
     return available_backends
 
 
+@result_verifier
 def is_torch_model(model: TModel) -> bool:
     """
     Returns True if the model is an instance of torch.nn.Module and not a torch.fx.GraphModule, otherwise False.
@@ -61,12 +72,10 @@ def is_torch_model(model: TModel) -> bool:
     import torch
     import torch.fx
 
-    try:
-        return not isinstance(model, torch.fx.GraphModule) and isinstance(model, torch.nn.Module)
-    except AttributeError:
-        return False
+    return not isinstance(model, torch.fx.GraphModule) and isinstance(model, torch.nn.Module)
 
 
+@result_verifier
 def is_torch_fx_model(model: TModel) -> bool:
     """
     Returns True if the model is an instance of torch.fx.GraphModule, otherwise False.
@@ -76,12 +85,10 @@ def is_torch_fx_model(model: TModel) -> bool:
     """
     import torch.fx
 
-    try:
-        return isinstance(model, torch.fx.GraphModule)
-    except AttributeError:
-        return False
+    return isinstance(model, torch.fx.GraphModule)
 
 
+@result_verifier
 def is_tensorflow_model(model: TModel) -> bool:
     """
     Returns True if the model is an instance of tensorflow.Module, otherwise False.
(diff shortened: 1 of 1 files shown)


KodiaqQ:
@openvinotoolkit/nncf-maintainers, can someone review this, please. Thanks.

nncf/common/utils/backend.py
```diff
@@ -85,7 +91,10 @@ def is_tensorflow_model(model: TModel) -> bool:
     """
     import tensorflow  # type: ignore
 
-    return isinstance(model, tensorflow.Module)
+    try:
+        return isinstance(model, tensorflow.Module)
```
alexsu52, line 95: I installed the latest `tensorflow` and `tensorflow.Module` is present there. I think `get_available_backends` is not working properly.

This fix is approved but I would suggest to check what nncf is reported in the env of bug about supported frameworks during `import nncf`. The output on my desktop:
```
INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, tensorflow, onnx, openvino 
```
If NNCF reports tensorflow then you should add fix in this part as well.

KodiaqQ: > If NNCF reports tensorflow then you should add fix in this part as well.

NNCF reports that tensorflow exists even if we place an empty tensorflow.py file into PYTHONPATH. How can we fix this? Could we verify each needed tensorflow part on our side to provide correct report?

alexsu52: What do you think about adding a model entity import check to the backend get function?

KodiaqQ: I don't get it, sorry. Did you mean to verify that model object may be imported from the package here - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/utils/backend.py#L29-L51?

AlexanderDokuchaev: @KodiaqQ  Would using https://docs.python.org/3.9/library/importlib.metadata.html#distributions instead of `import_module` inside `get_available_backends` help here?


```python
# pip install torch
# my_torch.py - script in cwd 

from importlib import metadata, import_module
import_module("torch")
# <module 'torch' from '/home/adokucha/work/nncf/venv/lib/python3.10/site-packages/torch/__init__.py'>
import_module("my_torch")
#<module 'my_torch' from '/home/adokucha/work/nncf/my_torch.py'>
metadata.distribution("torch")
# <importlib.metadata.PathDistribution object at 0x7b3d0bd537f0>
metadata.distribution("my_torch")
#   raise PackageNotFoundError(name)
#importlib.metadata.PackageNotFoundError: No package metadata was found for my_torch
```

alexsu52: @AlexanderDokuchaev will this work if I install a package that builds locally?

alexsu52: > I don't get it, sorry. Did you mean to verify that model object may be imported from the package here - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/utils/backend.py#L29-L51?

Yes, that's what I meant.

KodiaqQ: @alexsu52, @AlexanderDokuchaev I've updated the proposal a little, but this message will still be here.

> INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, tensorflow, onnx, openvino

alexsu52: > @alexsu52, @AlexanderDokuchaev I've updated the proposal a little, but this message will still be here.
> 
> > INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, tensorflow, onnx, openvino

Do you have suggestions on how to fix this?

AlexanderDokuchaev:
```python
frameworks = [
    ("torch", BackendType.TORCH),
    ("torch", BackendType.TORCH_FX),
    ("tensorflow", BackendType.TENSORFLOW),
    ("onnx", BackendType.ONNX),
    ("openvino", BackendType.OPENVINO),
]

try:
    distribution(module_name)
except ...
```

Think that it will be enough.
`distribution` check installed packages by name, not namespaces of modules
`fx` is part of `torch`, i dont know easy way to broke the check (without creating fake package with same name). 
For openvino looks same, but if possible to install openvino package without runtime, we can combine `distribution(module_name.split('.')[0]` and `import_module(module_name)` checks

KodiaqQ: > Do you have suggestions on how to fix this?

No. From the user's perspective, the presence of the `tensorflow` looks correct, even as an empty file.
Another approach is to remove the notion of the supported frameworks.

KodiaqQ: @alexsu52, @AlexanderDokuchaev, was a final solution found? How should I update this PR? Do we need it?

alexsu52: I agree with removing the message about supported frameworks during import if you are not able to implement it correctly.

alexsu52 (approved):

AlexanderDokuchaev (approved):

nncf/common/utils/backend.py
alexsu52: Agreed that this issue will be fixed in the follow-up PR.

PR #3153, opened 2024-12-16
Add to pyguide.md point about checking symbolic links

AlexanderDokuchaev:
### Changes

Add to pyguide.md point about checking symbolic links


docs/styleguide/PyGuide.md
```diff
@@ -37,6 +37,8 @@
   - [6.3 Folder structure](#63-folder-structure)
   - [6.4 Test runtime considerations](#64-test-runtime-considerations)
   - [6.5 BKC management](#65-bkc-management)
+- [7 SDL rules](#s7-sdl-rules)
```
MaximProshin, line 40: I don't think external developers know what is SDL. Either explain the abbreviation or just call it "Security rules"

MaximProshin (approved):

PR #3154, opened 2024-12-16
[mypy] extend mypy check

AlexanderDokuchaev:
### Changes

Added some files to mypy check and fix errors:

- nncf/quantization/passes.py
- nncf/quantization/advanced_parameters.py
- nncf/quantization/range_estimator.py
- nncf/quantization/telemetry_extractors.py


nncf/quantization/advanced_parameters.py
```diff
@@ -551,13 +551,15 @@ def apply_advanced_parameters_to_config(
         initializer["batchnorm_adaptation"] = {"num_bn_adaptation_samples": 0}
         config["initializer"] = initializer
 
-    activations_config = convert_quantization_parameters_to_dict(params.activations_quantization_params)
-    if activations_config:
-        config["activations"] = activations_config
+    if isinstance(params.activations_quantization_params, QuantizationParameters):
```
AlexanderDokuchaev, line 554: @alexsu52 is it valid condition?
quantization_params got annotation `QuantizationParameters | FP8QuantizationParameters | None` but convert_quantization_parameters_to_dict expects only `QuantizationParameters | None`.
It's looks like used only for TF, so it does not support FP8 quantization.
May be need add extra check before and raise exception?

nncf/common/graph/graph.py
```diff
@@ -11,7 +11,21 @@
 import pathlib
 from collections import defaultdict
 from copy import deepcopy
-from typing import Any, Callable, Dict, Generator, KeysView, List, Optional, Tuple, Type, Union, ValuesView, cast
+from typing import (
```
alexsu52, line 14: ?

nncf/common/graph/graph.py
```diff
@@ -68,8 +82,8 @@ def node_name(self) -> NNCFNodeName:
         return cast(NNCFNodeName, self._attributes[NNCFNode.NODE_NAME_ATTR])
 
     @property
-    def metatype(self) -> Type[OperatorMetatype]:
-        return cast(Type[OperatorMetatype], self._attributes[NNCFNode.METATYPE_ATTR])
+    def metatype(self) -> OperatorMetatype:
```
alexsu52, line 85: Could you briefly explain of these changes. I think, that the previous version is correct.

nncf/quantization/advanced_parameters.py
alexsu52: Yes, you are right.

nncf/common/graph/graph.py
AlexanderDokuchaev: reverted

nncf/common/graph/graph.py
AlexanderDokuchaev: combination of length of line 120 and isort `single_line_exclusions = "typing"`, if you are talking about formatting

alexsu52 (approved):
LGTM

nncf/quantization/advanced_parameters.py
```diff
@@ -232,16 +232,16 @@ class AdvancedQuantizationParameters:
     """
 
     # General parameters
-    overflow_fix: OverflowFix = None
+    overflow_fix: Optional[OverflowFix] = None
```
alexsu52, line 235: Please, update type in the docstring.

PR #3171, opened 2024-12-30
[mypy] Enable check for some files in nncf/common

AlexanderDokuchaev:
### Changes

1) Enable mypy check for:
    - nncf/api
    - nncf/common/*.py (exlude some files with many error)
    - nncf/common/initialization

2) Remove `extension_is_loading_info_log` - torch specific function that used only ones.


nncf/api/compression.py
```diff
@@ -127,7 +128,7 @@ class CompressionStage(IntEnum):
     PARTIALLY_COMPRESSED = 1
     FULLY_COMPRESSED = 2
 
-    def __add__(self, other: "CompressionStage") -> "CompressionStage":
+    def __add__(self, other: int) -> CompressionStage:
```
alexsu52, line 131: I understand that `CompressionStage` inherits from `int`, but now the type hint is confusing.

nncf/common/scopes.py
```diff
@@ -73,7 +73,7 @@ def should_consider_scope(
     )
 
 
-def get_not_matched_scopes(scope: Union[List[str], str, IgnoredScope], nodes: List[NNCFNode]) -> List[str]:
+def get_not_matched_scopes(scope: Optional[Union[List[str], str, IgnoredScope]], nodes: List[NNCFNode]) -> List[str]:
```
alexsu52, line 76:
```suggestion
def get_not_matched_scopes(scope: Union[List[str], str, IgnoredScope, None], nodes: List[NNCFNode]) -> List[str]:
```

nncf/tensorflow/sparsity/rb/algorithm.py
```diff
@@ -190,3 +191,6 @@ def compression_rate(self, compression_rate: float) -> None:
 
     def disable_scheduler(self):
         self._scheduler = StubCompressionScheduler()
+
+    def compression_stage(self) -> CompressionStage:
```
alexsu52, line 195: Could you briefly explain why you implemented `compression_stage` in `RBSparsityController` this way?

nncf/api/compression.py
AlexanderDokuchaev: Yes, it's problem exists, but
```
nncf/api/compression.py:131: error: Argument 1 of "__add__" is incompatible with supertype "int"; supertype defines the argument type as "int"  [override]
nncf/api/compression.py:131: note: This violates the Liskov substitution principle
nncf/api/compression.py:131: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
```
Shortly, unsafe to override a method with a more specific argument type and unsafe to override a method with a more general return type.

nncf/common/scopes.py
AlexanderDokuchaev: done

nncf/tensorflow/sparsity/rb/algorithm.py
AlexanderDokuchaev: compression_stage was not abstract method, but it should be
And two controllers does not have implementation for this method, as result `compression_stage` returns `None` for them.
- `NoCompressionAlgorithmController` - added return CompressionStage.UNCOMPRESSED
- `RBSparsityController` - keep return None as on develop (looks like there should be a logic to return stage that depends from sparsity_level)

alexsu52: Is it possible to throw a runtime error and not return None?

AlexanderDokuchaev: It breaks tests https://github.com/openvinotoolkit/nncf/actions/runs/12651525603/job/35252297218
Created jira ticket 160174

alexsu52 (commented):
LGTM

alexsu52 (approved):
LGTM

Issue #3129, opened 2024-12-03
[Good First Issue][NNCF][Tests]: Refactor weights compression tests into a template tests.


daniil-lyakhov:
### Context

Majority of the NNCF PTQ tests are using templates: for example tests for [Fast Bias Correction](https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_fast_bias_correction.py) or [MinMax algorithm](https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_min_max.py).
The structure is following: there is a template file like [this](https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/test_fast_bias_correction.py), and each backend implements abstract methods of the template file like [this](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/test_fast_bias_correction.py).

### What needs to be done?

The task is to:
To extend the [TemplateWeightCompression](https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/test_templates/template_test_weights_compression.py#L65) class and Refactor this files
** https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/ptq/test_weights_compression.py
** https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/fx/test_compress_weights.py
to implement the following common tests:
* test_compress_weights
* test_compress_shared_weights / test_compress_weights_shared_weights
* test_compress_weights_functional_model / test_compress_weights_functional_model
* test_raise_error_with_unsupported_params_for_int8
* test_raise_error_with_unsupported_params_for_int4
* test_raise_error_with_not_int8
* test_raise_error_for_statistics_caching
* test_get_dtype_attribute_of_parameter
* test_model_devices_and_precisions

Please use this two template files as a reference:
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/fx/test_fast_bias_correction.py
https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/ptq/test_fast_bias_correction.py

 

### Example Pull Requests

https://github.com/openvinotoolkit/nncf/pull/2856

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers
[...]

PR #3002, opened 2024-10-01
Fix GPTQ for inputs with batch size != 1 and with seq len == 1

ljaljushkin:
### Changes

GPTQ correctly processes inputs with batch size != 1 and with batch size and sequence length equal 1.
Also changed the errors we are raising in NNCF from built-in Python errors to NNCF-specific ones.

### Reason for changes

Stable-diffusion models, e.g. `runwayml/stable-diffusion-v1-5` has as an input for linear layers with the following shapes: `[2*num_images_in_prompt, text_embedding_size, hidden_dimension]`. 

https://github.com/openvinotoolkit/nncf/blob/develop/examples/llm_compression/openvino/tiny_llama/main.py
uses not filtered data from `wikitext` that leads to the corner case with sequence length == 1.

### Related tickets

150851, 155538

### Tests

- test_compression_with_transposed_activations
- test_compression_with_different_algo_combinations
- test_raise_error_with_unsupported_params_for_e2m1
- test_raise_error_with_unsupported_params_for_empty_dataset

CI

- [ ] weight compression conformance

PR #3179, opened 2025-01-08
[Torch][WeightCompression] Add Scale Estimation data-aware support

kshpv:
### Changes

Added data-aware support for the Torch backend for WeightCompression with Scale Estimation.
Introduced support for MeanVarianceReducer, MaxVarianceReducer, and MeanAbsMaxReducer.
Incorporated `torch.inference_mode()` context for WeightCompression.

### Reason for changes

These changes enable the utilization of data-aware Scale Estimation for the Torch backend, specifically leveraging CUDA devices for improved performance.

### Related tickets

Ticket ID: 158974

### Tests

Added a template for WeightCompression tests for both Torch and OV backends, covering data-aware and Scale Estimation scenarios.
Extended the test scope to include `tinyllama_data_aware` and `tinyllama_scale_estimation_per_channel` for Torch.
Added a new test case `tinyllama_scale_estimation_group_size_64` for both Torch and OV backends.

### Performance Metrics

Note: All CUDA results are obtained locally on a single RTX 3090.

Model | Backend | Metric Name | Metric Value | Num int4 | Num int8 | Compression Time (from Performance Job) | RAM MiB (from Performance Job)
-- | -- | -- | -- | -- | -- | -- | --
tinyllama_data_aware | OV | Similarity | 0.8577 | 94 | 124 | 0:01:28 | 8545
tinyllama_data_aware | TORCH | Similarity | 0.8577 | 94 | 124 | 0:02:15 | 1225
tinyllama_data_aware | TORCH (CUDA) | Similarity | 0.8577 | 94 | 124 | 0:00:28 | -
tinyllama_scale_estimation_per_channel | OV | Similarity | 0.8139 | 188 | 124 | 0:02:57 | 8681
tinyllama_scale_estimation_per_channel | TORCH | Similarity | 0.8139 | 188 | 124 | 0:03:25 | 5472
tinyllama_scale_estimation_per_channel | TORCH (CUDA) | Similarity | 0.8139 | 188 | 124 | 0:00:35 | -
tinyllama_scale_estimation_group_size_64 | OV | Similarity | 0.8566 | 94 | 124 | 0:04:17 | 8681
tinyllama_scale_estimation_group_size_64 | TORCH | Similarity | 0.8566 | 94 | 124 | 0:04:01 | 5575
tinyllama_scale_estimation_group_size_64 | TORCH (CUDA) | Similarity | 0.8566 | 94 | 124 | 0:00:36 | -


kshpv:
weight compression build - 291

tests/common/experimental/test_reducers_and_aggregators.py
```diff
@@ -235,6 +235,54 @@ def test_quantile_reducers(self, reducer_name, ref, reducers):
         for i, ref_ in enumerate(ref):
             assert fns.allclose(val[i], self.get_nncf_tensor(ref_))
 
+    @pytest.mark.parametrize(
+        "axes, np_data, reference",
+        [
+            [None, np.array([[[1, 2, 0], [1, -3, 10]], [[-1, 2, -3], [4, 5, -6]]]), 16.1666],
+            [(0,), np.array([[[1, 2, 0], [1, -3, 10]], [[-1, 2, -3], [4, 5, -6]]]), 14.25],
+            [(0, 1), np.array([[[1, 2, 0], [1, -3, 10]], [[-1, 2, -3], [4, 5, -6]]]), 15.875],
```
daniil-lyakhov, line 243: As each ref input is equal to each other, I believe we can create a global var like `DEFAULT_VARIANCE_REF_INPUT` 😀

kshpv:
conformance job 292

nncf/torch/quantization/quantize_functions.py
```diff
@@ -292,6 +292,8 @@ def pack_uint4(tensor: torch.Tensor) -> torch.Tensor:
     if tensor.dtype != torch.uint8:
         raise ValidationError(f"Invalid tensor dtype {tensor.type}. torch.uint8 type is supported.")
     packed_tensor = tensor.contiguous()
+    # packed_tensor = packed_tensor.split(2, dim=-1)
+    # packed_tensor = packed_tensor
```
daniil-lyakhov, line 296: Please remove debug code

tests/common/experimental/test_reducers_and_aggregators.py
```diff
@@ -232,6 +236,54 @@ def test_quantile_reducers(self, reducer_name, ref, reducers):
...
+            [None, np.array(WEIGHT_COMPRESSION_REDUCERS_DATA), 16.1666],
+            [(0,), np.array(WEIGHT_COMPRESSION_REDUCERS_DATA), 14.25],
+            [(0, 1), np.array(WEIGHT_COMPRESSION_REDUCERS_DATA), 15.875],
+            [(0, 1, 2), np.array(WEIGHT_COMPRESSION_REDUCERS_DATA), 16.1666],
+        ],
+    )
+    def test_mean_variance_reducer(self, axes, np_data, reference):
+        reducer = MeanVarianceReducer(reduction_axes=axes)
+        nncf_data = self.get_nncf_tensor(np_data, dtype=Dtype.FLOAT)
+        result = reducer._reduce_out_of_place([nncf_data])
```
daniil-lyakhov, line 251:
```suggestion
    def test_mean_variance_reducer(self, axes, reference):
        reducer = MeanVarianceReducer(reduction_axes=axes)
        nncf_data = self.get_nncf_tensor(WEIGHTS_COMPRESSION_REDUCERS_DATA, dtype=Dtype.FLOAT)
        result = reducer._reduce_out_of_place([nncf_data])
```

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -0,0 +1,130 @@
...
+NON_ZERO_ROW = [-4, 1, 2]
+ACTIVATION = [[NON_ZERO_ROW, [0, 0, 0], [0, 0, 0]]]
+MAX_VAR = 3.555555  # np.max(np.var(ACTIVATION, 1))
+MEAN_VAR = 1.555555  # np.mean(np.var(ACTIVATION, 1))
+MEAN_MAX = 2.333333  # np.mean(np.max(np.abs(ACTIVATION), 1))
+HESSIAN_TRACE = (16 + 1 + 4) * 2 / 9  # sum(i*i for i in NON_ZERO_ROW) * 2 / ACTIVATION.size
+MAX_BASELINE_SCORE = 1 / 1.1920928955078125e-07
+
+
+class TemplateWeightCompression(ABC):
```
daniil-lyakhov, line 38: Nice! Actually, this should close this GFI https://github.com/openvinotoolkit/nncf/issues/3129

daniil-lyakhov (approved):

nncf/quantization/algorithms/weight_compression/torch_backend.py
```diff
@@ -49,6 +61,37 @@
...
+            tensor, scale, zero_point = Tensor(tensor), Tensor(scale), Tensor(zero_point)
+        else:
+            tensor, scale = inputs
+            tensor, scale = Tensor(tensor), Tensor(scale)
+            zero_point = None
+        quantized = calculate_quantized_weight(tensor, scale=scale, zero_point=zero_point, config=config)
+        dequantized = do_int_dequantization(quantized, scale=scale, zero_point=zero_point)
+        return dequantized.data
+
+    return _forward_fn
```
kshpv, line 92: Should be removed in https://github.com/openvinotoolkit/nncf/pull/2727

kshpv:
Results from conformance job:
Model | Backend | Metric name | Metric value | Metric diff | Num int4 | Num int8 | Compr. time | Total time | RAM MiB
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
tinyllama_data_aware | TORCH | Similarity | 0.8577 | -0.1423 | 94 | 124 | 1:54:26 | 2:18:31 | 6912
tinyllama_data_aware | OV | Similarity | 0.8577 | -0.1423 | 94 | 124 | 0:02:36 | 0:17:57 | 13828
tinyllama_scale_estimation_per_channel | OV | Similarity | 0.8139 | -0.1861 | 188 | 124 | 0:07:56 | 0:24:14 | 13881
tinyllama_scale_estimation_per_channel | TORCH | Similarity | 0.8139 | -0.1861 | 188 | 124 | 2:40:47 | 3:05:47 | 55553

kshpv:
ptq performance job - 36

kshpv:
Model | Num int4 | Num int8 | Compr. time | Stat. collection time | Mixed-Precision search time | Apply Compression time | Total time | RAM MiB | RAM MiB System
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
tinyllama_data_aware | 94 | 124 | 0:01:26 | 0:01:16 | 0:00:03 | 0:00:02 | 0:02:15 | 1225 | 1201
tinyllama_int4_data_free | 114 | 84 | 0:00:09 |   | 0:00:03 | 0:00:02 | 0:00:47 | 5617 | 1401
tinyllama_int8_data_free | 0 | 312 | 0:00:06 |   |   | 0:00:02 | 0:00:39 | 5585 | 1356
tinyllama_scale_estimation_per_channel | 188 | 124 | 0:02:50 | 0:01:20 | 0:00:03 | 0:00:01 | 0:03:25 | 5472 | 1311

kshpv:
Locally, I tested the PR on cuda and it works and demonstrates the same accuracy as the CPU for tinyllama with reduced compression time a lot. Some changes are needed to enable `cuda` for conformance as well as enabling jobs with GPUs.
So, I think it can be done in the following PRs.

andreyanufr (approved):

nncf/experimental/torch/fx/quantization/quantize_model.py
```diff
@@ -135,25 +135,25 @@ def compress_weights_impl(
...
-        gptq,
-        lora_correction,
-        backup_mode,
-        advanced_parameters,
-    )
-    graph = NNCFGraphFactory.create(model)
-    compressed_model = compression_algorithm.apply(model, graph, dataset=dataset)
-    compressed_model = GraphModule(compressed_model, compressed_model.graph)
-    compressed_model = _disallow_eval_train(compressed_model)
+    with torch.no_grad():
```
ljaljushkin, line 138: What about `with torch.inference_mode():`? It might be even faster. 
https://pytorch.org/docs/stable/generated/torch.autograd.grad_mode.inference_mode.html
> Code run under this mode gets better performance by disabling view tracking and version counter bumps.

tests/post_training/data/wc_reference_data.yaml
```diff
@@ -31,6 +35,10 @@ tinyllama_scale_estimation_per_channel_backend_OV:
   metric_value: 0.81389
   num_int4: 188
   num_int8: 124
+tinyllama_scale_estimation_per_channel_backend_TORCH:
```
ljaljushkin, line 38: What about adding test for groupped quantization, since in most cases we use group_size=128 or 64?

ljaljushkin: You can copy this configuration with AWQ and remove `awq: True` for now. After adding AWQ for Torch this test would be identical for OV and Torch.
```python
{
        "reported_name": "tinyllama_data_aware_awq_scale_estimation",
        "model_id": "tinyllama/tinyllama-1.1b-step-50k-105b",
        "pipeline_cls": LMWeightCompression,
        "compression_params": {
            "group_size": 64,
            "ratio": 0.8,
            "mode": CompressWeightsMode.INT4_SYM,
            "awq": True,
            "scale_estimation": True,
            "advanced_parameters": AdvancedCompressionParameters(
                scale_estimation_params=AdvancedScaleEstimationParameters(32, 5, 10, 1.0)
            ),
        },
```

nncf/torch/quantization/layers.py
```diff
@@ -1094,7 +1094,7 @@ def __init__(self, scale: torch.Tensor, zero_point: torch.Tensor, result_dtype:
         """
         super().__init__()
         self.register_buffer("_scale", scale.type(dtype=torch.float16))
-        self.register_buffer("_zero_point", self.pack_weight(zero_point))
+        self.register_buffer("_zero_point", self.pack_weight(zero_point.type(dtype=torch.uint8)))
```
ljaljushkin, line 1097: @alexsu52 please confirm the fix

tests/openvino/native/models.py
```diff
@@ -1185,3 +1185,18 @@ def _create_ov_model(self):
 
         model = ov.Model([sin_result, cos_result], [position_ids])
         return model
+
+
+class MLP(OVReferenceModel):
```
andreyanufr, line 1190: To be honest it is not MLP. There is one layer.

tests/torch/ptq/test_weights_compression.py
```diff
@@ -381,3 +380,46 @@ def check_weights(model: torch.nn.Module, ref_ids: List[int]) -> None:
                 assert torch.numel(op.weight) == 8  # workaround to detect uint4 weights
             else:
                 assert torch.numel(op.weight) == 16
+
+    @staticmethod
+    def get_model_for_test_scale_estimation():
+        return LinearModel(torch.arange(0, 32 * 32, dtype=torch.float32).reshape(32, 32))
+
+    @staticmethod
+    def get_scale_estimation_ref():
```
andreyanufr, line 389: is it possible to move reference to json file like int this test: https://github.com/openvinotoolkit/nncf/blob/0931072ac03117f4dcc07b45dbd3b1d263815a29/tests/openvino/native/quantization/test_weights_compression.py#L255 ?

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -128,3 +127,39 @@ def test_mixed_precision(self, mode, all_layers, ratio, ref_ids):
...
+        Returns the reference output of calculate_quantization_params of ScaleEstimation.
+        """
+
+    def test_scale_estimation(self, mocker):
+        calc_q_params_spy = mocker.spy(ScaleEstimation, "calculate_quantization_params")
+        model = self.get_model_for_test_scale_estimation()
+
+        # prepare dataset with one input tensor
+        input = np.arange(0, 32 * 32, dtype=np.float32).reshape(1, 32, 32)
+        input[0, 15] *= 100  # make one channel relatively higher.
```
ljaljushkin, line 151: I would expect this to be reflected in the references, but I don't see anything outstanding. Maybe the value should be higher.

tests/torch/ptq/test_weights_compression.py
```diff
@@ -327,3 +354,72 @@ def test_pack_int4():
...
+    def cast_to(x: torch.Tensor, dtype: TensorDataType) -> torch.Tensor:
+        return cast_to(x, dtype)
+
+    @staticmethod
+    def check_weights(model: torch.nn.Module, ref_ids: List[int]) -> None:
+        for i, op in enumerate(model.layers):
+            if i in ref_ids:
+                assert torch.numel(op.weight) == 8  # workaround to detect uint4 weights
+            else:
+                assert torch.numel(op.weight) == 16
```
ljaljushkin, line 382: This function looks too hardcoded for a particular model. 
What about converting torch model to OV here and explicitly check for ov.Type.i4?

ljaljushkin (changes requested):

tests/post_training/data/wc_reference_data.yaml
kshpv: ptq - 296
performance - 40

nncf/experimental/torch/fx/quantization/quantize_model.py
kshpv: done

tests/cross_fw/test_templates/template_test_weights_compression.py
kshpv: I think that this test intends to check the difference between torch and OV backends. It does not aim to check the algorithm's correctness. However, I agree that your proposal is good. I can add a new test which will check the error after quantization and demonstrate that the outlier channel has the lowest error against others

tests/torch/ptq/test_weights_compression.py
kshpv: Reduce the scales number. Should look much better

nncf/torch/quantization/layers.py
kshpv: Motivation is that `zero_point` is used in ScaleEstimation quantization formula as float type, not unit type. See these:
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/scale_estimation.py#L226
https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/openvino_backend.py#L377
This type conversion is hidden for OV because it applies type transformation from float to dtype in the transformation directly - https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/openvino_backend.py#L252
So, I basically did the same for Torch and transform zero_point to expected dtype in model transformation step.

tests/torch/ptq/test_weights_compression.py
kshpv: updated, please check

tests/post_training/data/wc_reference_data.yaml
kshpv: The results of the jobs:
Model | Backend | Metric name | Metric value | Metric diff | Num int4 | Num int8 | Compr. time (from Performance job) | RAM MiB (from Performance job)
-- | -- | -- | -- | -- | -- | -- | -- | --
tinyllama_scale_estimation_group_size_64 | OV | Similarity | 0.8566 | -0.1434 | 94 | 124 | 0:04:17 | 8681
tinyllama_scale_estimation_group_size_64 | TORCH | Similarity | 0.8566 | -0.1434 | 94 | 124 | 0:04:01 | 5575

tests/torch/ptq/test_weights_compression.py
```diff
@@ -327,3 +354,48 @@ def test_pack_int4():
...
+    def to_tensor(t) -> torch.Tensor:
+        return torch.tensor(t)
+
+    @staticmethod
+    def cast_to(x: torch.Tensor, dtype: TensorDataType) -> torch.Tensor:
+        return cast_to(x, dtype)
+
+    @staticmethod
+    def check_weights(model: torch.nn.Module, ref_ids: List[int]) -> None:
+        low_precision_nodes = {f"{i}_weight" for i in ref_ids}
```
ljaljushkin, line 378: Please move model names from the testing code to model. 
```python
all_names = model.get_weight_names_in_exec_order()
low_precision_nodes = list(map(lambda i: all_names[i], ref_ids))
for op_name, op in model.nncf.external_op.items():
    for name in low_precision_nodes:
          if name in op_name:
              assert isinstance(op, INT4SymmetricWeightsDecompressor)
```

ljaljushkin (changes requested):

nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -464,18 +464,27 @@ def _reduce_out_of_place(self, x: List[Tensor]) -> List[Tensor]:
 
 
 class MeanVarianceReducer(TensorReducerBase):
-    def _reduce_out_of_place(self, x: List[TensorType]) -> List[TensorType]:
-        raise NotImplementedError()
+    def _reduce_out_of_place(self, x: List[Tensor]) -> List[Tensor]:
+        x = x[0]
```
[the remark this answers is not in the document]
daniil-lyakhov: Reducers was designed to receive several inputs and output several outputs as well. For now, there is no such reducers, but the possible use case - quantization error (fp32 input, int8 input) -> diff

We can create a class for that, it will make hierarchy tree more complicated, we than should introduce method like `_reduce_out_of_place_one_input` method, and I don't think this will make code more readable

tests/torch/ptq/test_weights_compression.py
kshpv: done

nncf/experimental/common/tensor_statistics/collectors.py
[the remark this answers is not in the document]
ljaljushkin: Since all reducers use one input, it should be relatively easy, isn't? 
1) define in base class `_reduce_out_of_place_general` for many inputs
2) subclass an intermediate class with `_reduce_out_of_place_one_input`
3) rename inheritance from base class to intermediate class
4) rename `_reduce_out_of_place` to `_reduce_out_of_place_one_input`.

ljaljushkin:
@kshpv Have you run the performance job after merging with develop? Could you share it?

tests/torch/ptq/test_weights_compression.py
ljaljushkin: I can't find where it's done: f"{i}_weight" is still there

tests/cross_fw/test_templates/template_test_weights_compression.py
ljaljushkin: Are you going to add this new test in the follow-up PR?

kshpv:
> @kshpv Have you run the performance job after merging with develop? Could you share it?

Sure, performance build 41

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -0,0 +1,212 @@
...
+            all_layers=True,
+            dataset=dataset,
+        )
+
+        original_weight = self.get_orig_weight(model)
+        decompressed_weight_before_se = quantize_dequantize_weight(
+            original_weight, config=WeightCompressionConfig(CompressWeightsMode.INT4_ASYM, -1), reduction_axes=1
+        )
+        decompressed_weight_after_se = self.get_decompressed_weight(compressed_model, input)
+        error_before_se = get_realtive_error(original_weight, decompressed_weight_before_se)
```
ljaljushkin, line 209:
```suggestion
        error_before_se = get_relative_error(original_weight, decompressed_weight_before_se)
```

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -0,0 +1,212 @@
...
+        model = self.get_model_for_test_scale_estimation()
+
+        # prepare dataset with one input tensor
+        input = np.arange(0, 4 * 8, dtype=np.float32).reshape(1, 4, 8)
+        input[:, :, OUTLIER_CHANNEL] *= 1000  # make one channel relatively higher, should have lowest error.
+        input = self.to_tensor(input)
+        dataset = Dataset([input])
+
+        compressed_model = compress_weights(
+            deepcopy(model),
```
ljaljushkin, line 195: you can avoid copy by taking original weights before compression

ljaljushkin (changes requested):

ljaljushkin:
If compare 40 and 41 performance build, one can notice that AWQ is sometimes slowly. 
40 build doesn't include #2727
image
41 build includes #2727
image

Although time for mixed-precision, applying compession and total time are better with 41 build, I wonder why there is some slowness. Is it measurement error or expected numbers? @nikita-savelyevv

nikita-savelyevv:
> If compare 40 and 41 performance build, one can notice that AWQ is sometimes slowly. 40 build doesn't include #2727 image 41 build includes #2727 image
> 
> Although time for mixed-precision, applying compession and total time are better with 41 build, I wonder why there is some slowness. Is it measurement error or expected numbers? @nikita-savelyevv

It is possible that AWQ part could become a bit slower for a small model like tiny-llama because the added compiled functions are the most effective for compressing large tensors. When operating with small tensors, compilation overhead can overshadow the speedup from compiled computation. 

That's why for example int8 tiny-llama compression is a bit slower after #2727 
image

My assumption is that the same effect makes AWQ a bit slower in this case. It should not happen for larger models though. I will check this. Thanks for the observation.

ljaljushkin:
> t the same effect makes AWQ a bit slower in this case. It should not happen for larger models though. I will check this. Thanks for the observa

Thanks for confirmation!

ljaljushkin (approved):

nncf/quantization/quantize_model.py
```diff
@@ -546,7 +539,8 @@ def compress_weights(
         else:
             example_input = next(iter(dataset.get_inference_data()))
             model = wrap_model(model, example_input=example_input, trace_parameters=True)
-        dataset = None
+        if mode in (CompressWeightsMode.INT8, CompressWeightsMode.INT8_ASYM, CompressWeightsMode.INT8_SYM):
+            dataset = None  # workaround as INT8 mode still does not support dataset
```
alexsu52, line 543:
```suggestion
            if mode in (CompressWeightsMode.INT8, CompressWeightsMode.INT8_ASYM, CompressWeightsMode.INT8_SYM):
                dataset = None  # data-aware methods don't support INT8 modes
```

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -109,6 +110,8 @@ def mean_statistic_collector(
 
     @staticmethod
     def get_activation_port_id(node: NNCFNode, nncf_graph: NNCFGraph) -> int:
+        if node.layer_attributes.input_attributes["transpose"]:
+            raise nncf.UnsupportedModelError("Transposed input is not supported")
```
alexsu52, line 114: Why did you add runtime error instead of introduce a fix? Is there any plan?

kshpv: This was introduce there - https://github.com/openvinotoolkit/nncf/pull/3002. I did not find any ticket on that. Probably, @ljaljushkin can have it

nncf/torch/quantization/quantize_model.py
```diff
@@ -101,21 +101,21 @@ def compress_weights_impl(
...
-        subset_size,
-        scale_estimation,
-        gptq,
-        lora_correction,
-        backup_mode,
-        advanced_parameters,
-    )
-    graph = NNCFGraphFactory.create(model)
-    return compression_algorithm.apply(model, graph, dataset=dataset)
+    with torch.inference_mode():
```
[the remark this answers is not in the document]
kshpv: Memory leak should be fixed in https://github.com/openvinotoolkit/nncf/pull/3199.  I did not see significant difference using `torch.inference_mode()`. To not introduce requirements on model training in algorithm I removed `torch.inference_mode()`.

kshpv:
The perfomance run after rebase including fix for memory leak - 43 - do not see any significant difference with the previous runs.

Accuracy job is passed - https://github.com/openvinotoolkit/nncf/actions/runs/13052429517

nncf/quantization/algorithms/weight_compression/openvino_backend.py
ljaljushkin: https://github.com/openvinotoolkit/nncf/issues/3230

nncf/quantization/quantize_model.py
```diff
@@ -583,7 +577,7 @@ def compress_weights(
 
         if dataset:
             raise nncf.ParameterNotSupportedError(
-                "TorchFX only supports data-free weights compression," "Set the 'dataset' option to None"
+                "TorchFX only supports data-free weights compression,Set the 'dataset' option to None"
```
alexsu52, line 580:
```suggestion
                "TorchFX only supports data-free weights compression. Set the 'dataset' option to None"
```

alexsu52 (approved):
LGTM, please address the last my comments.

nncf/quantization/quantize_model.py
kshpv: done

PR #3181, opened 2025-01-08
[mypy] nncf/common/logging

AlexanderDokuchaev:
### Changes

Enable mypy for `nncf/common/logging`
Add method `update` to `track` to use `pbar.update(advance=1)` instead of `pbar.progress.update(pbar.task, advance=1)`


nncf/common/logging/track_progress.py
```diff
@@ -116,13 +117,13 @@ def get_weighted_completed(task: Task, completed: float) -> float:
...
-        return sum(task.fields["weights"][: int(completed)])
+        return float(sum(task.fields["weights"][: int(completed)]))
 
 
-class track:
+class track(Generic[ProgressType]):
     def __init__(
         self,
-        sequence: Optional[Union[Sequence[ProgressType], Iterable[ProgressType]]] = None,
+        sequence: Union[Sequence[ProgressType], Iterable[ProgressType], None] = None,
```
AlexanderDokuchaev, line 126: There is some problem in case of sequence is None, in this case ProgressType is unknown, as result mypy shows var-annotated error.

```python
with track(iterable, description="Processing...") as pbar:
 # error: Need type annotation for "pbar"  [var-annotated]
```

In python3.13 TypeVar have default argument https://peps.python.org/pep-0696/
But now the simplest solution use `with track[None](...)` or add `# type: ignore[var-annotated]`

nikita-savelyevv: Is `# type: ignore[var-annotated]` would be needed every time `with track(...)` is used? Is there a possibility to ignore this annotation at definition?

nncf/common/logging/track_progress.py
```diff
@@ -221,16 +237,22 @@ def __iter__(self) -> Iterable[ProgressType]:
...
         self.progress.__exit__(*args)
-        self.progress.remove_task(self.task)
-        self.task = None
+        if self.task is not None:
+            self.progress.remove_task(self.task)
+            self.task = None
+
+    def update(self, advance: float, **kwargs: Any) -> None:
+        if self.task is None:
+            raise RuntimeError("update available only inside context manager.")
```
nikita-savelyevv, line 257:
```suggestion
            raise RuntimeError("update is available only inside context manager.")
```

tests/common/utils/test_progress_tracking.py
```diff
@@ -83,4 +83,4 @@ def test_track_context_manager(n, is_weighted):
     with track(total=n, description="Progress...", weights=weights if is_weighted else None) as pbar:
         for i in range(n):
             assert pbar.progress._tasks[pbar.task].completed == (sum(weights[:i]) if is_weighted else i)
-            pbar.progress.update(pbar.task, advance=1)
+            pbar.update(advance=1)
```
nikita-savelyevv, line 86: Why in this case `[None]` is not required for `with track(...)`?

AlexanderDokuchaev: it's test, [none] or type ignore need to pass mypy

nikita-savelyevv: ok

nncf/common/logging/track_progress.py
AlexanderDokuchaev: updated

nncf/common/logging/track_progress.py
AlexanderDokuchaev: > Would # type: ignore[var-annotated] be needed every time with track(...) is used?
Yes, comment or [none] 

>  Is there a possibility to ignore this annotation at definition?
It's correctly will works in python3.13 but 3.12 will supported until 2028 😞 
But for python3.9 i dont know how to resolve it

nikita-savelyevv (approved):

PR #3190, opened 2025-01-13
[GHA] run tests with numpy<2

AlexanderDokuchaev:
### Changes

- Run precommit tests with numpy-1.24.0 on python3.9
- Add `override_requirements` argument to call-precommit.yml
- Use one line install `pip install . -r <path>/requarements.txt` to correctly resolve versions of install packages
- Bump minimal version of numpy to 1.24.0

### Reason for changes

- numpy <1.19.3 - fail on installation https://github.com/openvinotoolkit/nncf/actions/runs/12757498767/job/35557883523
- numpy <1.24 - falls by using numpy.typing

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/12767723338/job/35586673309?pr=3190
sdl/job/nightly_trigger/407/


.github/workflows/weekly.yml
```diff
@@ -12,10 +12,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.9", "3.11", "3.12"]
+        include:
+          - python_version: "3.9"
+            override_requirements: "numpy==1.24.0"
+          - python_version: "3.11"
```
MaximProshin, line 18: Why do you have 3.11 version here? I would avoid it.

.github/scripts/override_constraints.py
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2025 Intel Corporation
```
MaximProshin, line 1: Please run sdl checks due to a new file.

.github/workflows/weekly.yml
MaximProshin: discussed offline

MaximProshin (approved):

PR #3194, opened 2025-01-16
[GHA] windows precommit

AlexanderDokuchaev:
### Changes

Run test from precommit test scope in nightly trigger

- pt2: use fixed parameters for .T, _saved_dim0/1 depends from os, but actual value always -1, -2
- onnx: build model only inside test function, not in collect parameters, to more effective use memory 
- common: use `1.2` s for test timer, to be sure that will be catch 00:00:01

NOTE: pt test falls in case of parallel run on build quantization extension.

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/12835288745/job/35794407928?pr=3194

### DIFF
--- a/tests/common/utils/test_timer.py
+++ b/tests/common/utils/test_timer.py
@@ -17,7 +17,7 @@ from nncf.common.utils.timer import timer
 
 def test_timer(nncf_caplog):
     with timer() as t:
-        time.sleep(1)
+        time.sleep(1.2)
 
     t()
 
--- a/tests/onnx/quantization/test_classification_models_graph.py
+++ b/tests/onnx/quantization/test_classification_models_graph.py
@@ -22,72 +22,56 @@ from tests.onnx.quantization.common import min_max_quantize_model
 from tests.onnx.quantization.common import mock_collect_statistics
 from tests.onnx.weightless_model import load_model_topology_with_zeros_weights
 
+
+def model_builder(model_name):
+    if model_name == "resnet18":
+        return models.resnet18(weights=models.ResNet18_Weights.IMAGENET1K_V1)
+    if model_name == "resnet50_cpu_spr":
+        return models.resnet50(weights=models.ResNet50_Weights.IMAGENET1K_V1)
+    if model_name == "mobilenet_v2":
+        return models.mobilenet_v2(weights=models.MobileNet_V2_Weights.IMAGENET1K_V1)
+    if model_name == "mobilenet_v3_small":
+        return models.mobilenet_v3_small(weights=models.MobileNet_V3_Small_Weights.IMAGENET1K_V1)
+    if model_name == "inception_v3":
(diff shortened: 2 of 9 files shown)


tests/common/utils/test_timer.py
```diff
@@ -17,7 +17,7 @@
 
 def test_timer(nncf_caplog):
     with timer() as t:
-        time.sleep(1)
```
kshpv, line 20: curious - why? :)

tests/onnx/quantization/test_classification_models_graph.py
```diff
@@ -22,72 +22,56 @@
 from tests.onnx.quantization.common import mock_collect_statistics
 from tests.onnx.weightless_model import load_model_topology_with_zeros_weights
 
+
+def model_builder(model_name):
+    if model_name == "resnet18":
```
kshpv, line 27: Could we use `getattr` here? Also please reuse one func in all tests

tests/common/utils/test_timer.py
AlexanderDokuchaev: Look at description of PR

tests/onnx/quantization/test_classification_models_graph.py
AlexanderDokuchaev: How?

kshpv:
```
def model_builder(model_name: str) -> Callable:
    model_weights_map = {
        "resnet18": "ResNet18_Weights.IMAGENET1K_V1",
        "mobilenet_v2": "MobileNet_V2_Weights.IMAGENET1K_V1",
        "mobilenet_v3_small": "MobileNet_V3_Small_Weights.IMAGENET1K_V1",
        "inception_v3": "Inception_V3_Weights.IMAGENET1K_V1",
        "googlenet": "GoogLeNet_Weights.IMAGENET1K_V1",
        "vgg16": "VGG16_Weights.IMAGENET1K_V1",
        "shufflenet_v2_x1_0": "ShuffleNet_V2_X1_0_Weights.IMAGENET1K_V1",
        "squeezenet1_0": "SqueezeNet1_0_Weights.IMAGENET1K_V1",
        "densenet121": "DenseNet121_Weights.IMAGENET1K_V1",
        "mnasnet0_5": "MNASNet0_5_Weights.IMAGENET1K_V1",
    }

    if model_name not in model_weights_map:
        raise ValueError(f"Unknown model name {model_name}")

    model_func = getattr(models, model_name)
    weights = getattr(models, model_weights_map[model_name])
    return model_func(weights=weights)
```

AlexanderDokuchaev: Looks the same.
Still required hardcode definition of special parameter for each model, still works for limited list of model

kshpv (approved):

MaximProshin (approved):

PR #3199, opened 2025-01-19
[PT] detach tensor in get_const_data

AlexanderDokuchaev:
### Changes

Return detached tensor from `get_const_data` function

### Reason for changes

Memory leak


AlexanderDokuchaev:
Reproducer

```python
import torch
results = []
for _  in range(100):
    a = torch.rand((1,3,1000, 1000),requires_grad=True)
    b = torch.rand((1,3,1000, 1000),requires_grad=True)
    r = torch.sum(a + b)
    results.append(r)
```

image

```python
import torch
results = []
for _  in range(100):
    a = torch.rand((1,3,1000, 1000),requires_grad=True)
    b = torch.rand((1,3,1000, 1000),requires_grad=True)
    r = torch.sum(a.detach() + b.detach())
    results.append(r)
```
image

alexsu52 (commented):
Do you have idea how to check that you covered all places in NNCF code?

AlexanderDokuchaev:
No, but problem actual only for case when original tensor is replaced to new in the model which was calculated by using original.

kshpv (approved):

alexsu52 (approved):

PR #3209, opened 2025-01-23
OV nf4 support

KodiaqQ:
### Changes

- Added `nf4` precision for OV `GraphConverter`.

### Reason for changes

- `nf4` precision support.

### Related tickets

- 153357

### Tests

- Added


KodiaqQ:
@openvinotoolkit/nncf-maintainers, can someone review this? Thanks.

ljaljushkin:
Could you elaborate which scenario is supported now? Can it be covered by some test?

KodiaqQ:
> Could you elaborate which scenario is supported now?

A model with partially quantized weights in `nf4` format, as an example.

> Can it be covered by some test?

I'm not sure that non-quantized models with these types exist. Only possible with the custom synthetic case. Do we really need it?

ljaljushkin (approved):

PR #3223, opened 2025-01-29
check year in copyright

AlexanderDokuchaev:
### Changes

Added check to 2025 year in copyrights and 2026 to avoid falls until copyrights will not updated in new year.
Remove commits with updating copyrights from .git-blame-ignore-revs, it was useless my pr.


daniil-lyakhov (approved):
Nice!

PR #3225, opened 2025-01-29
[FX] Dynamic Shapes Support

anzr299:
### Changes

Modify NNCF Graph Builder for FX backend to correctly get and insert the dynamic shapes into NNCFGraph

### Reason for changes

To support quantization of Torch FX models exported with dynamic shapes

### Tests

test is added to `tests/torch/fx/test_models.py` in test_quantized_models(). Currently only the synthetic transformer is tested because torch.export.dynamic_shapes.Dim.DYNAMIC is not supported in pytorch but is supported in upcoming releases. https://pytorch.org/tutorials/intermediate/torch_export_tutorial.html#constraints-dynamic-shapes

`test_dynamic_edge()` is also added in `tests/torch/fx/test_models.py` to check that the tensor shape in NNCF Graph edge has values only of type int or str and not SymInt.


tests/torch/fx/test_models.py
```diff
@@ -171,18 +176,29 @@ def test_model(test_case: ModelCase):
 )
```
daniil-lyakhov, line 178: Looks like some test cases are duplicated: non synthetic transformer models still are being checked twice with the static shapes. Please refactor the test to prevent redundant cases and please try to avoid pytest.skip call as well

daniil-lyakhov (changes requested):

tests/torch/fx/test_models.py
anzr299: Would you suggest I rather remove the graph tests and only retain `test_dynamic_edge()` test since most of the transformer models in this test are not correctly traced with dynamic shapes

daniil-lyakhov:
NNCFGraphEdge and NNCF algorithms expect the edge shape to be a `List[int]` https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/graph/graph.py. The use of an str as a shape dimension could lead to unexpected errors in the algorithms: what would happened if a channel dim will be "s1" for a per-channel quantization?

Could you please show an example with a quantization of a dynamic model? Perhaps we can cover the case with less drastic change (using -1 as a shape dim, for example)

anzr299:
> NNCFGraphEdge and NNCF algorithms expect the edge shape to be a `List[int]` https://github.com/openvinotoolkit/nncf/blob/develop/nncf/common/graph/graph.py. The use of an str as a shape dimension could lead to unexpected errors in the algorithms: what would happened if a channel dim will be "s1" for a per-channel quantization?
> 
> Could you please show an example with a quantization of a dynamic model? Perhaps we can cover the case with less drastic change (using -1 as a shape dim, for example)
I can add an exception for the case where the in_channel has a non-static .
using integer such as -1 could cause more complex dynamic shape definition (for example `s0//2*3+s1`) in int to resolve to another value. Would you recommend simply catching the known cases? or to brainstorm on other ways to represent the dynamic shapes

anzr299:
1. I have tried with -1 as dynamic shape. It seems to work fine and pass all tests. I also tested with a random constant shape  for all the tensor shape values except for channel shape. For example, [1,3,224,224] -> [1000, 3, 1000, 1000] where the random constant is 1000. This also did not cause any issues for torch FX tests. I tested this for all the shapes in every model. So it looks like there is no issue with the tensor shapes.
2. I am looking into ways to obtain the maximum value. I did find [this](https://pytorch.org/docs/stable/torch.compiler_dynamic_shapes.html#the-guard-model) which talks about how torch utilizes the `hint` value of symbolic data types. I think this could also be a good replacement for dynamic shape in nncf graph.

tests/torch/fx/test_models.py
daniil-lyakhov: Yes, that would solve the duplicates. You can create a helper function in which you can keep the duplicated body of the test

nncf/experimental/torch/fx/nncf_graph_builder.py
```diff
@@ -197,6 +197,9 @@ def get_edge_params(
                 tensor = source_node.meta["val"]
             if isinstance(tensor, torch.Tensor):
                 tensor_shape = tuple(tensor.shape)
+                tensor_shape = tuple(i.node.hint if isinstance(i, torch.SymInt) else i for i in tensor_shape)
```
daniil-lyakhov, line 200:
```suggestion
                tensor_shape = tuple(i.node.hint if isinstance(i, torch.SymInt) else i for i in tensor.shape)
```

tests/torch/fx/test_models.py
```diff
@@ -208,6 +235,28 @@ def transform_fn(data_item):
...
+        )
+    ]
+    fx_model = get_torch_fx_model(model, ex_inputs, dynamic_shapes=dynamic_shapes)
+    nncf_graph = GraphConverter.create_nncf_graph(fx_model)
+
+    for edge in nncf_graph.get_all_edges():
+        edge_shape = edge.tensor_shape
+        assert isinstance(edge_shape, tuple)
+        for dim in edge_shape:
+            assert isinstance(dim, (int, str))
```
daniil-lyakhov, line 256:
```suggestion
            assert isinstance(dim, int)
```

tests/torch/fx/test_models.py
```diff
@@ -157,32 +166,46 @@ def test_model(test_case: ModelCase):
             (250, 250),
             (149, 149),
         ],
+        [True, False, True, True],
     ),
     (
         ModelCase(partial(ShortTransformer, 5, 10), "synthetic_transformer", [5]),
         {"model_type": nncf.ModelType.TRANSFORMER},
         [(4, 4), (2, 2)],
+        [True],
```
daniil-lyakhov, line 175: Can we build dynamic shapes directly here, without bool -> dim conversion?
```suggestion
        [Dim.Auto],
```

daniil-lyakhov (commented):
Can we reuse existing reference dot files? AFAIK they look identical

tests/torch/fx/test_models.py
anzr299: Yes, that is possible. I used bool since it was simpler type and to avoid repetition. I will change it as suggested

nncf/experimental/torch/fx/nncf_graph_builder.py
anzr299: Done

tests/torch/fx/test_models.py
anzr299: Done

tests/torch/fx/test_models.py
anzr299: Done

tests/torch/fx/test_statistics_aggregator.py
```diff
@@ -59,7 +60,15 @@ def get_fast_bias_correction_algo_backend_cls(self) -> Type[FXFastBiasCorrection
...
         conv_w = self.dataset_samples_to_conv_w(np.array(sample))
-        return get_torch_fx_model(IdentityConv(conv_w), torch.ones(INPUT_SHAPE))
+        dynamic_shapes = [
+            (
+                Dim.AUTO,
+                Dim.STATIC,
+                Dim.AUTO,
+                Dim.AUTO,
+            )
+        ]
```
daniil-lyakhov, line 70: What's the reason behind this change?

tests/torch/fx/test_models.py
```diff
@@ -208,6 +235,28 @@ def transform_fn(data_item):
...
+    ]
+    fx_model = get_torch_fx_model(model, ex_inputs, dynamic_shapes=dynamic_shapes)
+    nncf_graph = GraphConverter.create_nncf_graph(fx_model)
+
+    for edge in nncf_graph.get_all_edges():
+        edge_shape = edge.tensor_shape
+        assert isinstance(edge_shape, tuple)
+        for dim in edge_shape:
+            assert isinstance(dim, int)
+            assert not isinstance(dim, torch.SymInt)
```
daniil-lyakhov, line 257:
```suggestion
        assert all(isinstance(dim, int) for dim in edge_shape)
```

daniil-lyakhov (approved):

nncf/experimental/torch/fx/nncf_graph_builder.py
```diff
@@ -196,7 +196,9 @@ def get_edge_params(
             else:
                 tensor = source_node.meta["val"]
             if isinstance(tensor, torch.Tensor):
-                tensor_shape = tuple(tensor.shape)
+                tensor_shape = tuple(-1 if isinstance(i, torch.SymInt) else i for i in tensor.shape)
+            if isinstance(tensor, torch.SymInt):
```
AlexanderDokuchaev, line 200:
```suggestion
            elif isinstance(tensor, torch.SymInt):
```

AlexanderDokuchaev (changes requested):

AlexanderDokuchaev (approved):

PR #3227, opened 2025-01-30
Bump PyTorch to 2.6.0

AlexanderDokuchaev:


anzr299:
Can you please change all instances of the reference value `output_1` to `output` in `tests/torch/fx/test_model_transformer.py` in order to pass the torch fx tests in Pytorch 2.6.0.
Also, you can update the reference graphs and reference metatypes for FX backend

Issue #3231, opened 2025-01-30
[TorchFX] Support quantization with custom torch.ao quantizers via `nncf.quantize_pt2e` API


daniil-lyakhov:
### 🚀 Feature request

The task is to support quantization with custom torch.ao quantizers via `nncf.quantize_pt2e` API. 

Custom quantizers list:
* https://github.com/pytorch/executorch/tree/main/backends/apple/coreml
* https://github.com/pytorch/executorch/tree/main/backends/arm
* https://github.com/pytorch/executorch/tree/main/backends/qualcomm
* https://github.com/pytorch/executorch/tree/main/backends/xnnpack

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #3232, opened 2025-01-30
[PTQ] Remove HWConfig hardcode from calculate_quantizer_parameters

daniil-lyakhov:
### Changes

* `symmetric_range` and  `asymmetric_range` functions were refactored to avoid dependency on quantization group
* input configs of `test_calculate_quantizer_parameters` were updated to align with real cases computed by the solver with `cpu.json` HW config (all symmetric weights configs have `signedness_to_force=True` in actual code)

### Reason for changes

`symmetric_range` and  `asymmetric_range` functions are using prior knowledge about HW config (if symmetric and weights -> input_low = -input_high) but could just use the qconfig attributes (signed == True and narrow_range == True -> input_low = -input_high) . New quantizers from the #3231 does not obey hardcoded OpenVINO HW config rules (X86 quantizer require narrow_range false for symmetric weights, for example), thus support of the new quantizers requires more general `calculate_quantizer_parameters` function

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/3231

### Tests

* test_calculate_quantizer_parameters is updated

### DIFF
--- a/nncf/quantization/fake_quantize.py
+++ b/nncf/quantization/fake_quantize.py
@@ -148,7 +148,7 @@ def symmetric_range(
     max_values: Tensor,
     levels: int,
     quantizer_config: QuantizerConfig,
-    q_group: QuantizerGroup,
+    narrow_range: bool,
 ) -> Tuple[Tensor, Tensor]:
     """
     Calculates the numbers of the low and high quant for the symmetric quantization scheme.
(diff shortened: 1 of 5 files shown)


nncf/quantization/fake_quantize.py
```diff
@@ -328,7 +323,7 @@ def _calculate_scaled_parameters(
     num_bits = quantizer_config.num_bits
     level_low, level_high = calculate_symmetric_level_ranges(num_bits - 1, signed=True, narrow_range=False)
     levels = get_num_levels(level_low, level_high)
-    input_low, input_high = symmetric_range(min_values, max_values, levels, quantizer_config, quant_group)
+    input_low, input_high = symmetric_range(min_values, max_values, levels, quantizer_config, narrow_range)
```
daniil-lyakhov, line 326: @kshpv, please confirm `half_range` calculations are correct. input_low = -input_hight before despite the value of the narrow_range before (due to quant_group == WEIGHTS). And now, for a synthetic case, impossible with `cpu.json` config (symmetric weights + narrow_range==False), calculations were altered

l-bat (approved):
Thanks!

PR #3234, opened 2025-01-31
[OV][Tests] Merge test_fq_params_calculation.py and test_calculation_quantizer_params.py

daniil-lyakhov:
### Changes

test_fq_params_calculation.py and test_calculation_quantizer_params.py are merged to one file

### Reason for changes

To decrease amount of test files with similar content


AlexanderDokuchaev:
Looks like it actual for PT and ONNX too.
And please check may be possible add some tests to template.

PR #3244, opened 2025-02-03
[ruff] INP rule

AlexanderDokuchaev:

### DIFF
--- a/tests/cross_fw/install/install_checks_torch.py
+++ b/tests/cross_fw/install/install_checks_torch.py
@@ -14,6 +14,7 @@ import sys
 import torch
 
 import nncf
+from tests.cross_fw.install.common import load_nncf_modules
 
 if len(sys.argv) != 3:
     msg = "Must be run with an execution type as argument (either 'cpu' or 'gpu') and package type"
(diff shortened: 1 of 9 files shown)


tests/cross_fw/install/install_checks_torch.py
```diff
@@ -54,3 +55,17 @@
...
     raise nncf.ValidationError(msg)
+
+EXCLUDED_MODULES_PATTERNS = (
+    "nncf\\.openvino.*",
+    "nncf\\.tensorflow.*",
+    "nncf\\.onnx.*",
+    "nncf\\.experimental\\.tensorflow.*",
+    "nncf\\.experimental\\.openvino.*",
+    "nncf\\.experimental\\.onnx.*",
+    ".*?openvino_[^\\.]*",
```
alexsu52, line 66: Logically, we should not exclude files from the target backend, even if they contain names of other backends. I would suggest to modify such patterns the following way:
```
"^(?!nncf(?:\\.experimental)*\\.torch.*?\\.).*?openvino_[^\.]*\\.py"
```

alexsu52 (changes requested):
The major comment about renaming `openvino_quantizer.py`:

Taking into account, the tutorial https://github.com/daniil-lyakhov/tutorials/blob/acf1647329e72de081a4035f42b74fadc311cd44/prototype_source/openvino_quantizer.rst I would suggest to make imports more clear. I mean OpenVINOQuantizer is not NNCF quantizer, but it's the implementation of torch.ao quantizer. Thus OpenVINOQuantizer should be placed in Torch FX backend instead of common part.

The import is now looked like that:
```
from nncf.experimental.quantization.quantizers.openvino_quantizer import OpenVINOQuantizer
```
I would suggest to implement the following:
```
from nncf.experimental.torch.fx.quantization.qunatizer.openvino_quantizer import OpenVINOQuantizer
```

Also I would also suggest to move adapters to `nncf.experimental.torch.fx.quantization.qunatizer` and move `nncf.experimental.quantization.quantizers.quantizer.py` -> `nncf.experimental.quantization.quantizer.py`

@daniil-lyakhov, @AlexanderDokuchaev, let's discuss this comment.

alexsu52 (approved):
LGTM, please provide GA for the last changes.

daniil-lyakhov (approved):

AlexanderDokuchaev:
> LGTM, please provide GA for the last changes. 

https://github.com/openvinotoolkit/nncf/actions/runs/13137591065

PR #3233, opened 2025-01-31
[Release] Set environment variable on LNL CPUs before compression model compilation

nikita-savelyevv:
### Changes

Set `export DNNL_MAX_CPU_ISA=AVX2_VNNI` during weights compression on LNL CPUs.

### Reason for changes

Fix compression on LNL CPUs.

### Related tickets

161336

### Tests
- https://github.com/openvinotoolkit/nncf/actions/runs/13075940656 ✅ 
- https://github.com/openvinotoolkit/nncf/actions/runs/13075938396 ✅ 
- NNCF/job/nightly/job/sdl/job/nightly_trigger/426/ ✅

PR #3246, opened 2025-02-04
Set environment variable on LNL CPUs before compression model compilation

nikita-savelyevv:
Copy of #3233 and #3247 to develop branch


nikita-savelyevv:
> @nikita-savelyevv , please run install checks for this PR

Install checks will fail for the same reason they failed on the release branch. Please see ticket 161808.

PR #3260, opened 2025-02-06
[Fixed]: Onnx function transform error in onnx ptq

thinvy:
### Changes
Added the function member transfer when constructing the _insert_outputs onnx model.

### Reason for changes
Onnx function is a widely used new feature. In 'torch.onnx.dynamo_export' and 'torch.onnx.export' ('export_modules_as_functions' param), some user modules will be converted to onnx functions. However, the current nncf onnx quantization does not do special processing for it, resulting in the onnx exported by this function converting the user's onnx function into an unimplemented custom operator, causing errors in the subsequent ptq process.

image

### DIFF
--- a/nncf/onnx/graph/model_transformer.py
+++ b/nncf/onnx/graph/model_transformer.py
@@ -179,6 +179,8 @@ class ONNXModelTransformer(ModelTransformer):
             model_version=model.model_version,
             doc_string=model.doc_string,
         )
+        if hasattr(model, "functions") and getattr(model, "functions"):
+            new_model.functions.extend(model.functions)
         if model.metadata_props:
             values = {p.key: p.value for p in model.metadata_props}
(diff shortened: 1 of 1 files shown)


MaximProshin:
@kshpv , please take a look

kshpv:
@thinvy thanks for raising the issue. Could you provide more details on the scenario where you encountered it? Did this fix completely resolve the problem? I am wondering whether the model that you mentioned having function inside would be handled correctly.

MaximProshin:
@thinvy , do you have any input requested above? Otherwise I would close this PR.

alexsu52 (commented):
Thanks for the contribution!

@thinvy, please, add a test to [test_nncf_graph_builder.py](https://github.com/alexsu52/nncf/blob/develop/tests/onnx/test_nncf_graph_builder.py). After PR will be merged.

PR #3263, opened 2025-02-07
[OpenVINOQuantizer] return annotated model

daniil-lyakhov:
### Changes

Mode return is added to `annotate` method of the OpenVINOQuantizer

### Reason for changes

To align with the signature


alexsu52 (approved):
LGTM

PR #3265, opened 2025-02-10
[TESTS] Make a common entry point for conformance tests

kshpv:
### Changes
Refactors confromance test suite to enhance maintainability and reduce potential bugs.

1) Unified test entry point
2) Centralized Fail/XFail logic

### Reason for changes

1) Reduce code complexity
2) Reduce potential number of bugs

### Tests

WC run - https://github.com/openvinotoolkit/nncf/actions/runs/13239414168 - passed
ptq run -job/manual/job/post_training_quantization/608/ - in progress
WC CI - job/manual/job/post_training_weight_compression/317 - passed


tests/post_training/test_quantize_conformance.py
```diff
@@ -293,50 +390,49 @@ def test_ptq_quantization(
     capsys: pytest.CaptureFixture,
     extra_columns: bool,
     memory_monitor: bool,
+    use_avx2: Optional[bool] = None,
```
AlexanderDokuchaev, line 393: use_avx2 used only as fixture to set env variable, no need to pass it to pipeline

tests/post_training/data/ptq_reference_data.yaml
```diff
@@ -18,7 +18,7 @@ hf/hf-internal-testing/tiny-random-GPTNeoXForCausalLM_statefull_backend_OPTIMUM:
   metric_value: null
 hf/hf-internal-testing/tiny-random-GPTNeoXForCausalLM_stateless_backend_OPTIMUM:
   metric_value: null
-  xfail_reason: "Issue-161969"
+  exception_xfail_reason: "Issue-161969"
```
AlexanderDokuchaev, line 21: For exception need to know exception type and msg to set xfail only for expected error
```
exception_xfail_reason:
  msg: "Issie-123"
  class: RuntimeError
  errmsg: "some error"
```

tests/post_training/test_quantize_conformance.py
```diff
@@ -174,6 +178,9 @@ def fixture_ptq_report_data(output_dir, run_benchmark_app, pytestconfig):
         if not run_benchmark_app:
             df = df.drop(columns=["FPS"])
 
+        df = df.drop(columns=["Num sparse activations"])
```
AlexanderDokuchaev, line 181: looks like report_dat can be use one function to generate csv,
some thing like `save_results(daf, droped_columns)`

AlexanderDokuchaev (changes requested):

tests/post_training/pipelines/base.py
```diff
@@ -92,9 +92,10 @@ def fill(self, stdout: str) -> None:
 
 @dataclass
 class NumCompressNodes:
-    num_fq_nodes: Optional[int] = None
-    num_int8: Optional[int] = None
-    num_int4: Optional[int] = None
+    num_fq_nodes: int = 0
+    num_int8: int = 0
+    num_int4: int = 0
+    num_sparse_activations: int = 0
```
nikita-savelyevv, line 98: Does it have to be defined in the `base.py`? I believe a test pipeline should define it's own fields here.

tests/post_training/pipelines/base.py
```diff
@@ -196,6 +197,7 @@ def get_result_dict(self):
             "Num FQ": self.num_compress_nodes.num_fq_nodes,
             "Num int4": self.num_compress_nodes.num_int4,
             "Num int8": self.num_compress_nodes.num_int8,
+            "Num sparse activations": self.num_compress_nodes.num_sparse_activations,
```
nikita-savelyevv, line 200: Same comment as for `NumCompressNodes.num_sparse_activations`

tests/post_training/pipelines/base.py
kshpv: I think so, it makes code common. If any test does not need any field, that's fine, only values in reference data matter

tests/post_training/test_quantize_conformance.py
```diff
@@ -164,50 +168,41 @@ def fixture_wc_reference_data():
...
-        output_file = output_dir / "results.csv"
-
-        if pytestconfig.getoption("forked") and output_file.exists():
-            # When run test with --forked to run test in separate process
-            # Used in post_training_performance jobs
-            df.to_csv(output_file, index=False, mode="a", header=False)
-        else:
-            df.to_csv(output_file, index=False)
+    columns_to_drop = ["Num sparse activations", "Num int4"]
+    yield from create_fixture_report_data(output_dir, run_benchmark_app, pytestconfig, columns_to_drop)
```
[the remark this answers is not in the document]
kshpv: You are right. But I do not see any problem with omitting some columns (it is already done on the develop in the same manner), it is much easier to implement in that way than to have several different classes with different logic.
What do you think?

nikita-savelyevv (commented):
Activation sparsity conformance test can't be run in the CI because of some infrastructure problems. Could you please run it on the develop and your branches locally and share the results?

nikita-savelyevv (approved):

AlexanderDokuchaev (approved):

PR #3292, opened 2025-02-18
OV data-free mixed precision assignment running time improvement

nikita-savelyevv:
### Changes

After implementation of openvino optimized functions #2727 int4 compression part of the compression pipeline is now faster compared to mixed precision search part.

In this PR mixed precision assignment runtime is improved using the same approach. Namely, the function `weight_lowering.get_integer_quantization_error()` is optimized.

Below are results before/after for several BF16 models:
| Model        | Compression         | Compression Time Develop (sec.) | Compression Time Branch (sec.) | Peak Memory Develop (MiB) | Peak Memory Branch (MiB) 
|--------------|---------------------|----------------------|------------------|---------------------------|--------------------------|
| tiny-llama   | int4_asym, ratio=0.8 | 22.94                           | 17.03 (-26%)                   | 1406                      | 1085 (-23%)            |
| phi3-mini    | int4_asym, ratio=0.8 | 56.40                           | 32.06 (-43%)                   | 3669                    | 2754 (-25%)      |   
| llama-3.1-8b | int4_asym, ratio=0.8 | 97.27                           | 56.80 (-42%)                   | 6660              | 5505 (-17%)      |   

### Related tickets

161921

### Tests

Extended
- tests/openvino/optimized_functions/test_compression_functions.py
- tests/openvino/optimized_functions/test_ov_model_parameters.py

https://github.com/openvinotoolkit/nncf/actions/runs/13412231315


nncf/quantization/algorithms/weight_compression/mixed_precision.py
```diff
@@ -161,7 +161,11 @@ def _calc_weight_sensitivity(
         graph: NNCFGraph,
     ) -> float:
         weight = self._backend_entity.get_weight(
-            weight_param.node_with_weight, weight_param.weight_port_id, model, graph
+            weight_param.node_with_weight,
+            weight_param.weight_port_id,
+            model,
+            graph,
+            as_ov_tensor=True,
```
alexsu52, line 168: I think introducing a backend-specific argument to backend-agnostic code is bad practice. Is it possible to avoid this? Maybe it makes sense to always return `ov.Tensor` and cast the tensor to numpy if needed.

A more general question is implementing functions for the OpenVINO backend in NNCF `Tensor` via numpy. What do you think?

nncf/openvino/optimized_functions/models.py
```diff
@@ -245,6 +245,7 @@ def get_compress_decompress_weight_model(
     zero_point_shape: Optional[Tuple] = None,
     reduction_axes: Optional[ReductionAxes] = None,
     return_compressed_weight: Optional[bool] = False,
+    return_nodes: Optional[bool] = False,
```
alexsu52, line 248: Do you really need to introduce this argument? Where do you use it?

nncf/openvino/optimized_functions/functions.py
```diff
@@ -168,6 +169,39 @@ def quantize_dequantize_weight(
         return decompressed_weight
 
 
+def get_integer_quantization_error(
+    weight: Tensor,
+    reduction_axes: ReductionAxes,
+    config: WeightCompressionConfig,
+) -> float:
+    """
+    Calculates a quantity characterizing the difference between floating point weights and fake quantized
```
alexsu52, line 178: Could you please add a clearer description because the code with building the openvino model is difficult to understand how "a quantity" is calculated. If you add the formula it will greatly improve the clarity.

nncf/openvino/optimized_functions/models.py
alexsu52: Thanks for the answer! Sorry for being persistent, but I asked because I couldn't find where `get_compress_decompress_weight_model` was used with `return_nodes=True`.

nncf/quantization/algorithms/weight_compression/mixed_precision.py
nikita-savelyevv: > I think introducing a backend-specific argument to backend-agnostic code is bad practice. Is it possible to avoid this? 

I've now introduced `OVTensorWeightCompressionAlgoBackend`. For it, `get_weight()` returns an `ov.Tensor`.

> Maybe it makes sense to always return ov.Tensor and cast the tensor to numpy if needed.

This is possible, but I don't like the idea that 

```python
if weight.backend == TensorBackend.ov:
    weight = weight.as_numpy_tensor()
```

will need to be called every time after a `get_weight()` call besides some exceptions where `as_ov_tensor=True` is passed right now. I can imagine that this will be a surprise to other developers. 

> A more general question is implementing functions for the OpenVINO backend in NNCF Tensor via numpy. What do you think?

This is unfortunately not straightforward to do. The problem is that ov.Tensor can contain data in i4, u4, bf16 precisions which will actually be represented by i8, u8 and fp16 numpy arrays respectively. So we can't for example simply add numpy arrays (`a.data + b.data`) if we wanted to implement addition for nncf.Tensor OV backend. Supporting operations for these data types will require some additional logic, for example as it is done for `as_numpy_tensor`: https://github.com/openvinotoolkit/nncf/blob/develop/nncf/tensor/functions/openvino_numeric.py#L79

nncf/openvino/optimized_functions/models.py
nikita-savelyevv: Thanks. I see what you mean. Removed.

nncf/openvino/optimized_functions/functions.py
nikita-savelyevv: Added

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -335,6 +335,15 @@ def get_integer_quantization_error(
...
     """
+    # Optimized implementation
+    if _can_run_optimized(weight.backend):
+        from nncf.openvino.optimized_functions import (
+            get_integer_quantization_error as get_integer_quantization_error_ov,
+        )
+
+        return get_integer_quantization_error_ov(weight, reduction_axes, config)
+
+    weight = weight.as_numpy_tensor()
```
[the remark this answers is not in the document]
nikita-savelyevv: Thanks, fixed

nncf/openvino/graph/node_utils.py
```diff
@@ -110,19 +110,21 @@ def cnt_if_op(model: ov.Model, cnt: int) -> int:
     return cnt_if_op(model, 0)
 
 
-def get_const_value(const_node: ov.Node, cast_bf16_to_fp32: bool = True) -> np.ndarray:
+def get_const_value(const_node: ov.Node, as_ov_tensor: bool = False) -> Union[np.ndarray, ov.Tensor]:
```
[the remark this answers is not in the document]
nikita-savelyevv: Changed

alexsu52 (approved):
LGTM

AlexanderDokuchaev (approved):

PR #3294, opened 2025-02-18
[ruff] isort

AlexanderDokuchaev:
### Changes

Use ruff isort instead of direct isort


alexsu52 (commented):
Please double check that you updated all places in the repository such as Python Guide and etc. https://github.com/openvinotoolkit/nncf/blob/9bd8ddf61712bbdc9317015926a7ebb10ae97af1/docs/styleguide/PyGuide.md?plain=1#L64

AlexanderDokuchaev:
@alexsu52  updated

alexsu52 (approved):

PR #3309, opened 2025-02-25
Refactoring of minmax algorithm parameter setup and checking

alexsu52:
### Changes

- Fixed `_override_device` function https://github.com/openvinotoolkit/nncf/blob/5c75e22c2888ebde2a87534c8cb204497899b0b7/nncf/quantization/algorithms/min_max/algorithm.py#L253 
- Apply overflow fix only for 8-bit quantization by default https://github.com/openvinotoolkit/nncf/blob/5c75e22c2888ebde2a87534c8cb204497899b0b7/nncf/quantization/advanced_parameters.py#L36
### Reason for changes

The overflow fix should only be applied for 8 bit quantization

### Related tickets

None

### Tests

test_target_device
test_npu_target_device
test_overflow_fix


andrey-churkin (approved):
LGTM

PR #3310, opened 2025-02-25
[PT2] check overridden forward in _replicate_for_data_parallel

AlexanderDokuchaev:
### Changes

Move check on overridden forward to _replicate_for_data_parallel, because it's actual only for case with using DataParallel.
For other cases overridden forward can correctly works, depends from user implementation


AlexanderDokuchaev:
Reproducer of fail with overridden forward
```python
example_input = torch.ones([10,1,3,3], dtype=torch.float32).cuda()
model = nn.Conv2d(1, 1, 1).cuda()
model.forward = model.forward
model(example_input)
parallel_model = torch.nn.DataParallel(model)
parallel_model(example_input)  # RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0!
```

alexsu52 (approved):
LGTM

PR #3315, opened 2025-02-26
Aanuf/data free awq

andreyanufr:
### Changes

Data free AWQ: smooth down_proj input channels and merge extra scale to up_proj output channels.

New method is enabled when AWQ is enabled but the dataset is not available. 

### Reason for changes
CVS-160032

image

tests
https://github.com/openvinotoolkit/nncf/actions/runs/14864351054


nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -132,19 +132,161 @@ def apply(
...
+
+            if is_data_free:
+                scale = self._data_free_step(weight)
+            else:
+                scale = self._data_aware_step(wp, weight, statistics[k])
+
+            w_scale = fns.unsqueeze(scale, 1 - wp.reduction_axes[0])
+            a_scale = fns.unsqueeze(1.0 / scale, wp.reduction_axes[0])
+
+            scaled_weight = weight * w_scale
```
ljaljushkin, line 168: Please rebase from develop, I've made some casting for float16 models

nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -132,19 +132,161 @@ def apply(
...
 
+        awq_data = self._get_awq_data(graph, all_weight_params, nodes_to_compress)
+        if len(awq_data) == 0:
+            return model
+
+        transformation_layout = TransformationLayout()
+        model_transformer = ModelTransformerFactory.create(model, inplace=True)
+
+        is_data_free = True #statistics is None
+        description = "Applying data-free AWQ" if is_data_free else "Applying AWQ"
```
ljaljushkin, line 144:
```suggestion
        description = "Applying data-free AWQ" if is_data_free else "Applying data-aware AWQ"
```
maybe more clear

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -506,7 +506,7 @@ def apply(
         nodes_to_compress = self.get_nodes_to_compress(graph)
 
         statistics = None
-        if self._data_aware_mixed_precision or self._data_aware_compression:
+        if (self._data_aware_mixed_precision or self._data_aware_compression) and dataset:
```
ljaljushkin, line 509: I'd redefine 
```python
self._data_aware_compression = (self._awq and dataset) or (...)
```

nncf/quantization/algorithms/weight_compression/awq.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/awq.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: dataset not defined in __init__

nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -132,156 +132,40 @@ def apply(
...
-            target_node_names = []
-            for weight_op_friendly_name, _ in self._backend_entity.get_weight_names_and_port_ids(nncf_node, graph):
-                target_node_names.append(weight_op_friendly_name)
-
-            # skip node if it is in IgnoredScope or should not be compressed
-            if target_node_names[-1] not in name_mapping:
-                continue
-
-            weight_params = all_weight_params[name_mapping[target_node_names[-1]]]
+        is_data_free = statistics is None
```
nikita-savelyevv, line 143: The description and type hint for `statistics` parameter should probably also be updated.

nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: My concern is that with this approach we won't be able to run data-free AWQ together with another data-aware method, e.g. Scale Estimation.

I suppose we have the following options:
1. Introduce another argument like `data_free_awq` and expose it to `nncf.compress_weights`. 
2. Remove data-aware AWQ altogether. As I remember according to your experiments data-free AWQ shows better accuracy results anyway.

Have you considered this?

alexsu52: @nikita-savelyevv I suggest to use subset_size=-1 from AdvancedAWQParameters to run data-free AWQ together with another data-aware methods. What do you think?

nikita-savelyevv: > @nikita-savelyevv I suggest to use subset_size=-1 from AdvancedAWQParameters to run data-free AWQ together with another data-aware methods. What do you think?

In my opinion this is not very intuitive, but I suppose it could work, especially if we set `subset_size=-1` as the new default value instead of the current `subset_size=32`. This way data-free AWQ will run by default, while data-aware AWQ will require some additional actions from a user.

One thing is that if we consider integration with optimum-intel, it won't be possible to expose `AdvancedAWQParameters` to optimum cli. This way the current `--awq` argument will induce execution of data-free AWQ. I think it's ok to leave data-aware AWQ as an advanced option to be applied only through python API considering the fact that the accuracy is higher for data-free AWQ.

alexsu52 (commented):
@andreyanufr, please fill the PR description

nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: > 2\. Remove data-aware AWQ altogether. As I remember according to your experiments data-free AWQ shows better accuracy results anyway.

It shows sometimes accuracy degradation in comparison with data-aware, buy this reason I am not going to enable data-free awq by default.

nikita-savelyevv: Another proposal: introduce `AdvancedAWQParameters.is_data_aware` which is True by default. A user will need to manually set it to False in order to run data-aware SE + data-free AWQ.

nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -132,156 +135,41 @@ def apply(
...
-            target_node_names = []
-            for weight_op_friendly_name, _ in self._backend_entity.get_weight_names_and_port_ids(nncf_node, graph):
-                target_node_names.append(weight_op_friendly_name)
-
-            # skip node if it is in IgnoredScope or should not be compressed
-            if target_node_names[-1] not in name_mapping:
-                continue
-
-            weight_params = all_weight_params[name_mapping[target_node_names[-1]]]
+        is_data_free = statistics is None or not self._is_data_aware
```
nikita-savelyevv, line 146: We can remove `is_data_free` variable and use `self._is_data_aware` instead.

If `self._is_data_aware` is True, but `statistics` is None, we should raise an exception.

tests/openvino/native/quantization/test_weights_compression.py
```diff
@@ -1509,6 +1509,30 @@ def run_compression():
...
+        awq=True,
+        advanced_parameters=CompressionParams(
+            awq_params=AWQParams(
+                is_data_aware=is_data_aware,
+            )
+        ),
+    )
+
+    n_awq = TestOVTemplateWeightCompression.get_num_multiply_from_awq(compressed_model)
+    assert n_awq == n_layers - 1
```
nikita-savelyevv, line 1533: Can you please add an additional check to actually validate that data-aware or data-free AWQ was applied?

One way this can be done is by using `unittest.mock.patch` for the `AWQ._data_aware_step` method. As an example: https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L1502

nncf/quantization/quantize_model.py
```diff
@@ -601,13 +601,12 @@ def compress_weights(
...
 
-        if any((awq, scale_estimation, gptq, lora_correction)) and (
-            dataset is None or mode == CompressWeightsMode.E2M1
-        ):
-            msg = (
-                "Scale estimation, AWQ, GPTQ or Lora Correction algorithm is defined, "
-                "but dataset is None or mode is E2M1."
-            )
+        if any((scale_estimation, gptq, lora_correction)) and dataset is None:
+            msg = "Scale estimation, GPTQ or Lora Correction algorithm is defined, but dataset is None."
```
nikita-savelyevv, line 605:
```suggestion
        if any((awq and advanced_parameters.awq_params.is_data_aware, scale_estimation, gptq, lora_correction)) and dataset is None:
            msg = "Data-aware AWQ, scale estimation, GPTQ or Lora Correction algorithm is defined, but dataset is None."
```

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -542,7 +543,7 @@ def apply(
...
```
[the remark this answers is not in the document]
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/awq.py
andreyanufr: New method is enabled when AWQ is enabled but the dataset is not available.

tests/openvino/native/quantization/test_weights_compression.py
andreyanufr: Done

nncf/quantization/quantize_model.py
andreyanufr: New method is enabled when AWQ is enabled but the dataset is not available.

nncf/quantization/advanced_parameters.py
```diff
@@ -276,13 +276,16 @@ class AdvancedAWQParameters:
...
+    :param prefer_data_aware: Determines whether to use activations to calculate scales if activations are presented.
+    :type prefer_data_aware: bool
     """
 
     subset_size: int = 32
     percent_to_apply: float = 0.002
     alpha_min: float = 0.0
     alpha_max: float = 1.0
     steps: int = 100
+    prefer_data_aware: bool = True
```
alexsu52, line 288:
```suggestion
    :param use_data_aware_scaling: Whether to use activation data for scale calculation when available.
    :type use_data_aware_scaling: bool
    """

    subset_size: int = 32
    percent_to_apply: float = 0.002
    alpha_min: float = 0.0
    alpha_max: float = 1.0
    steps: int = 100
    use_data_aware_scaling: bool = True
```
What do you think about this?

nikita-savelyevv: In my opinion, the word "prefer" conveys exactly what we want: when there's a dataset, preference will be given to the data-aware scenario.
@alexsu52  what do you think about `prefer_data_aware_scaling`?

nncf/quantization/algorithms/weight_compression/algorithm.py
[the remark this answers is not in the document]
nikita-savelyevv: I believe this is not the case if we change the check to this one:
```python
data_aware_awq = dataset and self._awq and self._advanced_parameters.awq_params.prefer_data_aware
if self._data_aware_mixed_precision or self._data_aware_compression or data_aware_awq:
```

But I'm ok with the current implementation too.

alexsu52 (approved):
LGTM

nikita-savelyevv (approved):

tests/torch2/fx/test_compress_weights.py
```diff
@@ -369,6 +369,13 @@ def check_weights(model: torch.fx.GraphModule, ref_ids: list[int]) -> None:
     def get_not_supported_algorithms() -> list[str]:
         return ["lora_correction", "gptq"]
 
+    @staticmethod
+    def supports_data_free() -> bool:
+        """
+        Returns True if data-free compression is supported, False otherwise.
+        """
+        return False
```
ljaljushkin, line 377: Why not supported for torch fx?

ljaljushkin (changes requested):

ljaljushkin (approved):

PR #3316, opened 2025-02-26
[gha] trivy

AlexanderDokuchaev:
### Changes

Add trivy check
Custom template .github/scripts/trivy_html.tpl (like in jenkins), original is not show path to checked file

### Tests

https://github.com/AlexanderDokuchaev/nncf/actions/runs/13554528994


MaximProshin (approved):
LGTM but I would still prefer to have all SDL checks in one trigger or so, even manual, because it's more convenient to run them once if I want to make sure we're ok from SDL standpoint

PR #3317, opened 2025-02-27
Add kwargs to quantize calls

nikita-savelyevv:
### Changes

Add `**kwargs` to `nncf.quantize()` and other call signatures.

### Reason for changes

Helpful in integration with optimum-intel.


alexsu52 (changes requested):
Could you provide an example?

PR #3320, opened 2025-02-28
[MinMax] Embedding nodes as input nodes for inference graph

daniil-lyakhov:
Reopen of the #2862

### Changes

* Embedding nodes are used as input nodes for the inference graph (with that embedding nodes are being included to the inference_nncf_graph)
*both nncf_graph and  inference_nncf_graph is used to identify weighted nodes 
* PT/FX MinMax `get_weight_nodes` method is updated to work with the graph / inference graph pair
* Constant folding is removed from the `OpenVINOQuantizer` and FX `nncf.quantize` implementation 
* New symbolic types are introduced for the TorchFX NNCFGraphBuilder

### Reason for changes

* To prevent quantization of the constant branches without constant folding (https://github.com/pytorch/executorch/pull/8573#discussion_r1965848318)

### Related tickets

163025

### Tests

* tests/cross_fw/test_templates/test_quantizer_config.py is updated with shape_of /constant embedding model and  conv model with constant branches
* TorchFX reference graphs for VIT and Swin were updated: constant branches are present in the quantized graph but they don't have quantizers inside
* conformance test post_training_quantization/625/ - passed
* References are updated for models with a constant branches


nncf/torch/utils.py
```diff
@@ -467,3 +470,46 @@ def get_model_dtype(model: torch.nn.Module) -> torch.dtype:
...
+    :param node: A NNCFNode instance.
+    :param inference_nncf_graph: An inference NNCFGraph instance.
+    :return: True if given node is a matmul with a constant input, False otherwise.
+    """
+    if node.metatype == om.PTLinearMetatype:
+        return True
+
+    # Inference graph does not containt constans, so
+    # any missed input edge means it is a constant branch.
+    return node.metatype in [om.PTMatMulMetatype, om.PTAddmmMetatype] and len(
```
AlexanderDokuchaev, line 513: Please use variables to make it more readable

nncf/torch/utils.py
```diff
@@ -467,3 +470,46 @@ def get_model_dtype(model: torch.nn.Module) -> torch.dtype:
...
+        inference_nncf_graph.get_input_edges(node)
+    ) < len(node.metatype.weight_port_ids)
```
AlexanderDokuchaev, line 515: is it work for possible missed inputs, that determinate in get_nodes_with_missed_input_edges?

nncf/torch/utils.py
```diff
@@ -467,3 +470,46 @@ def get_model_dtype(model: torch.nn.Module) -> torch.dtype:
         # The model had no parameters at all, assume FP32
         dtype = torch.float32
     return dtype
+
+
+def get_weight_nodes_in_inference_grpah(
+    inference_nncf_graph: NNCFGraph, mat_mul_metatypes: List[om.PTOperatorMetatype]
```
AlexanderDokuchaev, line 476: Looks like mat_mul_metatypes expected always same value, mat_mul_metatypes should not be used as argument instead use reusable constant variable

AlexanderDokuchaev (changes requested):

nncf/torch/utils.py
daniil-lyakhov: Done

nncf/torch/utils.py
daniil-lyakhov: Done, please check

AlexanderDokuchaev (approved):

PR #3321, opened 2025-02-28
[FX][Conformance] Enable Conformance Test for FX Backend

anzr299:
### Changes

1. Save torch exported compressed model instead of OV coverted model. Save OV model which is obtained using torch.compile and dumping cache. The OV model obtained by dumping cache is used for counting FQ nodes.
2. Save torch exported model using torch.export.save instead of converting to OV and saving OV model when saving fp32 model for debugging.
3. Make torch compile validation default path for torch fx models.

Above changes were made to obtain the validation path:
image

### Related tickets

163422

### Tests

-- | -- | -- | -- | --
Manual | post_training_quantization | Pass | 652 |  
GH WC | WC test on GHA | Pass | <a href="https://github.com/openvinotoolkit/nncf/actions/runs/14587543146">67</a> |
Example | Example | Pass | <a href="https://github.com/openvinotoolkit/nncf/actions/runs/14587556157">287</a> |


tests/post_training/pipelines/image_classification_base.py
```diff
@@ -94,7 +94,7 @@ def _validate(self) -> None:
         predictions = np.zeros(dataset_size)
         references = -1 * np.ones(dataset_size)
 
-        if self.backend in FX_BACKENDS and self.torch_compile_validation:
```
daniil-lyakhov, line 97: Please do not remove the `self.torch_compile_validation` option and made it False by default

tests/post_training/pipelines/base.py
```diff
@@ -485,6 +485,25 @@ def compress(self) -> None:
...
 
+    def _rename_files(self, folder_path, new_name):
+        model_folder = folder_path / "model"
+        bin_file = None
+        xml_file = None
+        for file in os.listdir(model_folder):
+            if file.endswith(".bin"):
+                bin_file = file
+            elif file.endswith(".xml"):
+                xml_file = file
```
daniil-lyakhov, line 496: What if there are several submodels in this dir?

tests/post_training/pipelines/image_classification_base.py
anzr299: but then the default path for FX backend models will be using OV validation

daniil-lyakhov: Typo, my bad, I mead True by default

tests/post_training/pipelines/base.py
anzr299: submodels of the same model? or other models? 
I can save in a different folder by the model name if the problem is the latter like this:
`torch.compile(exported_model.module(), backend="openvino", options = {"model_caching" : True, "cache_dir": str(self.output_model_dir / self.model_name)})`
instead of just `"cache_dir": str(self.output_model_dir)`

daniil-lyakhov: I mean one model is being cut on several part, as it was with the Yolo 11 model. As far as I remember, this means there are several IRs generated for one model which are run sequentially

anzr299: hm, but the models with graph break should not be supported right?

daniil-lyakhov: They should have one graph, but it is possible due to bugs in ov/nncf that there are several IRs. I wonder, what will be the result? Perhaps we shouldn't analyze the parts of the model separatly

anzr299: Then Maybe I can raise an error after a check for multiple bin and xml files in the location.

alexsu52 (commented):
Could you update description with more details? I mean, it's not entirely clear what exactly was done.

daniil-lyakhov:
@anzr299, issue 162009 was fixed in the openvino 2025.1, please consider it

daniil-lyakhov (approved):

alexsu52 (approved):
LGTM

tests/post_training/pipelines/base.py
```diff
@@ -500,9 +524,14 @@ def save_compressed_model(self) -> None:
...
-            ov_model.reshape(self.input_size)
-            ov.serialize(ov_model, self.path_compressed_ir)
+            torch.export.save(exported_model, self.output_model_dir / "model.pt2")
+            mod = torch.compile(
+                exported_model.module(),
+                backend="openvino",
+                options={"model_caching": True, "cache_dir": str(self.output_model_dir)},
+            )
+            mod(self.dummy_tensor)
+            self._rename_files(self.output_model_dir, "model")
```
AlexanderDokuchaev, line 534:
```suggestion
            self.path_compressed_ir = list(Path(self.output_model_dir / "model").glob("*.xml"))[0]
```
Just set path_compressed_ir to saved model

AlexanderDokuchaev (changes requested):

tests/post_training/pipelines/image_classification_base.py
```diff
@@ -75,7 +75,7 @@ def process_result(request, userdata):
     def _validate_torch_compile(
         self, val_loader: torch.utils.data.DataLoader, predictions: np.ndarray, references: np.ndarray
     ):
-        compiled_model = torch.compile(self.compressed_model.cpu(), backend="openvino")
+        compiled_model = torch.compile(self.compressed_model.cpu(), backend="openvino", options={"aot_autograd": True})
```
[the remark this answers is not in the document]
anzr299: This is used to achieve the correct graph for swin models
ref CVS-163462

tests/post_training/pipelines/image_classification_torchvision.py
```diff
@@ -126,8 +126,7 @@ def _dump_model_fp32(self) -> None:
 
         if self.backend in FX_BACKENDS:
             exported_model = torch.export.export(self.model.cpu(), (self.dummy_tensor.cpu(),))
-            ov_model = ov.convert_model(exported_model, example_input=self.dummy_tensor, input=self.input_size)
-            ov.serialize(ov_model, self.fp32_model_dir / "fx_model_fp32.xml")
+            torch.export.save(exported_model, self.fp32_model_dir / "fx_model_fp32.pt2")
```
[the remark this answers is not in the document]
anzr299: It can be used to check the FX graph but requires to be loaded from a script for visualizing. Should I dump OV model instead?

tests/post_training/pipelines/base.py
anzr299: Oh, this is clever. I will add it to the function since the function also raises error for graph breaks and OV not dumping the graph due to error in torch compile OV backend.

tests/post_training/pipelines/base.py
```diff
@@ -500,9 +524,14 @@ def save_compressed_model(self) -> None:
             ov.serialize(ov_model, self.path_compressed_ir)
         elif self.backend in FX_BACKENDS:
             exported_model = torch.export.export(self.compressed_model.cpu(), (self.dummy_tensor.cpu(),))
-            ov_model = ov.convert_model(exported_model, example_input=self.dummy_tensor.cpu(), input=self.input_size)
-            ov_model.reshape(self.input_size)
-            ov.serialize(ov_model, self.path_compressed_ir)
+            torch.export.save(exported_model, self.output_model_dir / "model.pt2")
+            mod = torch.compile(
```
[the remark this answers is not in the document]
anzr299: This is to ensure that the torch compile cache model has the correct no. of FQ nodes. Since this is the default user journey I believe.
Also, because PT FE in OV for convert_model and the translations used for torch.compile sometimes differ. This method can help ensure that the correct OV graph is being checked.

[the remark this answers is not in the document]
AlexanderDokuchaev: Please add comment about it
Now it's not clear why used this method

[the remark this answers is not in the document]
anzr299: Alright

tests/post_training/pipelines/image_classification_base.py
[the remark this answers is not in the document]
anzr299: I have tested the `reproducer.py` locally from [temporary URL](https://www.temporary-url.com/B1BD5)
It works with the following packages:
Pytorch: 2.6.0
Openvino: 2025.1

tests/post_training/test_quantize_conformance.py
```diff
@@ -276,10 +276,10 @@ def test_ptq_quantization(
     run_torch_cuda_backend: bool,
     subset_size: Optional[int],
     run_benchmark_app: bool,
-    torch_compile_validation: bool,
     capsys: pytest.CaptureFixture,
     extra_columns: bool,
     memory_monitor: bool,
+    torch_compile_validation: bool = True,
```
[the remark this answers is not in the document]
anzr299: Done

tests/post_training/pipelines/base.py
anzr299: Done

tests/post_training/pipelines/base.py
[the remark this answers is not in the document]
anzr299: Done

tests/post_training/pipelines/base.py
```diff
@@ -485,6 +485,30 @@ def compress(self) -> None:
...
+                bin_file = file
+            elif file.endswith(".xml"):
+                xml_file = file
+        if bin_file is None or xml_file is None:
+            msg = "Openvino Model Files Not Found!"
+            raise FileNotFoundError(msg)
+        bin_new_path = folder_path / f"{new_name}.bin"
+        xml_new_path = folder_path / f"{new_name}.xml"
+
+        os.rename(os.path.join(model_folder, bin_file), bin_new_path)
```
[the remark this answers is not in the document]
anzr299: Done

tests/post_training/test_quantize_conformance.py
```diff
@@ -294,7 +294,7 @@ def test_ptq_quantization(
         run_torch_cuda_backend,
         subset_size,
         run_benchmark_app,
-        torch_compile_validation,
+        True, # Torch compile validation is true by default for the FX backend to have torch compile testing path as default.
```
AlexanderDokuchaev: @daniil-lyakhov https://github.com/openvinotoolkit/nncf/blob/develop/tests/post_training/conftest.py#L27 now is no need anymore?

AlexanderDokuchaev (changes requested):

tests/post_training/test_quantize_conformance.py
daniil-lyakhov: You are right. @anzr299, could you please remove this parameter from the test?

anzr299: okay

anzr299: Done

daniil-lyakhov (approved):

PR #3333, opened 2025-03-07
Use wikitext-2-raw-v1 in the examples

alexsu52:
### Changes

Use wikitext-2-raw-v1 in the examples

### Reason for changes
Minimize number downloads.

### Related tickets

N/A

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/13715504451


MaximProshin:
@alexsu52 , what do you mean by "Minimize number downloads"? The dataset size?

l-bat:
@alexsu52, have you checked whether the compressed parameters are the same for both datasets in example?

alexsu52:
> @alexsu52 , what do you mean by "Minimize number downloads"? The dataset size?

If examples run on same instance then the dataset will be downloaded once using the caching mechanism.

l-bat (approved):

Issue #3206, opened 2025-01-22
[TorchFX] Support weight quantization for operations with `weight_port_id != 1`


daniil-lyakhov:
### 🚀 Feature request

The task is to support weight quantization for operations with `weight_port_id != 1`.
  To do so, one should update the method `_get_input_scale_shape `of the MinMax `torch_fx_backend.py` https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/torch_fx_backend.py#L181-L182 as it done for the torch backend https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/min_max/torch_backend.py#L197-L202

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

tej-1916:
Hi @daniil-lyakhov, I’d like to work on this. I’ll add a regression test, update `_get_input_scale_shape`, and run the relevant tests. Is the issue still available?

PR #3334, opened 2025-03-08
[FX] Support weight quantization for operations where `weight_port_id` != 1

siddhant-0707:
### Changes

Updated the FX backend’s `_get_input_scale_shape` to use the FX insertion point shape and, when available, the actual weight tensor’s shape to compute the per‑channel scale shape.
Adjusted statistics collector in `_get_stat_collector` so that the reduction and aggregation axes are derived using the same channel axes as used for scale shape computation.

### Related tickets

Issue #3206 

### Tests
All tests run successfully

### DIFF
--- a/nncf/quantization/algorithms/min_max/algorithm.py
+++ b/nncf/quantization/algorithms/min_max/algorithm.py
@@ -439,29 +439,29 @@ class MinMaxQuantization(Algorithm):
 
     def _get_stat_collector(
         self,
-        nncf_graph: NNCFGraph,
+        graph: NNCFGraph,
         target_point: TargetPoint,
-        quantizer_config: QuantizerConfig,
+        qconfig: QuantizerConfig,
         batchwise_statistics: bool,
     ) -> TensorCollector:
         """
         Creates and returns a statistic collector based on the quantizer's configuration.
 
-        :param nncf_graph: NNCFGraph instance.
+        :param graph: NNCFGraph instance.
         :param target_point: Target point indicates where statistics should be collected.
-        :param quantizer_config: Configuration of a quantizer layer,
+        :param qconfig: Configuration of a quantizer layer,
         defining the configuration of created statistic collector.
         :param batchwise_statistics: Determines whether quantizer statistics should be calculated
             for each item of the batch or for the entire batch.
-        :return: TensorCollector for the statistics calculation.
+        :return: Statistic Collector.
         """
         is_weight = target_point.is_weight_target_point()
-        node = nncf_graph.get_node_by_name(target_point.target_node_name)
-        shape = self._backend_entity.get_target_point_shape(nncf_graph, node, target_point)
+        node = graph.get_node_by_name(target_point.target_node_name)
+        shape = self._backend_entity.get_target_point_shape(graph, node, target_point)
         
         # Get channel axes considering ConvTranspose layers
         channel_axes = ()
-        if quantizer_config.per_channel:
+        if qconfig.per_channel:
             if is_weight:
                 if node.metatype.__name__.startswith("PTConvTranspose"):
                     channel_axes = (1,)  # Output channels for transpose conv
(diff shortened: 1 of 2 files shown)


nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -458,15 +458,27 @@ def _get_stat_collector(
...
-            )
+            if is_weight:
+                if node.metatype.__name__.startswith("PTConvTranspose"):
+                    channel_axes = (1,)  # Output channels for transpose conv
+                else:
+                    channel_axes = self._backend_entity.get_weight_quantization_axes(
+                        node, target_point, len(shape)
+                    )
+            else:
+                channel_axes = (1,)
```
anzr299, line 473: You can take care of these cases inside the backend `get_weight_quantization_axes` implementation.
It could be more robust if you reuse the Torch backend function for `get_weight_quantization_axes` in Torch FX backend

nncf/quantization/algorithms/min_max/algorithm.py
```diff
@@ -458,15 +458,27 @@ def _get_stat_collector(
...
+
+        # Align statistics collection with scale shape
+        reduction_axes, aggregation_axes = None, None
+        if shape is not None and channel_axes:
+            all_axes = set(range(len(shape)))
+            reduction_axes = tuple(all_axes - set(channel_axes))
```
anzr299, line 479: Why do you need to write the reduction axes logic again here?

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -177,16 +178,35 @@ def _get_input_scale_shape(
...
+                channel_idx = channel_axes[0]
+                scale_shape = tuple(get_scale_shape(
+                    input_shape, 
+                    is_weights=True, 
+                    per_channel=per_channel, 
+                    channel_idx=channel_idx
+                ))
+            else:
+                scale_shape = (1,)
+                channel_idx = 0
```
anzr299, line 201: You can move this if-else outside the parent if-else. To look something like this:

```
if is_weights:
    ...
else:
    ...

if channel_axes:
    ....
else:
    ....
```

anzr299 (changes requested):

nncf/quantization/algorithms/min_max/algorithm.py
siddhant-0707: Removed manual reduction axes calculation.

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -177,16 +177,25 @@ def _get_input_scale_shape(
...
-            channel_idx = 1  # channel dim for activations
+            channel_axes = [1]
 
-        input_shape = nncf_graph.get_input_shape_for_insertion_point(target_point)
-        scale_shape = tuple(
-            get_scale_shape(input_shape, is_weights=is_weights, per_channel=per_channel, channel_idx=channel_idx)
-        )
+        channel_idx = channel_axes[0] if channel_axes else 0
+
+        if is_weights and not channel_axes:
```
anzr299, line 190:
```suggestion
        if not len(channel_axes):
```
to cover the case of vector weights which are being quantized per channel

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -177,16 +177,25 @@ def _get_input_scale_shape(
...
+            channel_axes = get_weight_channel_axes(node.metatype, len(input_shape), target_point.input_port_id)
         else:
-            channel_idx = 1  # channel dim for activations
+            channel_axes = [1]
 
-        input_shape = nncf_graph.get_input_shape_for_insertion_point(target_point)
-        scale_shape = tuple(
-            get_scale_shape(input_shape, is_weights=is_weights, per_channel=per_channel, channel_idx=channel_idx)
-        )
+        channel_idx = channel_axes[0] if channel_axes else 0
```
anzr299, line 188:
```suggestion
```
Since channel axes is already being checked and handled in the `if-else` block below. `channel_axes[0]` can directly be passed to `channel_idx` parameter of `get_scale_shape`

anzr299 (changes requested):

nncf/quantization/algorithms/min_max/torch_fx_backend.py
anzr299: Okay

anzr299:
Please check the failing pre-commit test.

PR #3336, opened 2025-03-11
Change decompressor result dtype

nikita-malininn:
### Changes

- As stated in the title

### Reason for changes

- To allow exporting fp16/bf16 models after NNCF

### Related tickets

- 159708

### Tests

- Updated

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/torch_backend.py
+++ b/nncf/quantization/algorithms/weight_compression/torch_backend.py
@@ -291,17 +291,16 @@ class PTWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
 
             # creates weight decompressor
             if compression_config.mode == CompressWeightsMode.INT8_SYM:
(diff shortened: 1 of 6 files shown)


tests/torch2/requirements.txt
```diff
@@ -13,3 +13,6 @@ torchvision
 addict>=2.4.0
 efficientnet_pytorch==0.7.1
 transformers==4.48.3
+
+optimum-intel==1.22.0
+optimum==1.24.0
```
nikita-malininn, line 18: This was added only because of the misalignment of the requirements of the `torch` and `torch2` tests.
For example, imports like https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch2/function_hook/quantization/test_weights_compression.py#L40-L41, requires the same packages as the https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/ptq/test_weights_compression.py file.

PR #3330, opened 2025-03-05
Fixed weight compression for float16/bfloat16 Torch models

ljaljushkin:
### Changes

Adapted AWQ and Scale Estimation algorithm for the case when weights and activations are float16 and bfloat16.

### Reason for changes

Otherwise, compression fails with errors like that:
`
RuntimeError: expected mat1 and mat2 to have the same dtype, but got: c10::BFloat16 != float
`

### Related tickets

n/a

### Tests

- tests/torch/ptq/test_weights_compression.py::test_half_precision_models
- PTWC https://github.com/openvinotoolkit/nncf/actions/runs/13680175191
- PTWC Performance 

| 51 build on develop |  52 build on PR |
:-------------------------:|:-------------------------:
 image | image

PR #3337, opened 2025-03-11
Set weight without overriding convert node

ljaljushkin:
### Changes

Assign the weight to the constant operation, rather than the first operation following the matrix multiplication.

### Reason for changes

In the past, the weights were assigned to the first operation after the Matmul, and this approach functioned properly when the weights were in float32 format. However, with the introduction of the ability to use float16/bfloat16 weights in AWQ, as described in the pull request #3330, the set_weights function no longer operates correctly. The first operation in the process is a `Convert`, which is replaced by a constant, and this causes an error:

```
RuntimeError: Check 'element::Type::merge(result_et, get_input_element_type(0), get_input_element_type(1))' failed at src\core\src\op\matmul.cpp:113:
While validating node 'opset1::MatMul MatMul_1657124 (opset1::Multiply __module.model.layers.0.post_attention_layernorm/aten::mul/Multiply_1[0]:f32[?,?,4096], opset1::Reshape Reshape_511646[0]:f16[14336,4096]) -&gt; (dynamic[...])' with friendly_name 'MatMul_1657124':
```

image

### Related tickets

163980
 should also fix tests in Optimum: 

### Tests

- [x] test_set_weight
- [x] `tests/openvino/test_quantization.py::OVWeightCompressionTest::test_ovmodel_4bit_auto_compression_with_config_07` and `tests/openvino/test_quantization.py::OVWeightCompressionTest::test_ovmodel_4bit_auto_compression_with_config_08` in Optimum have started to work, but failed before: https://github.com/huggingface/optimum-intel/actions/runs/13713350054/job/38353813210

### DIFF
--- a/nncf/quantization/algorithms/weight_compression/openvino_backend.py
+++ b/nncf/quantization/algorithms/weight_compression/openvino_backend.py
@@ -149,28 +149,19 @@ class OVWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
     def set_weight(
         self, node_with_weight: NNCFNode, weight_port_id: int, model: ov.Model, graph: NNCFGraph, weight: Tensor
     ):
-        node_with_const = self.name_to_node_mapping[node_with_weight.node_name]
+        const_op_friendly_name = node_with_weight.layer_attributes.constant_attributes[weight_port_id]["name"]
+        const_op = self.name_to_node_mapping[const_op_friendly_name]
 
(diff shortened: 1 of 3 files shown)


nikita-malininn (approved):

andreyanufr (approved):

alexsu52:
WC conformance test: https://github.com/openvinotoolkit/nncf/actions/runs/13805673573

PR #3348, opened 2025-03-17
[QAT Lora 4/N] Strip for LoRA modules

ljaljushkin:
### Changes

- Introduced strip method for LoRA modules as a new strip format = `StripFormat.DQ`. The old format named `StripFormat.NATIVE`. 
- Continuation of https://github.com/openvinotoolkit/nncf/pull/3331

### Reason for changes

- To able IR model conversion

### Related tickets

- 159708

### Tests

- test_fq_lora_tuning
- test_nncf_strip_lora_model

### DIFF
--- a/nncf/parameters.py
+++ b/nncf/parameters.py
@@ -119,6 +119,23 @@ class CompressionFormat(StrEnum):
     FQ_LORA = "fake_quantize_with_lora"
 
 
+@api(canonical_alias="nncf.StripFormat")
+class StripFormat(StrEnum):
+    """
+    Describes the format in which model is saved after strip: operation that removes auxiliary layers and
+    operations added during the compression process, resulting in a clean model ready for deployment.
+    The functionality of the model object is still preserved as a compressed model.
+
+    :param NATIVE: Returns the model with as much custom NNCF additions as possible,
+    :param DQ: Replaces FakeQuantize operations with dequantization subgraph and compressed weights in low-bit
+        precision using fake quantize parameters. This is the default format for deployment of models with compressed
+        weights.
+    """
+
+    NATIVE = "native"
+    DQ = "dequantize"
+
+
 @api(canonical_alias="nncf.BackupMode")
 class BackupMode(StrEnum):
     """
--- a/nncf/parameters.py
+++ b/nncf/parameters.py
@@ -126,7 +126,7 @@ class StripFormat(StrEnum):
     operations added during the compression process, resulting in a clean model ready for deployment.
     The functionality of the model object is still preserved as a compressed model.
 
-    :param NATIVE: Returns the model with as much custom NNCF additions as possible,
+    :param NATIVE: Returns the model with as much custom NNCF additions as possible.
     :param DQ: Replaces FakeQuantize operations with dequantization subgraph and compressed weights in low-bit
         precision using fake quantize parameters. This is the default format for deployment of models with compressed
         weights.
--- a/nncf/torch/quantization/strip.py
+++ b/nncf/torch/quantization/strip.py
@@ -10,6 +10,8 @@
 # limitations under the License.
 
 
+from typing import Tuple
+
 import numpy as np
 import torch
 from torch.quantization.fake_quantize import FakeQuantize
@@ -18,10 +20,10 @@ import nncf
 from nncf.common.graph.transformations.commands import TargetType
 from nncf.common.graph.transformations.layout import TransformationLayout
 from nncf.parameters import StripFormat
-from nncf.torch.dynamic_graph.scope import Scope
 from nncf.torch.graph.transformations.commands import ExtraCompressionModuleType
 from nncf.torch.graph.transformations.commands import PTSharedFnInsertionCommand
 from nncf.torch.graph.transformations.commands import PTTargetPoint
+from nncf.torch.model_graph_manager import get_const_data
 from nncf.torch.model_graph_manager import get_const_node
 from nncf.torch.model_graph_manager import get_module_by_name
 from nncf.torch.model_graph_manager import split_const_name
@@ -29,6 +31,7 @@ from nncf.torch.model_transformer import PTModelTransformer
 from nncf.torch.nncf_network import NNCFNetwork
 from nncf.torch.quantization.layers import AsymmetricQuantizer
 from nncf.torch.quantization.layers import BaseQuantizer
+from nncf.torch.quantization.layers import BaseWeightsDecompressor
 from nncf.torch.quantization.layers import INT4AsymmetricWeightsDecompressor
 from nncf.torch.quantization.layers import INT4SymmetricWeightsDecompressor
 from nncf.torch.quantization.layers import INT8AsymmetricWeightsDecompressor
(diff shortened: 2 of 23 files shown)


nncf/parameters.py
```diff
@@ -119,6 +119,23 @@ class CompressionFormat(StrEnum):
...
 
 
+@api(canonical_alias="nncf.StripFormat")
+class StripFormat(StrEnum):
+    """
+    Describes the format in which model is saved after strip: operation that removes auxiliary layers and
+    operations added during the compression process, resulting in a clean model ready for deployment.
+    The functionality of the model object is still preserved as a compressed model.
+
+    :param NATIVE: Returns the model with as much custom NNCF additions as possible,
```
daniil-lyakhov, line 129:
```suggestion
    :param NATIVE: Returns the model with as much custom NNCF additions as possible.
```

nncf/parameters.py
```diff
@@ -119,6 +119,23 @@ class CompressionFormat(StrEnum):
     FQ_LORA = "fake_quantize_with_lora"
 
 
+@api(canonical_alias="nncf.StripFormat")
```
daniil-lyakhov, line 122: This should appear in the nncf API docs after the merge, please check everything is shown properly after the merge

daniil-lyakhov (approved):
LGTM, minor comments

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +181,161 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+    transformation_layout = TransformationLayout()
+    model = model.nncf.get_clean_shallow_copy()
+    graph = model.nncf.get_graph()
+
+    for command in transformations:
+        quantizer = command.fn
+
+        msg = None
+        if not isinstance(quantizer, (SymmetricQuantizer, AsymmetricQuantizer)):
+            msg = f"Unexpected compression module on strip: {quantizer.__class__}"
```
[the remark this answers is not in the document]
ljaljushkin: Done

nncf/parameters.py
ljaljushkin: Done

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +178,159 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+
+    The FQ format may round these values to one "quant", while the DQ format rounds them to another "quant".
+    To avoid these issues, the compressed representation should be provided not by directly quantizing the input,
+    but by quantizing a pre-processed, fake-quantized, floating-point representation.
+
+    :param model: Compressed model with Decompressors.
+    :return: The modified NNCF network.
+    """
+    transformation_layout = TransformationLayout()
+    transformations = model.nncf.transformation_layout().transformations
```
AlexanderDokuchaev, line 220: As fast solution is it may be ok, but using transformation is WA to replacer module. 
It's not portable to new tracing, potentially problems other problems in changes in other parts of code that not related to strip.

As i see need to do two steps: replace quantizer and pack wights.

As fast WA to fast merge works is it ok, but it should be reworked after.

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +178,159 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+    model = model.nncf.get_clean_shallow_copy()
+    graph = model.nncf.get_graph()
+    for command in transformations:
+        quantizer = command.fn
+
+        msg = ""
+        if not isinstance(quantizer, (SymmetricQuantizer, AsymmetricQuantizer)):
```
AlexanderDokuchaev, line 227: Why raise error in this case?
as i understand if modules is not Quantizer, it should be skipped, and keep module as it.

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +178,159 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+        original_shape = original_weight.shape
+        original_eps = torch.finfo(original_dtype).eps
+
+        qdq_weight = quantizer.quantize(original_weight)
+        if hasattr(quantizer, "_lspec"):
+            # Reshape for group-wise quantization, implemented for classes with lora spec only
+            qdq_weight = qdq_weight.reshape(quantizer._lspec.weight_shape)
+        qdq_weight = qdq_weight.to(original_dtype)
+
+        if isinstance(quantizer, AsymmetricQuantizer):
```
AlexanderDokuchaev, line 256: add convert_fq_to_decompressor functions to make easy to support this feature and add test that will only convert FQ to decompressors.

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +178,159 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+            input_range_safe = abs(quantizer.input_range) + quantizer.eps
+            input_low, input_range = TuneRange.apply(quantizer.input_low, input_range_safe, quantizer.levels)
```
AlexanderDokuchaev, line 258: Please check that is not possible to reuse existed function to get zp and scales? 
It can lead to differ results on training and after strip.

nncf/torch/quantization/strip.py
```diff
@@ -163,14 +178,159 @@ def remove_disabled_quantizers(model: NNCFNetwork) -> NNCFNetwork:
...
+                    scale=scale,
+                    compressed_weight_shape=q_weight.shape,
+                    result_shape=original_shape,
+                    result_dtype=original_dtype,
+                )
+
+        packed_tensor = decompressor.pack_weight(q_weight)
+
+        # sets compressed tensor
+        compressed_parameter = torch.nn.Parameter(packed_tensor, requires_grad=False)
```
AlexanderDokuchaev, line 316: Make updating weights like in WC, where changed only weight.data = packed_tensor

AlexanderDokuchaev (changes requested):

nncf/torch/quantization/strip.py
ljaljushkin: copy pasted the part from WC with TODO :)

nncf/torch/quantization/strip.py
ljaljushkin: `calculate_quantized_weight` requires `WeightCompressionConfig`
`calculate_scale_zero_point` would require creation of nncf.Tensor and do unnecessary calculations with input_high, input_low, when for symmetric case it just needs `scale = quantizer.scale / abs(quantizer.level_low)`
It could lead to bigger error, because of different order of operations.

nncf/torch/quantization/strip.py
ljaljushkin: Splitted into 2 functions, 
IMO this test should be enough, doing the same as you said, but without manual initialization of quantizer: 
https://github.com/openvinotoolkit/nncf/pull/3348/files#diff-aab032c69dafe50ea4db308ec7308d3012ba792e30196e145fd8cad3f7d97e64R346

nncf/torch/quantization/strip.py
ljaljushkin: Good catch! Fixed

nncf/torch/quantization/strip.py
ljaljushkin: Strip will have slightly different implementation with new tracing. Decided to support with old tracing first.

AlexanderDokuchaev (approved):

PR #3350, opened 2025-03-18
Add `is_arm_cpu` check

nikita-savelyevv:
### Changes

Avoid calling `cpuinfo.get_cpu_info()["brand_raw"])` on ARM CPUs because an error can be raised in such case.

Related to https://github.com/openvinotoolkit/openvino_notebooks/issues/2810 .


nikita-savelyevv:
Opened https://github.com/openvinotoolkit/nncf/pull/3357 instead

PR #3357, opened 2025-03-19
Use openvino instead of cpuinfo to get CPU full name

nikita-savelyevv:
### Changes

Use openvino instead of cpuinfo to get CPU full name. Using cpuinfo can result in issues.

image

Related to https://github.com/openvinotoolkit/openvino_notebooks/issues/2810

### DIFF
--- a/nncf/openvino/cpu_info.py
+++ b/nncf/openvino/cpu_info.py
@@ -17,6 +17,10 @@ _IS_LNL_CPU = None
 
 
 def is_lnl_cpu() -> bool:
+    """
+    Checks whether current CPU is an Intel Lunar Lake generation or not.
+    :return: True if current CPU is an Intel Lunar Lake generation, False othwerwise.
+    """
     global _IS_LNL_CPU
     if _IS_LNL_CPU is None:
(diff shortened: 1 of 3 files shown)


eaidova (approved):

nncf/openvino/cpu_info.py
```diff
@@ -11,13 +11,14 @@
...
+import openvino as ov
 
 _IS_LNL_CPU = None
 
 
 def is_lnl_cpu() -> bool:
     global _IS_LNL_CPU
     if _IS_LNL_CPU is None:
-        _IS_LNL_CPU = re.search(r"Ultra \d 2\d{2}", cpuinfo.get_cpu_info()["brand_raw"]) is not None
+        cpu_name = ov.Core().get_property("CPU", ov.properties.device.full_name)
```
AlexanderDokuchaev, line 22: Please add docstring for is_lnl_cpu

AlexanderDokuchaev (changes requested):

nncf/openvino/cpu_info.py
nikita-savelyevv: Added

AlexanderDokuchaev (approved):

PR #3361, opened 2025-03-19
[PT2] Context manager to disable tracing

AlexanderDokuchaev:
### Changes

Add context manager to disable tracing for some part of code in a model

### DIFF
--- a/tests/torch2/function_hook/test_disable_tracing.py
+++ b/tests/torch2/function_hook/test_disable_tracing.py
@@ -1,0 +1,50 @@
+# Copyright (c) 2025 Intel Corporation
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#      http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


tests/torch2/function_hook/test_disable_tracing.py
```diff
@@ -0,0 +1,50 @@
...
+        assert isinstance(mode[0], FunctionHookMode)
+        assert not mode[0].enabled
+        return x - 1
+
+
+disable_tracing(ModelNoTrace.foo)
+
+
+def test_build_graph_with_disable_tracing(regen_ref_data):
+    model = wrap_model(ModelNoTrace())
```
daniil-lyakhov, line 47: What do you think about updating this test in the following way:
1) You check tracing is enabled
2) You wraps the model with the `disable_tracing`, check the tracing is disabled
3) You check that tracing is once again enabled without the `disable_tracing` context

AlexanderDokuchaev: > You check tracing is enabled

Tracing is not available for not wrapped model (in user's scenarios of using).
There is no any global tracing context. Instance of FunctionHookMode creates in patched forward https://github.com/openvinotoolkit/nncf/blob/d614c1ed9f2e143895a74d21a75f6369c1decff7/nncf/experimental/torch2/function_hook/wrapper.py#L53

But as i understand you suggestion:
- added test_disable_function_hook_mode that check FunctionHookMode inside and outside disble context.
- add operation after function with disabled tracing

PR #3362, opened 2025-03-19
[PT2] rework ForwardWithHooks

AlexanderDokuchaev:
### Changes

Rework ForwardWithHooks to be more flexible to patching forward function. 
Now model collects inside ForwardWithHooks and not used original forward to find model.
Add `_has_compatible_shallow_copy_type` and `_parse_to` to ignored function names.
Fix tests on cuda.

### Reason for changes

```python 
model = AutoModelForCausalLM.from_pretrained("hf-internal-testing/tiny-random-GPTNeoXForCausalLM", torch_dtype=torch.bfloat16, device_map="auto")
model = wrap_model(model)

#    model = cast(nn.Module, self._func.__self__)  # type: ignore[attr-defined]
# AttributeError: 'functools.partial' object has no attribute '__self__'. Did you mean: '__call__'?
```

### DIFF
--- a/nncf/experimental/torch2/function_hook/wrapper.py
+++ b/nncf/experimental/torch2/function_hook/wrapper.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import inspect
 import types
+from functools import partial
 from types import MethodType
 from typing import Any, Callable, Dict, Tuple, TypeVar, cast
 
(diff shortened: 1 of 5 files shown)


nncf/experimental/torch2/function_hook/wrapper.py
```diff
@@ -31,70 +31,76 @@
...
-    __slots__ = "_func", "__dict__", "__weakref__"
-    _func: Callable[..., Any]
-
-    def __new__(cls, orig_forward: Callable[..., Any]) -> ForwardWithHooks:
-        if not callable(orig_forward):
-            msg = "the first argument must be callable"
-            raise TypeError(msg)
+    __slots__ = "_orig_forward", "_model", "__dict__", "__weakref__"
+    _orig_forward: Callable[..., Any]
+    _model: nn.Module
```
ljaljushkin, line 36: Could the storing a reference to the module in hooks potentially result in issues with the removal of the model?

alexsu52: I would suggest checking the case with the model's coping.

AlexanderDokuchaev: The reference to module contains in `orig_forward` too under `__self__` attributes (in case of bound method) and collected in arg[0] in patched forward by `functools.partial` (for accelerate).
This changes only about getting model only ones on wrapping, on develop model gets from `orig_forward` each forward.

alexsu52: Thanks for the answer.

ljaljushkin (approved):

PR #3369, opened 2025-03-21
Optimized weights compression for NF4 data type

nikita-savelyevv:
### Changes

- Added optimized NF4 weights compression.
- Refactored function names inside `weight_lowering.py` so they are better structured.
- A temporary workaround had to be added in GPTQ algorithm. Will address it in a separate PR.

The comparison table is below. The weight compression part of the pipeline has become significantly faster. The rest should be roughly the same as before.

| Model         | Compression     | Compression Time Develop (sec.) | Compression Time Branch (sec.)     | Peak Memory Develop (MiB) | Peak Memory Branch (MiB)     |
|---------------|-----------------|----------------------------------|-------------------------------------|-----------------------------|-------------------------------|
| tiny-llama    | nf4             | 60.54                            | 10.58 (-82.52%)                     | 860                        | 868 (+0.93%)                  |
| phi4-mini     | nf4             | 178.00                           | 14.52 (-91.85%)                     | 5448                       | 5424 (-0.44%)                 |
| llama-3.1-8b  | nf4             | 380.39                           | 25.20 (-93.37%)                     | 5714                       | 5965 (+4.39%)                 |
 | | | | | |
| tiny-llama    | nf4, AWQ        | 268.16                           | 122.07 (-54.48%)                    | 2754                       | 2760 (+0.22%)                 |
| tiny-llama    | nf4, SE         | 694.79                           | 247.63 (-64.37%)                    | 2763                       | 2763 (+0.00%)                 |
| tiny-llama    | nf4, LoRA       | 967.92                           | 853.80 (-11.80%)                    | 2759                       | 2754 (-0.18%)                 |
| tiny-llama    | nf4, GPTQ       | 1404.00                          | 1367.76 (-2.58%)                    | 4098.63                       | 4100.73(+0.01%)                 |

Measured on Core i9-10980XE.

F4_E2M1 could also be improved, but ticket 164851 currently blocks that.

### Reason for changes

Improving UX during NF4 weights compression.

### Related tickets

164717

### Tests

- Expanded compression alignment tests to NF4 data type.
- Added an end-to-end compression alignment test (tests/openvino/optimized_functions/test_compression_functions.py::test_end_to_end_alignment) that checks the whole compression pipeline alignment including data-aware compression.
- https://github.com/openvinotoolkit/nncf/actions/runs/14926646556


alexsu52 (changes requested):

nncf/tensor/functions/numeric.py
```diff
@@ -806,9 +806,19 @@ def as_numpy_tensor(a: Tensor) -> Tensor:
     Convert tensor to numpy.
     In certain cases, this conversion may involve data copying, depending on the
     data type or device. Specifically:
-      - OV: if tensors data type is bf16, u4 or i4.
+      - OV: if tensors data type is bf16, fp8, u4, i4, nf4 or f4_e2m1.
```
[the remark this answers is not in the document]
nikita-savelyevv: Updated

nncf/tensor/tensor.py
```diff
@@ -205,6 +205,9 @@ def clone(self) -> Tensor:
     def as_numpy_tensor(self) -> Tensor:
         return cast(Tensor, _call_function("as_numpy_tensor", self))
 
+    def as_openvino_tensor(self) -> Tensor:
```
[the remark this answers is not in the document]
nikita-savelyevv: I've added a conversion from torch to openvino via a numpy array. It should be efficient since numpy -> ov conversion does not allocate additional memory.

nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -284,16 +285,24 @@ def _quantize_weights(
...
-                    quantized_col = do_nf4_dequantization(compressed_weights, scales[-1], reduction_axis=-1)
-                else:
-                    compressed_weights = calculate_quantized_weight(
-                        fns.unsqueeze(weight_col, 1), block_compression_config, scales[-1], zero_points[-1]
-                    )
-                    quantized_col = do_int_dequantization(compressed_weights, scales[-1], zero_points[-1])
+                with set_env_variable("NNCF_DISABLE_OPTIMIZED_COMPRESSION", "1"):
+                    # Because of the fact that compression if performed per-column, weight size is very small and
+                    # optimized OV compression performs worse than numpy compression.
+                    # TODO: Remove this workaround by introducing logic that will control whether to execute
```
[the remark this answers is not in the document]
nikita-savelyevv: Added

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -137,136 +143,126 @@ def calculate_nf4_scale(weight: Tensor, reduction_axes: ReductionAxes) -> Tensor
...
     """
-    Calculates the scale for e2m1 quantization.
+    Dequantize the float-quantized weight tensor.
 
-    :param weight: Weight array to compress.
-    :param reduction_axes: Axes along which to reduce (collect) different statistics (e.g., min, max).
-    :param max_val: Maximal value of e2m1 type.
-    :param to_e8m0: Defines convert scale to e8m0 or not.
-    :return: Scale tensor of float32 type for e2m1 quantization.
+    :param compressed_weight: Tensor with floating-point values, where each of them corresponds to 1 out of 16 quants.
```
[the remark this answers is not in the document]
nikita-savelyevv: Removed. This is just the case now because this function is currently supported only for nf4 and f4e2m1. In the future this statement indeed will not be true.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -137,136 +143,126 @@ def calculate_nf4_scale(weight: Tensor, reduction_axes: ReductionAxes) -> Tensor
...
+    :param reduction_axes: Axes along which to reduce statistics. Not required if precomputed scale are provided.
+    :param precomputed_scale: Optional precomputed scale tensor.
+    :param return_compressed_weight: If True, besides decompressed weight will also return compressed weight and scale.
+    :return: Dequantized weight tensor or a tuple containing the decompressed weight, compressed weight and scale.
     """
-    assert mode in [CompressWeightsMode.NF4, CompressWeightsMode.E2M1]
-    if weight.dtype != TensorDataType.float32:
-        weight = weight.astype(TensorDataType.float32)
+    assert config.mode == CompressWeightsMode.NF4
+    # TODO: add support for f4e2m1 once ticket 164851 is resolved
```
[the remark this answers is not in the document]
nikita-savelyevv: Added

nncf/tensor/tensor.py
```diff
@@ -205,6 +205,25 @@ def clone(self) -> Tensor:
...
+        if x.backend == TensorBackend.numpy:
+            try:
+                import openvino as ov  # type: ignore
+            except ImportError:
+                msg = "OpenVINO is not installed. Please install OpenVINO to cast a tensor to OpenVINO."
+                raise ImportError(msg)
+
+            from nncf.tensor.functions.openvino_numeric import DTYPE_MAP
+
+            x = Tensor(ov.Tensor(x.data, x.data.shape, DTYPE_MAP[x.dtype]))
```
alexsu52, line 221: I would suggest don't add support for PyTorch because the implementation is not optimal.

My suggestion for numpy backend:
```suggestion
        if x.backend == TensorBackend.numpy:
            x = cast(Tensor, _call_function("from_numpy", x.data, BackendType.OPENVINO))
```

nikita-savelyevv: Got it, done.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -137,136 +143,126 @@ def calculate_nf4_scale(weight: Tensor, reduction_axes: ReductionAxes) -> Tensor
...
-
-    eps = fns.finfo(scale).eps
-    scale = fns.where(fns.abs(scale) < eps, eps, scale)
+    if config.group_size != -1 and reduction_axes is not None:
+        # weights are reshaped: [a1, r, a2] -> [a1, r//gs, gs, a2]
+        weight, reduction_axes = reshape_weight_for_grouped_quantization(weight, reduction_axes, config.group_size)
 
-    return scale
+    # Optimized implementation
+    if config.mode == CompressWeightsMode.NF4 and _can_run_optimized(weight.backend):
```
ljaljushkin, line 197: What do you think about caching result of `_can_run_optimized`, e.g. by using `@lru_cache(maxsize=None)`?
It's called quite often and incurs a system call to retrieve an environment variable, but the result shouldn't change for a given backend.

nncf/openvino/optimized_functions/models.py
```diff
@@ -233,7 +234,72 @@ def get_compress_weight_model(
...
+    ov_model_params: OVModelParameters,
+    config: WeightCompressionConfig,
+    weight_shape: tuple,
+    scale_shape: Optional[tuple] = None,
+    reduction_axes: Optional[ReductionAxes] = None,
+    return_compressed_weight: Optional[bool] = False,
+) -> ModelCallable:
+    """
+    Get a model that performs float (currently only nf4) compression and decompression of the given weight.
+    :param ov_model_params: OV model parameters.
```
ljaljushkin, line 277:
```suggestion
    Get a model that performs float (currently only nf4) compression and decompression of the given weight.
    
    :param ov_model_params: OV model parameters.
```

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -526,7 +481,7 @@ def quantize_dequantize_weight(
     return_compressed_weight: Optional[bool] = False,
 ) -> Union[Tensor, tuple[Tensor, Tensor, Tensor, Tensor]]:
     """
-    First quantizes the given weight tensor and then dequantizes it back to obtain float32 values.
+    First quantizes the given weight tensor to integer dtype and then dequantizes it back to obtain float32 values.
     :param weight: The weight tensor to quantize-dequantize.
```
ljaljushkin, line 485:
```suggestion
    First quantizes the given weight tensor to integer dtype and then dequantizes it back to obtain float32 values.
    
    :param weight: The weight tensor to quantize-dequantize.
```

nncf/openvino/optimized_functions/models.py
```diff
@@ -233,7 +234,72 @@ def get_compress_weight_model(
...
+def get_float_quantization_model(
+    ov_model_params: OVModelParameters,
+    config: WeightCompressionConfig,
+    weight_shape: tuple,
+    scale_shape: Optional[tuple] = None,
+    reduction_axes: Optional[ReductionAxes] = None,
+) -> Union[ModelCallable, ModelAsNodes]:
+    """
+    Get a model that compresses weights to float (currently only nf4) destination type using the given configuration.
+    :param ov_model_params: OV model parameters.
```
ljaljushkin, line 246:
```suggestion
    Get a model that compresses weights to float (currently only nf4) destination type using the given configuration.
    
    :param ov_model_params: OV model parameters.
```

ljaljushkin (changes requested):
Nice work - clean and effective!

nncf/quantization/algorithms/weight_compression/weight_lowering.py
nikita-savelyevv: Unfortunately, there is an additional check for `NNCF_DISABLE_OPTIMIZED_COMPRESSION` environment variable inside `_can_run_optimized`, so the result is not dependent solely on input backend. There is currently a workaround that depends on this: https://github.com/nikita-savelyevv/nncf/blob/ns/ov-nf4-optimized-compression/nncf/quantization/algorithms/weight_compression/gptq.py#L288 . 


Also, I believe the time of this function execution should not be significant. For example, for the code below

```python
import time
import numpy as np
from nncf.quantization.algorithms.weight_compression.weight_lowering import _can_run_optimized
from nncf.tensor import TensorBackend
import nncf.tensor.functions as fns

N = 1000
size = (10,)
tensors = []
np.random.seed(0)
for _ in range(N):
    x = np.random.randn(*size)
    k = np.random.randint(3)
    t = fns.from_numpy(x, backend=TensorBackend.torch if k == 0 else TensorBackend.ov if k == 1 else TensorBackend.numpy)
    tensors.append(t)

start_time = time.perf_counter()
for t in tensors:
    _can_run_optimized(t.backend)
end_time = time.perf_counter()
print(f"Average time for _can_run_optimized: {(end_time - start_time) / N:.6f} seconds")

```

I get `Average time for _can_run_optimized: 0.000062 seconds`.

nncf/openvino/optimized_functions/models.py
nikita-savelyevv: Done

nncf/openvino/optimized_functions/models.py
nikita-savelyevv: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
ljaljushkin: My main concern was about env variable, not other part of the condition. I supposed that env variable shouldn't change within run and caching potentially should be functional. 
But if it's a workaround that is going to be resolved in the next PR, we can keep it as is for now.

ljaljushkin (approved):

alexsu52 (approved):

PR #3391, opened 2025-03-28
More helpful error message in case of invalid group size

ljaljushkin:
### Changes

If the number of weight channels is not dividable by group size, the compression will fail with the following message:
image

### Reason for changes

previous message wasn't informative

### Related tickets

n/a

### Tests

- test_raise_error_with_unsupported_params_for_int8

### DIFF
--- a/tests/cross_fw/test_templates/template_test_weights_compression.py
+++ b/tests/cross_fw/test_templates/template_test_weights_compression.py
@@ -9,6 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import math
+import re
 from abc import ABC
 from abc import abstractmethod
 from typing import Dict, List, TypeVar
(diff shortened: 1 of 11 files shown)


tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -316,3 +325,37 @@ def test_awq_scale_reference(self, monkeypatch):
...
+            model=model,
+            mode=CompressWeightsMode.INT4_ASYM,
+            ratio=1.0,
+            group_size=invalid_group_size,
+            all_layers=True,
+            **algorithm_dict,
+            dataset=dataset,
+        )
+
+        with pytest.raises(InvalidGroupSizeError) as exc_info:
```
AlexanderDokuchaev, line 354:
```suggestion
        with pytest.raises(InvalidGroupSizeError, match="...") as exc_info:
```

AlexanderDokuchaev (approved):

andreyanufr (approved):

PR #3395, opened 2025-03-31
fix len() of unsized object

XinyuWuu:
### Changes

Added new logic when `hasattr(tensor_collector.get_statistics().mean_values, "__len__") == False`.

### Reason for changes

if `tensor_collector.get_statistics().mean_values` is a `nncf.tensor` with shape `()` of dim `0`, `extend` function call will fail.

### Tests

`tests/onnx/quantization/test_bias_correction.py` is tested.
I am not sure how to test it as I encountered this bug during quantization of a network from one of my robotics projects. I guess the bug will only be triggled by certain networks and certain calibration datasets, therefor no new tests are added to varify the fix.

### DIFF
--- a/nncf/quantization/algorithms/bias_correction/algorithm.py
+++ b/nncf/quantization/algorithms/bias_correction/algorithm.py
@@ -550,7 +550,10 @@ class BiasCorrection(Algorithm):
         for tensor_collector in statistic_points.get_algo_statistics_for_node(
             node_name, output_filter_func, self._algorithm_key
         ):
-            output_fp.extend(tensor_collector.get_statistics().mean_values)
+            if not hasattr(tensor_collector.get_statistics().mean_values, "__len__"):
+                output_fp.append(tensor_collector.get_statistics().mean_values)
+            else:
+                output_fp.extend(tensor_collector.get_statistics().mean_values)
         return output_fp
 
     def get_statistic_points(self, model: TModel, graph: NNCFGraph) -> StatisticPointsContainer:


daniil-lyakhov:
Greetings, @XinyuWuu! I afraid we need to find a root cause for this bug, looks like the effect you are describing is consequence of the deeper error. `MeanTensorStatistic`, which unexpectedly returns a scalar, indicate a problems inside. 
It is combined only with the `MeanAggregator` which should ALWAYS return a NNCFTensor. 

So please answer following question to clarify the problem:
* Which NNCF backend did you use?
* Could you please attach a minimal reproducer?

We indeed need to test it, I'll help you with that

XinyuWuu:
Glad to spot something real. 
I am using the OpenVino backend. I recorded some data and zipped them with the model for reproduce:
[test.zip](https://github.com/user-attachments/files/19553467/test.zip)
test code:
<details>

```py
import pickle

import nncf
import openvino as ov
import torch

ovcore = ov.Core()
ovmodel = ovcore.read_model("test.xml")
with open("test.pkl", "rb") as f:
    calibration_dataset = pickle.load(f)
calibration_dataset = nncf.Dataset(
    calibration_dataset,
    lambda x: {k: torch.tensor(v).to(torch.float32) for k, v in x.items()},
)
quantized_model = nncf.quantize(
    ovmodel,
    calibration_dataset,
    fast_bias_correction=False,
    target_device=nncf.TargetDevice.CPU,
    model_type=nncf.ModelType.TRANSFORMER,
)
```

</details>

The return type of `MeanTensorStatistic` is actully correct `<class 'nncf.tensor.tensor.Tensor'>` but the shape is `()`.

MaximProshin:
@daniil-lyakhov , please clarify the next step here.

daniil-lyakhov:
Hello @XinyuWuu,

Thank you for your help! I was able to reproduce the problem and have found issues in the nncf.Tensor, BiasCorrection and FastBiasCorrection algorithms: the first and the second issue are resolved by the following PR: #3479, but looks like FastBiasCorrection just does not support models without a batch dimension

Sanity test to check BiasCorrection algorithm is working: https://colab.research.google.com/drive/1fk7DJOCtOirFLFH35Z1kasgLwPKD-bl_?usp=sharing

daniil-lyakhov:
Moved to #3479 https://github.com/openvinotoolkit/nncf/issues/3481

PR #3386, opened 2025-03-27
Added qat lora sample to test_examples

ljaljushkin:
### Changes

Added sample with QAT + absorbable LoRA to test example on CUDA.
new job was introduced, since previous tests were launched on CPU only.

### Reason for changes

validation of the tuning sample

### Related tickets
164755

### Tests
https://github.com/openvinotoolkit/nncf/actions/runs/14261388047/attempts/1#summary-39973579272

PR #3399, opened 2025-04-01
Test examples job for CUDA

ljaljushkin:
### Changes

Added job for testing examples on CUDA

### Reason for changes

Previously, all examples were tested on CPU only.
But "QAT with absorbable LORA" hangs on AMD CPU in CI (https://github.com/openvinotoolkit/nncf/actions/runs/14109007969/job/39522492543), that's why need job for testing on CUDA.

### Related tickets

https://github.com/openvinotoolkit/nncf/pull/3386

### Tests

n/a


.github/workflows/examples.yml
```diff
@@ -55,9 +55,79 @@ jobs:
...
+      - name: Print installed modules
+        run: pip list
+      - name: Run examples test scope
+        run: |
+          set +e
+          python -m pytest -s -m "cuda" -ra tests/cross_fw/examples \
+            --junit-xml=pytest-results.xml \
+            --durations-path=tests/cross_fw/examples/.test_durations_cuda \
             --splitting-algorithm=least_duration \
             --splits 4 \
```
AlexanderDokuchaev, line 132: You set only one test on cuda, i thinks split, group and split algorithm is not need.
or use 1 splits

ljaljushkin: utilized 1 split for scalability

PR #3402, opened 2025-04-02
Minor strip updates

nikita-malininn:
### Changes

- Added test for strip sym part.
- Updated asym part with common formula.
- Updated `BaseWeightsDecompressor` for comparison possibility.

### Reason for changes

- Backlog

### Related tickets

- 164645

### Tests

- Updated

### DIFF
--- a/nncf/torch/quantization/strip.py
+++ b/nncf/torch/quantization/strip.py
@@ -20,6 +20,8 @@ import nncf
 from nncf.common.graph.transformations.commands import TargetType
 from nncf.common.graph.transformations.layout import TransformationLayout
 from nncf.parameters import StripFormat
+from nncf.quantization.fake_quantize import calculate_scale_zero_point
+from nncf.tensor import Tensor
 from nncf.torch.graph.transformations.commands import ExtraCompressionModuleType
 from nncf.torch.graph.transformations.commands import PTSharedFnInsertionCommand
 from nncf.torch.graph.transformations.commands import PTTargetPoint
@@ -233,9 +235,16 @@ def asym_fq_to_decompressor(
     scale = torch.where(torch.abs(scale) < eps, eps, scale)
     scale = scale.to(weight_dtype)
 
-    zero_point = quantizer.level_low - torch.round(input_low / scale)
-    zero_point = torch.clip(zero_point, quantizer.level_low, quantizer.level_high)
-    zero_point = zero_point.to(integer_dtype)
+    input_high = input_range - input_low
+    scale, zero_point = calculate_scale_zero_point(
+        input_low=Tensor(input_low),
+        input_high=Tensor(input_high),
+        level_low=quantizer.level_low,
+        level_high=quantizer.level_high,
+        narrow_range=False,
+    )
+    scale = scale.data.to(weight_dtype)
+    zero_point = zero_point.data.to(integer_dtype)
 
     q_weight = qdq_weight / scale
     q_weight = q_weight + zero_point
--- a/tests/torch/quantization/test_strip.py
+++ b/tests/torch/quantization/test_strip.py
@@ -33,9 +33,12 @@ from nncf.torch.quantization.layers import INT4AsymmetricWeightsDecompressor as 
 from nncf.torch.quantization.layers import INT4SymmetricWeightsDecompressor as INT4SymDQ
 from nncf.torch.quantization.layers import INT8AsymmetricWeightsDecompressor as INT8AsymDQ
 from nncf.torch.quantization.layers import INT8SymmetricWeightsDecompressor as INT8SymDQ
+from nncf.torch.quantization.layers import PTLoraSpec
 from nncf.torch.quantization.layers import PTQuantizerSpec
+from nncf.torch.quantization.layers import SymmetricLoraQuantizer
 from nncf.torch.quantization.layers import SymmetricQuantizer
 from nncf.torch.quantization.strip import convert_to_torch_fakequantizer
+from nncf.torch.quantization.strip import sym_fq_to_decompressor
 from tests.common.quantization.data_generators import check_outputs
 from tests.common.quantization.data_generators import generate_lazy_sweep_data
 from tests.common.quantization.data_generators import generate_random_low_and_range_by_input_size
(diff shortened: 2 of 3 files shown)


tests/torch/quantization/test_strip.py
```diff
@@ -409,3 +414,129 @@ def test_nncf_strip_lora_model(mode, decompressor_class, torch_dtype, atol, mock
...
+
+
+SIGNED_WEIGHT_SAMPLE = [-1.0, -0.75, -0.5, -0.25, 0.0, 0.25, 0.5, 0.75]
+SCALE_SAMPLE = [2.0]
+
+
+@pytest.mark.parametrize(
+    ("num_bits", "ref_scale", "torch_dtype"),
+    (
+        (4, [0.25], torch.float32),
```
ljaljushkin, line 426: Would you kindly provide a mathematical formula with substituted values for both tests, indicating how these values should be determined? In case of changed test parameters or a new test case, give a hint to the developer to calculate these references.

nikita-malininn: Done

tests/torch/quantization/test_strip.py
```diff
@@ -409,3 +414,140 @@ def test_nncf_strip_lora_model(mode, decompressor_class, torch_dtype, atol, mock
...
+
+@pytest.mark.parametrize(
+    ("num_bits", "ref_scale", "ref_zero_point", "torch_dtype"),
+    (
+        (4, 0.03333, 0.0, torch.float32),
+        (8, 0.00196, 0.0, torch.float32),
+        (4, 0.03333, 0.0, torch.float16),
+        (8, 0.00196, 0.0, torch.float16),
+        (4, 0.03345, 0.0, torch.bfloat16),
+        (8, 0.007812, 0.0, torch.bfloat16),
```
ljaljushkin, line 494: It's so strange that the reference scales for 8-bit are much different from those for 4-bit.

tests/torch/quantization/test_strip.py
```diff
@@ -409,3 +414,140 @@ def test_nncf_strip_lora_model(mode, decompressor_class, torch_dtype, atol, mock
...
+    quantizer.input_low.data = input_low
+    quantizer.input_range.data = input_range
+
+    with torch.no_grad():
+        decompressor, q_weight = asym_fq_to_decompressor(
+            quantizer,
+            weight,
+        )
+
+    qdq_weight = (q_weight - ref_zero_point) * ref_scale
```
ljaljushkin, line 548: uint8 value - uint8 value likely will overflow. need a casting to float dtype

tests/torch/quantization/test_strip.py
```diff
@@ -409,3 +414,140 @@ def test_nncf_strip_lora_model(mode, decompressor_class, torch_dtype, atol, mock
...
+    # reference scale calculates with this formula:
+    # levels = (2 ** num_bits)
+    # level_high = levels - 1
+    # ref_scale = INPUT_RANGE_SAMPLE / level_high
+    ref_scale = torch.tensor(ref_scale)
+    ref_scale = ref_scale.expand(scale_shape).to(torch.float16)
+
+    # reference zero point calculates with this formula:
+    # level_low = 0
+    # ref_scale = level_low - round(INPUT_LOW_SAMPLE / ref_scale)
```
ljaljushkin, line 512:
```suggestion
    # ref_zero_point = level_low - round(INPUT_LOW_SAMPLE / ref_scale)
```

tests/torch/quantization/test_strip.py
ljaljushkin: Why not just call `decompressor` for `q_weight `?

tests/torch/quantization/test_strip.py
nikita-malininn: This is normal. For the provided numbers, in bfloat16, the scale is calculated lower than bf16.eps, so it was replaced.

tests/torch/quantization/test_strip.py
nikita-malininn: > uint8 value - uint8 value likely will overflow.

It will not since zero point value here and everywhere is 0.0.

> Why not just call decompressor for q_weight to get qdq_weight, instead of implementing its functional

You asked me to use a reference for scale and zero points here. How should it be finally?

ljaljushkin: ZP is not always zero. The formula is incorrect. I advise against using it in testing. Just call `decompressor`, its formula is correct and tested separately.  
 
Where is the contradiction with using reference points for scale and zeros? 
I meant the following:
```python
    with torch.no_grad():
        decompressor, q_weight = asym_fq_to_decompressor(
            quantizer,
            weight,
        )
        fq_weight = quantizer(weight)
    # qdq_weight = (q_weight - ref_zero_point) * ref_scale
    # qdq_weight = qdq_weight.to(torch_dtype)
    packed_tensor = decompressor.pack_weight(q_weight)
    ref_zero_point = decompressor.pack_weight(ref_zero_point)
    qdq_weight = decompressor(packed_tensor)

    assert torch.allclose(fq_weight, qdq_weight)
    assert torch.allclose(qdq_weight, weight)
    assert torch.allclose(decompressor._zero_point, ref_zero_point)
    assert torch.allclose(decompressor._scale, ref_scale)
```

nikita-malininn: As you wish. Done.

nncf/torch/quantization/strip.py
```diff
@@ -269,22 +272,24 @@ def sym_fq_to_decompressor(
...
     qdq_weight = quantizer.quantize(weight)
     if hasattr(quantizer, "_lspec"):
         # Reshape for group-wise quantization, implemented for classes with lora spec only
         qdq_weight = qdq_weight.reshape(quantizer._lspec.weight_shape)
-    qdq_weight = qdq_weight.to(weight_dtype)
-
-    integer_dtype = torch.int8
+    qdq_weight = qdq_weight.to(float_dtype)
 
     scale = quantizer.scale / abs(quantizer.level_low)
```
ljaljushkin, line 285:
```suggestion
    scale = quantizer.scale.to(float_dtype) / abs(quantizer.level_low)
```

ljaljushkin:
Passed manual checks on test example https://github.com/openvinotoolkit/nncf/pull/3386 and validation on wikitext.

ljaljushkin (approved):

PR #3421, opened 2025-04-09
disable tests for macos

AlexanderDokuchaev:


tests/openvino/native/quantization/test_weights_compression.py
```diff
@@ -1485,6 +1486,7 @@ def test_compression_with_transposed_activations(kwargs):
         )
 
 
+@pytest.mark.skipif(is_macos(), reason="Not relevant for MacOS")
```
nikita-savelyevv, line 1489: Could you please instead add an xfail and mention ticket 164135?

nikita-savelyevv (approved):

PR #3427, opened 2025-04-10
set xfail for 164135 

AlexanderDokuchaev:
### Changes

Add xfail for tests which starts fall after bump openvino 

### Reason for changes

https://github.com/openvinotoolkit/nncf/actions/runs/14381385720/job/40325786908

### Related tickets

164135


nikita-savelyevv (approved):
Thanks!

PR #3429, opened 2025-04-11
[ONNX] Add support for ONNX models larger than 2GB

andrey-churkin:
### Changes

The main objective of this PR is to extract tensors (`raw_data` attribute for initializer tensors) from the `onnx.ModelProto`  and store them within the `OnnxModel` wrapper. After the extraction, the `onnx.ModelProto` model will no longer contain the weights, resulting in a serialized form with a size of less than 2GB.

Extracted tensors can be explicitly passed into the inference session.

```python
sees_options = ort.SessionOptions()
sees_options.add_external_initializers(names, tensors)  # names: List[str], tensors: List[ort.OrtValue]
sess = ort.InferenceSession(serialized_model, sees_options, providers=["CPUExecutionProvider"])
```

### Reason for changes

Any proto in serialized form must be <2GB ([Total Size of the Message](https://protobuf.dev/programming-guides/proto-limits/))

To use `ort.InferenceSession()`, we call `model.SerializeToString()` to pass the model as bytes. However, for large models (greater than 2GB), the serialized form also exceeds 2GB, which is not supported. As a result, passing such a model to `ort.InferenceSession()` causes an error during inference.

### Related tickets

Ref: 164211

### DIFF
--- a/nncf/onnx/graph/model_utils.py
+++ b/nncf/onnx/graph/model_utils.py
@@ -8,12 +8,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 from collections import deque
 from typing import Dict
 
 import numpy as np
 import onnx
 from onnx.external_data_helper import _get_initializer_tensors
+from onnx.external_data_helper import set_external_data
 
 from nncf.common.factory import ModelTransformerFactory
 from nncf.common.graph.graph import NNCFGraph
@@ -78,6 +80,10 @@ def extract_raw_data_from_model(model: onnx.ModelProto) -> Dict[str, np.ndarray]
         # TODO(andrey-churkin): Probably, we should convert the NumPy array into an `ort.OrtValue`
         # here as follows: `OrtValue.ortvalue_from_numpy(numpy_tensor)`.
         data[tensor.name] = onnx.numpy_helper.to_array(tensor)
+
+        # We should call the `set_external_data()`` method here; otherwise, we will get an error during
+        # session creation because we can't replace a non-external initializer with external data.
+        set_external_data(tensor, location="foo.bin")
         tensor.ClearField("raw_data")
 
     return data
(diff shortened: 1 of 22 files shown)


andrey-churkin:
@alexsu52 @kshpv Before I begin fixing the tests, could you please take a moment to review and provide some quick feedback?

nncf/onnx/graph/model_utils.py
```diff
@@ -51,3 +56,101 @@ def remove_fq_from_inputs(model: onnx.ModelProto, nncf_graph: NNCFGraph) -> onnx
...
+    tensors = _get_initializer_tensors(model)
+    for tensor in tensors:
+        numpy_array = data.get(tensor.name, None)
+        if numpy_array is not None:
+            # TODO(andrey-churkin): Should we preserve the external data options here?
+            tensor_proto = onnx.numpy_helper.from_array(numpy_array, tensor.name)
+            tensor.CopyFrom(tensor_proto)
+
+
+class OnnxModel:
```
alexsu52, line 110:
```suggestion
class ONNXModel:
```

alexsu52, line 110: What do you think about moving `ONNXModel` in nncf/onnx/model.py?

andrey-churkin: That sounds good. I will do that.

nncf/onnx/quantization/quantize_model.py
```diff
@@ -80,10 +81,11 @@ def quantize_impl(
         model_type=model_type,
         advanced_parameters=advanced_parameters,
     )
-
-    graph = GraphConverter.create_nncf_graph(model)
+    onnx_model = ONNXModel.from_model(model)
```
alexsu52, line 84: Please check this for compress_weights

nncf/onnx/model.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Dict
+
+import numpy as np
+import onnx
+
+from nncf.onnx.graph.model_utils import extract_raw_data_from_model
+from nncf.onnx.graph.model_utils import insert_raw_data_into_model
+
+
+class ONNXModel:
```
kshpv, line 22: I have a general comment regarding the new class/approach:

When working with large ONNX models, a common issue is that all weights are loaded into memory, which can complicate debugging. To address this, I typically save model weights externally and work with the model topology. Before saving the compressed model, I use load_external_data_for_model().
I wonder if we could implement a similar strategy here, as the current approach seems inefficient in terms of memory consumption. This could potentially improve performance and make handling larger models more manageable.
Here's an example of how the code can be structured:
```
model = onnx.load_model(
    path,
    load_external_data=False,
)
model = compress_weights(model, CompressWeightsMode.INT8_ASYM)
load_external_data_for_model(model, base_dir)
onnx.save_model(
    model,
    "model.onnx",
    save_as_external_data=True,
    location="model.onnx.data",
    all_tensors_to_one_file=True,
    convert_attribute=True,
)

```

PR #3443, opened 2025-04-18
[ONNX] Add external_data_dir backend parameter

andrey-churkin:
### Changes

- Added the `BackendParameters.EXTERNAL_DATA_DIR` parameter for the ONNX backend. This parameter specifies the absolute path to the directory where the model’s external data files are stored. All external data files must be located in the same directory. It should be used when the model is loaded without external data using `onnx.load("model.onnx", load_external_data=False)`, and the external data files are not in the current working directory of the process. This parameter can be omitted if the external data files are located in the current working directory of the process.

- Added `check_model_protobuf_size()` and `check_external_data_location()` validation checks. After this PR, calling `nncf.quantize()` on an ONNX model with a protobuf size exceeding 2GB will raise an `nncf.ValidationError`. The `check_external_data_location()` method ensures that the external data files can be successfully located.

### Reason for changes

Any proto in serialized form must be <2GB ([Total Size of the Message](https://protobuf.dev/programming-guides/proto-limits/))

To use `ort.InferenceSession()`, we call `model.SerializeToString()` to pass the model as bytes. However, for large models (greater than 2GB), the serialized form also exceeds 2GB, which is not supported. As a result, passing such a model to `ort.InferenceSession()` causes an error during inference.

### Related tickets

Ref: 164211

### Tests

tests/onnx/models.py
tests/onnx/quantization/test_backend_parameters.py
tests/onnx/test_engine.py
tests/onnx/test_external_data.py
tests/onnx/test_model_metadata.py
tests/onnx/test_onnx_helper.py

### DIFF
--- a/nncf/onnx/quantization/quantize_model.py
+++ b/nncf/onnx/quantization/quantize_model.py
@@ -72,7 +72,7 @@ def check_model_protobuf_size(model: onnx.ModelProto) -> None:
         raise nncf.ValidationError(msg)
 
 
-def check_external_data_location(model: onnx.ModelProto, external_data_dir: Optional[str] = None) -> None:
+def check_external_data_location(model: onnx.ModelProto, external_data_dir: Optional[str] = None) -> Optional[str]:
     """
     Raises `nncf.ValidationError` if any referenced external data file does not exist, is not a regular file,
     or is a symlink.
@@ -80,6 +80,8 @@ def check_external_data_location(model: onnx.ModelProto, external_data_dir: Opti
     :param model: The ONNX model to validate.
     :param external_data_dir: Path to the directory where the external data files are expected to be located.
         If None, the current working directory is used.
+    :return: Path to the directory where external data files are located, if the model uses external data.
+        Returns None if no external data is used.
     """
     # If external_data_dir is not provided, we should test against the current working directory.
     external_data_dir = Path.cwd() if external_data_dir is None else Path(external_data_dir)
@@ -109,6 +111,9 @@ def check_external_data_location(model: onnx.ModelProto, external_data_dir: Opti
                 "but it doesn't exist or is not accessible."
             )
             raise nncf.ValidationError(msg)
+
+    # If len(data_path) == 0, it means there are no tensors that use external data.
+    return str(external_data_dir) if data_paths else None
 
 
 def quantize_impl(
@@ -147,7 +152,7 @@ def quantize_impl(
 
     check_model_protobuf_size(model)
     external_data_dir = get_external_data_dir(advanced_parameters)
-    check_external_data_location(model, external_data_dir)
+    external_data_dir = check_external_data_location(model, external_data_dir)
     if external_data_dir:
         set_metadata(model, MetadataKey.EXTERNAL_DATA_DIR, external_data_dir)
 
@@ -314,7 +319,7 @@ def compress_weights_impl(
         raise nncf.ValidationError(msg)
 
     external_data_dir = get_external_data_dir(advanced_parameters)
-    check_external_data_location(model, external_data_dir)
+    external_data_dir = check_external_data_location(model, external_data_dir)
     if external_data_dir:
         set_metadata(model, MetadataKey.EXTERNAL_DATA_DIR, external_data_dir)
 
--- a/nncf/onnx/quantization/quantize_model.py
+++ b/nncf/onnx/quantization/quantize_model.py
@@ -91,6 +91,7 @@ def check_external_data_location(model: onnx.ModelProto, external_data_dir: Opti
         )
         raise nncf.ValidationError(msg)
 
+    data_paths = set()
     for tensor in _get_all_tensors(model):
         if uses_external_data(tensor):
             info = ExternalDataInfo(tensor)
@@ -100,12 +101,14 @@ def check_external_data_location(model: onnx.ModelProto, external_data_dir: Opti
             # Source: https://onnx.ai/onnx/repo-docs/ExternalData.html
             external_data_file_name = Path(info.location).name  # Extract only the filename
             data_path = external_data_dir / external_data_file_name
-            if not data_path.exists() or not data_path.is_file() or data_path.is_symlink():
-                msg = (
-                    f"Data of TensorProto (tensor name: {tensor.name}) should be stored in {str(data_path)}, "
-                    "but it doesn't exist or is not accessible."
-                )
-                raise nncf.ValidationError(msg)
+            data_paths.add(data_path)
+    for data_path in data_paths:
+        if not data_path.exists() or not data_path.is_file() or data_path.is_symlink():
+            msg = (
+                f"Data of TensorProto (tensor name: {tensor.name}) should be stored in {str(data_path)}, "
+                "but it doesn't exist or is not accessible."
+            )
+            raise nncf.ValidationError(msg)
 
 
 def quantize_impl(
--- a/nncf/onnx/quantization/quantize_model.py
+++ b/nncf/onnx/quantization/quantize_model.py
@@ -16,6 +16,7 @@ from typing import Any, Callable, Iterable, Optional, TypeVar, Union
 import onnx
 from onnx.external_data_helper import ExternalDataInfo
 from onnx.external_data_helper import _get_all_tensors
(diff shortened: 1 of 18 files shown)


nncf/onnx/quantization/quantize_model.py
```diff
@@ -234,6 +239,11 @@ def compress_weights_impl(
     if model.opset_import[0].version < 21 and group_size > 0:
         msg = "ONNX models with opset version < 21 do not support block-wise quantization."
         raise nncf.ValidationError(msg)
+
+    external_data_dir = get_external_data_dir(advanced_parameters)
+    if external_data_dir:
+        model.metadata_props.add(key="nncf.external_data_dir", value=external_data_dir)
```
alexsu52, line 245: Will NNCF return a model with metadata_props `nncf.external_data_dir`?

andrey-churkin: This is a good question. I will check. We probably shouldn't do that.

Adding this parameter to `metadata_props` is just one possible solution (the simplest one that requires minimal changes). We could also pass such a parameter explicitly.

andrey-churkin: Fixed

andrey-churkin:
@alexsu52 Could you please take another look?

nncf/onnx/quantization/quantize_model.py
```diff
@@ -9,17 +9,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+from pathlib import Path
 from typing import Any, Callable, Iterable, Optional, TypeVar, Union
 
 import onnx
+from onnx.external_data_helper import ExternalDataInfo
+from onnx.external_data_helper import _get_all_tensors
```
alexsu52, line 18: Please, do not use private functions.

nncf/onnx/quantization/quantize_model.py
```diff
@@ -44,6 +54,60 @@
...
+    for tensor in _get_all_tensors(model):
+        if uses_external_data(tensor):
+            info = ExternalDataInfo(tensor)
+            # `info.location` field stores file path relative to the filesystem directory where
+            # the ONNX protobuf model was stored. Note, that up-directory path components such
+            # .. are disallowed and should be stripped when parsing.
+            # Source: https://onnx.ai/onnx/repo-docs/ExternalData.html
+            external_data_file_name = Path(info.location).name  # Extract only the filename
+            data_path = external_data_dir / external_data_file_name
+            if not data_path.exists() or not data_path.is_file() or data_path.is_symlink():
```
alexsu52, line 103: ONNX model often has one data file. You will check the same file for each tensor, am I right? If this statement is correct, this is not efficient approach.

nncf/onnx/quantization/quantize_model.py
```diff
@@ -44,6 +54,60 @@
...
 
 
+def check_model_protobuf_size(model: onnx.ModelProto) -> None:
+    """
+    Checks whether the serialized ONNX model exceeds the 2GB protobuf size limit.
+    If the size limit is exceeded, a `nncf.ValidationError` is raised.
+
+    :param model: The ONNX model to be checked.
+    """
+    MAXIMUM_PROTOBUF = 2000000000  # Limitation of single protobuf file is 2GB
```
alexsu52, line 64: Why do you use capital letters to create a constant in a function? Why do you use 2000000000  instead of 2^31?

nncf/onnx/quantization/quantize_model.py
```diff
@@ -44,6 +54,60 @@
...
+    protobuf_string = model.SerializeToString()
```
alexsu52, line 65: Will a copy of the model be created by `model.SerializeToString()`?

nncf/onnx/quantization/quantize_model.py
```diff
@@ -251,4 +332,11 @@ def compress_weights_impl(
         advanced_parameters,
     )
     graph = NNCFGraphFactory.create(model)
-    return compression_algorithm.apply(model, graph, dataset=dataset)
+
+    compressed_model = compression_algorithm.apply(model, graph, dataset=dataset)
+
+    if external_data_dir:
+        remove_metadata(compressed_model, MetadataKey.EXTERNAL_DATA_DIR)
```
alexsu52, line 339: Very fragile logic, since any algorithm inside the weight compression pipeline can make a copy.

alexsu52 (changes requested):
Please add tests.

nncf/onnx/quantization/quantize_model.py
andrey-churkin: Fixed

nncf/onnx/quantization/quantize_model.py
```diff
@@ -251,4 +332,11 @@ def compress_weights_impl(
...
+        load_external_data_for_model(compressed_model, external_data_dir)
```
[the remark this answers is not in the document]
andrey-churkin: Fixed

nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -185,9 +186,12 @@ def _check_arguments_for_transform_model(
         if compression_format != CompressionFormat.DQ:
             msg = "Compression format is not supported for the ONNX backend"
             raise nncf.ValidationError(msg)
-        if advanced_parameters != AdvancedCompressionParameters():
-            msg = "Advanced parameters are not supported for the ONNX backend"
-            raise nncf.ValidationError(msg)
+        default = AdvancedCompressionParameters()
+        for field in dataclasses.fields(advanced_parameters):
+            name = field.name
+            if name != "backend_params" and getattr(default, name) != getattr(advanced_parameters, name):
```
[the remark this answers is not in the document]
andrey-churkin: `statistics_path` is checked here

https://github.com/openvinotoolkit/nncf/blob/3f7d9e483cf518f75559c09161ea8ff39f875ac8/nncf/quantization/algorithms/weight_compression/algorithm.py#L147-L149

So I've removed those checks."

nncf/onnx/quantization/quantize_model.py
andrey-churkin: Yes, you are right. But ONNX doesn't provide an API to properly handle all tensors. So, it is a trade-off between private functionality and copying all that functionality into our repo. I'd be happy to copy it if you'd like.

nncf/onnx/quantization/quantize_model.py
andrey-churkin: I've followed the ONNX approach to handle it.

https://github.com/onnx/onnx/blob/main/onnx/checker.py#L176-L179

nncf/onnx/quantization/quantize_model.py
andrey-churkin: I guess so. There are no other solutions for resolving it (as far as I know). We can use `model.ByteSize(`) (which returns the serialized size of this message), but it also uses extra memory.

nncf/onnx/quantization/quantize_model.py
andrey-churkin: To resolve this, we can wrap the model in our custom class, but it requires significantly more changes to achieve the same result. Shall we discuss and choose a more optimal solution?

nncf/onnx/quantization/quantize_model.py
alexsu52: Pay attention on size of constant: https://github.com/onnx/onnx/blob/94e820742ea3f9716a62d486611e46c61355d7f5/onnx/checker.py#L47

nncf/onnx/quantization/quantize_model.py
alexsu52: Please add ```remove_metadata(model, MetadataKey.EXTERNAL_DATA_DIR)```

andrey-churkin: Done

nncf/onnx/quantization/quantize_model.py
andrey-churkin: Done

anzr299 (approved):

alexsu52 (approved):

PR #3451, opened 2025-04-22
Import backend modules in nncf/__init__.py

AlexanderDokuchaev:
### Changes

Import backend modules in nncf/__init__.py
```
import nncf
nncf.torch.load_from_config(...)
```

### Related tickets

164656

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/14603877559

### DIFF
--- a/nncf/torch/dynamic_graph/patch_pytorch.py
+++ b/nncf/torch/dynamic_graph/patch_pytorch.py
@@ -283,6 +283,8 @@ class OriginalOpInfo:
 ORIGINAL_OPERATORS: List[OriginalOpInfo] = []
 ORIGINAL_CALL = torch.nn.Module.__call__
 _ORIG_JIT_SCRIPT = None
+_ORIG_JIT_SCRIPT_IF_TRACE = None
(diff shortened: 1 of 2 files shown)


nncf/__init__.py
```diff
@@ -90,3 +90,15 @@
         "Please install one of the supported frameworks above in order to use NNCF on top of it.\n"
         "See the installation guide at https://github.com/openvinotoolkit/nncf#installation-guide for help."
     )
+
+if _AVAILABLE_FRAMEWORKS["torch"]:
+    from nncf import torch as torch
```
alexsu52, line 95: Do PyTorch extensions compile during this call?

AlexanderDokuchaev: No, extensions compiles on first using

alexsu52 (changes requested):
{A78B60FA-38A9-4570-9937-C2490BD27858}
I tried using nncf from this PR and got lots of warnings in my environment, but if I use develop, nncf is silent.

PR #3455, opened 2025-04-23
Bump PyTorch to 2.7.0

AlexanderDokuchaev:
### Changes

Update pytorch to 2.7.0
Regenerate ref data for test_generate_text_data_functional, changed output of `hf-internal-testing/tiny-random-gpt2`
Bump cuda version to 12.6

### Tests
precommits:
 - test_compilation_wrapper - Issue-166894

manual/job/post_training_quantization/665/ - fail FX on save compressed models (issue - 166847)
nightly/job/TriggerBetta/1032/ - pass
[examples](https://github.com/openvinotoolkit/nncf/actions/runs/14714621634) : 
 - llm_compression_qat_with_lora assert 0.034 == 0.027 ± 2.0e-03
 - llm_compression_synthetic - AssertionError: metric word_count: 81 != 83

[wc](https://github.com/openvinotoolkit/nncf/actions/runs/14711846338) - pass 
[Test_Install](https://github.com/openvinotoolkit/nncf/actions/runs/14648657133) - pass


ljaljushkin:
It appears that F.linear() is the source of indeterminism.

When model is on the `CUDA`, the code below
```
model = AutoModelForCausalLM.from_pretrained(BASE_TEST_MODEL_ID, device_map="cuda")
torch.use_deterministic_algorithms(True)
```

leads to an error:

> ../../env/nncf-py/lib/python3.10/site-packages/torch/nn/modules/linear.py:125: in forward
>     return F.linear(input, self.weight, self.bias)
> 
> RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#results-reproducibility
> 

Then, with `export CUBLAS_WORKSPACE_CONFIG = :4096 :8`, the results are the same with torch 2.6.0 and 2.7.0

The same thing may happen with qat-lora sample on CPU. 
When I switched to GPU, wwb reference were different:
image
`use_deterministic_algorithms` + `CUBLAS_WORKSPACE_CONFIG` aligns it.

Propose updating references, since with `use_deterministic_algorithms` the test would be slower, but the results were the same.

anzr299 (approved):

daniil-lyakhov (approved):

PR #3459, opened 2025-04-24
[PTQ Conformance][FX] Fix Torch Compile OV Model Caching

anzr299:
### Changes

Save Torch Compile using torch compile with `aot_autograd` tracing.

### Reason for changes

To have OV model cached with torch.compile and `aot_autograd` tracing mechanism similar to the torch.compile used in inference.

### Tests
-- | -- | -- | -- | --
PTQ | PTQ | Running | 663 |


daniil-lyakhov:
@anzr299, does it relate to #3455 ?

PR #3469, opened 2025-04-29
Added logging for compilation trigger

nikita-malininn:
### Changes

- As stated in the title. It is likely that Python side-effects like printing a message would not be compiled with `torch.compile` and would trigger `is_compiling`.

### Reason for changes

- Fix for Torch 2.7.0.

### Related tickets

- 166894

### Tests

- N/A

### DIFF
--- a/tests/torch/test_utils.py
+++ b/tests/torch/test_utils.py
@@ -14,6 +14,7 @@ import torch
 from torch import nn
 
 import nncf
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev:
Still falls with torch 2.7.0

AlexanderDokuchaev (approved):

PR #3472, opened 2025-04-30
[FX] move test to tests/torch2 

AlexanderDokuchaev:
### Changes

Move fx tests to torch2 dir

### Reason for changes

To avoid disable patching


daniil-lyakhov (approved):
Looks like tracing affected the X86Quantizer with the Swin_v2 model. I've checked the reference quantization (via torch.ao) against the nncf quantization and confirm that the nncf quantization is correct

P.S. Please fix the pre-commit

PR #3479, opened 2025-05-06
[BC] Use `atleast_1d` to enable models with no batch dimention

daniil-lyakhov:
### Changes

* nncf.Tensor is updated to use appropriate method in the iterator instead of the `len()` function: it does not defined for the scalars in the numpy, so it is replaced with `.shape[0]` if it is exist
* BC is updated to call `atleast_1d` function while collecting fp_outputs: the extend method with a scalar tensor does not add anything to the output list, `atleast_1d` fixes that
* Shape reducer is returning a tuple instead of a nncf.Tensor of a tuple: nncf.Tensor does not expect to wrap a tuple thus fails to work properly (could not get shape of a tuple, for example)

### Reason for changes

#3395 

### Tests

* test_bias_correction is updated with a case with a scalar activation tensor
* torch.Tensor iterator and the `atleast_1d` function are tested in the torch tensor template test


tests/cross_fw/test_templates/test_bias_correction.py
```diff
@@ -143,9 +144,14 @@ def quantized_test_model(self, tmpdir) -> TModel:
...
             (DepthwiseConvTestModel, {"/conv/Conv": [-1.1229, -0.1863]}),
             (TransposeConvTestModel, {"/conv/ConvTranspose": [0.66797173, -0.7070703]}),
+            (OneDimMM, {"/linear/MatMul": [0.95773065, 1.3218939, 0.81694865]}),
         ),
     )
     def test_update_bias(self, model_cls, ref_biases, tmpdir):
+        from nncf.quantization.algorithms.bias_correction.onnx_backend import ONNXBiasCorrectionAlgoBackend
+
+        if self.get_backend() is ONNXBiasCorrectionAlgoBackend and model_cls is OneDimMM:
+            pytest.skip("ONNX does not support BC with MM ops")
```
AlexanderDokuchaev, line 154: Please check backend specific parts not in template, there is no guaranty that backend can be imported here

nncf/tensor/tensor.py
```diff
@@ -65,7 +65,7 @@ def __bool__(self) -> bool:
         return bool(self.data)
 
     def __iter__(self) -> Iterator[Tensor]:
-        return TensorIterator(self.data)
+        return TensorIterator(self)
```
AlexanderDokuchaev, line 68: Could you check if use nncf.Tensor without TensorIterator, 
Looks like it have same behavior 
```python
    def __iter__(self) -> Iterator[Tensor]:
        if not self.shape:
            raise TypeError("iteration over a 0-d tensor")
        return iter(self)
```

AlexanderDokuchaev (changes requested):

nncf/tensor/tensor.py
```diff
@@ -227,10 +227,11 @@ def __init__(self, tensor: Tensor) -> None:
         self._index = 0
 
     def __next__(self) -> Tensor:
-        if self._index < len(self._tensor):
+        tensor_shape = self._tensor.shape
+        if tensor_shape and self._index < tensor_shape[0]:
```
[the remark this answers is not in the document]
daniil-lyakhov: Good catch, done

AlexanderDokuchaev (approved):

PR #3487, opened 2025-05-08
[TorchFX] `quantize_pt2e` custom quantizers support

daniil-lyakhov:
### Changes

* TorchAOAdapter is updated with new entity: the `ExtendedQuantizerSetup`. It contains additional info about dtypes to use in q->dq pairs
* TorchFX MinMax algo backend migrated from restrained stip procedure to flexible custom quantization parameters assignment code

 
### Reason for changes

To fully support quantization via `quantize_pt2e` with custom quantizers (like `XNNPACKQuantizer`)

### Related tickets

#3231 

### Tests

* flexible custom quantization parameters assignment is tested by `tests/torch/fx/test_calculation_quantizer_params.py`
* Conformance test is updated with 2 new configurations:
`OV_QUANTIZER_NNCF`, `OV_QUANTIZER_AO`
* conformance: post_training_quantization/682/


nncf/experimental/quantization/quantizer.py
```diff
@@ -11,14 +11,95 @@
...
+class IntDtype(Enum):
+    """
+    Enum of possible integer types.
+    """
+
+    INT8 = "INT8"
+    UINT8 = "UINT8"
+
+
+class ExtendedQuantizerSetup(ABC, SingleConfigQuantizerSetup):
```
alexsu52: This change requires offline discussion.

alexsu52 (changes requested):

nncf/experimental/quantization/quantizer.py
daniil-lyakhov: Discussed offline, `ExtendedQuantizerSetup` is replaced with `ExtendedQuantizerConfig`

nncf/common/quantization/quantizer_setup.py
```diff
@@ -193,9 +194,15 @@ def from_state(cls, state: dict[str, Any]) -> "SingleConfigQuantizationPoint":
         insertion_point_cls_name = state[cls._state_names.INSERTION_POINT_CLASS_NAME]
         insertion_point_cls = CommonStatefulClassesRegistry.get_registered_class(insertion_point_cls_name)
         insertion_point = insertion_point_cls.from_state(state[cls._state_names.INSERTION_POINT])  # type: ignore
+        qconfig_state = state[cls._state_names.QCONFIG]
+        if QuantizerConfig().__dict__.keys() == qconfig_state.keys():
```
AlexanderDokuchaev: There is any tests to restore ExtendedQuantizerConfig, is anywhere used?
If no, it can raise error, or looks like `get_state` should save type of config

nncf/experimental/quantization/structs.py
```diff
@@ -0,0 +1,51 @@
...
+class IntDtype(StrEnum):
+    """
+    Enum of possible integer types.
+    """
+
+    INT8 = "INT8"
+    UINT8 = "UINT8"
+
+
+class ExtendedQuantizerConfig(QuantizerConfig):
```
AlexanderDokuchaev: Docstring

nncf/experimental/quantization/structs.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(
+        self,
+        num_bits: int = QUANTIZATION_BITS,
+        mode: QuantizationScheme = QuantizationScheme.SYMMETRIC,
+        signedness_to_force: Optional[bool] = None,
+        per_channel: bool = QUANTIZATION_PER_CHANNEL,
+        narrow_range: bool = QUANTIZATION_NARROW_RANGE,
+        dest_dtype: IntDtype = IntDtype.INT8,
```
AlexanderDokuchaev:
```suggestion
        dest_dtype: Literal[TensorDataType.int8, TensorDataType.uint8] = TensorDataType.int8,
```
Is it possible to use TensorDataType?

tests/torch2/fx/test_quantizer.py
```diff
@@ -156,26 +168,98 @@ def test_quantized_model(
...
+)
+def test_quantizer_setup(
+    quantizer_builder: Callable[[tuple[Any, ...]], Quantizer],
+    model_case: ModelCase,
+    quantizer_params,
+):
+    fx_model, _ = _build_torch_fx_model(model_case)
+    quantizer = quantizer_builder(**quantizer_params)
+    ref_qconfig_filename = (
+        FX_QUANTIZED_DIR_NAME / str(quantizer.__class__.__name__) / get_qconf_filename(model_case.model_id)
```
AlexanderDokuchaev: quantizer.__class__.__name__ is already str

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -179,49 +179,87 @@ def get_weight_config(config: QuantizerConfig, model: NNCFNetwork) -> QuantizerC
...
+        else:
+            observer = torch.ao.quantization.observer.MinMaxObserver
+
+        levels = 256
+        if dtype is IntDtype.INT8:
+            level_high = (levels // 2) - 1
+            level_low = -(levels // 2)
+        else:
+            level_high = levels - 1
+            level_low = 0
```
AlexanderDokuchaev:
```suggestion
        if dtype is IntDtype.INT8:
            level_high = 127
            level_low = -128
        else:
            level_high = 255
            level_low = 0
```

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -179,49 +179,87 @@ def get_weight_config(config: QuantizerConfig, model: NNCFNetwork) -> QuantizerC
...
+        scale = scale.view(-1)
+        zero_point = zero_point.view(-1)
+
+        fakequantizer = FakeQuantize(
+            observer=observer,
+            quant_max=level_high,
+            quant_min=level_low,
+            dtype=torch.qint8 if dtype is IntDtype.INT8 else torch.quint8,
+            qscheme=qscheme,
+            eps=eps,
```
AlexanderDokuchaev:
```suggestion
            eps=1e-16,
```

nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -179,49 +179,87 @@ def get_weight_config(config: QuantizerConfig, model: NNCFNetwork) -> QuantizerC
...
+        eps = 1e-16
+        if dtype is None:
+            if quantizer_config.mode != QuantizationScheme.SYMMETRIC:
+                dtype = IntDtype.UINT8
+            else:
+                dtype = (
+                    IntDtype.INT8
+                    if quantizer_config.signedness_to_force or torch.any(parameters.input_low.data < 0.0)
+                    else IntDtype.UINT8
+                )
```
AlexanderDokuchaev:
```suggestion
        if isinstance(quantizer_config, ExtendedQuantizerConfig):
            dtype = quantizer_config.dest_dtype
        elif quantizer_config.mode != QuantizationScheme.SYMMETRIC:
            dtype = IntDtype.UINT8
        else:
            dtype = (
                IntDtype.INT8
                if quantizer_config.signedness_to_force or torch.any(parameters.input_low.data < 0.0)
                else IntDtype.UINT8
            )
```

AlexanderDokuchaev (changes requested):

nncf/common/quantization/quantizer_setup.py
daniil-lyakhov: Yes, there is a test for that https://github.com/openvinotoolkit/nncf/pull/3487/files#diff-b18aaa822e7f8fcaa85c4aa291d7c452e8720f58c2c8290dbe37f22e6ad987bfR98-R101
I'm not sure we should bloat nncf by one more registry as we are working on the flexible quantizer api (ticket 170078)

nncf/experimental/quantization/structs.py
daniil-lyakhov: Done

nncf/experimental/quantization/structs.py
daniil-lyakhov: Yes, done

nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: Done

nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: Done

nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: Done

tests/torch2/fx/test_quantizer.py
daniil-lyakhov: Done

daniil-lyakhov:
X86Quantizer validation is disabled by default in the conformance test.
Test: job/post_training_quantization/682/

src/nncf/tensor/definitions.py
```diff
@@ -31,7 +33,7 @@ class TensorBackend(Enum):
     ov = auto()
 
 
-class TensorDataType(Enum):
+class TensorDataType(StrEnum):
```
AlexanderDokuchaev, line 36: auto() add integer values for enum members, starts from 0.
As result if will added new members, serialization objects in different version of nncf can be incorect. Because one value will be associated with different datatype.

In this case auto() can be used, as it.
Need to define each value manually, as in nncf/parameters.py.
Or as option define `_generate_next_value_` to generate values automatically from name (it can be defined in StrEnum)
https://docs.python.org/3/library/enum.html#enum.Enum._generate_next_value_
```
    @staticmethod
    def _generate_next_value_(name, start, count, last_values):
        return name.lower()
```

tests/post_training/pipelines/image_classification_base.py
```diff
@@ -103,3 +127,98 @@ def _validate(self) -> None:
...
+        pt2e_kwargs["smooth_quant_params"] = sq_params
+        pt2e_kwargs["bias_correction_params"] = advanced_parameters.bias_correction_params
+        pt2e_kwargs["activations_range_estimator_params"] = advanced_parameters.activations_range_estimator_params
+        pt2e_kwargs["weights_range_estimator_params"] = advanced_parameters.weights_range_estimator_params
+
+        smooth_quant = False
+        if self.compression_params.get("model_type", False):
+            smooth_quant = self.compression_params["model_type"] == nncf.ModelType.TRANSFORMER
+
+        with disable_patching(), torch.no_grad():
```
AlexanderDokuchaev: no need to use disable_patching()

tests/torch2/conftest.py
```diff
@@ -50,6 +50,7 @@ def pytest_configure(config: Config) -> None:
     regen_dot = config.getoption("--regen-ref-data", False)
     if regen_dot:
         os.environ["NNCF_TEST_REGEN_DOT"] = "1"
+        os.environ["NNCF_TEST_REGEN_JSON"] = "1"
```
AlexanderDokuchaev: please use regen_ref_data fixture instead to adding new env variables

tests/post_training/data/ptq_reference_data.yaml
```diff
@@ -82,6 +98,14 @@ torchvision/vit_b_16_backend_CUDA_FX_TORCH:
     type: "FileNotFoundError"
     error_message: "Openvino Model Files Not Found!"
     message: "Issue-166847"
+torchvision/vit_b_16_backend_OV_QUANTIZER_NNCF:
+  metric_value: 0.80922
+torchvision/vit_b_16_backend_OV_QUANTIZER_AO:
+  metric_value: 0.2429
+torchvision/vit_b_16_backend_X86_QUANTIZER_NNCF:
+  metric_value: 0.80922
+torchvision/vit_b_16_backend_X86_QUANTIZER_AO:
```
AlexanderDokuchaev: we are really need to add tests cases that got low accuracy and runs several days
post_training_quantization/681/ - still in progress after 6 days

src/nncf/common/quantization/quantizer_setup.py
```diff
@@ -193,9 +194,19 @@ def from_state(cls, state: dict[str, Any]) -> "SingleConfigQuantizationPoint":
         insertion_point_cls_name = state[cls._state_names.INSERTION_POINT_CLASS_NAME]
         insertion_point_cls = CommonStatefulClassesRegistry.get_registered_class(insertion_point_cls_name)
         insertion_point = insertion_point_cls.from_state(state[cls._state_names.INSERTION_POINT])  # type: ignore
+        qconfig_state = state[cls._state_names.QCONFIG]
+        # Need to instantiate ExtendedQuantizerConfig
+        # to support additional fields used by ExecuTorch-specific quantizer configs.
+        # TODO (dlyakhov): Refactor and generalize quantizer config deserialization to cleanly handle both
+        # standard and extended formats without relying on manual key comparison (ticket 170078).
+        if QuantizerConfig().__dict__.keys() == qconfig_state.keys():
```
AlexanderDokuchaev, line 202: We got common class CommonStatefulClassesRegistry, can it be reused for it, without adding new registry?

tests/common/experimental/test_structs.py
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+from nncf.experimental.quantization.structs import ExtendedQuantizerConfig
+from nncf.tensor.definitions import TensorDataType
+
+
+def test_extended_q_config_non_supported_dest_dtype():
```
AlexanderDokuchaev: If you adding this tests, you should check all TensorDataType,
```
    for dtype in list(TensorDataType) + [None]:
        if dtype not in [TensorDataType.int8, TensorDataType.uint8]:
            with pytest.raises(RuntimeError):
                ExtendedQuantizerConfig(dest_dtype=TensorDataType)
```
And please use exceptions from nncf.error

tests/post_training/data/ptq_reference_data.yaml
daniil-lyakhov: They are disabled by default, but the pipepline is present. I would keep it for testing purposes, they don't run without a key anyway

src/nncf/tensor/definitions.py
daniil-lyakhov: ~~Let's get IntTypes back then maybe? And make a Good First Issue to do this fun exercise outside of this PR~~

src/nncf/common/quantization/quantizer_setup.py
daniil-lyakhov: We can try, you are right. Can we do it in a separate PR please? I made a separate ticket for that https://github.com/openvinotoolkit/nncf/issues/3574

tests/post_training/pipelines/image_classification_base.py
daniil-lyakhov: Removed

tests/torch2/conftest.py
daniil-lyakhov: removed

tests/common/experimental/test_structs.py
daniil-lyakhov: The test structure you are suggesting is more complicated than the code we are testing. I've manually added all possible dtypes to check all the possibilities. Hope it will work

daniil-lyakhov:
@AlexanderDokuchaev, I hope we can merge this PR without the refactoring of TensorDataType and a registry for the new ExtendedQuantizerConfig class. Please let me know ASAP if it is not possible

tests/torch2/conftest.py
AlexanderDokuchaev: But it still used in tests

daniil-lyakhov: Ok, I don't understand what I need to do, please elaborate. In the first revision I was using `regen_ref_data` and was setting both regen dot and regen json

src/nncf/tensor/definitions.py
daniil-lyakhov: Done

tests/torch2/conftest.py
AlexanderDokuchaev: Dont use any extra env variables, just use only `regen_ref_data` fixture.
Not found revision where used `regen_ref_data`

AlexanderDokuchaev: Just replace 
```
def test_quantizer_setup(
    quantizer_builder: Callable[[tuple[Any, ...]], Quantizer],
    model_case: ModelCase,
    quantizer_params,
):
 ....
    if os.getenv("NNCF_TEST_REGEN_JSON") is not None:
        with safe_open(ref_qconfig_filename, "w") as file:
            json.dump(qsetup_config, file, indent=4)
```
to
```
def test_quantizer_setup(
    quantizer_builder: Callable[[tuple[Any, ...]], Quantizer],
    model_case: ModelCase,
    quantizer_params,
    regen_ref_data: bool
):
 ....
    if regen_ref_data:
        with safe_open(ref_qconfig_filename, "w") as file:
            json.dump(qsetup_config, file, indent=4)
```

src/nncf/tensor/definitions.py
AlexanderDokuchaev: Please add _generate_next_value_ to StrEnum, to get behavior like in enum.StrEnum in python>=3.11

src/nncf/experimental/quantization/structs.py
```diff
@@ -0,0 +1,63 @@
...
+        :param signedness_to_force: True if the quantizer *must* be signed, False if *must* be unsigned,
+            None if the signed/unsigned attribute should be determined based on the incoming activation
+            statistics during range initialization.
+        :param per_channel: True for per-channel quantization, False for per-tensor.
+        :param narrow_range: True if the range of quantized values should be narrowed as compared to the
+            naive case, False if all 2^`num_bits` quantizations should be used.
+        :param dest_dtype: Target integer data type for quantized values.
+        """
+        super().__init__(num_bits, mode, signedness_to_force, per_channel, narrow_range)
+        if dest_dtype not in [TensorDataType.int8, TensorDataType.uint8]:
```
AlexanderDokuchaev: General question, why ExtendedQuantizerConfig responsible to check type?
It's responsibilities of algorithms to use only valid type, it's just structure that in general case can contain any type.
Suggest to remove this check, check type in algorithms if need, and remove test_structs.py

tests/post_training/pipelines/image_classification_base.py
```diff
@@ -12,16 +12,29 @@
 import copy
 import os
 
+os.environ["TORCHINDUCTOR_FREEZING"] = "1"
```
AlexanderDokuchaev, line 18: Please add comment why it need and why it defined before another imports

tests/post_training/data/ptq_reference_data.yaml
AlexanderDokuchaev: I will not merge tests that are running for several days, even if they are disabled by default.
post_training_quantization/681/ - still in progress after 7 days

What a reason that is takes long time? 
Did you check that there is no any bugs?

tests/torch2/conftest.py
daniil-lyakhov: But half of the tests are using NNCF_TEST_REGEN_DOT
image
Should I really do it in this pr? Or is it better to update all the tests at once?

tests/post_training/data/ptq_reference_data.yaml
daniil-lyakhov: Test number is 682, please

I disabled them here https://github.com/openvinotoolkit/nncf/pull/3487/commits/e227f738ba8f8713d75b320a57249127a2ea244e#diff-a6044884bf87069635cb96de0928ebd1c2d0a1a849693480684139737e35420f

src/nncf/tensor/definitions.py
daniil-lyakhov: ~~I bet it will fail other things in other places after that, let's see~~

src/nncf/experimental/quantization/structs.py
daniil-lyakhov: Because other types are not supported, no? I don't want to check types in all other places, I want to check them at the beginning. Config with dest_type bfloat16 or int4 does not make any sense to have right now

tests/post_training/pipelines/image_classification_base.py
daniil-lyakhov: Done

tests/torch2/conftest.py
AlexanderDokuchaev: Just replace it in new tests from this PR

daniil-lyakhov: Done, please take a look

src/nncf/experimental/quantization/structs.py
AlexanderDokuchaev: Validation type is responsibility of algorithms, not a structure that just contains parameters.
Config will be valid with all dest_type, but quantization algorithm supports only uint8 and int8

tests/post_training/data/ptq_reference_data.yaml
AlexanderDokuchaev: I found the issue: validation with using PyTorch runs on all available CPU cores by default, but there's a limit set in the CI environment. As a result, validation runs slower due to CPU throttling.
Tried to fix it by set TORCH_NUM_THREADS env variable in CI.
post_training_quantization/685/

AlexanderDokuchaev (changes requested):

tests/post_training/data/ptq_reference_data.yaml
daniil-lyakhov: Performance of X86Quantizer quantization is out of scope of this PR. I've removed everything related to it from the conformance test and stopped your job. Job post_training_quantization/686/ is running

src/nncf/experimental/quantization/structs.py
daniil-lyakhov: Done

src/nncf/experimental/quantization/structs.py
```diff
@@ -0,0 +1,59 @@
...
+from nncf.common.quantization.structs import QuantizerConfig
+from nncf.config.schemata.defaults import QUANTIZATION_BITS
+from nncf.config.schemata.defaults import QUANTIZATION_NARROW_RANGE
+from nncf.config.schemata.defaults import QUANTIZATION_PER_CHANNEL
+from nncf.tensor.definitions import TensorDataType
+
+IntDtype = Literal[TensorDataType.int8, TensorDataType.uint8]
+
+
+class ExtendedQuantizerConfig(QuantizerConfig):
```
alexsu52: What do you think?
```suggestion
class TypedQuantizerConfig(QuantizerConfig):
```

src/nncf/experimental/quantization/structs.py
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2025 Intel Corporation
```
alexsu52: I believe you can introduce this class in the `nncf/common/quantization/structs.py`.

daniil-lyakhov: Great idea, done

src/nncf/experimental/quantization/structs.py
daniil-lyakhov: Done

src/nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -179,49 +179,84 @@ def get_weight_config(config: QuantizerConfig, model: NNCFNetwork) -> QuantizerC
...
-        quantizer = quantizer_cls(quantizer_spec)
 
-        # Fill it with minmax
-        # TODO(dlyakhov) Prevent creation of intermediate objects like nncf quantizer.
-        FXMinMaxAlgoBackend._fill_quantizer_parameters(quantizer, parameters, quantizer_spec.scale_shape)
-        # Convert to the torch fake quantizer
-        torch_fq = convert_to_torch_fakequantizer(quantizer)
-        return torch_fq
+        fakequantizer.scale = scale
+        fakequantizer.ch_axis = channel_axis
```
alexsu52: `ch_axis` is used only for `per_channel` quantization scheme. Do you assign `ch_axis` for any quantization schemes?

src/nncf/quantization/algorithms/min_max/torch_fx_backend.py
```diff
@@ -179,49 +179,84 @@ def get_weight_config(config: QuantizerConfig, model: NNCFNetwork) -> QuantizerC
...
 
     @staticmethod
-    def _get_input_scale_shape(
-        nncf_graph: NNCFGraph, target_point: PTTargetPoint, per_channel: bool
-    ) -> tuple[tuple[int, ...], tuple[int, ...], int]:
-        is_weights = target_point.is_weight_target_point()
-        if is_weights:
+    def _get_channel_axis(target_point: PTTargetPoint, per_channel: bool) -> int:
+        if not per_channel:
+            return 0
```
alexsu52: Could you explain the meaning of this condition?

src/nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: It is always assigned despite FQ being per-tensor. This logic came from the torch strip script https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/torch/quantization/strip.py#L134

Fixed, now the ch_axis is only set to per-channel quantizers

src/nncf/quantization/algorithms/min_max/torch_fx_backend.py
daniil-lyakhov: That's strange, agree, fixed

PR #3490, opened 2025-05-14
use stateful model in the tiny_llama example

alexsu52:
### Changes

Remove  `stateful=False` from `OVModelForCausalLM.from_pretrained(MODEL_ID, export=True, load_in_8bit=False, compile=False)`

### Reason for changes

Use default behavior in the example

### Related tickets

N/A

### Tests

TBD

### DIFF
--- a/examples/llm_compression/openvino/tiny_llama/main.py
+++ b/examples/llm_compression/openvino/tiny_llama/main.py
@@ -12,7 +12,6 @@ import time
 from functools import partial
 
 import numpy as np
-import openvino as ov
 from datasets import load_dataset
 from optimum.intel.openvino import OVModelForCausalLM
 from transformers import AutoTokenizer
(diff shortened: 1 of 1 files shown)


examples/llm_compression/openvino/tiny_llama/main.py
```diff
@@ -27,9 +26,9 @@ def main():
     dataset = load_dataset("wikitext", "wikitext-2-raw-v1", split="test")
 
     tokenizer = AutoTokenizer.from_pretrained(MODEL_ID)
-    model = OVModelForCausalLM.from_pretrained(MODEL_ID, export=True, load_in_8bit=False, compile=False, stateful=False)
+    model = OVModelForCausalLM.from_pretrained(MODEL_ID, export=True, load_in_8bit=False, compile=False)
```
ljaljushkin, line 29: There are 3 more samples using it and conformance test

alexsu52: GFI for examples https://github.com/openvinotoolkit/nncf/issues/3491. Taking into account that stateless and stateful models have to produce the same accuracy and I would suggest to postpone changes in the conformance test.

ljaljushkin (approved):

PR #3493, opened 2025-05-14
Aligned bf16 tuning vs f32 inference for 4bit compression

ljaljushkin:
### Changes

Always cast input to float32 inside FQ + LoRA.

Benchmark results with new schema on the https://github.com/ljaljushkin/nncf_pytorch/tree/nl/ref_benchmark with small modifications from @nikita-malininn's branch https://github.com/nikita-malininn/nncf/tree/nm/ref_benchmark:  

device | dtype | exec_type | tensor_type | granularity | symmetric | narrow_range | timing_mode | num_runs | input_size
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
cuda | bfloat16 | ExecutionType.REGULAR | TensorType.WEIGHTS | GranularityType.PER_CHANNEL | TRUE | FALSE | TimingMode.KERNEL | 1000 | [2048, 128256]

name | Mode | forward_avg, ms | backward_avg, ms | memory, Gb
-- | -- | -- | -- | --
compile (PR) | sym | 6.5 | 10.7 | 3.9
compile (PR) | asym | 6.8 | 10.7 | 3.9
compile (before) | sym | 1.6 | 9.5 | 4.2
compile (before) | asym | 1.9 | 9.5 | 3.9
not compiled (PR) | sym | 19.0 | 46.6 | 5.9
not compiled (PR) | asym | 19.6 | 47.0 | 5.9
not compiled (before)  | sym | 9.2 | 37.0 | 5.4
not compiled (before) | asym | 9.6 | 37.0 | 5.4

There's an overhead on forward, but it's leveled up by using torch.compile.

There's a 1-6% overhead on RTX per epoch, and on A100, depending on the setup, there can even be a boost of 6% or a slowdown of 3%.
image
image

### Reason for changes

Minimize the disparity in precision between the Torch model and its exported OV equivalent.
The full alignment would be very inefficient, so here's a compromise: align accuracy with minimal overhead on the forward pass.

e2e test on `facebook/opt-125m` proves that output is the same now within default absolute tolerance (1e-8) instead of 1e-2 one:
https://github.com/openvinotoolkit/nncf/pull/3493/files#diff-7a4f90fe4f07d515df355d6fb618112d7d3fe88eb8ba777e502c695a7c715010R170

Previously, there were 3 problematic models with significant difference in accuracy, now it's much more aligned:
image

### Related tickets

166195

### Tests

test examples - https://github.com/openvinotoolkit/nncf/actions/runs/15024278726/job/42221028011

### DIFF
--- a/nncf/torch/quantization/quantize_functions.py
+++ b/nncf/torch/quantization/quantize_functions.py
@@ -126,15 +126,11 @@ class QuantizeSymmetricTorch(torch.autograd.Function):
         input_low = torch.where(scale > 0, -scale, -scale / level_low * level_high)
         # 15/8 * scale or (2-1/8) * scale
         input_range = torch.abs((2 + 1 / level_low) * scale)
-
-        if input_.dtype in [torch.bfloat16, torch.float16]:
-            input_low = input_low.type(input_.dtype)
-            input_range = input_range.type(input_.dtype)
-
+        dtype = input_.dtype
         original_shape = input_.shape
         input_ = input_.reshape(input_shape)
 
-        output = RQ.Quantize_forward(input_, input_low, input_range, levels)
+        output = RQ.Quantize_forward(input_.type(torch.float32), input_low, input_range, levels)
 
         ctx.save_for_backward(input_, input_low, input_range)
         ctx.level_low = level_low
@@ -142,7 +138,7 @@ class QuantizeSymmetricTorch(torch.autograd.Function):
         ctx.levels = levels
 
         output = output.reshape(original_shape)
-        return output
+        return output.type(dtype)
 
     @staticmethod
     def backward(ctx, grad_output):
@@ -168,14 +164,11 @@ class QuantizeSymmetricTorch(torch.autograd.Function):
 class QuantizeAsymmetricTorch(torch.autograd.Function):
     @staticmethod
     def forward(ctx, input_, input_shape, input_low, input_range, level_low, level_high, levels):
-        if input_.dtype in [torch.bfloat16, torch.float16]:
-            input_low = input_low.type(input_.dtype)
-            input_range = input_range.type(input_.dtype)
-
+        dtype = input_.dtype
         original_shape = input_.shape
         input_ = input_.reshape(input_shape)
 
-        output = RQ.Quantize_forward(input_, input_low, input_range, levels)
+        output = RQ.Quantize_forward(input_.type(torch.float32), input_low, input_range, levels)
 
         # Save tensors for backward pass
         ctx.save_for_backward(input_, input_low, input_range)
@@ -184,7 +177,7 @@ class QuantizeAsymmetricTorch(torch.autograd.Function):
         ctx.levels = levels
 
         output = output.reshape(original_shape)
-        return output
+        return output.type(dtype)
 
     @staticmethod
     def backward(ctx, grad_output):
--- a/tests/torch2/function_hook/quantization/strip/test_strip_dequantize.py
+++ b/tests/torch2/function_hook/quantization/strip/test_strip_dequantize.py
@@ -12,19 +12,24 @@
 from dataclasses import dataclass
 from typing import Any
 
+import openvino as ov
 import pytest
 import torch
+from openvino._pyopenvino.properties.hint import inference_precision
+from openvino.tools.ovc import convert_model
 from pytest_mock import MockerFixture
 from torch import nn
 
 import nncf
 import nncf.torch
 from nncf.common.quantization.structs import QuantizationScheme
+from nncf.openvino.optimized_functions.models import _compile_ov_model
 from nncf.parameters import CompressWeightsMode
 from nncf.parameters import StripFormat
 from nncf.torch.function_hook.wrapper import get_hook_storage
 from nncf.torch.quantization.layers import AsymmetricLoraQuantizer
 from nncf.torch.quantization.layers import BaseQuantizer
+from nncf.torch.quantization.layers import BaseWeightsDecompressor
 from nncf.torch.quantization.layers import INT4AsymmetricWeightsDecompressor as INT4AsymDQ
 from nncf.torch.quantization.layers import INT4SymmetricWeightsDecompressor as INT4SymDQ
 from nncf.torch.quantization.layers import INT8AsymmetricWeightsDecompressor as INT8AsymDQ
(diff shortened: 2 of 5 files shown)


tests/torch2/function_hook/quantization/strip/test_strip_dequantize.py
```diff
@@ -155,7 +170,7 @@ def test_sym_fq_to_decompressor(param: ParamSymFQ):
 
     scale_shape = (1, 1)
     scale = torch.tensor(SCALE_SAMPLE)
-    scale = scale.expand(scale_shape).to(torch.float16)
+    scale = scale.expand(scale_shape)
```
ljaljushkin, line 173: Doesn't influence the result, just more aligned with default precision in FQ (float32).

tests/torch2/function_hook/quantization/strip/test_strip_dequantize.py
```diff
@@ -76,17 +82,14 @@ def num_call_pack_weight(self) -> int:
...
-        # int8 uses per-channel vs int4 group-wise
-        ParamStripLora(CompressWeightsMode.INT8_ASYM, INT8AsymDQ, torch.bfloat16, 1e-8, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_ASYM, INT4AsymDQ, torch.float32, 1e-3, 1e-3, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_ASYM, INT4AsymDQ, torch.float16, 1e-3, 1e-3, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_ASYM, INT4AsymDQ, torch.bfloat16, 1e-8, 1e-1, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.float32, 1e-3, 1e-3, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.float16, 1e-8, 1e-3, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.bfloat16, 1e-8, 1e-2, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT8_SYM, INT8SymDQ, torch.bfloat16, 1e-2, 1e-3, torch.int8),
+        ParamStripLora(CompressWeightsMode.INT8_ASYM, INT8AsymDQ, torch.bfloat16, 1e-8, 1e-3, torch.uint8),
```
ljaljushkin, line 92: Note: It's expected that the ov_tol value is higher than the torch_tol value, since the OV model executes in f32 but the torch model has activations in bf16 or f16 on tuning. Even though the ov_tol isn't very small, in a few cases it was larger before the PR.

AlexanderDokuchaev (approved):

nncf/torch/quantization/quantize_functions.py
```diff
@@ -126,23 +126,19 @@ def forward(ctx, input_, input_shape, scale, level_low, level_high, levels):
...
-        if input_.dtype in [torch.bfloat16, torch.float16]:
-            input_low = input_low.type(input_.dtype)
-            input_range = input_range.type(input_.dtype)
-
+        dtype = input_.dtype
         original_shape = input_.shape
         input_ = input_.reshape(input_shape)
 
-        output = RQ.Quantize_forward(input_, input_low, input_range, levels)
+        output = RQ.Quantize_forward(input_.type(torch.float32), input_low, input_range, levels)
```
nikita-malininn, line 133: Why not pass all arguments in float32 precision?

nikita-malininn:
I don't get the results from the performance/memory table. Or this PR slows down the reference and increases memory consumption or the table filled somehow wrong.

nncf/torch/quantization/quantize_functions.py
ljaljushkin: input_low, input_range is already in float32

ljaljushkin:
> I don't get the results from the performance/memory table. Or this PR slows down the reference and increases memory consumption or the table filled somehow wrong.

Yes, your understanding is correct. This is the price for the accuracy. 
As I mentioned in the description, increase is not that bad in comparison with not compiled version and backward is not strongly affected.

nikita-malininn (approved):

ljaljushkin:
Updated description with total time required for 1 epoch before and with PR.
There's a 1-6% overhead on RTX per epoch, and on A100, depending on the setup, there can even be a boost of 6% or a slowdown of 3%.

tests/torch2/function_hook/quantization/strip/test_strip_dequantize.py
```diff
@@ -76,17 +82,14 @@ def num_call_pack_weight(self) -> int:
...
-        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.float32, 1e-3, torch.uint8),
-        # torch.compile introduces bigger diff for sym
-        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.float16, 1e-3, torch.uint8),
-        ParamStripLora(CompressWeightsMode.INT4_SYM, INT4SymDQ, torch.bfloat16, 1e-2, torch.uint8),
-        # int8 uses per-channel vs int4 group-wise
-        ParamStripLora(CompressWeightsMode.INT8_SYM, INT8SymDQ, torch.bfloat16, 1e-2, torch.int8),
-        # int8 uses per-channel vs int4 group-wise
-        ParamStripLora(CompressWeightsMode.INT8_ASYM, INT8AsymDQ, torch.bfloat16, 1e-8, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_ASYM, INT4AsymDQ, torch.float32, 1e-3, 1e-3, torch.uint8),
+        ParamStripLora(CompressWeightsMode.INT4_ASYM, INT4AsymDQ, torch.float16, 1e-3, 1e-3, torch.uint8),
```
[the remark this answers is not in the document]
ljaljushkin: This is only for one configuration with float16 asym. I suppose it shows a risk of some differences in accuracy for fp16, but since
- torch vs ov (the second tolerance) shows the same result for f16 on develop and PR, 
- all others and especially bf16 show improvement from (1e-2 to 1e-8) or the same accuracy. 
- evaluation on 8 models, including 3 problematic ones show the small difference. 
this risk is small. 

Propose to leave this test and this atol=1e-3 for f16, and when we tune in f16 and see some significant differences in accuracy, this would be a good starting point for debugging.

P.S. eps=0 in all places didn't help.

PR #3495, opened 2025-05-15
use default dynamic quantization group size

ljaljushkin:


andreyanufr (approved):

ljaljushkin:
image
llm_bench also shows expected performance benefit for 1B model

PR #3496, opened 2025-05-15
Aanuf/lut

andreyanufr:
### Changes

Implementation of compression to fixed codebook (LUT) values .

### Reason for changes

CVS-167084

### Related tickets

CVS-167084

### Tests

tests/openvino/native/quantization/test_weights_compression.py


ljaljushkin:
Do you have a verification that it correctly quantizes values to the specified codebook?

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,40 @@ class AdvancedLoraCorrectionParameters:
     use_int8_adapters: bool = True
 
 
+@api()
+@dataclass
+class AdvancedCodebookParameters:
```
alexsu52, line 364:
```suggestion
class CodebookParameters:
```

alexsu52 (changes requested):
Some comments from my side:
- How will this feature expose in optimum-intel?
- How will NNCF provide the best recipes for definition of codebook?
- For what reason compression algorithms cannot be applied to codebook datatype?
- What is difference in nf4 and codebook flow?

nncf/version.py
```diff
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.17.0"
+__version__ = "2.17.0.dev0+6ccd252b3dirty"
```
ljaljushkin: please revert

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,40 @@ class AdvancedLoraCorrectionParameters:
...
+    """
+    Contains advanced parameters for codebook compression algorithm.
+    :param codebook: The codebook (LUT) for the weight compression.
```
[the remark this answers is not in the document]
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+    model = OVModelForCausalLM.from_pretrained(
+        OUTPUT_DIR, ov_config={"DYNAMIC_QUANTIZATION_GROUP_SIZE": "0", "INFERENCE_PRECISION_HINT": "f32"}
+    )
+    answers_by_questions = generate_answers(questions, model, tokenizer)
+    print(f"Optimized model outputs:\n{answers_by_questions}\n")
+
+
+def main():
+    MODEL_ID = "HuggingFaceTB/SmolLM2-360M-Instruct"
```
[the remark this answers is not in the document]
andreyanufr: Done.

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,40 @@ class AdvancedLoraCorrectionParameters:
...
+        Applicable for vector quantization.
+    :type codebook: list[Any]
+    :param dts_type: The type of the codebook.
+    """
+
+    codebook: list[Any] = field(
+        default_factory=lambda: [
```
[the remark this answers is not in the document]
andreyanufr: Done.

nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -84,7 +85,7 @@ def apply(
         weight_compression_parameters: list[WeightCompressionParameters],
         statistic_points: Optional[StatisticPointsContainer] = None,
         backend_entity: Optional[WeightCompressionAlgoBackend] = None,
-    ) -> tuple[TModel, dict[str, Tensor], dict[str, Tensor]]:
+    ) -> tuple[TModel, dict[str, CompressedWeight]]:
```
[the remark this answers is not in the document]
andreyanufr: Done.

nncf/quantization/algorithms/weight_compression/common.py
```diff
@@ -0,0 +1,53 @@
...
+    zero_point: Optional[Tensor] = None
+    codebook: Optional[Codebook] = None
+
+    def is_codebook(self):
+        """
+        Check if the compressed weight is a codebook.
+
+        :return: True if the compressed weight is a codebook, False otherwise.
+        """
+        return not (self.codebook is None or self.tensor is None or self.scale is None)
```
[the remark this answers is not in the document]
andreyanufr: Done.

nncf/quantization/algorithms/weight_compression/torch_backend.py
```diff
@@ -432,8 +432,7 @@ def transform_model(
         model: Union[GraphModelWrapper, torch.nn.Module],
         graph: NNCFGraph,
         weight_compression_parameters: Iterable[WeightCompressionParameters],
-        precomputed_scales: dict[str, Tensor] = None,
-        precomputed_zero_points: dict[str, Tensor] = None,
+        compressed_weights: dict[str, CompressedWeight] = None,
```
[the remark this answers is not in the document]
andreyanufr: Done.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -537,6 +560,30 @@ def _calculate_nf4_quantized_weight(norm_weight: Tensor) -> Tensor:
...
 
 
+def _calculate_codebook_quantized_weight(
+    norm_weight: Tensor, quantiles: Tensor = None, center_of_quantiles: Tensor = None
+) -> Tensor:
+    """
+    Performs quantization by quantiles (if center_of_quantiles is None). Look-up table is used to
+    "round" or "quantize" to the closest quant.
+
+    :param norm_weight: Weight tensor to quantize already normalized to quantiles range.
```
ljaljushkin: 2 params are missing

nncf/quantization/algorithms/weight_compression/config.py
```diff
@@ -32,6 +32,7 @@ class WeightCompressionConfig:
 
     mode: Optional[CompressWeightsMode] = CompressWeightsMode.INT8_ASYM
     group_size: Optional[int] = -1
+    user_data: Optional[Any] = None
```
ljaljushkin: `user_data` may sound too general. What about `codebook_values` or something more specific?
Please also add new parameter to the docstring.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,121 @@
...
+        mode=nncf.CompressWeightsMode.CODEBOOK,
+        ratio=1.0,
+        group_size=-1,
+        advanced_parameters=nncf.AdvancedCompressionParameters(codebook_params=codebook_params),
+    )
+    model.save_pretrained(output_dir)
+    tokenizer.save_pretrained(output_dir)
+
+    model = OVModelForCausalLM.from_pretrained(
+        output_dir, ov_config={"DYNAMIC_QUANTIZATION_GROUP_SIZE": "64", "INFERENCE_PRECISION_HINT": "f32"}
```
ljaljushkin: We are returning to the default config: 
https://github.com/openvinotoolkit/nncf/pull/3515
https://github.com/openvinotoolkit/nncf/pull/3495

Suggest removing `DYNAMIC_QUANTIZATION_GROUP_SIZE`. 
Do you see it's mandatory for some reason in this specific case?

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,121 @@
...
+    answers_by_questions = generate_answers(QUESTIONS, model, tokenizer)
+    print(f"Optimized model outputs:\n{answers_by_questions}\n")
+
+
+def main():
+    model_id = "HuggingFaceTB/SmolLM2-360M-Instruct"
+    output_dir = "smollm2_360m_compressed_codebook"
+
+    default_codebook_example(model_id, output_dir)
+    custom_codebook_example(model_id, output_dir + "_custom")
```
ljaljushkin: Usually, we validate sample's output to make sure it's functionally correct.
What about adding the same test as for `smollm2_360m_fp8`: https://github.com/openvinotoolkit/nncf/blob/develop/tests/cross_fw/examples/run_example.py#L200?

nncf/quantization/advanced_parameters.py
```diff
@@ -390,6 +406,7 @@ class AdvancedCompressionParameters:
     lora_correction_params: AdvancedLoraCorrectionParameters = field(default_factory=AdvancedLoraCorrectionParameters)
     lora_adapter_rank: int = 256
     backend_params: dict[str, Any] = field(default_factory=dict)
+    codebook_params: AdvancedCodebookParameters = field(default_factory=AdvancedCodebookParameters)
```
ljaljushkin: docstring for new param is missing

nncf/openvino/graph/node_utils.py
```diff
@@ -685,3 +685,27 @@ def create_ov_const_from_tensor(x: Tensor, dtype: ov.Type, name: Optional[str] =
...
+    :param name: Optional name of the constant.
+    :return: OpenVINO subgraph.
+    """
+    codebook_const = opset.constant(codebook.data, dtype=codebook_dtype)
+    if codebook_dtype != ov.Type.f16:
+        codebook_const = opset.convert(codebook_const, destination_type=ov.Type.f16)
+
+    codebook_indexes = opset.constant(indexes.data, dtype=dtype)
+    if dtype == ov.Type.u4:
+        codebook_indexes = opset.convert(codebook_indexes, destination_type=ov.Type.u8)
```
ljaljushkin: I'd extend or add similar [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L262C1-L269C49) for the graph for different use cases (custom, predefined codebook, different precisions), like it was done for all compression types: nf4, int4, int8
https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L205-L229

or maybe would extend this [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L805-L813)

I think it's important to verify the compression graph, and signalize by failing when someone is doing some refactoring in this part. What do you think?

nncf/quantization/algorithms/weight_compression/config.py
andreyanufr: Done

nncf/version.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done

nncf/quantization/advanced_parameters.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/common.py
```diff
@@ -0,0 +1,53 @@
...
+    Compressed weight and decompression parameters.
+
+    :param tensor: The tensor with compressed weight.
+    :param scale: The decompression scale, in practice it is dequantization scale for the quantization.
+    :param zero_point: The zero-point, it is the value of the compression type corresponding to the value 0
+        in the non-compression realm. Applicable for INT quantization.
+    :param codebook: The codebook (LUT) for the weight compression. Applicable for vector quantization
+    """
+
+    tensor: Optional[Tensor] = None
```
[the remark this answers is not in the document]
andreyanufr: It means that some algorithm computed scale/zp , but not computed compressed weights. For example AWQ, SE, GPTQ return scale and zp, but not compressed weights.

nncf/openvino/graph/node_utils.py
```diff
@@ -685,3 +685,27 @@ def create_ov_const_from_tensor(x: Tensor, dtype: ov.Type, name: Optional[str] =
         return opset.constant(x.data, name=name, shared_memory=True)
     const = opset.constant(x.data, dtype=dtype, name=name)
     return const
+
+
+def create_ov_codebook_subgraph(
+    codebook: Tensor, indexes: Tensor, dtype: ov.Type, codebook_dtype: ov.Type, name: Optional[str] = None
+) -> op.Constant:
+    """
+    Create an OpenVINO subgraph with gather from the given codebook and indexes tensors.
```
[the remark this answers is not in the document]
andreyanufr: Done

nncf/quantization/advanced_parameters.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -50,6 +50,28 @@
     dtype=np.float32,
 )
 
+CB4_QUANTILES = np.array(
```
alexsu52: @xiao1228, please review.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -50,6 +50,28 @@
...
+    [
+        -3.5,
+        -2.5,
+        -1.875,
+        -1.375,
+        -1.0,
+        -0.625,
+        -0.3125,
+        0.0,
+        0.2812,
```
alexsu52:
```suggestion
        0.28125,
```
mapping to fp8 without casting

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,22 @@ class AdvancedLoraCorrectionParameters:
     use_int8_adapters: bool = True
 
 
+@api()
+@dataclass
+class CodebookParameters:
+    """
+    Contains advanced parameters for codebook compression algorithm.
```
alexsu52:
```suggestion
    Contains parameters for codebook compression algorithm.
```

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,22 @@ class AdvancedLoraCorrectionParameters:
...
+
+    :param codebook: The codebook (LUT) for the weight compression.
+        Applicable for vector quantization.
+    :type codebook: list[Any]
+    :param dts_type: The type of the codebook.
+    """
+
+    codebook: Optional[list[Any]] = None
+    dst_type: Any = None
```
alexsu52: I would suggest to remove dst_type because any tensor has type of data. LUT can be defined the following:
```
codebook = np.array([0, 1, 2, 3], dtype=np.float16)
codebook = ov.Tensor([0, 1, 2, 3], [4], ov.Type.f8e4m3)
codebook = fns.tensor([0, 1, 2, 3], backend=TensorBackend.numpy, dtype=TensorDataType.float16)
```

```suggestion
    TTensor = Any

    codebook: Optioinal[TTensor] = None
```

nncf/openvino/graph/metatypes/openvino_metatypes.py
```diff
@@ -817,7 +817,16 @@ def _is_embedding(node: ov.Node) -> bool:
     allowed_types_list = ["f16", "f32", "f64"]
     const_port_id = 0
     input_tensor = node.input_value(const_port_id)
-    if input_tensor.get_element_type().get_type_name() in allowed_types_list:
+    input_type = input_tensor.get_element_type().get_type_name()
+
+    try:
+        input_node = node.input(const_port_id).get_source_output().node
```
alexsu52: What is happened if LUT is f16? What do you think about check pattern:
{3E61D414-F93C-497B-BA80-44010BE9D40B}

nncf/parameters.py
```diff
@@ -94,6 +96,8 @@ class CompressWeightsMode(StrEnum):
     NF4 = "nf4"
     INT8 = "int8"  # Deprecated mode
     E2M1 = "e2m1"
+    CODEBOOK = "codebook"
+    CB4_F8E4M3 = "cb4_f8e4m3"
```
alexsu52:
```suggestion
    NF4 = "nf4"
    CB4_F8E4M3 = "cb4_f8e4m3"
    INT8 = "int8"  # Deprecated mode
    E2M1 = "e2m1"
    CODEBOOK = "codebook"
```

nncf/quantization/algorithms/weight_compression/common.py
```diff
@@ -0,0 +1,53 @@
...
+@dataclass
+class Codebook:
+    """
+    Codebook parameters for weight compression.
+    :param codebook: The initial codebook for compression.
+    :param dst_type: The destination type for the codebook.
+    """
+
+    codebook: Optional[Tensor] = None
+    dst_type: Optional[Any] = None
```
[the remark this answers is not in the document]
alexsu52: I suggest to remove

nncf/parameters.py
andreyanufr: Done

nncf/openvino/graph/metatypes/openvino_metatypes.py
andreyanufr: Done by name matching (postfix _nncf_codebook).

nncf/quantization/advanced_parameters.py
andreyanufr: Does not work for ov.Tensor:

```
import openvino as ov
import numpy as np

from nncf.tensor import Tensor



arr = np.ones(shape=(100), dtype=np.uint8)
t = ov.Tensor(arr, ov.Shape([100, 8]), ov.Type.u1)

CB4_QUANTILES = np.array([
        -3.5,
        -2.5,
        -1.875,
        -1.375,
        -1.0,
        -0.625,
        -0.3125,
        0.0,
        0.2812,
        0.5625,
        0.875,
        1.125,
        1.5,
        2.0,
        2.5,
        3.5,
    ],
    dtype=np.float32)

CB4_QUANTILES = ov.Tensor(
    CB4_QUANTILES,
    ov.Shape([16]),
    ov.Type.f8e4m3
)

print(CB4_QUANTILES)
print(CB4_QUANTILES.data)

nncf_tensor = Tensor(CB4_QUANTILES)

print(nncf_tensor.as_numpy_tensor())
```

nncf/quantization/advanced_parameters.py
andreyanufr: Done

xiao1228:
LGTM

nncf/openvino/optimized_functions/functions.py
```diff
@@ -151,7 +151,7 @@ def do_float_quantization(
         compressed_weight = model([weight, precomputed_scale])[0]
         scale = precomputed_scale
 
-    return compressed_weight, scale
+    return compressed_weight, scale, None
```
AlexanderDokuchaev: ?

AlexanderDokuchaev (changes requested):

nncf/openvino/optimized_functions/functions.py
andreyanufr: Done

nncf/quantization/advanced_parameters.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

nncf/openvino/graph/node_utils.py
andreyanufr: Done: extend [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L262C1-L269C49) for default codebook.

nncf/quantization/algorithms/weight_compression/common.py
```diff
@@ -0,0 +1,89 @@
...
+
+from nncf.errors import InvalidGroupSizeError
+from nncf.errors import UnsupportedModelError
+from nncf.tensor import Tensor
+
+ReductionAxes = Union[int, tuple[int, ...]]
+
+
+@dataclass
+class Codebook:
```
ljaljushkin: suggest moving `Codebook` and `CompressedWeight` to parameters.py
`reshape_weight_for_grouped_quantization` to 'group_quantization.py'.  
@nikita-savelyevv can extract functionality for flexible_group_size to this file as well.

nncf/openvino/graph/node_utils.py
ljaljushkin: I meant something like that in a single test.
```
@pytest.mark.parametrize(
    "codebook_dtype, dtype",
    [
        (ov.Type.f32, ov.Type.f32),
        (ov.Type.f16, ov.Type.f16),
        (ov.Type.f16, ov.Type.u4),
        (ov.Type.i8, ov.Type.u4),
...
    ],
)
def test_compression_with_сcdebook_for_different_dtypes(codebook_dtype, dtype):
    // verify graph and convert
```

nncf/quantization/algorithms/weight_compression/common.py
nikita-savelyevv: @ljaljushkin Possibly, this could make sense. But I would suggest to wait for which PR is going to be merged first. Because if we consider only the changes in PR, I don't think creation of 'group_quantization.py' is justified taking into account that the only method there would be `reshape_weight_for_grouped_quantization`.

nikita-savelyevv: At the same time I haven't yet looked at this PR so I don't know what are the reasons to move this function to the newly created common.py. Let me take a look.

andreyanufr: @ljaljushkin @nikita-savelyevv 
Moved Codebook and CompressedWeight to parameters.py
Moved reshape_weight_for_grouped_quantization back to the weight_lowering.py

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -177,20 +122,23 @@ def do_float_quantization(
     config: WeightCompressionConfig,
     reduction_axes: Optional[ReductionAxes] = None,
     precomputed_scale: Optional[Tensor] = None,
-) -> tuple[Tensor, Tensor]:
+    max_val: float = 6.0,
```
nikita-savelyevv: My suggestion would be to:
1. Remove `max_val` argument from do_float_quantization and calculate_float_quantization_params functions
2. Compute `max_val` inside calculate_float_quantization_params as:
```python
if config.mode in [CompressWeightsMode.E2M1, CompressWeightsMode.CODEBOOK, CompressWeightsMode.CB4_F8E4M3]:
    max_val = 6.0 if config.mode == CompressWeightsMode.E2M1 else max(np.abs(config.get_numpy_codebook()))
    scale = scale / max_val
```

This makes more sense in my opinion since we don't really have a need in providing `max_val` values to do_float_quantization dynamically.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -537,6 +503,33 @@ def _calculate_nf4_quantized_weight(norm_weight: Tensor) -> Tensor:
     return quantized_weight
 
 
+def _calculate_codebook_quantized_weight(
+    norm_weight: Tensor, quantiles: Tensor = None, center_of_quantiles: Tensor = None
+) -> Tensor:
```
nikita-savelyevv:
```suggestion
) -> tuple[Tensor, Tensor]:
```

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -210,18 +158,25 @@ def do_float_quantization(
...
         if original_weight_backend == TensorBackend.ov:
             # Can convert through OpenVINO and return OpenVINO-native NF4 tensor
             compressed_weight = norm_weight.as_openvino_tensor().astype(TensorDataType.nf4)
         else:
             compressed_weight = _calculate_nf4_quantized_weight(norm_weight)
+    elif config.is_codebook:
+        compressed_weight, indexes = _calculate_codebook_quantized_weight(
+            norm_weight, quantiles=config.get_numpy_codebook()
+        )
+        return compressed_weight, scale, indexes
```
nikita-savelyevv: Can _calculate_codebook_quantized_weight be used for nf4 compression too?

```suggestion
            compressed_weight = _calculate_codebook_quantized_weight(norm_weight, NF4_QUANTILES, CENTER_OF_NF4_QUANTILES)
    elif config.is_codebook:
        compressed_weight, indexes = _calculate_codebook_quantized_weight(
            norm_weight, quantiles=config.get_numpy_codebook()
        )
        return compressed_weight, scale, indexes
```

nncf/quantization/quantize_model.py
```diff
@@ -597,8 +607,10 @@ def compress_weights(
             msg = "Scale estimation, GPTQ or Lora Correction algorithm is defined, but dataset is None."
             raise nncf.ParameterNotSupportedError(msg)
 
-        if any((awq, scale_estimation, gptq, lora_correction)) and mode == CompressWeightsMode.E2M1:
-            msg = "AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode is E2M1."
+        if any((awq, scale_estimation, gptq, lora_correction)) and mode in [
+            CompressWeightsMode.E2M1,
+        ]:
+            msg = f"AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode is {mode}."
```
nikita-savelyevv: Looks like this change can be reverted

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -217,8 +219,8 @@ def _create_compression_subgraph(
         weight_port_id: int,
         const_dtype,
         should_add_convert_node: bool,
-        layer_scales: Optional[Tensor] = None,
-        layer_zero_points: Optional[Tensor] = None,
+        compressed_weight: Optional[CompressedWeight] = None,
+        advanced_parameters: Optional[AdvancedCompressionParameters] = None,
```
nikita-savelyevv: `advanced_parameters` argument is not used

nncf/quantization/algorithms/weight_compression/constants.py
```diff
@@ -0,0 +1,87 @@
...
+            0.875,
+            1.125,
+            1.5,
+            2.0,
+            2.5,
+            3.5,
+        ],
+        dtype=np.float32,
+    )
+    return Tensor(CB4_QUANTILES).as_openvino_tensor().astype(TensorDataType.f8e4m3)
```
nikita-savelyevv: Can we avoid returning this as openvino tensor? Is it really needed?

I see that later codebook values are casted to numpy anyways:
```python
def get_numpy_codebook(self):
    return self.codebook_values.as_numpy_tensor().data
```

If NP -> OV -> NP conversion is what is achieved here, then we can do this manually beforehand and paste the resulting values to `CB4_QUANTILES` constant.

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +361,21 @@ class AdvancedLoraCorrectionParameters:
...
+@api()
+@dataclass
+class CodebookParameters:
+    """
+    Contains parameters for codebook compression algorithm.
+
+    :param codebook: The codebook (LUT) for the weight compression.
+        Applicable for vector quantization. Must be a numpy array, ov Tensor, or torch Tensor.
+    :type codebook: TTensor
+    :param dts_type: The type of the codebook.
```
nikita-savelyevv:
```suggestion
```

nncf/quantization/algorithms/weight_compression/config.py
```diff
@@ -32,6 +32,7 @@ class WeightCompressionConfig:
 
     mode: Optional[CompressWeightsMode] = CompressWeightsMode.INT8_ASYM
     group_size: Optional[int] = -1
+    codebook_values: Optional[Any] = None
```
nikita-savelyevv: `Any` looks too vague, can we describe the expected type more precisely?

nncf/quantization/algorithms/weight_compression/config.py
```diff
@@ -49,7 +50,22 @@ def is_integer(self):
...
+
+    @property
+    def is_codebook(self):
+        """
+        :return: True if compression type is codebook, else False.
+        """
+        return self.mode in [CompressWeightsMode.CODEBOOK, CompressWeightsMode.CB4_F8E4M3]
+
+    def get_numpy_codebook(self):
+        return self.codebook_values.as_numpy_tensor().data
```
nikita-savelyevv: I don't like the idea that we may store some other Tensor other than numpy type, e.g. torch or openvino, and only then cast to numpy.

As I understand it is possible to override the `__init__` method for dataclasses. There, we could cast to numpy on initialization and store a ready-to-return numpy tensor. What do you think?

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -653,13 +686,13 @@ def apply(
             # del is used to prematurely mark non-necessary data as free for garbage collection
             del self.awq_algo
 
-        scales = {}
-        zero_points = {}
+        compressed_weights = None
```
nikita-savelyevv: After replacing of `scales` and `zero_points` with `compressed_weights` it looks a bit misleading, because the actual weight compression happens later, and here we only compute some compression parameters. 

I remember that this change was required in your last implementation, but is this change really needed now?

nncf/openvino/optimized_functions/functions.py
nikita-savelyevv: Please update the return type hint accordingly

nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -234,32 +236,45 @@ def _create_compression_subgraph(
...
-            compressed_weight.tensor, compression_dtype, name=const_node_name
-        )
-        converted_const = opset.convert(compressed_const, ov.Type.f16)
 
-        if compressed_weight.zero_point is not None:
-            zero_point_const = create_ov_const_from_tensor(
-                compressed_weight.zero_point, compression_dtype, name=f"{const_node_name}/zero_point"
+        if compression_config.is_codebook:
+            n_quants = compressed_weight.tensor.max()
+            compression_dtype = ov.Type.u16 if n_quants > 255 else (ov.Type.u8 if n_quants > 15 else ov.Type.u4)
```
nikita-savelyevv: Does it mean that in theory it is possible that different weights will be saved in different precision despite having the same grids? Is this expected?

tests/openvino/native/quantization/test_weights_compression.py
```diff
@@ -1045,6 +1220,30 @@ def test_compressed_weighs_range(mode, data):
...
+        ([-8.0, -7.0, -6.0, -5.0, -4.0, -3.0, -2.0, -1.0, 0.0]),
+        ([0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0]),
+        ([-8.0, -7.0, -6.0, -5.0, -4.0, -3.0, -2.0, -1.0, 0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0]),
+        ([-1.5, -1.0, -0.5, 0.0, 0.5, 1.0, 1.5]),
+    ),
+)
+def test_codebook_weighs_range(data):
+    data = np.array(data).astype(np.float32)
+    codebook = data
+    max_diff = 0.1
```
nikita-savelyevv, line 1233: Considering the size of the arrays, I think it would be better to specify precise input weight values and references instead of random data and `max_diff` approach

ljaljushkin (approved):

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +361,21 @@ class AdvancedLoraCorrectionParameters:
...
 
 
+@api()
+@dataclass
+class CodebookParameters:
+    """
+    Contains parameters for codebook compression algorithm.
+
+    :param codebook: The codebook (LUT) for the weight compression.
+        Applicable for vector quantization. Must be a numpy array, ov Tensor, or torch Tensor.
```
nikita-savelyevv: I haven't found a test where codebook values are provided as torch tensors. Could you please add such cases?

nncf/quantization/algorithms/weight_compression/constants.py
andreyanufr: No, because we need to know destination type for codebook.

nncf/quantization/algorithms/weight_compression/config.py
andreyanufr: We need to know destination data type for codebook. If we cast fp8/fp4 tensor to numpy we will lost destination data type.

nikita-savelyevv: I see, thanks for the explanation. Could you please then extend docstring of `WeightCompressionConfig` for `codebook_values` argument and describe this logic there?

nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -537,6 +499,33 @@ def _calculate_nf4_quantized_weight(norm_weight: Tensor) -> Tensor:
     return quantized_weight
 
 
+def _calculate_codebook_quantized_weight(
+    norm_weight: Tensor, quantiles: Tensor = None, center_of_quantiles: Tensor = None
```
nikita-savelyevv:
```suggestion
    norm_weight: Tensor, quantiles: np.ndarray = None, center_of_quantiles: np.ndarray = None
```

nncf/quantization/algorithms/weight_compression/openvino_backend.py
andreyanufr: No, we always will have max value from our codebook in compressed weights. And n_quants = len(codebook) - 1.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: It is possible, but I suggest doing it at the next PR.

nncf/quantization/quantize_model.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/openvino_backend.py
andreyanufr: Done

nncf/quantization/advanced_parameters.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/config.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: I think we need it now because it is more convenient to store compression parameters in one structure. In the next step we will add activation-based codebook calculation, and it is easier to implement this approach with this implementation.

nncf/openvino/optimized_functions/functions.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: Ok, in such case I think it would be better to rename this kind of variable from `compressed_weights` to `precomputed_compressed_weights` to at least distinguish it from the actual compressed weight which is computed inside backend logic

nncf/quantization/algorithms/weight_compression/openvino_backend.py
nikita-savelyevv: > And n_quants = len(codebook) - 1

I see, thanks! Then I would suggest to compute n_quants this way

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

nikita-savelyevv: I believe it should be `np.ndarray` and not `np.array` because the first one is a type and the second one is a function.

nncf/quantization/algorithms/weight_compression/openvino_backend.py
andreyanufr: Done

nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: Done

nikita-savelyevv: Could you please also do this renaming elsewhere? For example methods `_create_compression_subgraph()`, `transform_model()` at <backend_name>_backend.py or `compress_weight()` at weight_lowering.py still have `compressed_weight` as an argument.

nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def default_codebook_example(model_id, output_dir):
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+    model = OVModelForCausalLM.from_pretrained(
+        model_id,
+        export=True,
+        load_in_8bit=False,
+        compile=False,
+        stateful=False,
```
alexsu52: Why is the default value (stateful=True) not used?

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,122 @@
...
+        ov_config={"INFERENCE_PRECISION_HINT": "f32"},
```
alexsu52: By default, CPU use f32 inference precision. What is reason of explicit definition of the inference precision?

src/nncf/parameters.py
```diff
@@ -85,15 +85,19 @@ class CompressWeightsMode(StrEnum):
     :param NF4: The the same as INT4_SYM mode, but primary precision is NF4 data type without zero point.
     :param INT8: Mode is deprecated and will be removed in future releases. Please use `INT8_ASYM` instead.
     :param E2M1: FP4 format from "OCP Microscaling Formats (MX) Specification" Version 1.0.
+    :param CODEBOOK: Codebook (LUT) quantization format.
+    :param CB4_F8E4M3: Codebook (LUT) format with 16 fixed fp8 values.
```
alexsu52:
```suggestion
    :param CB4_F8E4M3: Codebook (LUT) format with 16 fixed fp8 values in E4M3 format.
```

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -214,12 +213,13 @@ def transform_model(
             compression_config = wc_params.compression_config
             node = wc_params.node_with_weight
             weight = self.get_weight(node, wc_params.weight_port_id, model, graph)
-            compressed_weight = compress_weight(
+            precomputed_compressed_weight = compress_weight(
```
nikita-savelyevv:
```suggestion
            compressed_weight = compress_weight(
```

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -231,7 +231,7 @@ def transform_model(
             # See https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md
             if opset_version < 21 and dequantize_block_size > 0:
                 compressed_weight, scale, zero_point = self._preprocess_compressed_weight(
-                    compressed_weight, weight.shape, dequantize_block_size=None, apply_transpose=True
+                    precomputed_compressed_weight, weight.shape, dequantize_block_size=None, apply_transpose=True
```
nikita-savelyevv:
```suggestion
                    compressed_weight, weight.shape, dequantize_block_size=None, apply_transpose=True
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -350,22 +301,34 @@ def compress_weight(
...
 ) -> CompressedWeight:
     """
     Compress weight using compression configuration.
 
     :param weight: The weight to compress.
     :param reduction_axes: Axes, along which to reduce (collect) different statistics (e.g. min, max).
     :param config: Compression configuration.
-    :param precomputed_scale: Precomputed scale.
-    :param precomputed_zero_point: Precomputed zero point.
+    :param precomputed_compressed_weight: precomputed scale and zero point.
```
nikita-savelyevv:
```suggestion
    :param precomputed_compressed_weight: Contains precomputed scale and zero point.
```

src/nncf/quantization/algorithms/weight_compression/config.py
```diff
@@ -28,10 +29,13 @@ class WeightCompressionConfig:
     :param mode: Defines a mode for weight compression. Defaults to INT8_ASYM mode.
     :param group_size: Number of weights (e.g. 128) in the channel dimension that share quantization parameters (scale).
         The value -1 means no grouping. Defaults to -1.
+    :param codebook_values: Optional codebook values for CODEBOOK compression mode.
+        Must be fns.Tensor which wraps numpy array or ov tensor.
```
nikita-savelyevv:
```suggestion
    :param codebook_values: Optional codebook values for CODEBOOK compression mode.
        Must be fns.Tensor which wraps numpy array or ov tensor. Storing ov tensor is useful for having
        destination data type information available.

File: examples/llm_compression/openvino/smollm2_360m_codebook/main.py (hunk 40 above)
Comment (andreyanufr) [reply examples/llm_compression/openvino/smollm2_360m_codebook/main.py]: Done.

File: examples/llm_compression/openvino/smollm2_360m_codebook/main.py (hunk 41 above)
Comment (andreyanufr) [reply examples/llm_compression/openvino/smollm2_360m_codebook/main.py]: Done

File: src/nncf/quantization/algorithms/weight_compression/onnx_backend.py (hunk 43 above)
Comment (andreyanufr) [reply src/nncf/quantization/algorithms/weight_compression/onnx_backend.py]: Done.

File: src/nncf/quantization/algorithms/weight_compression/onnx_backend.py (hunk 44 above)
Comment (andreyanufr) [reply src/nncf/quantization/algorithms/weight_compression/onnx_backend.py]: Done.

File: src/nncf/quantization/algorithms/weight_compression/weight_lowering.py (hunk 45 above)
Comment (andreyanufr) [reply src/nncf/quantization/algorithms/weight_compression/weight_lowering.py]: Done.

**Review by nikita-savelyevv — APPROVED** [review]
LGTM!

File: examples/llm_compression/openvino/smollm2_360m_codebook/main.py [hunk 47]
```diff
@@ -0,0 +1,115 @@
...
+# limitations under the License.
+
+import numpy as np
+from optimum.intel.openvino import OVModelForCausalLM
+from transformers import AutoTokenizer
+
+import nncf
+
+
+def generate_answers(questions, model, tokenizer, max_new_tokens=50):
```
AlexanderDokuchaev: Please add type hints
And little docstring to describe what each function do?

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+    messages = [
+        {"role": "system", "content": "You are a chatbot who always responds as short as possible."},
+        {"role": "user", "content": "What is the capital of Spain?"},
+        {"role": "assistant", "content": "Madrid."},
+    ]
+    answers_by_questions = {}
+    model.request = None
```
AlexanderDokuchaev: Add comment why need to set request None

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+def default_codebook_example(model_id, output_dir):
+    model, tokenizer = load_model_and_tokenizer(model_id)
+    answers_by_questions = generate_answers(QUESTIONS, model, tokenizer)
+    print(f"Non-optimized model outputs:\n{answers_by_questions}\n")
+
+    model.model = nncf.compress_weights(model.model, mode=nncf.CompressWeightsMode.CB4_F8E4M3, ratio=1.0, group_size=64)
+    model.save_pretrained(output_dir)
+    tokenizer.save_pretrained(output_dir)
+
+    model, tokenizer = load_model_and_tokenizer(output_dir, False)
```
AlexanderDokuchaev: Why the model saves and load after? Can it be loaded only ones?
If it's need better saves files in own directory for each *_example.

And it is not clear why load_model expect model_id by got output_dir.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+    "What is the capital of France?",
+    "What is the highest peak in the Alps?",
+    "What is the largest city in Canada?",
+    "What is the most visited city in Japan?",
+]
+
+
+def load_model_and_tokenizer(model_id, export=True):
+    tokenizer = AutoTokenizer.from_pretrained(model_id)
+    model = OVModelForCausalLM.from_pretrained(
```
AlexanderDokuchaev, line 93: Suggest to add warnings.filterwarnings("ignore", category=TracerWarning) like in https://github.com/openvinotoolkit/nncf/blob/develop/examples/llm_compression/torch/distillation_qat_with_lora/main.py

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+        export=export,
+        load_in_8bit=False,
+    )
+    return model, tokenizer
+
+
+def default_codebook_example(model_id, output_dir):
+    model, tokenizer = load_model_and_tokenizer(model_id)
+    answers_by_questions = generate_answers(QUESTIONS, model, tokenizer)
+    print(f"Non-optimized model outputs:\n{answers_by_questions}\n")
```
AlexanderDokuchaev: May be add function to print answering?
```
{'What is the capital of France?': 'Paris.', 'What is the highest peak in the Alps?': 'Mont Blanc.', 'What is the largest city in Canada?': 'Toronto.', 'What is the most visited city in Japan?': 'Tokyo.'}
```
It is not readable.

```python
def print_answers(answers_by_questions):
    for question, answer in answers_by_questions.items():
        print(f"Q: {question}\nA: {answer}\n")
```

```
Non-optimized model outputs:

Q: What is the capital of France?
A: Paris.

Q: What is the highest peak in the Alps?
A: Mont Blanc.

Q: What is the largest city in Canada?
A: Toronto.

Q: What is the most visited city in Japan?
A: Tokyo.
```

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+    model = OVModelForCausalLM.from_pretrained(
+        model_id,
+        export=export,
+        load_in_8bit=False,
+    )
+    return model, tokenizer
+
+
+def default_codebook_example(model_id, output_dir):
+    model, tokenizer = load_model_and_tokenizer(model_id)
```
AlexanderDokuchaev, line 109: Add some log in example that currently run example with default codebook for CB4_F8E4M3 and in another custom, to separate output logs.

AlexanderDokuchaev:
```
`loss_type=None` was set in the config but it is unrecognised.Using the default loss: `ForCausalLMLoss`.
The attention mask is not set and cannot be inferred from input because pad token is same as eos token. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
```
Can it be fixed or suppressed in ouput?

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,115 @@
...
+
+    model, tokenizer = load_model_and_tokenizer(output_dir, False)
+    answers_by_questions = generate_answers(QUESTIONS, model, tokenizer)
+    print(f"Optimized model outputs:\n{answers_by_questions}\n")
+
+    return list(answers_by_questions.values())
+
+
+def main():
+    model_id = "HuggingFaceTB/SmolLM2-360M-Instruct"
```
AlexanderDokuchaev: I thinks better to use global variable to set model id and output dir, to decrease functions signatures, to simplify example

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -179,6 +181,22 @@ def check_user_compression_configuration(
             ]
         )
         ranks = [advanced_parameters.lora_adapter_rank, advanced_parameters.lora_correction_params.adapter_rank]
+
+        if advanced_parameters.codebook_params.codebook is not None:
+            codebook = Tensor(advanced_parameters.codebook_params.codebook).as_numpy_tensor().data
```
AlexanderDokuchaev: For what is it need `Tensor(advanced_parameters.codebook_params.codebook).as_numpy_tensor().data`?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -429,6 +453,15 @@ def _get_ratio_defining_params(
...
         return ratio_defining_params
 
+    def _get_primary_config(self):
+        return WeightCompressionConfig(
+            mode=self._mode,
+            group_size=self._group_size,
+            codebook_values=get_cb4_quantiles()
+            if self._mode == CompressWeightsMode.CB4_F8E4M3
+            else Tensor(self._advanced_parameters.codebook_params.codebook),
+        )
```
AlexanderDokuchaev:
```suggestion
        codebook_values = get_cb4_quantiles() if self._mode == CompressWeightsMode.CB4_F8E4M3 else Tensor(self._advanced_parameters.codebook_params.codebook)
        return WeightCompressionConfig(
            mode=self._mode,
            group_size=self._group_size,
            codebook_values=codebook_values,
        )
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -124,30 +68,31 @@ def reshape_weight_for_grouped_quantization(
...
     """
-    assert config.mode in [CompressWeightsMode.NF4, CompressWeightsMode.E2M1]
+    assert not config.is_integer
 
     if weight.dtype != TensorDataType.float32:
         weight = weight.astype(TensorDataType.float32)
 
     scale = fns.max(fns.abs(weight), axis=reduction_axes, keepdims=True)
+    if config.mode in [CompressWeightsMode.E2M1, CompressWeightsMode.CODEBOOK, CompressWeightsMode.CB4_F8E4M3]:
+        max_val = 6.0 if config.mode == CompressWeightsMode.E2M1 else max(np.abs(config.get_numpy_codebook()))
```
AlexanderDokuchaev: Use nncf.Tensor in common code

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -537,6 +500,33 @@ def _calculate_nf4_quantized_weight(norm_weight: Tensor) -> Tensor:
...
+    :param center_of_quantiles: Center of quantiles to use for quantization. If None, it is calculated as the average
+        of adjacent quantiles.
+    :return: Tensor with floating-point values, where each of them corresponds to elements from quantiles.
+    """
+    assert quantiles is not None or center_of_quantiles is not None, (
+        "Either quantiles or center_of_quantiles should be provided"
+    )
+
+    if center_of_quantiles is None:
+        quantiles = np.array(quantiles)
```
AlexanderDokuchaev: Please don't combine operations with backend specific types and nncf.Tensor in common code.

src/nncf/quantization/algorithms/weight_compression/constants.py
```diff
@@ -0,0 +1,87 @@
...
+    ],
+    dtype=np.float32,
+)
+
+
+def get_cb4_quantiles() -> Tensor:
+    """
+    Returns the quantiles for the CB4 codebook.
+    """
+    CB4_QUANTILES = np.array(
```
AlexanderDokuchaev: Define global vraible as global or use lower case in function 
https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#47-naming

src/nncf/quantization/algorithms/weight_compression/constants.py
```diff
@@ -0,0 +1,87 @@
...
+            0.875,
+            1.125,
+            1.5,
+            2.0,
+            2.5,
+            3.5,
+        ],
+        dtype=np.float32,
+    )
+    return Tensor(CB4_QUANTILES).as_openvino_tensor().astype(TensorDataType.f8e4m3)
```
AlexanderDokuchaev: Dont use backend specific code in common parts.

AlexanderDokuchaev (changes requested):

src/nncf/openvino/graph/metatypes/openvino_metatypes.py
```diff
@@ -817,7 +817,12 @@ def _is_embedding(node: ov.Node) -> bool:
     allowed_types_list = ["f16", "f32", "f64"]
     const_port_id = 0
     input_tensor = node.input_value(const_port_id)
-    if input_tensor.get_element_type().get_type_name() in allowed_types_list:
+    input_type = input_tensor.get_element_type().get_type_name()
+
+    if node.friendly_name.endswith("nncf_codebook"):
```
AlexanderDokuchaev, line 823: My mind it's bad design when graph builder know about any algorithms. 
May be it should be done via auto generated ignored scope, like ignored patterns.
@alexsu52

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Removed.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: I think that It is usual behavior: compress (once) -> save (once) -> load and run (multiply time)

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: To be sure that 
if (codebook[:-1] >= codebook[1:]).any():
works correctly for all data types.

src/nncf/quantization/algorithms/weight_compression/constants.py
andreyanufr: Removed. Done.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

src/nncf/openvino/graph/metatypes/openvino_metatypes.py
AlexanderDokuchaev, line 823: Please add TODO

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+    return model, tokenizer
+
+
+def default_codebook_example(model_id: str, compressed_model_id: str) -> list[str]:
+    """
+    Example of using the default codebook compression.
+
+    :param  model_id (str): The identifier of the model to load.
+    :param compressed_model_id (str): The identifier for the compressed model to save.
```
AlexanderDokuchaev:
```suggestion
    :param  model_id: The identifier of the model to load.
    :param compressed_model_id: The identifier for the compressed model to save.
```

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,163 @@
...
+    "What is the most visited city in Japan?",
+]
+
+
+def load_model_and_tokenizer(model_id: str, export=True) -> tuple[OVModelForCausalLM, AutoTokenizer]:
+    """
+    Load the model and tokenizer from the specified model ID.
+
+    :param model_id (str): The identifier of the model to load.
+    :param  export (bool): Whether to export the model for OpenVINO. Defaults to True.
```
AlexanderDokuchaev:
```suggestion
    :param model_id: The identifier of the model to load.
    :param export: Whether to export the model for OpenVINO. Defaults to True.
```

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,163 @@
...
+
+    return answers_by_questions
+
+
+def print_answers(header: str, answers_by_questions: list[str]) -> None:
+    """
+    Print the answers to the console.
+
+    :param header (str): Header to print before the answers.
+    :param answers_by_questions (dict): Dictionary mapping questions to their answers.
```
AlexanderDokuchaev:
```suggestion
    :param header: Header to print before the answers.
    :param answers_by_questions: Dictionary mapping questions to their answers.
```

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,163 @@
...
+def generate_answers(
+    questions: list[str], model: OVModelForCausalLM, tokenizer: AutoTokenizer, max_new_tokens: int = 50
+) -> dict[str, str]:
+    """
+    Generate answers for a list of questions using the provided model and tokenizer.
+
+    :param questions : List of questions to be answered.
+    :param model : The model to use for generating answers.
+    :param tokenizer : The tokenizer to use for processing the input and output.
+    :param max_new_tokens (int, optional): Maximum number of new tokens to generate for each answer. Defaults to 50.
```
AlexanderDokuchaev:
```suggestion
    :param questions: List of questions to be answered.
    :param model: The model to use for generating answers.
    :param tokenizer: The tokenizer to use for processing the input and output.
    :param max_new_tokens: Maximum number of new tokens to generate for each answer. Defaults to 50.
```

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
```diff
@@ -0,0 +1,163 @@
...
+
+    return list(answers_by_questions.values())
+
+
+def custom_codebook_example(model_id: str, compressed_model_id: str) -> list[str]:
+    """
+    Example of using the custom codebook compression.
+
+    :param model_id (str): The identifier of the model to load.
+    :param compressed_model_id (str): The identifier for the compressed model to save.
```
AlexanderDokuchaev:
```suggestion
    :param model_id: The identifier of the model to load.
    :param compressed_model_id: The identifier for the compressed model to save.
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -179,6 +181,22 @@ def check_user_compression_configuration(
...
+                msg = "The codebook must be a 1D array, but a multi-dimensional array is given."
+            if codebook.size < 2:
+                msg = (
+                    "The codebook must contain at least two unique elements,"
+                    "but a single-element or empty array is given."
+                )
+            if (codebook[:-1] >= codebook[1:]).any():
+                msg = "The codebook must be a sorted 1D array with unique elements, but an unsorted array is given."
+            if msg:
+                raise nncf.ValidationError(msg)
```
AlexanderDokuchaev:
```suggestion
        codebook = advanced_parameters.codebook_params.codebook
        if codebook is not None:
            # OpenVINO Tensor is not support functions to validate codebook
            np_codebook = Tensor(codebook).as_numpy_tensor()
            msg = None
            if np_codebook.ndim != 1:
                msg = "The codebook must be a 1D array, but a multi-dimensional array is given."
            elif np_codebook.size < 2:
                msg = (
                    "The codebook must contain at least two unique elements,"
                    "but a single-element or empty array is given."
                )
            elif fns.any(np_codebook[:-1] >= np_codebook[1:]):
                msg = "The codebook must be a sorted 1D array with unique elements, but an unsorted array is given."
            if msg:
                raise nncf.ValidationError(msg)
```

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -218,8 +217,9 @@ def transform_model(
                 Tensor(weight),
                 wc_params.reduction_axes,
                 compression_config,
-                None if precomputed_scales is None else precomputed_scales.get(wc_params.weight_name),
-                None if precomputed_zero_points is None else precomputed_zero_points.get(wc_params.weight_name),
+                None
```
AlexanderDokuchaev:
```python
precomputed_compressed_weights = precomputed_compressed_weights or {}
...

compress_weight(
    Tensor(weight),
    wc_params.reduction_axes,
    compression_config,
    precomputed_compressed_weights.get(wc_params.weight_name),
)
```

Multiline condition in arguments is hard to read

src/nncf/openvino/graph/metatypes/openvino_metatypes.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

examples/llm_compression/openvino/smollm2_360m_codebook/main.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
andreyanufr: Done.

AlexanderDokuchaev (approved):

src/nncf/quantization/quantize_model.py
```diff
@@ -598,7 +608,7 @@ def compress_weights(
             raise nncf.ParameterNotSupportedError(msg)
 
         if any((awq, scale_estimation, gptq, lora_correction)) and mode == CompressWeightsMode.E2M1:
-            msg = "AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode is E2M1."
+            msg = f"AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode is {mode}."
```
alexsu52: ?

src/nncf/quantization/advanced_parameters.py
```diff
@@ -390,6 +408,7 @@ class AdvancedCompressionParameters:
     lora_correction_params: AdvancedLoraCorrectionParameters = field(default_factory=AdvancedLoraCorrectionParameters)
     lora_adapter_rank: int = 256
     backend_params: dict[str, Any] = field(default_factory=dict)
+    codebook_params: CodebookParameters = field(default_factory=CodebookParameters)
```
alexsu52: Simple is better than complex.

```suggestion
    codebook: Optional[TTensor] = None
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -429,6 +456,18 @@ def _get_ratio_defining_params(
 
         return ratio_defining_params
 
+    def _get_primary_config(self):
+        codebook_values = (
+            Tensor(CB4_QUANTILES)
+            if self._mode == CompressWeightsMode.CB4_F8E4M3
+            else Tensor(self._advanced_parameters.codebook_params.codebook)
```
alexsu52: I'm concerned about the code in else because self._advanced_parameters.codebook_params.codebook is None by default. Is this not mistake?

src/nncf/quantization/algorithms/weight_compression/parameters.py
```diff
@@ -0,0 +1,53 @@
...
+@dataclass
+class Codebook:
+    """
+    Codebook parameters for weight compression.
+    :param codebook: The initial codebook for compression.
+    :param dst_type: The destination type for the codebook.
+    """
+
+    codebook: Optional[Tensor] = None
+    dst_type: Optional[Any] = None
```
alexsu52: Please double check that this is not redundant code because I did not find where you uses it.

src/nncf/parameters.py
andreyanufr: Done.

src/nncf/quantization/quantize_model.py
andreyanufr: DOne.

src/nncf/quantization/advanced_parameters.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
andreyanufr: Done.

src/nncf/quantization/algorithms/weight_compression/parameters.py
andreyanufr: Done.

PR #3506, opened 2025-05-22
Made NLS sample self-contained

ljaljushkin:
### Changes

removed dependency from another sample

### Reason for changes

can run sample without setting PYTHONPATH

### Related tickets

n/a

### Tests

test_examples - https://github.com/openvinotoolkit/nncf/actions/runs/15190626862

CC: @jpablomch

### DIFF
--- a/examples/llm_compression/torch/qat_with_nls_downstream/main.py
+++ b/examples/llm_compression/torch/qat_with_nls_downstream/main.py
@@ -18,7 +18,7 @@ import warnings
 from collections import defaultdict
 from datetime import datetime
 from pathlib import Path
-from typing import Union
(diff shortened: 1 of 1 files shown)


nikita-savelyevv (commented):
Have you considered moving these to something like `examples/llm_compression/torch/common.py` or `examples/llm_compression/torch/utils.py`?

ljaljushkin:
> Have you considered moving these to something like `examples/llm_compression/torch/common.py` or `examples/llm_compression/torch/utils.py`?

Yes, there will be the same problem with dependencies. 
Common part is not that big for making a separation.

nikita-savelyevv (approved):

PR #3467, opened 2025-04-28
[Examples][FX] WC Example for TorchFX Backend

anzr299:
### Changes

Add example script for TorchFX backend in `examples/`.

### Reason for changes

This is to show a sample script for applying WC to LLMs which are exported using `torch.export()`

PR #3511, opened 2025-05-26
[TorchFX][WC Conformance] Fix Issues in WC Conformance for FX

anzr299:
### Changes

This PR achieves the following:
1. Fix an issue faced when using openvino nightly Ref. 167462
2. Fix performance issue where compressed FX models in conformance were running slow. Fixes discovered when working on #3467 allow the compressed FX model to run faster.

### Reason for changes

To resolve errors with openvino nightly and make the WC conformance test run faster.

### Related tickets

167462

-- | -- | -- | -- | --
WC Conformance | Nightly | Pass | 104 |


tests/post_training/pipelines/fx_modelling.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class FXAutoModelForCausalLM(OptimizedModel, GenerationMixin):
+    def __init__(
+        self,
+        model: torch.fx.GraphModule,
+        config: PretrainedConfig,
+        generation_config: GenerationConfig,
+        device: str = "cpu",
+        compile: bool = True,
```
AlexanderDokuchaev, line 31: compile=False, not used anywhere

tests/post_training/pipelines/fx_modelling.py
```diff
@@ -0,0 +1,113 @@
...
+    ):
+        super().__init__(model, config)
+        if compile:
+            self.model = torch.compile(model, backend="openvino", options={"aot_autograd": True})
+        self.generation_config = generation_config
+        self.main_input_name = "input_ids"
+        self._device = device.upper()
```
AlexanderDokuchaev, line 38: For what used upper and lower for device?

AlexanderDokuchaev (changes requested):

tests/post_training/pipelines/fx_modelling.py
anzr299: Alright, I will remove it

tests/post_training/pipelines/fx_modelling.py
anzr299: It was used in the old code to pass device to torch.compile(). I will remove it here since we are running with CPU only for now.

PR #3534, opened 2025-06-05
[gha] Remove coloring by latex in summary

AlexanderDokuchaev:
### Changes

Remove coloring by latex in summary

### Reason for changes

No supporting now 
image

### DIFF
--- a/.github/scripts/pytest_md_summary.py
+++ b/.github/scripts/pytest_md_summary.py
@@ -39,19 +39,19 @@ def parse_xml_report(xml_file) -> None:
         time_duration = float(testcase.get("time", "0"))
         message = ""
         if testcase.find("failure") is not None:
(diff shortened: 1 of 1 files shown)


.github/scripts/pytest_md_summary.py
```diff
@@ -39,19 +39,19 @@ def parse_xml_report(xml_file) -> None:
         time_duration = float(testcase.get("time", "0"))
         message = ""
         if testcase.find("failure") is not None:
-            status = "$${\color{red}Failed}$$"
```
ljaljushkin, line 42: Have you tried html tag?
```
<code style="color : green">text</code>
```

AlexanderDokuchaev: <code style="color : green">text</code> 

it does not work

PR #3556, opened 2025-06-24
[WC] Introduce flexible group size value search

nikita-savelyevv:
### Changes

Introduce flexible group size search logic as a part of mixed precision algorithm. When enabled, each weight for which the
channel size is not divisible by the general group size value will be compressed to a newly calculated group size.

The new group size value is the maximal power of two (i.e., 2^k) such that:
- channel size is divisible by it;
- it is less than the originally specified group size value;
- it is greater than or equal to `min_flexible_group_size` (16 by default).

If it's not possible to find a value satisfying these requirements, such weight is compressed to backup precision. If ratio < 1.0 and some weights have to be compressed to the backup precision because of group size issues, then these weights won't contribute to the ratio of backup mode group.

This method is disabled by default.

### Reason for changes

Some models may have channel size values that are not divisible by the default group size. In such case a user can now provide `nncf.AdvancedCompressionParameters(enable_flexible_group_size=True)` advanced parameter instead of an ignored scope.

Example models:
- `microsoft/Phi-4-multimodal-instruct` (lm_model and vision_embeddings_model)
- `HuggingFaceH4/Qwen2.5-Math-1.5B-Instruct-PRM-0.2`

### Metrics

Results for phi4-multimodal are below.

| Language Model Precision            | Vision Embed. Model Precision       | WWB Similarity | Time of image-to-text request (sec.) | Time of audio-to-text request (sec.) |
|------------------------------------|-------------------------------------|----------------|---------------------------------------|----------------------------------------|
| FP16                               | FP16                                | 99.19%         | 31.21                                | 17.76                                 |
| Mixed precision: int4 or bf16       | Mixed precision: int4 or bf16        | 77.51%         | 22.37                                | 10.93                                 |
| Mixed precision: int4 or int8       | Mixed precision: int4 or int8        | 79.03%         | 19.95                                | 9.47                                  |
| int4 with mixed group size: 128 or 64 | int4 with mixed group size: 128 or 16 | 81.36%         | 19.89                                | 9.16                                  |

Last row corresponds to `nncf.AdvancedCompressionParameters(enable_flexible_group_size=True)`. 

Third row corresponds to `nncf.AdvancedCompressionParameters(enable_flexible_group_size=True, min_flexible_group_size=128)` 

Second row corresponds to `nncf.AdvancedCompressionParameters(enable_flexible_group_size=True, min_flexible_group_size=128)` with `backup_mode="none"`.

Inference time results are expected. Similarity not so much, but still no degradation for group size 16 case.

### Related tickets

167337

### Tests

Added test cases which assert that the expected log messages are printed.

https://github.com/openvinotoolkit/nncf/actions/runs/15852358755


nikita-savelyevv:
During integration with optimum-intel I propose to introduce a boolean CLI argument called `--flexible-group-size` to expose this logic. `min_flexible_group_size` will only by available through Python API.

nncf/quantization/advanced_parameters.py
```diff
@@ -359,6 +359,28 @@ class AdvancedLoraCorrectionParameters:
...
+
+    :param enable_flexible_group_size: Whether to enable flexible group size searching.
+    :type enable_flexible_group_size: bool
+    :param min_flexible_group_size: Minimum group size for flexible group size searching. Defaults to 16. The reason
+        behind this argument is to avoid too small group size values, which may lead to performance issues.
+    :type min_flexible_group_size: int
+    """
+
+    enable_flexible_group_size: bool = False
+    min_flexible_group_size: int = 16
```
nikita-savelyevv: The value of 16 is open for debate. Possibly, it should be larger.

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -445,12 +448,25 @@ def _set_weight_compression_config(
...
+                failed_node_names.append(w_params.node_with_weight.node_name)
+        if len(failed_node_names) > 0:
+            names = ",".join(f'"{name}"' for name in failed_node_names)
+            msg = (
+                f"Failed to apply group-wise quantization with group size value {self._group_size}.\n"
+                "Ensure that the group size is divisible by the channel size, "
+                "or include this node and others with similar issues in the ignored scope:\n"
+                f"nncf.compress_weight(\n\t..., \n\tignored_scope=IgnoredScope(names=[{names}]\n\t)\n)"
+            )
+            raise nncf.InvalidGroupSizeError(msg)
```
nikita-savelyevv: With @alexsu52 decided to keep the message with ignored scope suggestion for now.

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -336,38 +339,72 @@ def test_awq_scale_reference(self, monkeypatch):
...
         model = self.get_awq_model()
         hidden_dim = 8
-        invalid_group_size = hidden_dim + 1
         input_example = self.to_tensor(np.ones([1, 4, hidden_dim], dtype=np.float32))
         dataset = Dataset([input_example])
         algorithm_dict = {algorithm: True} if algorithm else {}
         kwargs = dict(
             model=model,
             mode=CompressWeightsMode.INT4_ASYM,
             ratio=1.0,
```
ljaljushkin, line 385: I'd add a test for `enable_flexible_group_size` with ratio != 1.

nncf/quantization/algorithms/weight_compression/mixed_precision.py
```diff
@@ -124,6 +162,59 @@ def get_statistic_points(
...
+            )
+
+        if len(flexible_group_size_not_found_weight_params) > 0:
+            node_strings = [""] * len(flexible_group_size_not_found_weight_params)
+            for i, w_params in enumerate(flexible_group_size_not_found_weight_params):
+                weight_shape = w_params.weight_shape
+                reduction_channel_size, _ = get_reduction_channel_size(weight_shape, w_params.reduction_axes)
+                node_strings[i] = f"{w_params.node_with_weight.node_name} (weight shape: {weight_shape})"
+            nncf_logger.warning(
+                "Large enough flexible group size value cannot be found for some nodes. They will be compressed to the "
```
ljaljushkin:
```suggestion
                "Large enough flexible group size value cannot be found for some nodes. They will be compressed according to the "
```

nncf/quantization/algorithms/weight_compression/mixed_precision.py
```diff
@@ -85,20 +99,44 @@ def apply(
...
-        for index in indexes_of_layers_in_ascending_order_of_scores:
-            weight_param = weight_params[index]
-            current_ratio = (num_weights_in_4bit + weight_param.num_weights) / num_all_weights
-            if current_ratio >= self._ratio:
-                break
-            weight_param.compression_config = self._primary_config
-            num_weights_in_4bit += weight_param.num_weights
+        flexible_group_size_values = {}
+        valid_weight_params = weight_params
+        if self._enable_flexible_group_size and self._primary_config.group_size != -1:
```
ljaljushkin: I see that `flexible group size` is closely connected with mixed precision, but I don't feel it's responsibility of this class.
What about doing it outside and give a modified `ratio_defining_params`?

ljaljushkin (changes requested):

nncf/quantization/algorithms/weight_compression/mixed_precision.py
ljaljushkin: potentially, can be implemented in `group_quantization.py`: 
https://github.com/openvinotoolkit/nncf/pull/3496#discussion_r2169749898

nikita-savelyevv: > I see that flexible group size is closely connected with mixed precision, but I don't feel it's responsibility of this class.

I somewhat agree. This choice was not the first one of mine. But in the end I decided to make it a part of mixed precision algorithm. The new flexible group size logic decides which weights can't be compressed to int4, so it looks like a proper part of mixed precision assignment algorithm. Also, group size is a part of "precision", so selecting different group size values in a sense is also a part of mixed precision logic.

> What about doing it outside and give a modified ratio_defining_params? 

This unfortunately does no look so well if we want to keep this logic: "If ratio < 1.0 and some weights have to be compressed to the backup precision because of group size issues, then these weights contribute to the ratio of backup mode group.". 

This is because we can't just provide a filtered version of `ratio_defining_params` because mixed precision algo still needs to know how many weights there are in total to properly account for excluded weights when selecting primary precision nodes according to the ratio.

What do you think?

tests/cross_fw/test_templates/template_test_weights_compression.py
nikita-savelyevv: Added test_flexible_group_size

nncf/quantization/algorithms/weight_compression/mixed_precision.py
nikita-savelyevv: Done

nncf/quantization/algorithms/weight_compression/mixed_precision.py
```diff
@@ -78,27 +75,37 @@ def apply(
         graph: NNCFGraph,
         statistic_points: Optional[StatisticPointsContainer] = None,
         dataset: Optional[Dataset] = None,
-        weight_params: list[WeightCompressionParameters] = None,
-    ) -> None:
+        all_weight_params: list[WeightCompressionParameters] = None,
+        weight_param_candidates: list[WeightCompressionParameters] = None,
```
ljaljushkin: I'd add a docstring, because now it's not super clear why sum is not calculated for `weight_param_candidates`.

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -431,26 +438,118 @@ def _get_ratio_defining_params(
...
+        group_size_data = []
+        for w_params in weight_params:
+            reduction_channel_size, _ = get_reduction_channel_size(w_params.weight_shape, w_params.reduction_axes)
+            if reduction_channel_size % self._group_size == 0:
+                # The weight can be compressed with the given group size, nothing else to do
+                group_size_data.append((w_params, self._group_size))
+                continue
+
+            # Find the maximal power of two that divides reduction_channel_size
+            flexible_group_size = reduction_channel_size & (~reduction_channel_size + 1)
```
ljaljushkin: nice formula!

nncf/quantization/algorithms/weight_compression/mixed_precision.py
ljaljushkin: the revised version with lightweight mixed precision part looks good to me

nncf/quantization/algorithms/weight_compression/mixed_precision.py
nikita-savelyevv: Added

andreyanufr (approved):

src/nncf/quantization/advanced_parameters.py
```diff
@@ -390,6 +416,7 @@ class AdvancedCompressionParameters:
     lora_correction_params: AdvancedLoraCorrectionParameters = field(default_factory=AdvancedLoraCorrectionParameters)
     lora_adapter_rank: int = 256
     backend_params: dict[str, Any] = field(default_factory=dict)
+    group_size_params: AdvancedGroupSizeParameters = field(default_factory=AdvancedGroupSizeParameters)
```
alexsu52: Simple is better than complex. 

```suggestion
    enable_flexible_group_size: bool = False
    min_flexible_group_size: int = 16
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -431,26 +438,118 @@ def _get_ratio_defining_params(
...
-            for weight_param in ratio_defining_params:
-                weight_param.compression_config = primary_config
+            primary_precision_weight_params = valid_weight_params
         else:
-            self._mixed_precision_algo.apply(model, graph, statistics_points, weight_params=ratio_defining_params)
+            primary_precision_weight_params = self._mixed_precision_algo.apply(
+                model,
+                graph,
+                statistics_points,
+                all_weight_params=weight_params,
```
alexsu52: Could you clarify the ratio should be calculated by all model weights or selected weights for compression?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -431,26 +438,118 @@ def _get_ratio_defining_params(
...
+        :param weight_params: Information about weights that are used for calculating ratio between primary and
             backup precisions.
         :param model: The model.
         :param graph: The model graph associated with the model.
         :param statistics_points: Statistics points.
         """
-        primary_config = WeightCompressionConfig(mode=self._mode, group_size=self._group_size)
+        if self._enable_flexible_group_size and self._group_size != -1:
+            # Compute flexible group size values if enabled
+            flexible_group_size_data = self._get_flexible_group_size_data(weight_params)
```
alexsu52: Do you need all weights for calculation flexible group size? What do you think about simplification the function `_get_flexible_group_size_data` for one weight?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: The ratio is computed based on all model weight (ratio defining) parameters. This is described inside .apply() method documentation.

src/nncf/quantization/advanced_parameters.py
nikita-savelyevv: I agree that this could be a better option if there was a single argument. But starting from two it makes sense to create a separate parameter instance in my opinion. Also, in the follow-up PR I plan to introduce `ignore_invalid_group_size` parameter to ignore weights for which we can't set the default group size. The intention was to move it under `AdvancedGroupSizeParameters` too. What do you think?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: Not all the weights are needed for the actual flexible group size calculation. But there is some additional logic inside `_get_flexible_group_size_data`.

For example, providing comprehensive log messages to a user. If we make `_get_flexible_group_size_data` return a group size for a single weight, then all this logic has to go to inside `_set_weight_compression_config` method. Since all this logging is not required when flexible group size is not activated, I believe it shouldn't be there.

What do you think taking this into account?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
alexsu52: What is strategy for ignored scope and the first/last layers?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
alexsu52: ok

src/nncf/quantization/advanced_parameters.py
nikita-savelyevv: Removed AdvancedGroupSizeParameters

src/nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: Made it so that weights for which we can't find valid group size are excluded from `ratio_defining_parameters`. Resulting log table looks like this:
```
┍━━━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┑
│ Weight compression mode   │ % all parameters (layers)   │ % ratio-defining parameters (layers)   │
┝━━━━━━━━━━━━━━━━━━━━━━━━━━━┿━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┿━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┥
│ int8_asym                 │ 33% (28 / 158)              │ 0% (0 / 130)                           │
├───────────────────────────┼─────────────────────────────┼────────────────────────────────────────┤
│ int4_asym                 │ 67% (130 / 158)             │ 100% (130 / 130)                       │
┕━━━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┙
Applying Weight Compression ━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% • 0:00:07 • 0:00:00
```

PR #3558, opened 2025-06-24
[Common] Disable batchwise statistics for tensor with empty shape

daniil-lyakhov:
### Changes

Batchwise statistics collection is automatically disables when applies to a tensor with an empty shape

### Reason for changes

XNNPACKQuantizer with Swin_v2 model is requesting to quantize a scalar. It is working correctly with the batch size == 1, but during the validation with bigger batch size it raises the error as minmax is trying to aggregate by an non existing dimention
image

### Related tickets

#3231 

### Tests

test_stisticstics_no_batch_size is introduced

### DIFF
--- a/nncf/common/quantization/initialization/range.py
+++ b/nncf/common/quantization/initialization/range.py
@@ -215,8 +215,8 @@ class RangeInitCollectorParams:
     def use_means_of_maxs(self) -> bool:
         return not self._is_weights and not self._is_per_channel
 
+    @staticmethod
     def _get_reduction_axes(
-        self,
         shape_to_reduce: Union[tuple[int, ...], list[int]],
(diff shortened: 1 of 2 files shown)


AlexanderDokuchaev (approved):

AlexanderDokuchaev (changes requested):
Problem in incorrect type of quantization point, activation instead of weight.

daniil-lyakhov:
> Problem in incorrect type of quantization point, activation instead of weight.

To make the correct type of quantization one need to rewrite architecture of nncf graph / metatypes building. As it is not really feasible for such a small bug, I'll suggest to close this pr. Reference https://github.com/openvinotoolkit/nncf/issues/3576

PR #3559, opened 2025-06-24
[Solver] Move most of quantizer_config into the QuantizerConfig itself

daniil-lyakhov:
### Changes

Most QuantizerConfig logic is moved to the QuantizerConfig class

### Reason for changes

To separate logic of QuantizerConfig from the Solver to make it easier to process new QuantizerConfigs in future

### DIFF
--- a/nncf/common/quantization/structs.py
+++ b/nncf/common/quantization/structs.py
@@ -98,6 +98,71 @@ class QuantizerConfig:
     def __hash__(self) -> int:
         return hash(str(self))
 
+    def __lt__(self, other: "QuantizerConfig") -> bool:
+        # Prefer higher bitwidths, per-tensor, symmetrical
+        if self.num_bits > other.num_bits:
+            return True
+        if self.num_bits < other.num_bits:
+            return False
+        if self.per_channel is False and other.per_channel is True:
+            return True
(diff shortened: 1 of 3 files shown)


nncf/common/quantization/structs.py
```diff
@@ -98,6 +98,71 @@ def __str__(self) -> str:
...
+        is_redundant = is_redundant and (
+            (downstream_qconfig.per_channel == self.per_channel)
+            or (downstream_qconfig.per_channel is True and self.per_channel is False)
+        )
+
+        # Strictly prohibit merging of config with different narrow_range params
+        is_redundant = is_redundant and (downstream_qconfig.narrow_range == self.narrow_range)
+        return is_redundant
+
+    def is_compatible(self, other: "QuantizerConfig") -> bool:
```
AlexanderDokuchaev:
```suggestion
    def is_compatible_with(self, other: "QuantizerConfig") -> bool:
```

AlexanderDokuchaev (changes requested):

nncf/common/quantization/structs.py
daniil-lyakhov: Done

PR #3562, opened 2025-06-25
[WeightCompression] Ignored patterns introduction

daniil-lyakhov:
### Changes

Ignored patterns are being introduced to the WeightCompression algorithm

### Reason for changes

ROPE pattern breaks the logic of the WeightCompression algorithm, required to be automatically ignored 

### Related tickets
 https://github.com/huggingface/optimum-intel/issues/1295
 164548

### Tests
test_rope_weight_compression


nncf/common/graph/patterns/manager.py
```diff
@@ -91,6 +92,7 @@ def _filter_patterns(
         patterns_to_filter: dict[PatternNames, Callable[[], GraphPattern]],
         device: TargetDevice,
         model_type: Optional[ModelType] = None,
+        algorithm: Optional[AlgorithmType] = None,
```
nikita-savelyevv: What do you think about renaming `algorithm` variable to `algorithm_type` which is analogous to `model_type`?

nncf/parameters.py
```diff
@@ -47,6 +47,17 @@ class ModelType(StrEnum):
     TRANSFORMER = "transformer"
 
 
+class AlgorithmType(StrEnum):
+    """
+    Describes the algorithm applying to the model.
```
nikita-savelyevv:
```suggestion
    Type of an algorithm that is applied to the model.
```

nikita-savelyevv (commented):
Looks good!

nncf/common/graph/patterns/manager.py
daniil-lyakhov: Great idea, fixed

nikita-savelyevv (approved):

nncf/common/graph/patterns/manager.py
```diff
@@ -91,22 +92,26 @@ def _filter_patterns(
         patterns_to_filter: dict[PatternNames, Callable[[], GraphPattern]],
         device: TargetDevice,
         model_type: Optional[ModelType] = None,
+        algorithm_type: Optional[AlgorithmType] = None,
```
AlexanderDokuchaev: What means None for algorithms_type? and where it's used without set algorithm_type?
It looks like that it should be determinate always.

nncf/parameters.py
```diff
@@ -47,6 +47,17 @@ class ModelType(StrEnum):
...
 
 
+class AlgorithmType(StrEnum):
+    """
+    Type of an algorithm that is applied to the model.
+    :param MINMAX: The MinMaxQuantization algorithm.
+    :param WEIGHTS_COMPRESSION: The WeightCompression algorithm.
+    """
+
+    MINMAX = "MinMaxQuantization"
```
AlexanderDokuchaev:
```suggestion
    MINMAX_QUANTIZATION = "MinMaxQuantization"
```

nncf/common/graph/patterns/manager.py
```diff
@@ -91,22 +92,26 @@ def _filter_patterns(
...
     ) -> dict[PatternNames, Callable[[], GraphPattern]]:
         """
         Returns all patterns from patterns_to_filter that are satisfied device and model_type parameters.
 
         :param patterns_to_filter: Dictionary with the PatternNames instance as keys and creator function as a value.
         :param device: TargetDevice instance.
         :param model_type: ModelType instance.
+        :param algorithm_type: AlgorithmType instance.
```
AlexanderDokuchaev: Not sure that using annotation type as description is useful.

```suggestion
        :param algorithm_type: The algorithm type to apply to the model (e.g., MinMaxQuantization, WeightCompression).
```

nncf/common/graph/patterns/patterns.py
```diff
@@ -281,6 +282,7 @@ class PatternDesc:
     name: str
     devices: Optional[list[TargetDevice]] = None
     model_types: Optional[list[ModelType]] = None
+    algorithms: Optional[list[AlgorithmType]] = None
```
AlexanderDokuchaev:
```suggestion
    algorithm_types: list[AlgorithmType]
```

Using `None` as a default value is generally considered bad design:
- required additional None checks throughout the code, cluttering logic.
- increased the complexity of tests, as you must account for None cases explicitly.
- reduced readability by introducing ambiguity around expected types.

Better to use default_factory:
```
    algorithm_types: list[AlgorithmType] = field(default_factory=lambda: list(AlgorithmType))
```

Or use list to exclude pattern for selected algorithms
```
    excluded_algorithms: list[AlgorithmType] = field(default_factory=list)
```

and update docstring

nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -383,6 +389,17 @@ def get_nodes_to_compress(self, nncf_graph: NNCFGraph) -> list[NNCFNode]:
         ignored_names = get_ignored_node_names_from_ignored_scope(
             self._ignored_scope, nncf_graph, strict=self._ignored_scope.validate
         )
+
+        ignored_patterns = PatternsManager.get_full_ignored_pattern_graph(
+            backend=self._model_backend,
+            device=TargetDevice.ANY,
+            model_type=ModelType.TRANSFORMER,
```
AlexanderDokuchaev: Why is model_type hardcoded as ModelType.TRANSFORMER here? 
Why `device=TargetDevice.ANY`?
ModelType and TargetDevice  is not used for in `weight_compression` anywehere


Looks like current implementation of `PatternsManager` can not be used here.
`IgnoredPatternNames` looks is not valid to use for wc.
And more, there is no reason to use `PatternDesc`, because WC is not use model_type and device.

How about add `IgnoredPatternNamesWC` and `PT_IGNORED_PATTERNS_WC`
```
class PTMinMaxAlgoBackend:
def get_ignored_patterns()
        patterns = Patterns()
        patterns.register(create_rope(), "rope")
        return patterns.get_full_pattern_graph()
... 

autogenerated_ignored_names = get_ignored_names_by_ignored_patterns(nncf_graph, self._backend_entity.get_ignored_patterns())
```

In my mind, it can be done just in backend part, no using common parts or PatternManager. 
This ignored scope actual only for WC, not will be reused anywhere.

nncf/common/graph/patterns/manager.py
```diff
@@ -91,22 +92,26 @@ def _filter_patterns(
...
         """
         filtered_patterns = {}
         for pattern_desc, pattern_creator in patterns_to_filter.items():
             pattern_desc_devices = pattern_desc.value.devices
             pattern_desc_model_types = pattern_desc.value.model_types
+            pattern_desc_algos = pattern_desc.value.algorithms
             devices_condition = pattern_desc_devices is None or device in pattern_desc_devices
             model_types_condition = pattern_desc_model_types is None or model_type in pattern_desc_model_types
-            if devices_condition and model_types_condition:
+            algo_condition = pattern_desc_algos is None or algorithm_type in pattern_desc_algos
```
AlexanderDokuchaev: What is expected behavior for `algorithm_type = None` and `pattern_desc_algos = [MINMAX]`?

nncf/common/graph/patterns/patterns.py
```diff
@@ -403,8 +405,13 @@ class IgnoredPatternNames(Enum):
...
         devices=[TargetDevice.ANY, TargetDevice.CPU, TargetDevice.GPU, TargetDevice.NPU],
+        algorithms=[AlgorithmType.MINMAX],
+    )
+    SE_BLOCK = PatternDesc("se_block", algorithms=[AlgorithmType.MINMAX])
+    FC_BN_HSWISH_ACTIVATION = PatternDesc("fc_bn_hswish_activation", algorithms=[AlgorithmType.MINMAX])
+    EQUAL_LOGICALNOT = PatternDesc("equal_logicalnot", algorithms=[AlgorithmType.MINMAX])
+    ROPE = PatternDesc(
+        "rope",
+        model_types=[ModelType.TRANSFORMER],
+        algorithms=[AlgorithmType.MINMAX, AlgorithmType.WEIGHTS_COMPRESSION],
```
AlexanderDokuchaev:
```suggestion
```
Will have same behavior?

AlexanderDokuchaev (changes requested):

nncf/quantization/algorithms/weight_compression/algorithm.py
daniil-lyakhov:
```
Looks like current implementation of PatternsManager can not be used here.
```
The code you are suggesting will produce exactly the same patterns. I can create a method called `get_ignored_patterns` where I would describe, that currently only patterns for transformers and device agnostic are supported, so we are calling pattern manager to get this exact patterns. Perhaps in future WC will have target device, model type - so we have to use the pattern manager.

 I don't really understand why should we make another pattern manager for the WC when we already have one. It is not a good practice, maybe, but it doesn't break any logic (Pattern manager, please give me a pattern), give me an example when this code will fail
 
 PS
 I'll do it your way as I don't think this discussion will lead to anything productive

daniil-lyakhov: Done
hope + 4 backend-specific functions won't be a problem

daniil-lyakhov:
@AlexanderDokuchaev, I removed my approach and introduced your suggestion, please check

src/nncf/scopes.py
```diff
@@ -270,6 +271,21 @@ def get_ignored_node_names_from_ignored_scope(
     return {name for match in matches.values() for name in match}
 
 
+def get_ignored_names_by_ignored_patterns(nncf_graph: NNCFGraph, ignored_patterns: GraphPattern) -> set[str]:
```
AlexanderDokuchaev: why this function moved to scopes.py?

src/nncf/scopes.py
```diff
@@ -270,6 +271,21 @@ def get_ignored_node_names_from_ignored_scope(
...
+    """
+    Returns node names matched ignored_patterns.
+
+    :param nncf_graph: NNCFGraph instance.
+    :param ignored_patterns: Ignored patterns.
+    :return: IgnoredScope with all node names matched ignored_patterns.
```
AlexanderDokuchaev: Incorrect description

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -345,27 +349,37 @@ def set_backend_entity(self, model: TModel) -> None:
...
             raise nncf.UnsupportedBackendError(msg)
 
+    def get_ignored_patterns(self) -> GraphPattern:
+        """
+        Returns backend-specific ignored patterns for the Weight Quantization algorithm.
+
+        :return: Backend-specific ignored patterns for the Weight Quantization algorithm
+        """
+        patterns = Patterns()
+        patterns.register(self._backend_entity.create_rope_pattern(), "rope")
```
AlexanderDokuchaev: Add one backend specific function that will return list of patterns, to add patterns in them without adding new functions for each patterns.

daniil-lyakhov: Then we will have different set of patterns for different backends, do you think this way is sustainable?
 I did exactly as it was suggested by you in the last comment

src/nncf/scopes.py
daniil-lyakhov: Why not? `get_ignored_node_names_from_ignored_scope` is here, and `get_ignored_names_by_ignored_patterns` was a static method of MinMaxAlgorithm for some reasons despite being almost a copy of the `get_ignored_node_names_from_ignored_scope`

src/nncf/scopes.py
daniil-lyakhov: Done

src/nncf/scopes.py
AlexanderDokuchaev: The function `get_ignored_names_by_ignored_patterns` is not connected with this scopes.
Scope.py already collects function `get_ignored_node_names_from_subgraph` that used in `get_ignored_node_names_from_ignored_scope`.

What a reason add to add yet another function with the same functionality?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
AlexanderDokuchaev: I did not suggest to add function create_rope_pattern.
PatternManager is not guaranty that set of patterns is equal too

src/nncf/scopes.py
daniil-lyakhov: I was mistaken, they are different. I moved the `get_ignored_names_by_ignored_patterns ` to `/src/nncf/common/graph/graph.py`

src/nncf/quantization/algorithms/weight_compression/algorithm.py
daniil-lyakhov: Done

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -278,6 +279,7 @@ def __init__(
         self._ratio = ratio
         self._ignored_scope = ignored_scope
         self._backend_entity = None
+        self._model_backend = None
```
AlexanderDokuchaev: self._model_backend is unused as class member, set and used only in `set_backend_entity`

src/nncf/common/graph/graph.py
```diff
@@ -810,3 +810,18 @@ def find_matching_subgraphs(self, patterns: GraphPattern, strict: bool = True) -
...
         return output
+
+
+def get_ignored_names_by_ignored_patterns(nncf_graph: NNCFGraph, ignored_patterns: GraphPattern) -> set[str]:
+    """
+    Returns node names matched ignored_patterns.
+
+    :param nncf_graph: NNCFGraph instance.
+    :param ignored_patterns: Ignored patterns.
+    :return: A set of NNCF node names from the given NNCFGraph matched ignored_patterns.
```
AlexanderDokuchaev:
```suggestion
def get_node_names_matching_graph_pattern(nncf_graph: NNCFGraph, graph_pattern: GraphPattern) -> set[str]:
    """
    Returns the names of nodes in the given NNCFGraph that match the specified graph pattern.

    :param nncf_graph: An instance of NNCFGraph to search for matching subgraphs.
    :param graph_pattern: A GraphPattern instance used to identify matching subgraphs.
    :return: A set of node names from the NNCFGraph that match the given graph pattern.
    """
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -383,6 +385,12 @@ def get_nodes_to_compress(self, nncf_graph: NNCFGraph) -> list[NNCFNode]:
         ignored_names = get_ignored_node_names_from_ignored_scope(
             self._ignored_scope, nncf_graph, strict=self._ignored_scope.validate
         )
+
+        ignored_patterns_names = get_ignored_names_by_ignored_patterns(
+            nncf_graph, self._backend_entity.get_ignored_patterns()
+        )
+        ignored_names = ignored_names.union(ignored_patterns_names)
```
AlexanderDokuchaev:
```suggestion
        autogenerated_ignored_names = get_node_names_matching_graph_pattern(
            nncf_graph, self._backend_entity.get_ignored_patterns()
        )
        ignored_names = ignored_names.union(autogenerated_ignored_names)
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
daniil-lyakhov: Done

AlexanderDokuchaev (approved):

PR #3579, opened 2025-07-09
[Torch FX] Changes to Support OpenVINO Quantizer

anzr299:
### Changes

1. Add method to set ignored scope in Weight Compression algorithm. 

### Reason for changes

1. This is to allow the user of OpenVINO quantizer to set the ignored scope from the instance without passing it in the quantizer initialization.


alexsu52 (approved):
LGTM

PR #3587, opened 2025-07-17
[OpenVINO] A multihead_attention_output with ReadValue ignored pattern

daniil-lyakhov:
### Changes

* A new multihead_attention_output ignored pattern is introduced for the OpenVINO backend: attention output with a ReadVALUE branching node
* Redundant read_variable_metatypes method is removed from MinMax backends
### Reason for changes

To fix quantization scheme for stateful optimum model with self-attention blocks
https://github.com/huggingface/optimum-intel/blob/main/optimum/exporters/openvino/stateful.py#L261
Before
image

After
image

### Related tickets

170792

### Tests

* tests/openvino/native/quantization/test_quantizer_config.py::test_self_attn_output_with_read_value
* The tiny-random-t5 model (optimum-intel) quantization is checked manualy

### DIFF
--- a/src/nncf/version.py
+++ b/src/nncf/version.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.18.0.dev0+6687d1091"
+__version__ = "2.18.0"
 
 
 BKC_TORCH_SPEC = "==2.7.*"
--- a/src/nncf/version.py
+++ b/src/nncf/version.py
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.18.0"
+__version__ = "2.18.0.dev0+6687d1091"
 
 
 BKC_TORCH_SPEC = "==2.7.*"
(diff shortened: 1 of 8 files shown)


src/nncf/version.py
```diff
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.18.0"
+__version__ = "2.18.0.dev0+6687d1091"
```
nikita-savelyevv:
```suggestion
__version__ = "2.18.0"
```

tests/cross_fw/test_templates/test_quantizer_config.py
```diff
@@ -371,6 +371,10 @@ def test_model_type_transformer_quantization_config(self, transformer_nncf_graph
         state["quantization_points"][6]["directly_quantized_operator_node_names"] = sorted(
             state["quantization_points"][6]["directly_quantized_operator_node_names"]
         )
+        # helper to find the diff in qconfigs
+        # pip install dictdiffer
+        # from dictdiffer import diff
+        # diff_res = list(diff(state, self.REF_TRANSFORMER_SETUP_STATE))
```
nikita-savelyevv: Does this need to stay?

nikita-savelyevv (approved):
LGTM

src/nncf/version.py
daniil-lyakhov: Thanks! Done

tests/cross_fw/test_templates/test_quantizer_config.py
daniil-lyakhov: Removed

PR #3593, opened 2025-07-24
[TorchFX] TorchAOAdapter microfixes

daniil-lyakhov:
### Changes
In `TorchAOAdapter`:
* _get_node_args is used for the input quantization point branch
* A possible case with `quant_min` or `quant_max` are None is managed (faced with CoreMLQuantizer)

### Reason for changes

* To fix some issues of `quantize_pt2e` quantization with the `CoreMLQuantizer`

### Related tickets

169070

### Tests
in `test_quantizer.py`:
* test_none_q_min_q_max_quantizer
* test_adapter_inp_concat_idx

### DIFF
--- a/src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
+++ b/src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
@@ -98,7 +98,7 @@ class TorchAOQuantizerAdapter(Quantizer):
 
         qps = []
         for to_n_ in to_nodes:
-            input_port_id = to_n_.args.index(from_node)
+            input_port_id = TorchAOQuantizerAdapter._get_node_args(to_n_).index(from_node)
             qip = ActivationQuantizationInsertionPoint(to_n_.name, input_port_id)
             qp = SingleConfigQuantizationPoint(qip, qconfig, [to_n_.name])
             qps.append(qp)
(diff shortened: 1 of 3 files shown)


src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
```diff
@@ -166,7 +166,12 @@ def get_quantizer_config_from_annotated_model(annotated: torch.fx.GraphModule) -
                 if qspec.qscheme in [torch.per_channel_symmetric, torch.per_tensor_symmetric]
                 else QuantizationMode.ASYMMETRIC
             )
-            narrow_range = qspec.quant_max - qspec.quant_min == 254
+
+            if qspec.quant_min is None or qspec.quant_max is None:
```
AlexanderDokuchaev, line 163: Please, add a comment about when None is expected, None as quant_max/min is unusual case for NNCF

src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
```diff
@@ -98,7 +98,7 @@ def _get_quantization_points(
 
         qps = []
         for to_n_ in to_nodes:
-            input_port_id = to_n_.args.index(from_node)
+            input_port_id = TorchAOQuantizerAdapter._get_node_args(to_n_).index(from_node)
```
AlexanderDokuchaev: You got two same functions
https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py#L108-L117
https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/experimental/torch/fx/transformations.py#L448-L458
Looks like it should be one function in node_utils.py

And please rename `to_n_`, it's not clear what means `n_` https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#471-names-to-avoid

src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
daniil-lyakhov: Done

src/nncf/experimental/torch/fx/quantization/quantizer/torch_ao_adapter.py
daniil-lyakhov: Done

PR #3599, opened 2025-07-28
[TorchFX] quantize_pt2e Ignored Scope

daniil-lyakhov:
### Changes

New parameter: `ignored_scope` is introduced in the `quantize_pt2e` API

### Reason for changes

To make it possible to ignore some nodes for external quantization.
Particularly - to enable performant quantization of Yolo12 models with `XNNPACKQuantizer` ignoring the postprocessing branch of the models

### Related tickets

169070

### Tests

tests/torch2/fx/test_quantizer.py::test_quantize_pt2e_ignored_scope is introduced


alexsu52 (commented):
Do you have option to handle it outside quantize_pt2e?

PR #3616, opened 2025-08-12
[Torch FX] Introduce a New Method To Obtain Reduction Axes From Metatype

anzr299:
### Changes

abstract method `get_reduction_axes` accepts NNCF node as input. This limits its use in the OpenVINO quantizer when trying to re-use the method for obtaining reduction axes. To tackle this issue, the method `get_reduction_axes` calls another re-usable method `get_reduction_axes_from_node` such that this new method accepts the metatype only instead of the NNCFNode instance. This allows the newly created `get_reduction_axes_from_node` method to be re-used outside of NNCF without obtaining the NNCFNode. 

### Reason for changes

Re-use NNCF backend method outside of NNCF without the need for creating an NNCF graph everytime to obtain the NNCF Node instance. Instead, we directly obtain the metatype from the torch.fx.Node instance.


src/nncf/quantization/algorithms/weight_compression/torch_backend.py
```diff
@@ -82,19 +85,9 @@ class PTWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
...
-        om.PTDepthwiseConv1dSubtype,
-        om.PTDepthwiseConv2dSubtype,
-        om.PTDepthwiseConv3dSubtype,
-        om.PTConvTranspose1dMetatype,
-        om.PTConvTranspose2dMetatype,
-        om.PTConvTranspose3dMetatype,
-    ]
+    MATMUL_METATYPES = MATMUL_METATYPES
+    EMBEDDING_METATYPES = EMBEDDING_METATYPES
+    CONVOLUTION_METATYPES = CONVOLUTION_METATYPES
```
daniil-lyakhov:
```suggestion
```

src/nncf/quantization/algorithms/weight_compression/torch_fx_backend.py
```diff
@@ -62,9 +65,9 @@
 
 
 class FXWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
-    MATMUL_METATYPES = PTWeightCompressionAlgoBackend.MATMUL_METATYPES
-    EMBEDDING_METATYPES = PTWeightCompressionAlgoBackend.EMBEDDING_METATYPES
-    CONVOLUTION_METATYPES = PTWeightCompressionAlgoBackend.CONVOLUTION_METATYPES
+    MATMUL_METATYPES = MATMUL_METATYPES
+    EMBEDDING_METATYPES = EMBEDDING_METATYPES
+    CONVOLUTION_METATYPES = CONVOLUTION_METATYPES
```
daniil-lyakhov:
```suggestion
```

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
     return (0,)
 
 
+def get_reduction_axes_from_metatype(metatype: OperatorMetatype, weight_port_id: int, ndims: int):
```
daniil-lyakhov: Looks like this function is never used. Please remove previous implementation and use this function

daniil-lyakhov: Is it possible to reuse this function in the `torch_backend.py`?

daniil-lyakhov (changes requested):

src/nncf/quantization/algorithms/weight_compression/torch_backend.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/torch_fx_backend.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    """
+    Determine the axes along which weight tensor reduction should occur for a given operator metatype.
+
+    Args:
+        metatype: The metatype of the operator node containing the weight.
+        weight_port_id: The index of the input port corresponding to the weight tensor.
+        ndims: Number of dimensions in the weight tensor.
+
+    Returns:
+        List of axes to reduce over, or None if no reduction axes are determined.
```
daniil-lyakhov: wrong formating

src/nncf/torch/model_graph_manager.py
anzr299: > Looks like this function is never used. Please remove previous implementation and use this function

Oh Alright



> Is it possible to reuse this function in the `torch_backend.py`?

For FX backend, we have om.PTAtenEmbeddingMetatype which is different from the regular pytorch embedding metatype. I can add this exception in the torch_backend as a step after `get_reduction_axes_from_metatype` is called.

src/nncf/torch/model_graph_manager.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,41 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
 
 
+def get_reduction_axes_from_metatype(metatype: OperatorMetatype, weight_port_id: int, ndims: int):
+    """
+    Determine the axes along which weight tensor reduction should occur for a given operator metatype.
+
+    :param metatype: The metatype of the operator node containing the weight.
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+
```
daniil-lyakhov:
```suggestion
```

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,41 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
+    """
+    reduction_axes = None
+    if metatype == om.PTAtenEmbeddingMetatype:
```
daniil-lyakhov:
```suggestion
    if metatype in [om.PTAtenEmbeddingMetatype, PTEmbeddingMetatype]:
```

anzr299: 🤦 
That is much better, thanks!

src/nncf/torch/model_graph_manager.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,40 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
     return (0,)
 
 
+def get_reduction_axes_from_metatype(metatype: OperatorMetatype, weight_port_id: int, ndims: int):
```
daniil-lyakhov:
```suggestion
def get_weight_compression_reduction_axes(metatype: OperatorMetatype, weight_port_id: int, ndims: int) -> Optional[list[int]]:
```

tests/torch2/fx/test_compress_weights.py
```diff
@@ -314,6 +316,24 @@ def test_model_devices_and_precisions(use_cuda, dtype):
...
+        (om.PTLinearMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 1, 4, [2]),
+        (om.PTAddmmMetatype, 1, 4, [3]),
+        (om.PTAddmmMetatype, 2, 4, [2]),
+        (om.PTMatMulMetatype, 1, 1, [0]),
+        (om.PTMatMulMetatype, 1, 2, [0]),
+    ],
+)
+def test_basic_mappings(metatype, weight_port_id, ndims, expected_reduction_axes):
```
daniil-lyakhov:
```suggestion
def test_get_reduction_axes(metatype, weight_port_id, ndims, expected_reduction_axes):
```

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,40 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    """
+    Determine the axes along which weight tensor reduction should occur for a given operator metatype.
```
daniil-lyakhov:
```suggestion
    Returns reduction axes for the given parameters without axes that corresponds to weight channels of a node with the given metatype.
```

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,40 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+
+    :param metatype: The metatype of the operator node containing the weight.
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
```
daniil-lyakhov: Axis None means reduction along every dimension, or not? When does it occur?

anzr299: This is the case which should not be supported i.e all the if conditions should cover all the metatypes with weights.

src/nncf/torch/model_graph_manager.py
anzr299: Done

tests/torch2/fx/test_compress_weights.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
daniil-lyakhov: Let's raise nncf.InternalError than, to be precise here

anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,46 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    """
+    Returns reduction axes for the given parameters without axes that corresponds to weight channels of a node with the
+    given metatype.
+
+    :param metatype: The metatype of the operator node containing the weight.
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
+    """
+    reduction_axes = None
```
daniil-lyakhov: After a consideration, I think it would be clearer to use `return` statement instead of a variable here (and no eiif statements)
and please update the return typehint

anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    :param metatype: The metatype of the operator node containing the weight.
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
+    """
+    if metatype in [om.PTAtenEmbeddingMetatype, om.PTEmbeddingMetatype]:
+        return [1]
+    elif metatype == om.PTLinearMetatype:
+        return [ndims - 1]
+    elif metatype == om.PTMatMulMetatype:
```
daniil-lyakhov:
```suggestion
    if metatype == om.PTLinearMetatype:
        return [ndims - 1]
    if metatype == om.PTMatMulMetatype:
```
and so on...
Please don't use eilf / else + return, please use only if + return. 

It's strange that the pre-commit was not triggered by this code, pylint did trigger errors for such code before

anzr299: Done

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
+    """
+    if metatype in [om.PTAtenEmbeddingMetatype, om.PTEmbeddingMetatype]:
+        return [1]
+    if metatype == om.PTLinearMetatype:
+        return [ndims - 1]
+    if metatype == om.PTMatMulMetatype:
+        if weight_port_id == 0:
```
daniil-lyakhov, line 390: Sorry for a long review, but
what if weight_port_id not in `[0, 1]`?
Previously, code would fail doing `tuple(None)`, now it would return None. What is the desired behavior in such case?

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+        channel_idx = (
+            1
+            if metatype in [om.PTConvTranspose1dMetatype, om.PTConvTranspose2dMetatype, om.PTConvTranspose3dMetatype]
+            else 0
+        )
+        return [i for i in range(ndims) if i != channel_idx]
+    else:
+        msg = f"""The given metatype {metatype} with weight on {weight_port_id} 
+        does not map to a pre-defined reduction axes"""
+        raise nncf.InternalError(msg)
```
daniil-lyakhov:
```suggestion
    msg = f"""The given metatype {metatype} with weight on {weight_port_id} 
    does not map to a pre-defined reduction axes"""
    raise nncf.InternalError(msg)
```

daniil-lyakhov: Please move msg to the top and raise the error not only when there is an unknown metatype, but an unknown weight port id as well

daniil-lyakhov: And please use `""` instead of `""""""`

src/nncf/torch/model_graph_manager.py
anzr299: In this case, it returns an error and exits I believe (flowing to the else case.)
The desired behavior would be an error message to the user since it is not supported

src/nncf/torch/model_graph_manager.py
anzr299: > Please move msg to the top and raise the error not only when there is an unknown metatype, but an unknown weight port id as well

In this case, the error is thrown only when unknown metatype, port_id or an unknown combination of both is passed.



> And please use `""` instead of `""""""`

Done!

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
 
+def get_weight_compression_reduction_axes(metatype: OperatorMetatype, weight_port_id: int, ndims: int) -> list[int]:
+    """
+    Returns reduction axes for the given parameters without axes that corresponds to weight channels of a node with the
+    given metatype.
+
+    :param metatype: The metatype of the operator node containing the weight.
+    :param weight_port_id: The index of the input port corresponding to the weight tensor.
+    :param ndims: Number of dimensions in the weight tensor.
+    :return: list of axes to reduce over, or None if no reduction axes are determined.
```
AlexanderDokuchaev: None is never returns

src/nncf/torch/model_graph_manager.py
```diff
@@ -371,6 +372,43 @@ def get_weight_channel_axes(metatype: type[OperatorMetatype], ndims: int, input_
...
+    """
+    msg = f"The given metatype {metatype} with weight on {weight_port_id} does not map to a pre-defined reduction axes"
```
AlexanderDokuchaev: Define msg above of raise 
```
msg = "..."
raise nncf.InternalError(msg)
```

tests/torch2/fx/test_compress_weights.py
```diff
@@ -314,6 +316,24 @@ def test_model_devices_and_precisions(use_cuda, dtype):
...
+        (om.PTLinearMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 1, 4, [2]),
+        (om.PTAddmmMetatype, 1, 4, [3]),
+        (om.PTAddmmMetatype, 2, 4, [2]),
+        (om.PTMatMulMetatype, 1, 1, [0]),
+        (om.PTMatMulMetatype, 1, 2, [0]),
+    ],
+)
+def test_get_reduction_axes(metatype, weight_port_id, ndims, expected_reduction_axes):
```
AlexanderDokuchaev: Test should be defined in tests/torch/test_model_graph_manager.py as other tests for functions from src/nncf/torch/model_graph_manager.py
Test contain only part of metatypes that supports by get_weight_compression_reduction_axes

AlexanderDokuchaev (changes requested):

tests/torch2/fx/test_compress_weights.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
anzr299: Done

src/nncf/torch/model_graph_manager.py
anzr299: Removed the none case from docstring

tests/torch/test_model_graph_manager.py
```diff
@@ -370,6 +371,24 @@ def test_update_fused_bias(model_cls):
...
+    [
+        (om.PTAtenEmbeddingMetatype, 0, 4, [1]),
+        (om.PTLinearMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 0, 4, [3]),
+        (om.PTMatMulMetatype, 1, 4, [2]),
+        (om.PTAddmmMetatype, 1, 4, [3]),
+        (om.PTAddmmMetatype, 2, 4, [2]),
+        (om.PTMatMulMetatype, 1, 1, [0]),
+        (om.PTMatMulMetatype, 1, 2, [0]),
+    ],
```
AlexanderDokuchaev, line 409: Test contain only part of metatypes that supports by get_weight_compression_reduction_axes, add all other

AlexanderDokuchaev (changes requested):

tests/torch/test_model_graph_manager.py
anzr299: Oh
I missed that in the previous comment.
Done

AlexanderDokuchaev (approved):

PR #2708, opened 2024-05-29
Andreyan/awq extention

andreyanufr:
### Changes
Extended AWQ algorithms for patterns Act->MatMul and Act->Multiply->MatMul with insertion for extra scales after activation.

### Reason for changes
Support AWQ for wider family of LLMs

### Related tickets
CVS-141131

### Tests
Added unit tests

PR #3621, opened 2025-08-14
[AWQ] Remove Code to obtain Merge Node

anzr299:
### Changes

Remove Code to obtain Merge Node. Instead directly get the merge node from the match nodes.

### Reason for changes

`Matmul1 -> mul -> Matmul2` pattern was causing an error when the precision for `Matmul1` was FP32. The first matmul node was not being found in `all_weight_params`.

According to my experiments, in both the conditions for patterns `MatMul->Multiply->MatMul` and `Act->MatMul or Act->Multiply->MatMul` we essentially do the same task of obtaining the first node from the matched list of nodes.

### Tests

-- | -- | -- | -- | --
GH Examples | Examples | Pass| <a href="https://github.com/openvinotoolkit/nncf/actions/runs/16972167283">382</a> |
GH WC | Weights Compression | Pass | <a href="https://github.com/openvinotoolkit/nncf/actions/runs/16972170869">137</a> |

Perplexity Evaluation:
Compression Settings: Ratio=0.8, Hessian Input Activation, backup_mode NONE
Evaluation Settings: limit 10(evaluate with only 10 samples in lm_eval)
Model: microsoft/Phi-3-mini-128k-instruct

| Branch | AWQ | Accuracy |
|--------|--------|--------|
Develop Branch | No | 11.157 |
Develop Branch | Yes | 11.0583 |
After Change | Yes | 11.0583 |

### Related tickets

172326


daniil-lyakhov:
@anzr299, please check https://github.com/openvinotoolkit/nncf/pull/2708 and ticket 141131. Looks like this code was added to support some LLMs, we need to make sure that this models are compressed as expected

anzr299:
> @anzr299, please check #2708 and ticket 141131. Looks like this code was added to support some LLMs, we need to make sure that this models are compressed as expected

Ah I see, I will apply another workaround instead.

anzr299:
> @anzr299, please check #2708 and ticket 141131. Looks like this code was added to support some LLMs, we need to make sure that this models are compressed as expected

I have verified by checking the accuracy of the microsoft/Phi-3-mini-128k-instruct model before and after the change with AWQ enabled. It is the same. Looks like this change does not affect the model.

AlexanderDokuchaev (approved):

PR #3627, opened 2025-08-19
[pt] Fix process_model_inputs for python3.9

AlexanderDokuchaev:
### Changes

Filter signature of forward function to exclude self argument before binding arguments.

### Reason for changes

For python < 3.10 in case of combination of decorated forward method and overriding by partial `inspect.signature` got differ behavior:

```
# python 3.9 
signature=<Signature (self, x: int) -> int>

#python3.10
signature=<Signature (x: int) -> int>

#python3.13
signature=<Signature (x: int) -> int>
```

### Related tickets

171692

### Tests

tests/torch2/function_hook/test_wrapper.py::test_wrapper

### DIFF
--- a/src/nncf/torch/function_hook/hook_executor_mode.py
+++ b/src/nncf/torch/function_hook/hook_executor_mode.py
@@ -12,6 +12,7 @@
 from __future__ import annotations
 
 import inspect
+import sys
 import types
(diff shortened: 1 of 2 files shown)


andrey-churkin (approved):
LGTM

PR #3636, opened 2025-08-25
[Weight Compression] Rework WC Algorithm to Return WC Params

anzr299:
### Changes

Re-worked the `apply` method in WC algorithm to use an extra method to return weights compression params such that the apply method is more concise and only contains algorithm, quantization logic etc.

### Reason for changes

This is done so that OpenVINO quantizer can obtain the final collection of weights compression parameters for all the nodes so that are to be compressed.

### DIFF
--- a/src/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/src/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -780,7 +780,7 @@ class WeightCompression(Algorithm):
         Collects and processes weight compression parameters for all nodes in the model,
         applies all processing steps including group size fallback handling and compression
         configuration setting.
-        
+
         :param model: Backend-specific input model.
         :param graph: NNCFGraph instance.
         :param statistic_points: Optional pre-collected statistic points.
@@ -790,15 +790,15 @@ class WeightCompression(Algorithm):
             Statistics collected for data-aware compression. None if data-free compression.
         """
         nodes_to_compress = self.get_nodes_to_compress(graph)
-        
+
         initial_all_weight_params: list[WeightCompressionParameters] = []
         skipped_weight_params: list[WeightCompressionParameters] = []
-        
+
         weight_names = set()
         is_last_layer_skipped = False
         n = len(nodes_to_compress)
         ignored_names = self.get_ignored_node_names(graph)
-        
+
         for i, node in enumerate(nodes_to_compress):
             is_target_node = should_consider_scope(node.node_name, ignored_names)
             for weight_name, weight_port_id in self._backend_entity.get_weight_names_and_port_ids(node, graph):
@@ -840,7 +840,7 @@ class WeightCompression(Algorithm):
                         )
                         if self.is_weight_compression_supported(weight_dtype, mode):
                             wc_config = WeightCompressionConfig(mode=mode)
-                            
+
                     weight_params = WeightCompressionParameters(
                         weight_name, node, weight_port_id, weight_dtype, weight_shape, reduction_axes, wc_config
                     )
@@ -890,10 +890,11 @@ class WeightCompression(Algorithm):
             self._get_bitwidth_distribution_str(all_weight_params, ratio_defining_params, skipped_weight_params)
         )
 
-        final_all_weight_params = list(filter(lambda w_params: w_params.compression_config is not None, all_weight_params))
-        
-        return final_all_weight_params, skipped_weight_params, statistics
+        final_all_weight_params = list(
+            filter(lambda w_params: w_params.compression_config is not None, all_weight_params)
+        )
 
+        return final_all_weight_params, statistics
 
     def apply(
         self,
@@ -905,7 +906,7 @@ class WeightCompression(Algorithm):
         self.set_backend_entity(model)
 
         # Get processed weight compression parameters ready for compression
-        all_weight_params, skipped_weight_params, statistics = self.get_processed_weight_compression_parameters(
+        all_weight_params, statistics = self.get_processed_weight_compression_parameters(
             model, graph, statistic_points, dataset
         )
 
--- a/src/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/src/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -769,7 +769,43 @@ class WeightCompression(Algorithm):
 
         return is_supported_dtype and not no_bit_reduction
 
-    def get_processed_weight_compression_parameters(
+    def collect_weight_compression_statistics(
+        self,
+        model: TModel,
+        graph: NNCFGraph,
+        dataset: Dataset,
+        weight_params: list[WeightCompressionParameters],
+        statistic_points: Optional[StatisticPointsContainer] = None,
+    ) -> Optional[dict[str, Any]]:
+        """
+        Collects statistics for weight compression if data-aware compression or
+        mixed-precision is enabled.
+
+        :param model: Backend-specific input model.
+        :param graph: NNCFGraph instance.
+        :param dataset: Dataset for statistics collection.
+        :param weight_params: Weight parameters for which to collect statistics.
+        :param statistic_points: Optional pre-collected statistic points.
+        :return: Statistics and Statistic points container.
+        """
+        statistics = None
+        if not (self._data_aware_mixed_precision or self._data_aware_compression) and not dataset:
+            return statistics, statistic_points
+        matmul_nodes_to_compress = [
+            wp.node_with_weight
+            for wp in weight_params
+            if wp.node_with_weight.metatype in self._backend_entity.matmul_metatypes
+        ]
+        matmul_input_to_output_nodes_map = self.get_matmul_input_to_output_nodes_map(matmul_nodes_to_compress, graph)
+
+        if statistic_points is None:
+            statistic_points = self.get_statistic_points(model, graph, matmul_input_to_output_nodes_map.keys())
+            statistic_points = self._collect_statistics(dataset, graph, model, statistic_points)
+
+        statistics = self._get_statistics_for_weights_compression(matmul_input_to_output_nodes_map, statistic_points)
+        return statistics, statistic_points
+
+    def get_weight_compression_parameters(
         self,
         model: TModel,
         graph: NNCFGraph,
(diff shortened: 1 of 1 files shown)


src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,26 @@ def is_weight_compression_supported(
...
+        Collects and processes weight compression parameters for all nodes in the model,
+        applies all processing steps including mixed precision assignment, group size fallback
+        handling and compression configuration setting.
 
+        :param model: Backend-specific input model.
+        :param graph: NNCFGraph instance.
+        :param statistic_points: Optional pre-collected statistic points.
+        :param dataset: Optional dataset for statistics collection.
+        :return: A tuple containing Final processed list of weight parameters ready for compression,
+            List of weight parameters that were skipped
```
daniil-lyakhov, line 789:
```suggestion
            the list of weight parameters that were skipped
```
Could you please elaborate what does it mean? What does `skip weight parameters` mean?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,26 @@ def is_weight_compression_supported(
...
+            Statistics collected for data-aware compression. None if data-free compression.
```
daniil-lyakhov, line 790:
```suggestion
            statistics collected for data-aware compression. None if data-free compression.
```
Does None related to the last element in return tuple? It is not clear

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -884,6 +897,22 @@ def apply(
...
+        self,
+        model: TModel,
+        graph: NNCFGraph,
+        statistic_points: Optional[StatisticPointsContainer] = None,
+        dataset: Optional[Dataset] = None,
+    ) -> TModel:
+        self.set_backend_entity(model)
+
+        # Get processed weight compression parameters ready for compression
+        all_weight_params, statistics = self.get_processed_weight_compression_parameters(
```
daniil-lyakhov: I got from the docstring that there are three elements in the output tuple. But looks like it is only two. Please update the docstring

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,26 @@ def is_weight_compression_supported(
...
         graph: NNCFGraph,
         statistic_points: Optional[StatisticPointsContainer] = None,
         dataset: Optional[Dataset] = None,
-    ) -> TModel:
-        self.set_backend_entity(model)
+    ) -> tuple[list[WeightCompressionParameters], list[WeightCompressionParameters], dict[str, Any]]:
+        """
+        Collects and processes weight compression parameters for all nodes in the model,
+        applies all processing steps including mixed precision assignment, group size fallback
+        handling and compression configuration setting.
```
daniil-lyakhov: Please replace with
"""
Generates a list of weight compression parameters based on the Weight Compression algorithm configuration. Determines the appropriate precision, group size, and other parameters for each node eligible for weight compression.
"""

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,26 @@ def is_weight_compression_supported(
...
 
+        :param model: Backend-specific input model.
+        :param graph: NNCFGraph instance.
+        :param statistic_points: Optional pre-collected statistic points.
+        :param dataset: Optional dataset for statistics collection.
+        :return: A tuple containing Final processed list of weight parameters ready for compression,
```
```suggestion
        :return: A list of weight compression parameters based on the Weight Compression algorithm configuration.
```

As mentioned previously, let's return only the list of parameters from this function

daniil-lyakhov (changes requested):
Aamir, I think it would be beneficial to split this function on two: `get_weight_compression_parameters` and `get_weight_compression_statistics` as it seem like the statistic collection does not relate to the collection of weight parameters in any way and could be split without any effort. This way it would be clear: statistic points and datasets have nothing to do with the choice of the WC configuration.

Other thing - what `processed` parameters means? What is an unprocessed parameter than?

anzr299:
> Aamir, I think it would be beneficial to split this function on two: `get_weight_compression_parameters` and `get_weight_compression_statistics` as it seem like the statistic collection does not relate to the collection of weight parameters in any way and could be split without any effort. This way it would be clear: statistic points and datasets have nothing to do with the choice of the WC configuration.
> 
> Other thing - what a `processed` parameters means? What is an unprocessed parameter than?

Sure, that sounds good.
as for "processed" parameters, what I mean is that originally all the WC parameters are set to backup precision and in the later steps of this function, the parameters are changed as per user defined WC configurations like all_layers, backup_mode and so on.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: ah I missed that. Done

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: Done

daniil-lyakhov: Please make sure you pushed your local commits

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: 🤦 
Done now for real!

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -884,6 +918,20 @@ def apply(
...
+        self,
+        model: TModel,
+        graph: NNCFGraph,
+        statistic_points: Optional[StatisticPointsContainer] = None,
+        dataset: Optional[Dataset] = None,
+    ) -> TModel:
+        self.set_backend_entity(model)
+
+        # Get processed weight compression parameters ready for compression
+        all_weight_params, statistics = self.get_weight_compression_parameters(model, graph, statistic_points, dataset)
```
daniil-lyakhov, line 914:
```suggestion
        all_weight_params = self.get_weight_compression_parameters(model, graph, statistic_points, dataset)
        statistics = self.get_weight_compression_statistics(...)
```
I mean - can we do this? If it not possible - it's better to not separate the function

anzr299: It could be possible with some hacks but that would make it complicated. 
the statistic_points container is required when setting the compression config.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,24 @@ def is_weight_compression_supported(
...
+        """
+        Generates a list of weight compression parameters based on the Weight Compression algorithm
+        configuration. Determines the appropriate precision, group size, and other parameters for
+        each node eligible for weight compression.
 
+        :param model: Backend-specific input model.
+        :param graph: NNCFGraph instance.
+        :param statistic_points: Optional pre-collected statistic points.
+        :param dataset: Optional dataset for statistics collection.
+        :return: A list of weight compression parameters based on the Weight Compression algorithm configuration.
```
daniil-lyakhov: A tuple of ...

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,24 @@ def is_weight_compression_supported(
...
-    def apply(
+    def get_weight_compression_parameters(
         self,
         model: TModel,
         graph: NNCFGraph,
         statistic_points: Optional[StatisticPointsContainer] = None,
         dataset: Optional[Dataset] = None,
-    ) -> TModel:
-        self.set_backend_entity(model)
+    ) -> tuple[list[WeightCompressionParameters], dict[str, Any]]:
```
```suggestion
    ) -> tuple[list[WeightCompressionParameters], Optional[dict[str, WCTensorStatistic]]]:
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,24 @@ def is_weight_compression_supported(
...
+        """
+        Generates a list of weight compression parameters based on the Weight Compression algorithm
+        configuration. Determines the appropriate precision, group size, and other parameters for
+        each node eligible for weight compression.
 
+        :param model: Backend-specific input model.
+        :param graph: NNCFGraph instance.
+        :param statistic_points: Optional pre-collected statistic points.
+        :param dataset: Optional dataset for statistics collection.
+        :return: A tuple of weight compression parameters based on the Weight Compression algorithm configuration.
```
daniil-lyakhov: :return: A tuple consisting of a list of weight compression parameters, based on the Weight Compression algorithm configuration, and a mapping of target node names to the collected statistics.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -769,15 +769,24 @@ def is_weight_compression_supported(
...
         graph: NNCFGraph,
         statistic_points: Optional[StatisticPointsContainer] = None,
         dataset: Optional[Dataset] = None,
-    ) -> TModel:
-        self.set_backend_entity(model)
+    ) -> tuple[list[WeightCompressionParameters], dict[str, Any]]:
+        """
+        Generates a list of weight compression parameters based on the Weight Compression algorithm
+        configuration. Determines the appropriate precision, group size, and other parameters for
+        each node eligible for weight compression.
```
daniil-lyakhov: Please add:  Generates a mapping of target node names to the collected statistics based on the provided statistic_points. If statistic_points is None, collects required compression statistics on the given dataset.

daniil-lyakhov (changes requested):

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: But it generates a tuple of weight compression parameters(WeightCompressionParameters objects) and Optional statistics.
Would this be a good docstring instead?
"Generates a tuple containing a list of Weight compression parameters based on the weight compression algorithm configuration and a mapping of target node names to the collected statistics based on the statistic_points. If statistic_points is None, collects required compression statistics the given dataset." 

and since this is redundant with the :return: maybe return can be:
":return: A tuple consisting of a list of weight compression parameters and a mapping of target node names to the collected statistics."

daniil-lyakhov: Can we keep my suggestion, please?

daniil-lyakhov (approved):

AlexanderDokuchaev (approved):

PR #3637, opened 2025-08-25
[pt] update reference for sparsity 

AlexanderDokuchaev:
### Changes

Update expected ov_weight_ratio for some models

### Reason for changes

Change size of original an compressed model 

### Tests

nightly/job/torch_nightly/628/


ljaljushkin:
Could you elaborate on why the NNCF ratio stays the same while the openvino one is changing?

ljaljushkin:
> Could you elaborate on why the NNCF ratio stays the same while the openvino one is changing?

pruning transformation seems to be functional
image

the different ratio is coming from different size of float32 model with new OV
image

ljaljushkin (approved):

PR #3641, opened 2025-08-28
[ONNX] Fix the incorrect conversion of the Gemm into a MatMulNBits

andrey-churkin:
### Changes
- Fixed incorrect inputs for the `MatMulNBits` operation that led to a graph break.
- Fixed incorrect INT4 weight compression in the `Gemm` operation when the `transB` attribute is set to 1.

### Related tickets

Ref: 171011
Ref: 171016

### DIFF
--- a/src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
+++ b/src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
@@ -231,16 +231,27 @@ class ONNXWeightCompressionAlgoBackend(WeightCompressionAlgoBackend):
             node = wc_params.node_with_weight
             weight = self.get_weight(node, wc_params.weight_port_id, model, graph)
             precomputed_compressed_weights = precomputed_compressed_weights or {}
-            compressed_weight = compress_weight(
-                Tensor(weight),
-                wc_params.reduction_axes,
-                compression_config,
-                precomputed_compressed_weights.get(wc_params.weight_name),
-            )
+
             dequantize_block_size = max(compression_config.group_size, 0)  # 0 - is no block wise quantization
             dequantize_axis = (
                 get_weight_quantization_axis(node, wc_params.weight_port_id) if dequantize_block_size <= 0 else 0
             )  # axis = 0 when blockwise
+
+            reduction_axes = wc_params.reduction_axes
+            if node.metatype == onnx_metatypes.ONNXGemmMetatype and opset_version < 21 and dequantize_block_size > 0:
+                attr_name = "transB" if wc_params.weight_port_id == 1 else "transA"
+                transpose = node.layer_attributes.node_attrs[attr_name]
+                weight = fns.transpose(weight) if transpose else weight
+                (axis,) = reduction_axes
+                axis = (axis + 1) % 2 if transpose else axis
+                reduction_axes = (axis,)
+
+            compressed_weight = compress_weight(
+                Tensor(weight),
+                reduction_axes,
+                compression_config,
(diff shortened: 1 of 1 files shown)


src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -428,16 +439,21 @@ def _replace_matmul_with_matmulnbits(
...
-        for input_name in original_matmul.input:
-            if input_name != weight_name:
-                activation_input_name = input_name
-        assert activation_input_name is not None, "Activation input name not found in original matmul node"
+        # Composing operation inputs: A, B, scales, zero_points[optional], g_idx[optional, deprecated], bias
+        bias_name = None
+        if weight_compression_parameters.node_with_weight.layer_attributes.has_bias():
+            bias_name = weight_compression_parameters.node_with_weight.layer_attributes.bias_attrs["name"]
+
+        activation_input_name = next(name for name in original_matmul.input if name not in [weight_name, bias_name])
```
daniil-lyakhov, line 447: If the name wont' be available for some reason an unexplained `StopIteration` will be raised. Is that a possible case?

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -428,16 +439,21 @@ def _replace_matmul_with_matmulnbits(
...
 
-        activation_input_name = None
-        for input_name in original_matmul.input:
-            if input_name != weight_name:
-                activation_input_name = input_name
-        assert activation_input_name is not None, "Activation input name not found in original matmul node"
+        # Composing operation inputs: A, B, scales, zero_points[optional], g_idx[optional, deprecated], bias
+        bias_name = None
+        if weight_compression_parameters.node_with_weight.layer_attributes.has_bias():
+            bias_name = weight_compression_parameters.node_with_weight.layer_attributes.bias_attrs["name"]
```
daniil-lyakhov, line 445: What if the activation name will be None and there is no bias - thus the None activation input will be filtered.. I think a filter list without none would be clearer
```suggestion
        non_activation_inputs = [weight_name]
        if weight_compression_parameters.node_with_weight.layer_attributes.has_bias():
            bias_name = weight_compression_parameters.node_with_weight.layer_attributes.bias_attrs["name"]
            non_activation_inputs.append(bias_name)
```

daniil-lyakhov (commented):
Is it possible to add a test case for the changes?

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
andrey-churkin: We have `len(node.input) >= 2` for [MatMul](https://onnx.ai/onnx/operators/onnx__MatMul.html) and [Gemm](https://onnx.ai/onnx/operators/onnx__Gemm.html) opeartions. There is another input that is different from the weight and bias. So the your case is not possible here.

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
andrey-churkin: As far as I know, the name of an activation cannot be `None` because `node.input` contains only strings. So using `None` here is acceptable. I also use this check when adding other inputs.

```python
if bias_name:
    if zero_point is None:
        inputs.append("")
    inputs.append("")  # g_idx
    inputs.append(bias_name)
```

andrey-churkin:
> Is it possible to add a test case for the changes?

I've just added it.

PR #3657, opened 2025-09-11
[ONNX] Add MatMul for FBC/BC

andrey-churkin:
### Changes

Add support for the `MatMul` operation in the Fast Bias/Bias Correction algorithm.

### Reason for changes

The Fast Bias/Bias Correction algorithm is not applied to the `MatMul`->`Add` subgraph in which one of the inputs to the Add operation is a constant. In this case, the `MatMul` operation is not considered as having a bias, so the algorithm is not applied.

### Related tickets

Ref: 135433

### Tests

- test_update_bias_in_matmul_add
- text_examples: [build](https://github.com/openvinotoolkit/nncf/actions/runs/18529108467)

### DIFF
--- a/tests/onnx/common.py
+++ b/tests/onnx/common.py
@@ -155,6 +155,15 @@ class ModelBuilder:
         )
         return output
 
+    def add_add(self, input_a: str, input_b: str, output: Optional[str] = None) -> str:
+        i = len(self._nodes)
+
+        output = f"Add_{i}_output" if output is None else output
+        self._nodes.append(
+            onnx.helper.make_node(op_type="Add", inputs=[input_a, input_b], outputs=[output], name=f"Add_{i}")
+        )
+        return output
+
     def add_relu(self, input: str, output: Optional[str] = None) -> str:
         i = len(self._nodes)
 
--- a/tests/onnx/quantization/test_fast_bias_correction.py
+++ b/tests/onnx/quantization/test_fast_bias_correction.py
@@ -14,11 +14,15 @@ import numpy as np
 import onnx
 import torch
 
+from nncf.common.factory import ModelTransformerFactory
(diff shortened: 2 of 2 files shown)


src/nncf/onnx/graph/metatypes/onnx_metatypes.py
```diff
@@ -143,7 +143,6 @@ class ONNXMatMulMetatype(ONNXOpWithWeightsMetatype):
     op_names = ["MatMul"]
     hw_config_names = [HWConfigOpName.MATMUL]
     weight_channel_axis = -1  # For port_id=1
-    bias_port_id = 2
```
andrey-churkin, line 146: This does not apply to the `MatMul` ONNX operation, as it does not accept a bias input. Reference: https://onnx.ai/onnx/operators/onnx__MatMul.html

src/nncf/quantization/algorithms/fast_bias_correction/algorithm.py
```diff
@@ -175,13 +175,11 @@ def apply(
 
             output_channel_axis = node.metatype.output_channel_axis
             input_channel_axis = self._backend_entity.get_activation_channel_axis(node, input_port_id, input_shape)
-            if bias_value.ndim > 1:
-                # Make index positive
-                output_channel_axis = range(bias_value.ndim)[output_channel_axis]
-                input_channel_axis = range(bias_value.ndim)[input_channel_axis]
```
andrey-churkin, line 181: It was moved into `create_input_data()` because this method does not work properly with a negative `channel_axis`. The `output_channel_axis` is converted to positive inside `mean_per_channel()` method.

andrey-churkin:
Build post_training_quantization # 740

PR #3662, opened 2025-09-19
[ONNX] Compress quantize weights transformation

andrey-churkin:
### Changes

Added the `compress_quantize_weights_transformation()` method that transforms the model by folding `QuantizeLinear` nodes with constant inputs into precomputed, quantized initializers.

### Reason for changes

- Models after NNCF PTQ should be saved with INT8 weights for the ONNX backend.

### Related tickets

Ref: 101733

### Tests

-  tests/cross_fw/examples/test_examples.py[post_training_quantization_onnx_mobilenet_v2]
- tests/onnx/test_passes.py

### DIFF
--- a/src/nncf/onnx/graph/passes.py
+++ b/src/nncf/onnx/graph/passes.py
@@ -103,7 +103,9 @@ def compress_quantize_weights_transformation(model: onnx.ModelProto):
         if node.op_type != "QuantizeLinear":
             continue
 
-        x_name, y_scale_name, y_zero_point_name = node.input[:3]
+        x_name, y_scale_name = node.input[:2]
+        # `y_zero_point` is an optional input for the `QuantizeLinear` operation.
+        y_zero_point_name = node.input[2] if len(node.input) > 2 else None
 
         if x_name not in initializer:
             continue
@@ -113,7 +115,10 @@ def compress_quantize_weights_transformation(model: onnx.ModelProto):
         # Quantize
         x = onnx.numpy_helper.to_array(initializer[x_name])
         y_scale = onnx.numpy_helper.to_array(initializer[y_scale_name])
-        y_zero_point = onnx.numpy_helper.to_array(initializer[y_zero_point_name])
+
+        y_zero_point = None
+        if y_zero_point_name:
+            y_zero_point = onnx.numpy_helper.to_array(initializer[y_zero_point_name])
 
         axis = get_node_attr_value(node, "axis")
         if version < 21:
--- a/src/nncf/onnx/graph/passes.py
+++ b/src/nncf/onnx/graph/passes.py
@@ -83,11 +83,17 @@ def apply_preprocess_passes(model: onnx.ModelProto) -> onnx.ModelProto:
 
 def compress_quantize_weights_transformation(model: onnx.ModelProto):
     """
-    :param model:
+    Transforms the model by folding `QuantizeLinear` nodes with constant inputs
+    into precomputed, quantized initializers.
+
+    This transformation finds `QuantizeLinear` nodes with constant inputs
+    (i.e., inputs present in the model's initializers), precomputes their quantized values,
+    updates the initializer with these results, and removes the corresponding
+    `QuantizeLinear` nodes from the graph.
+
+    :param model: The model to be transformed.
     """
     initializer = {x.name: x for x in model.graph.initializer}
-    # value_info = {i.name: i for i in model.graph.value_info}
-
     nodes_to_remove = []
 
     version = model.opset_import[0].version
@@ -99,13 +105,12 @@ def compress_quantize_weights_transformation(model: onnx.ModelProto):
 
         x_name, y_scale_name, y_zero_point_name = node.input[:3]
 
-        # NOTE:
         if x_name not in initializer:
             continue
 
         nodes_to_remove.append(node)
 
-        # --- quantize x ---
+        # Quantize
         x = onnx.numpy_helper.to_array(initializer[x_name])
         y_scale = onnx.numpy_helper.to_array(initializer[y_scale_name])
         y_zero_point = onnx.numpy_helper.to_array(initializer[y_zero_point_name])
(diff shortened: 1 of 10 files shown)


andrey-churkin:
Test Examples: https://github.com/openvinotoolkit/nncf/actions/runs/18187232744

src/nncf/onnx/quantization/backend_parameters.py
```diff
@@ -19,11 +19,28 @@ class BackendParameters:
...
+def is_weight_compression_needed(advanced_parameters: Optional[AdvancedQuantizationParameters]) -> bool:
+    """
+    Determines whether weight compression is needed based on the provided
+    advanced quantization parameters.
+
+    :param advanced_parameters: Advanced quantization parameters.
+    :return: `True` if weight compression is needed, `False` otherwise.
+    """
+    if advanced_parameters is not None and advanced_parameters.backend_params is not None:
+        return advanced_parameters.backend_params.get(BackendParameters.COMPRESS_WEIGHTS, True)
```
daniil-lyakhov, line 40: I believe this function should be placed in common. I think that's a good GFI

src/nncf/onnx/quantization/quantize_model.py
```diff
@@ -202,8 +209,13 @@ def quantize_with_accuracy_control_impl(
...
         advanced_accuracy_restorer_parameters = AdvancedAccuracyRestorerParameters()
 
+    compress_weights = is_weight_compression_needed(advanced_quantization_parameters)
+
     if advanced_quantization_parameters is None:
-        advanced_quantization_parameters = AdvancedQuantizationParameters()
+        copied_parameters = AdvancedQuantizationParameters()
+    else:
+        copied_parameters = deepcopy(advanced_quantization_parameters)
+    copied_parameters.backend_params[BackendParameters.COMPRESS_WEIGHTS] = False
```
daniil-lyakhov, line 218: Why should we update this parameter here?

tests/onnx/test_passes.py
```diff
@@ -21,3 +27,43 @@ def test_apply_preprocess_passes():
...
+    quantized_model = nncf.quantize(
+        model,
+        calibration_dataset,
+        advanced_parameters=nncf.AdvancedQuantizationParameters(
+            backend_params={BackendParameters.COMPRESS_WEIGHTS: False}
+        ),
+    )
+
+    check_operation_count(quantized_model, {"QuantizeLinear": 2, "DequantizeLinear": 2})
+    compress_quantize_weights_transformation(quantized_model)
```
daniil-lyakhov, line 68: This test is covering the transformation, but I would suggest to additionally test the `nncf.quantize` with `COMPRESS_WEIGHTS: True` to check that the API is working as expected.

src/nncf/onnx/quantization/backend_parameters.py
daniil-lyakhov: https://github.com/openvinotoolkit/nncf/issues/3668

src/nncf/onnx/graph/passes.py
```diff
@@ -77,3 +79,56 @@ def apply_preprocess_passes(model: onnx.ModelProto) -> onnx.ModelProto:
...
+    nodes_to_remove = []
+
+    version = max(model.opset_import[0].version, 19)
+    QuantizeLinear = load_op("", "QuantizeLinear", version)
+
+    for node in model.graph.node:
+        if node.op_type != "QuantizeLinear":
+            continue
+
+        x_name, y_scale_name, y_zero_point_name = node.input[:3]
```
nikita-savelyevv: Will zero point always be present?

src/nncf/onnx/quantization/backend_parameters.py
```diff
@@ -19,11 +19,28 @@ class BackendParameters:
     """
     :param EXTERNAL_DATA_DIR: An absolute path to the directory where the external data
         files are stored. All external data files must be located in the same folder.
+    :param COMPRESS_WEIGHTS: If `True` compresses constant quantized weights by folding
+        `QuantizeLinear` nodes into pre-quantized initializers. If `False`, this transformation
+        is skipped.
```
nikita-savelyevv, line 24: What is the scenario when constants don't need to be compressed? As I understand it, the result will be the same in any case, just the model size is different, right?

src/nncf/onnx/quantization/quantize_model.py
andrey-churkin: We need to disable `COMPRESS_WEIGHT` here to properly remove Quantize-Dequantize pairs during the `quantize_with_accuracy_control()` pipeline. For reference, we do the same for the OpenVINO backend.

https://github.com/openvinotoolkit/nncf/blob/54aba4ad872090c7f43a5cb236ab1bcab781b197/src/nncf/openvino/quantization/quantize_model.py#L220

tests/onnx/test_passes.py
andrey-churkin: I don't think it is necessary here. We already have an end-to-end test (`tests/cross_fw/examples/test_examples.py[post_training_quantization_onnx_mobilenet_v2]` where we compare the model's compression rate with the reference. So we'll be able to catch the error there if `COMPRESS_WEIGHTS: True` doesn't work as expected.

src/nncf/onnx/graph/passes.py
andrey-churkin: Good catch! According to the [documentation](https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#inputs), the QuantizeLinear may contain between 2 and 3 inputs. I'll fix it.

andrey-churkin: I've just updated the code. Please check

src/nncf/onnx/quantization/backend_parameters.py
andrey-churkin: It seems like this description isn't clear enough. The `COMPRESS_WEIGHTS` parameter (`True` or `False`) determines the format (INT8 or FP) in which the weights of quantized operations will be saved. For a model in ONNX format, this is achieved by applying the `QuantizeLinear` operation to float-precision initializers (i.e., weights).

If you have any suggestions for how we can improve this, please let me know.

PR #3664, opened 2025-09-22
[OpenVINO][WC] E2M1 -> MXFP4 and MXFP8_E4M3 weights compression support

daniil-lyakhov:
### Changes
* E2M1 compression is renamed to MXFP4
* Support of MXFP8_E4M3 mode
* Default group size for `MXFP4` and `MXFP8_E4M3` is 32, other group sizes are failing with an error
* Documentation is updated with `MXFP4` and `MXFP8_E4M3`

### Reason for changes

* To expand compression precisions for the OpenVINO WC algorithm
* To restrict MX types: the definition of MX data types requires group size == 32

### Related tickets

164851

### Tests
* Tests on unsupported compression modes is expanded for non OV backends
* tests/openvino/native/quantization/test_weights_compression.py::test_mixed_precision_fp


src/nncf/quantization/algorithms/weight_compression/openvino_backend.py
```diff
@@ -228,6 +228,12 @@ def _create_compression_subgraph(
         elif compression_config.mode == CompressWeightsMode.E2M1:
             compression_dtype = ov.Type.f4e2m1
             scale_dtype = ov.Type.f8e8m0
+        elif compression_config.mode == CompressWeightsMode.E5M2:
+            compression_dtype = ov.Type.f8e5m2
+            scale_dtype = ov.Type.f8e8m0
```
ljaljushkin: | MODE          | Element   type       | Scale   type | Block   Size | Tensor   Scaling | Status                                                               |
|---------------|----------------------|--------------|--------------|------------------|----------------------------------------------------------------------|
| INT4_SYM      | INT4                 | FP16         | any          | -                | existing, not changed                                                |
| NF4           | LUT                  | FP16         | any          | -                | existing, not changed                                                |
| INT8_ASYM     | INT8                 | FP16         | -1           | -                | existing, not changed                                                |
| CB4_F8E4M3    | FP32   (NF4) -> E4M3 | FP16         | any          | -                | existing, not changed                                                |
| CODEBOOK      | Any                  | FP16         | any          | -                | existing, not changed                                                |
| MXFP4         | E2M1                 | E8M0         | 32           | -                | existing, need <br>to rename from E2M1 + <br>to check for group_size |
| NVFP4         | E2M1                 | E4M3         | 16           | F32              | new                                                                  |
| FP4_E2M1      | E2M1                 | FP16         | any          | -                | new                                                                  |
| MXFP8_E4M3    | E4M3                 | E8M0         | 32           | -                | new                                                                  |
| FP8_E4M3      | E4M3                 | FP16         | any          | -                | new, aligned with existing <br>QuantizationMode.FP8_E4M3             |
| FP8_E4M3_E8M0 | E4M3                 | E8M0         | any          | -                | new, possible extension                                              |
| FP4_E4M1_E8M0 | E2M1                 | E8M0         | any          | -                | new, possible extension                                              |

propose the following mode names and their parameters, and explicitly add this table to docstring of `CompressWeightsMode`.

1. MX-data types should follow the [specification](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf): E8M0 scale and group_size 32. That's why `nncf.compress_weights(mode=MXFP8_E4M3, group_size=64)` should throw an error something like that: `MXFP8 doesn't support custom group size, it's always 32`. 

2. MXFP4, NVFP4 types are uniquely defined, therefore suggest using a shorter name. 
 
3. MXFP8 and FP8 can represent elements in 2 different type: E5M2 and E4M3. But MXFP8_E5M2 and FP8_E5M2 is not needed in `nncf.compress_weight`, since it's used for gradients. 
We could use E4M3 for shorter names - `MXFP8` or `FP8`, but to be more clear and aligned with existing `QuantizationMode.FP8_E4M3` propose having a bit longer names: `MXFP8_E4M3` and `FP8_E4M3`. 
Potentially, E5M2 can be needed for QAT+LoRA in the future, but still most likely it will be E4M3 for the forward and E5M2 for backward and we can call it `FP8_HYBRID`.

3. FP8_E4M3 or FP4 with E8M0 scale is not likely needed, due to bad accuracy with quantized scale. But for potential extension in future, we can call it `FP8_E4M3_E8M0` or `FP4_E4M1_E8M0`. That's why we shouldn't have a simple `FP8` or `FP4`, otherwise `FP8_E8M0` or `FP4_E8M0` would be confusing. 

@daniil-lyakhov @andreyanufr @nikita-savelyevv  please share your feedback.

MaximProshin: I'm ok with that proposal. @nikita-savelyevv please specify exactly if any changes are required in optimum-intel.
@daniil-lyakhov @ljaljushkin , I would update NNCF documentation, not just docstrings. Is it possible?

ljaljushkin: @nikita-savelyevv 
on optimum-intel side just need to change `e2m1` to `mxfp4`, 
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L868
since it's already has mxfp4 data type
https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/configuration.py#L638

ljaljushkin: > I'm ok with that proposal. @nikita-savelyevv please specify exactly if any changes are required in optimum-intel. @daniil-lyakhov @ljaljushkin , I would update NNCF documentation, not just docstrings. Is it possible?

we have the "supported mode" section and it can be extended with a table above: 
https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/post_training_compression/weights_compression/Usage.md#supported-modes

nikita-savelyevv: @ljaljushkin The table looks good and makes it very structured 👍My comments are below.

**Comment 1**
> on optimum-intel side just need to change e2m1 to mxfp4

For optimum-intel users this can be a bit confusing, because before the default group size for MXFP4 dtype was 128 (as for all other 4-bit dtypes), but now it will be 32. We will need to take this into account, at least by adding a note about this in the corresponding optimum-intel release notes. Ideally, some accuracy measurements will also be needed to understand the effect of this change.

**Comment 2**
> But MXFP8_E5M2 and FP8_E5M2 is not needed in nncf.compress_weight, since it's used for gradients.

Do you mean *E5M2 dtypes are only used during training for backpropagation? We currently have fp8_e5m2 for PTQ. Does it make sense to keep it there?

ljaljushkin: @nikita-savelyevv, mxfp4 can't have group size 128 by specification. we need to fix it anyway
yes e5m2 used for backprop, cause it has a wider range at cost of lower precision of the numbers. we can keep in PTQ for compatibility, but not introduce in PTWC.

daniil-lyakhov:
@ljaljushkin, @nikita-savelyevv, @MaximProshin, I've renamed `E3M1` to `MXFP4`, have introduced `MXFP8_E4M3`, have added a check for MX types to have group_size == 32 (None group_size is translated to 32 for the MX types) and have updated the documentation accordingly

src/nncf/quantization/algorithms/weight_compression/openvino_backend.py
MaximProshin: @daniil-lyakhov , please create a task on required changes in optimum-intel.

docs/usage/post_training_compression/weights_compression/Usage.md
```diff
@@ -22,8 +22,8 @@ The Weights Compression algorithm is aimed at compressing the weights of the mod
...
+All 4-bit modes except MXFP4/MXFP8_E4M3 have a grouped quantization support, when small group of weights (e.g. 128) in the channel dimension share quantization parameters (scale). MXFP4 and MXFP8_E4M3 could be used only with group size == 32.
```
ljaljushkin: please add the table with supported formats:
https://github.com/openvinotoolkit/nncf/pull/3664#discussion_r2373257947

ljaljushkin (changes requested):

docs/usage/post_training_compression/weights_compression/Usage.md
```diff
@@ -191,13 +191,13 @@ from nncf import compress_weights, CompressWeightsMode
...
-  quantized to 8-bit asymmetric integer. Here's the example how to compress weights to e2m1 data type with group size = 32 (recommended).
-  Different `group_size` and `ratio` are also supported.
+- `MXFP4` or `MXFP8_E4M3` modes can be considered for improving accuracy, but currently models quantized to mxfp4/mxfp8_e4m3 should not be faster models
+  quantized to 8-bit asymmetric integer. Here's the example how to compress weights to mxfp4 data type with constant group size == 32.
+  Different `ratio` is also supported.
 
 ```python
 from nncf import compress_weights, CompressWeightsMode
-compressed_model = compress_weights(model, mode=CompressWeightsMode.E2M1, group_size=32, all_layers=True)
+compressed_model = compress_weights(model, mode=CompressWeightsMode.MXFP4, group_size=32, all_layers=True)
```
ljaljushkin:
```suggestion
compressed_model = compress_weights(model, mode=CompressWeightsMode.MXFP4, all_layers=True)
```
why not like that?

daniil-lyakhov: Agree, fixed

docs/usage/post_training_compression/weights_compression/Usage.md
daniil-lyakhov: Done, I've reformatted the document and added 2 tables, please check

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -194,7 +206,7 @@ def float_quantize_dequantize_weight(
 ) -> Union[Tensor, tuple[Tensor, Tensor, Tensor]]:
     """
     First quantizes the given weight tensor to float (nf4) dtype and then dequantizes it back to obtain float32 values.
-    E2M1 mode is currently not supported.
+    MXFP4 and MXFP8_E3M3 mode is currently not supported.
```
ljaljushkin:
```suggestion
    MXFP4 and MXFP8_E4M3 mode is currently not supported.
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -204,7 +216,7 @@ def float_quantize_dequantize_weight(
     :return: Dequantized weight tensor or a tuple containing the decompressed weight, compressed weight and scale.
     """
     assert config.mode in [CompressWeightsMode.NF4, CompressWeightsMode.CODEBOOK, CompressWeightsMode.CB4_F8E4M3]
-    # TODO(nikita-savelyevv): add support for f4e2m1 once ticket 164851 is resolved
+    # TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E3M3, once ticket 164851 is resolved
```
ljaljushkin:
```suggestion
    # TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E4M3, once ticket 164851 is resolved
```

src/nncf/quantization/quantize_model.py
```diff
@@ -458,7 +458,8 @@ def compress_weights(
         INT4_ASYM is the same as INT4_SYM mode, but weights are quantized to a primary precision asymmetrically
             with a typical non-fixed zero point.
         NF4 is the same as INT4_SYM mode, but primary precision is NF4 data type without zero point.
-        E2M1 is the same as INT4_SYM mode, but primary precision is E2M1 data type without zero point.
+        MXFP4 has E2M1 data dtype with E8M0 scale, group size 32 and no zero point.
```
ljaljushkin:
```suggestion
        MXFP4 is MX-compliant FP4 format with E2M1 values sharing group-level E8M0 scale. The size of group is 32.
```

src/nncf/quantization/quantize_model.py
```diff
@@ -458,7 +458,8 @@ def compress_weights(
...
+        MXFP8_E4M3 has E4M3 data dtype with E8M0 scale, group size 32 and no zero point.
```
ljaljushkin:
```suggestion
        MXFP8_E4M3 - is MX-compliant FP8 format with E4M3 values sharing a group-level E8M0 scale. The size of group is 32.
```

docs/usage/post_training_compression/weights_compression/Usage.md
```diff
@@ -21,9 +21,27 @@ The Weights Compression algorithm is aimed at compressing the weights of the mod
...
+| INT4_ASYM  | INT4                 | FP16         | Any          | The same as INT4_SYM mode, but weights are quantized to a primary precision asymmetrically with a typical non-fixed zero point. |
+| NF4        | NF4                  | FP16         | Any          | The the same as INT4_SYM mode, but primary precision is [NF4](https://arxiv.org/pdf/2305.14314v1.pdf) data type without zero point. |
+| MXFP4      | E2M1                 | E8M0         | 32           | FP4 format from [OCP Microscaling Formats (MX) Specification Version 1.0.](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) 2 bits exponent, 1 bit mantissa for the weight and 8 bit exponent, 0 bit mantissa scale. |
+| MXFP8_E4M3 | E4M3                 | E8M0         | 32           | FP8 format from [OCP Microscaling Formats (MX) Specification Version 1.0.](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) 4 bit exponent, 3 bit mantissa for the weight and 8 bit exponent, 0 bit mantissa scale.|
+
```
ljaljushkin:
```suggestion
#### INT8 modes

By default, the algorithm applies asymmetric 8-bit integer quantization (INT8_ASYM mode) to all weights. For symmetric quantization without zero point, the INT8_SYM mode is also available. Both modes typically preserve model accuracy while providing decent performance improvements.

| Compression Mode | Element type | Scale type | Granularity              | Description                |
|------------------|--------------|------------|--------------------------|----------------------------|
| INT8_ASYM        | INT8         | FP16       | Per-channel              | [Asymmetric quantization](/docs/usage/training_time_compression/other_algorithms/LegacyQuantization.md#asymmetric-quantization) |
| INT8_SYM         | INT8         | FP16       | Per-channel              | [Symmetric quantization](/docs/usage/training_time_compression/other_algorithms/LegacyQuantization.md#symmetric-quantization) |

#### Mixed precision modes

Mixed precision modes offer higher compression rates leading to faster inference, though potentially with greater accuracy loss. These modes utilize two precision types: **primary** and **backup**. The primary precision is determined by the compression mode, while backup precision refers to a higher precision format (default is INT8_ASYM, configurable via the `backup_mode` parameter).

By default, NNCF assigns backup precision to **special** quantization-sensitive layers: embeddings, convolutions, and the last linear layer. To compress these special layers using primary precision instead, set `all_layers=True`.

NNCF can automatically distribute precision assignments based on quantization sensitivity using the `ratio` parameter. For example, with `ratio=0.9`, layers (excluding special ones) accounting for 90% of model weights receive primary precision, while the remaining layers use backup precision. This distribution minimizes overall quality deterioration by prioritizing less sensitive layers for lower precision.

| Compression Mode | Element type | Scale type | Granularity              | Description |
|------------------|--------------|------------|--------------------------|-------------|
| INT4_SYM         | INT4         | FP16       | Per-channel / Group-wise | [Symmetric quantization](/docs/usage/training_time_compression/other_algorithms/LegacyQuantization.md#symmetric-quantization) |
| INT4_ASYM        | INT4         | FP16       | Per-channel / Group-wise | [Asymmetric quantization](/docs/usage/training_time_compression/other_algorithms/LegacyQuantization.md#asymmetric-quantization) |
| NF4              | FP32         | FP16       | Per-channel / Group-wise | [NormalFloat-4](https://arxiv.org/pdf/2305.14314v1.pdf) lookup table with 16 FP32 values |
| CODEBOOK         | Any          | FP16       | Per-channel / Group-wise | Arbitrary lookup table (codebook) |
| CB4_F8E4M3       | E4M3         | FP16       | Per-channel / Group-wise | A fixed lookup table with 16 E4M3 values based on NF4 values |
| MXFP4            | E2M1         | E8M0       | Group-wise (32)          | [MX-compliant FP4](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) |
| MXFP8_E4M3       | E4M3         | E8M0       | Group-wise (32)          | [MX-compliant FP8](https://www.opencompute.org/documents/ocp-microscaling-formats-mx-v1-0-spec-final-pdf) |

**Note**: Granularity refers to the scope of elements sharing quantization parameters. "Per-channel" applies different parameters for each output channel, while "Group-wise" divides weights into groups (e.g., group_size=128) that share the same parameters.

**Note**: ExMy is a notation for floating point formats with one sign bit, x exponent bits, and y mantissa bits. For example, E4M3 refers to an FP8 format with one sign bit, four exponent bits, and three mantissa bits. Formats like E8M0 don't include the mantissa bits.

```

docs/usage/post_training_compression/weights_compression/Usage.md
```diff
@@ -191,13 +209,13 @@ from nncf import compress_weights, CompressWeightsMode
 compressed_model = compress_weights(model, mode=CompressWeightsMode.NF4)
 ```
 
-- `E2M1` mode can be considered for improving accuracy, but currently models quantized to e2m1 should not be faster models
-  quantized to 8-bit asymmetric integer. Here's the example how to compress weights to e2m1 data type with group size = 32 (recommended).
-  Different `group_size` and `ratio` are also supported.
+- `MXFP4` or `MXFP8_E4M3` modes can be considered for improving accuracy, but currently models quantized to mxfp4/mxfp8_e4m3 should not be faster models
+  quantized to 8-bit asymmetric integer. Here's the example how to compress weights to mxfp4 data type with constant group size == 32.
+  Different `ratio` are also supported.
```
ljaljushkin:
```suggestion
- Here's the example how to compress weights to MXFP4. Different `ratio` are also supported.
```
not necessarily better accuracy, let's keep it simple for now

docs/Algorithms.md
```diff
@@ -11,7 +11,8 @@
   - Symmetric 8 bit compression mode
   - Symmetric and asymmetric 4 bit compression mode
   - NF4 compression mode
-  - E2M1 weights with E8M0 scales compression mode
+  - MXFP4 compression model with E2M1 data type, E8M0 scales and group size == 32
+  - MXFP8_E4M3 compression model with E4M3 data type, E8M0 scales and group size == 32
```
ljaljushkin:
```suggestion
  - Arbitrary look-up table (CODEBOOK) or predefined lookup table based on NF4 (CB4_F8E4M3) 
  - MX-compliant types - MXFP4 and MXFP8_E4M3
```

src/nncf/openvino/optimized_functions/functions.py
```diff
@@ -109,13 +109,13 @@ def do_float_quantization(
     """
     Computes quantization scale if not provided, and performs corresponding nf4 weight quantization.
     For NF4 quantization quantizes the weights to 16 levels on [-1, 1] interval.
-    TODO(nikita-savelyevv): add support for E2M1 once ticket 164851 is resolved
+    TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E3M3 once ticket 164851 is resolved
```
ljaljushkin:
```suggestion
    TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E4M3 once ticket 164851 is resolved
```

src/nncf/openvino/optimized_functions/functions.py
```diff
@@ -109,13 +109,13 @@ def do_float_quantization(
...
 
     :param weight: Weight array to compress.
     :param config: Weight compression configuration.
     :param reduction_axes: Axes, along which to reduce (collect) different statistics.
     :param precomputed_scale: Optional precomputed scale.
-    :return: Returns quantized (for e2m1 normalized) weight tensor and corresponding scale tensor.
+    :return: Returns quantized (for MXFP4 and MXFP8_E3M3 normalized) weight tensor and corresponding scale tensor.
```
ljaljushkin:
```suggestion
    :return: Returns quantized (for MXFP4 and MXFP8_E4M3 normalized) weight tensor and corresponding scale tensor.
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -244,6 +245,9 @@ def check_user_compression_configuration(
             f"Supported modes are: {[e.value for e in GroupSizeFallbackMode]}."
         )
         raise nncf.ValidationError(msg)
+    if mode in [CompressWeightsMode.MXFP4, CompressWeightsMode.MXFP8_E4M3] and group_size not in [None, 32]:
+        msg = f"MXFP4 and MXFP8_E4M3 types only support group size == 32, group size == {group_size} is given"
```
ljaljushkin:
```suggestion
        msg = f"MXFP4 and MXFP8_E4M3 types only support group size of 32, group size of {group_size} is given"
```

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -285,7 +289,8 @@ def __init__(
             INT4_ASYM is the same as INT4_SYM mode, but weights are quantized to a primary precision asymmetrically
                 with a typical non-fixed zero point.
             NF4 is the same as INT4_SYM mode, but primary precision is NF4 data type without zero point.
-            E2M1 is the same as INT4_SYM mode, but primary precision is E2M1 data type without zero point.
+            MXFP4 has E2M1 weight dtype with E8M0 scale, group size 32 and no zero point.
+            MXFP8_E4M3 has E4M3 weight dtype with E8M0 scale, group size 32 and no zero point.
```
ljaljushkin:
```suggestion
            MXFP4 is MX-compliant FP4 with E2M1 values sharing group-level E8M0 scale. The size of group is 32.
            MXFP8_E4M3 is MX-compliant FP8 with E4M3 values sharing group-level E8M0 scale. The size of group is 32.
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -134,16 +145,17 @@ def do_float_quantization(
...
 ) -> tuple[Tensor, Tensor, Tensor]:
     """
-    Computes quantization scale if not provided, and performs corresponding (nf4, e2m1) weight quantization.
+    Computes quantization scale if not provided,
+    and performs corresponding (nf4, MXFP4 and MXFP8_E3M3) weight quantization.
     For NF4 quantization quantizes the weights to 16 levels on [-1, 1] interval.
-    For E2M1 and CODEBOOK currently returns normalized weight without quantization.
-    TODO(nikita-savelyevv): add support for E2M1 once ticket 164851 is resolved
+    For MXFP4, MXFP8_E3M3 and CODEBOOK currently returns normalized weight without quantization.
+    TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E3M3 once ticket 164851 is resolved
```
ljaljushkin:
```suggestion
    For MXFP4, MXFP8_E4M3 and CODEBOOK currently returns normalized weight without quantization.
    TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E4M3 once ticket 164851 is resolved
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -134,16 +145,17 @@ def do_float_quantization(
...
 
     :param weight: Weight array to compress.
     :param config: Weight compression configuration.
     :param reduction_axes: Axes, along which to reduce (collect) different statistics.
     :param precomputed_scale: Optional precomputed scale.
-    :return: Returns quantized (for e2m1 normalized) weight tensor and corresponding scale tensor and
+    :return: Returns quantized (for MXFP4 and MXFP8_E3M3 normalized) weight tensor and corresponding scale tensor and
```
ljaljushkin:
```suggestion
    :return: Returns quantized (for MXFP4 and MXFP8_E4M3 normalized) weight tensor and corresponding scale tensor and
```

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -180,7 +192,7 @@ def do_float_quantization(
         )
         return compressed_weight, scale, indexes
     else:
-        # TODO(nikita-savelyevv): add support for E2M1 once ticket 164851 is resolved
+        # TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E3M3 once ticket 164851 is resolved
```
ljaljushkin:
```suggestion
        # TODO(nikita-savelyevv): add support for MXFP4 and MXFP8_E4M3 once ticket 164851 is resolved
```

src/nncf/quantization/quantize_model.py
```diff
@@ -518,11 +519,12 @@ def compress_weights(
...
         if mode in [
             CompressWeightsMode.NF4,
-            CompressWeightsMode.E2M1,
+            CompressWeightsMode.MXFP4,
+            CompressWeightsMode.MXFP8_E4M3,
             CompressWeightsMode.CODEBOOK,
             CompressWeightsMode.CB4_F8E4M3,
         ]:
-            msg = "Torch backend does not support NF4, E2M1 and CODEBOOK modes for weight compression."
+            msg = "Torch backend does not support NF4, MXFP4, MXFP8_E3M3 and CODEBOOK modes for weight compression."
```
ljaljushkin:
```suggestion
            msg = "Torch backend does not support NF4, MXFP4, MXFP8_E4M3 and CODEBOOK modes for weight compression."
```

src/nncf/quantization/quantize_model.py
```diff
@@ -567,11 +569,12 @@ def compress_weights(
...
```
ljaljushkin:
```suggestion
            msg = "Torch backend does not support NF4, MXFP4, MXFP8_E4M3 and CODEBOOK modes for weight compression."
```

src/nncf/quantization/quantize_model.py
```diff
@@ -607,8 +610,13 @@ def compress_weights(
...
             raise nncf.ParameterNotSupportedError(msg)
 
-        if any((awq, scale_estimation, gptq, lora_correction)) and mode == CompressWeightsMode.E2M1:
-            msg = "AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode is E2M1."
+        if any((awq, scale_estimation, gptq, lora_correction)) and mode in [
+            CompressWeightsMode.MXFP4,
+            CompressWeightsMode.MXFP8_E4M3,
+        ]:
+            msg = (
+                "AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode in [MXFP4, MXFP8_E3M3]."
```
ljaljushkin:
```suggestion
                "AWQ, Scale estimation, GPTQ or Lora Correction algorithm is defined, but mode in [MXFP4, MXFP8_E4M3]."
```

src/nncf/quantization/quantize_model.py
```diff
@@ -626,11 +634,12 @@ def compress_weights(
...
         if mode in [
             CompressWeightsMode.NF4,
-            CompressWeightsMode.E2M1,
+            CompressWeightsMode.MXFP4,
+            CompressWeightsMode.MXFP8_E4M3,
             CompressWeightsMode.CODEBOOK,
             CompressWeightsMode.CB4_F8E4M3,
         ]:
-            msg = "ONNX backend does not support NF4, E2M1 and CODEBOOK modes for weight compression."
+            msg = "ONNX backend does not support NF4, MXFP4, MXFP8_E3M3 and CODEBOOK modes for weight compression."
```
ljaljushkin:
```suggestion
            msg = "ONNX backend does not support NF4, MXFP4, MXFP8_E4M3 and CODEBOOK modes for weight compression."
```

src/nncf/parameters.py
```diff
@@ -90,7 +89,12 @@ class CompressWeightsMode(StrEnum):
         https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/training_time_compression/other_algorithms/LegacyQuantization.md#asymmetric-quantization
     :param NF4: The the same as INT4_SYM mode, but primary precision is NF4 data type without zero point.
     :param INT8: Mode is deprecated and will be removed in future releases. Please use `INT8_ASYM` instead.
-    :param E2M1: FP4 format from "OCP Microscaling Formats (MX) Specification" Version 1.0.
+    :param MXFP4: FP4 format from "OCP Microscaling Formats (MX) Specification" Version 1.0.
+        2 bits exponent, 1 bit mantissa for the weight and 8 bit exponent, 0 bit mantissa scale.
+        Group size is constant == 32.
+    :param MXFP8_E4M3: FP8 format from "OCP Microscaling Formats (MX) Specification" Version 1.0.
+        4 bits exponent, 3 bit mantissa for the weight and 8 bit exponent, 0 bit mantissa scale.
+        Group size is constant == 32.
```
ljaljushkin:
```suggestion
    :param MXFP4: MX-compliant FP4 format with E2M1 values sharing group-level E8M0 scale. The size of group is 32.
    :param MXFP8_E4M3: MX-compliant FP8 format with E4M3 values sharing group-level E8M0 scale. The size of group is 32.
```

ljaljushkin (changes requested):

daniil-lyakhov:
@ljaljushkin, thank you very much for the valuable comments! I've applied them and fixed other similar typos

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -94,15 +94,26 @@ def calculate_float_quantization_params(
...
-        max_val = 6.0 if config.mode == CompressWeightsMode.E2M1 else fns.max(fns.abs(config.get_numpy_codebook()))
+    FP_MAX_VALS = {
+        CompressWeightsMode.MXFP4: 6.0,
+        CompressWeightsMode.MXFP8_E4M3: 448.0,
+    }
+    if config.mode in [CompressWeightsMode.CODEBOOK, CompressWeightsMode.CB4_F8E4M3] + list(FP_MAX_VALS.keys()):
+        if config.mode in FP_MAX_VALS:
+            max_val = FP_MAX_VALS[config.mode]
+        else:
+            max_val = fns.max(fns.abs(config.get_numpy_codebook()))
```
nikita-savelyevv, line 105:
```suggestion
        max_val = FP_MAX_VALS.get(config.mode, fns.max(fns.abs(config.get_numpy_codebook())))
```

nikita-savelyevv (commented):
Overall looks good! 

One other thing we need to align is adjustable group size logic. 

[Here](https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L622) we compute a new group size if the given one does not fit the channel size. Currently if mxfp4 / mxfp8 data type is selected, group size fallback mode is set to ADJUST and there are some layers that have channel size not divisible by 32, but divisible by `min_adjusted_group_size` (equal to16 by default), then group size 16 will be selected. This contradicts mxfp4 / mxfp8 definition and should not be allowed. The desired outcome in such case is that such layers should be ignored.

I believe updating this condition to take into account compression mode should be enough https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L623 .

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
daniil-lyakhov: Unfortunately this statement is not a lazy one, it will try to calculate max abs of codebook before checking the map. I suggest to keep the code as it is

nikita-savelyevv:
> I believe updating this condition to take into account compression mode should be enough https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L623 

My bad, this suggestion actually does not work 🙂 At that point `weight_params` contains backup mode configs and primary config with MX dtype won't be set until https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L892 

Let's just not allow to set GroupSizeFallbackMode.ADJUST when MX data type is selected, I think it makes sense.

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
nikita-savelyevv: Good point 👍

daniil-lyakhov:
> > I believe updating this condition to take into account compression mode should be enough https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L623
> 
> My bad, this suggestion actually does not work 🙂 At that point `weight_params` contains backup mode configs and primary config with MX dtype won't be set until https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/algorithm.py#L892
> 
> Let's just not allow to set GroupSizeFallbackMode.ADJUST when MX data type is selected, I think it makes sense.

Agree, I've added a ValidationError for a WC with MX types and FallbackMode == ADJUST, please check

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -94,15 +94,26 @@ def calculate_float_quantization_params(
...
         scale = scale / max_val
 
     # NOTE: adding machine epsilon to avoid division by zero
     eps = fns.finfo(weight).eps
     scale = fns.where(fns.abs(scale) < eps, eps, scale)
 
-    if config.mode == CompressWeightsMode.E2M1:
+    if config.mode in [CompressWeightsMode.MXFP4, CompressWeightsMode.MXFP8_E4M3]:
+        # FP types are using E8M0 type scale.
```
andreyanufr:
```suggestion
        # MXFP types are using E8M0 type scale.
```

andreyanufr (approved):

PR #3674, opened 2025-10-01
Disable support of python 3.9

AlexanderDokuchaev:
### Changes

Set `requires-python = ">=3.10"` 

### Reason for changes

Python 3.9 reached end of life


MaximProshin:
@AlexanderDokuchaev , don't forget to update the docs.

PR #3678, opened 2025-10-06
Fix tests

andrey-churkin:

### DIFF
--- a/tests/onnx/models.py
+++ b/tests/onnx/models.py
@@ -1709,8 +1709,8 @@ class UnifiedEmbeddingModel(ONNXReferenceModel):
         reshape_tensor_name = "R"
         reshape_tensor = create_initializer_tensor(
             name=reshape_tensor_name,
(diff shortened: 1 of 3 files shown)


tests/onnx/models.py
```diff
@@ -1709,8 +1709,8 @@ def __init__(self):
         reshape_tensor_name = "R"
         reshape_tensor = create_initializer_tensor(
             name=reshape_tensor_name,
-            tensor_array=np.array([1, 3, 5]).astype(np.float32),
-            data_type=onnx.TensorProto.FLOAT,
+            tensor_array=np.array([1, 3, 5], dtype=np.int64),
+            data_type=onnx.TensorProto.INT64,
```
andrey-churkin, line 1713: Incorrect data type for the `shape` input of the `Reshape` operation. It should be `tensor(int64)`, according to the [documentation](https://onnx.ai/onnx/operators/onnx__Reshape.html).

PR #3687, opened 2025-10-13
[ONNX][SmoothQuant]  Introduce new axes and axes_mode parameters 

andrey-churkin:
### Changes

- This PR introduces a new `axes` and `axes_mode` parameters for `TensorReducerBase`. These parameters have the following meaning:
    - `axes`: The axes along which the reduction operation should be applied. If `None`, the operation will be applied to all axes (i.e., `tuple(range(tensor.ndim))`).
    - `axes_mode`: Determines how the specified `axes` are treated during the operation. Use `AxesMode.REDUCTION` to reduce over the given axes, or `AxesMode.KEEP` to preserve them.

    These parameters are used to calculate the reduction axes (`determine_reduction_axes()` method) during statistic collection, allowing us to avoid requiring the actual tensor shape (actually only number of dimensions `ndim` is required) before inference.

- Modifies the `SmoothQuant` algorithm to use the `axes` and `axes_mode` parameters for the ONNX backend instead of relying on the tensor shape from the NNCF graph, as this shape isn't always available.

### Related tickets

Ref: 173880, Ref: 174334

### Tests
- Build post_training_quantization # 735 (# 739)
- tests/onnx/test_nncf_graph_builder.py::test_unknown_shape

### DIFF
--- a/src/nncf/quantization/algorithms/smooth_quant/algorithm.py
+++ b/src/nncf/quantization/algorithms/smooth_quant/algorithm.py
@@ -256,7 +256,7 @@ class SmoothQuant(Algorithm):
             ):
                 stats = tensor_collector.get_statistics()
                 shape = stats[SHAPE_BRANCH_KEY]
-                shape = tuple() if shape is None else tuple(shape.tolist())
+                shape = tuple(shape.tolist())
 
             items.append((node, input_port, shape))
 
--- a/tests/cross_fw/test_templates/test_smooth_quant.py
+++ b/tests/cross_fw/test_templates/test_smooth_quant.py
@@ -237,7 +237,13 @@ class TemplateTestSQAlgorithm:
         algo._set_backend_entity = lambda model: backend_entity
 
         mocked_transformer = mocker.MagicMock()
+        empty_shapes = [
+            (node, port, ()) for node, port in algo._get_nodes_to_smooth_data(graph, algo._get_alpha_map().keys())
+        ]
         mocker.patch("nncf.common.factory.ModelTransformerFactory.create", return_value=mocked_transformer)
+        mocker.patch(
+            "nncf.quantization.algorithms.smooth_quant.algorithm.SmoothQuant._retrieve_shape", return_value=empty_shapes
+        )
         algo.apply(model, graph, algo_statistic_points)
 
         mocked_transformer.transform.assert_called_once()


src/nncf/quantization/algorithms/smooth_quant/openvino_backend.py
```diff
@@ -165,3 +154,11 @@ def filter_func(point: StatisticPoint) -> bool:
...
 
         return filter_func
+
+    @staticmethod
+    def get_abs_max_reducer_cls() -> type[OVAbsMaxReducer]:
+        return OVAbsMaxReducer
+
+    @staticmethod
+    def get_shape_reducer_cls() -> type[OVShapeReducer]:
+        return OVShapeReducer
```
andrey-churkin, line 164: We add the `get_abs_max_reducer_cls()` and `get_shape_reducer_cls(`) methods here because the OpenVINO backend uses the `OVAbsMaxReducer` and `OVShapeReducer` classes instead of `AbsMaxReducer` and `ShapeReducer` to enable in-place statistic collection.

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -98,6 +102,10 @@ def _set_backend_entity(self, model: TModel) -> None:
             msg = f"Cannot return backend-specific entity because {model_backend.value} is not supported!"
             raise nncf.UnsupportedBackendError(msg)
 
+        # Only the OpenVINO backend supports in-place statistics, so we should set this variable here.
+        if model_backend != BackendType.OPENVINO:
+            self._inplace_statistics = False
```
nikita-savelyevv:
```suggestion
        # Only the OpenVINO backend supports in-place statistics, so we should set this variable here.
        self._inplace_statistics = model_backend == BackendType.OPENVINO
```
Minor

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -204,27 +212,59 @@ def _calculate_scale_and_ratio(
...
+                node.node_name,
+                self._backend_entity.get_filter_fn_for_statistics(input_port, self._algorithm_key),
+                self._algorithm_key,
+            ):
+                stats = tensor_collector.get_statistics()
+                shape = stats[SHAPE_BRANCH_KEY]
+                if shape is not None:
+                    shape = tuple(shape.tolist())
+                else:
+                    shape = tuple()
```
nikita-savelyevv:
```suggestion
                shape = tuple() if shape is None else tuple(shape.tolist())
```
Minor

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -247,42 +287,90 @@ def _get_statistics_for_node(
...
+                    self._inplace_statistics,
+                    keep_axes=keep_axes,
+                )
+            else:
+                reduction_axes = self._calculate_input_reduction_axes(graph, node_to_smooth, input_act_port)
+                collector = self._create_tensor_collector(
+                    self._subset_size,
+                    self._inplace_statistics,
+                    reduction_axes=reduction_axes,
                 )
```
nikita-savelyevv: What do you think about moving this logic to `self._backend_entity.get_statistic_points()` or something like that?

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
nikita-savelyevv: `self._inplace_statistics` is originally set in the constructor and overwriting this variable like that may results in unexpected behavior. Perhaps, we should raise an exception if given value is not supported by the backend?

nikita-savelyevv: Also, I thought only ONNX backend is affected, but with such condition it seems that all backends other than OpenVINO do not support in-place statistics. Is that right?

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -247,42 +287,90 @@ def _get_statistics_for_node(
...
 
-    def _get_nodes_to_smooth_data(self, nncf_graph: NNCFGraph, node_metatypes: list[OperatorMetatype]) -> list[dict]:
+            container.add_statistic_point(StatisticPoint(target_point, collector, self._algorithm_key))
+
+        return container
+
+    def _create_tensor_collector(
+        self,
+        num_samples: int,
+        inplace: bool,
```
nikita-savelyevv: Can't we use `self._inplace_statistics` instead?

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
nikita-savelyevv:
```suggestion
            if model_backend == BackendType.ONNX:
                keep_axes = (self._backend_entity.get_activation_channel_axis(node_to_smooth, input_act_port),)
                reduction_axes = None
            else:
                keep_axes = None
                reduction_axes = self._calculate_input_reduction_axes(graph, node_to_smooth, input_act_port)
            collector = self._create_tensor_collector(
                self._subset_size,
                self._inplace_statistics,
                keep_axes=keep_axes,
                reduction_axes=reduction_axes,
            )
```

What do you think?

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
nikita-savelyevv, line 107: Shouldn't we raise an exception if `inplace_statistics=True` was provided in the constructor for any backend other that OpenVINO?

nikita-savelyevv (commented):
Should we perhaps add a test with an ONNX model for which ndim is not known beforehand to have an example of why keep_dims approach is introduced?

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: Good suggestion! Thanks.

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: We can. Fixed.

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: Looks good. I've fixed.

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: We should - but here's the thing: the `inplace_statistics` parameter has `True` as its default value. We probably can't change it, as that would break backward compatibility and force us to explicitly specify it for the `OpenVINO` backend.

https://github.com/openvinotoolkit/nncf/blob/8ef5ecd07361dbbdaa1aa7bcb958e756f8dd63c0/src/nncf/quantization/advanced_parameters.py#L264

andrey-churkin:
> Should we perhaps add a test with an ONNX model for which ndim is not known beforehand to have an example of why keep_dims approach is introduced?

Thank you for the suggestion. I’ll consider how to implement it.

UPD: This problem is reproduced on timm/visformer_small model from the ptq scope.

UPD: `tests/onnx/test_nncf_graph_builder.py::test_unknown_shape` test was added

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
nikita-savelyevv: I see, thanks for explanation. Seems like the default value should be backend-specific. But that's out of scope of this PR..

src/nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -43,13 +43,25 @@ class TensorReducerBase(ABC):
     the specified rule. Could handle tensors inplace or out of place.
     """
 
-    def __init__(self, reduction_axes: Optional[ReductionAxes] = None, inplace: bool = False):
+    def __init__(
+        self,
+        reduction_axes: Optional[ReductionAxes] = None,
```
daniil-lyakhov: Should we forward this parameter in the children of  the `TensorReducerBase`?

src/nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -43,13 +43,25 @@ class TensorReducerBase(ABC):
...
+        keep_axes: Optional[tuple[int, ...]] = None,
```
daniil-lyakhov:
```suggestion
        keep_axes: Optional[Axes] = None,
```
Perhaps we could rename `ReductionAxes` and reuse them there?

src/nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -99,14 +111,24 @@ def __eq__(self, __o: object) -> bool:
             isinstance(__o, self.__class__)
             and self._reduction_axes == __o._reduction_axes
             and self._inplace == __o.inplace
+            and self._keep_axes == __o._keep_axes
         )
 
     def __hash__(self) -> int:
-        return hash((self.__class__.__name__, self.inplace, self._reduction_axes))
+        return hash((self.__class__.__name__, self.inplace, self._reduction_axes, self._keep_axes))
```
daniil-lyakhov: Perhaps we should update  `__hash__` methods for some of the `TensorReducerBase` as well

tests/cross_fw/test_templates/test_smooth_quant.py
```diff
@@ -170,28 +168,6 @@ def test_smooth_quant_algo(self, model_cls, reference_values, tmpdir):
 
         self.check_scales(quantized_model, reference_values, model_cls)
 
-    def test_get_abs_max_channel_collector(self, inplace_statistics: bool):
-        backend = self.get_backend()
-        reduction_axes = (3, 2, 1)
```
daniil-lyakhov, line 175: Please test `self._backend_entity.get_abs_max_reducer_cls()` and `_backend_entity.get_shape_reducer_cls`

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -247,42 +284,84 @@ def _get_statistics_for_node(
...
+            # However, when using `keep_axes`, the reduction axes are determined during statistics collection.
+            # Therefore, `keep_axes` and `inplace` cannot be used together with the OpenVINO backend.
+            # For the ONNX backend, we can't calculate reduction axes before inference because the tensor shape
+            # (actually, only the number of dimensions (ndim) is required) is unknown for some operations.
+            if model_backend == BackendType.ONNX:
+                keep_axes = (self._backend_entity.get_activation_channel_axis(node_to_smooth, input_act_port),)
+                reduction_axes = None
+            else:
+                keep_axes = None
+                reduction_axes = self._calculate_input_reduction_axes(graph, node_to_smooth, input_act_port)
```
daniil-lyakhov: Usually we create a method in the backend to resolve such situation, why don't you introduce a method in the backend? The comment could be placed as a docstring for the method

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -204,27 +212,56 @@ def _calculate_scale_and_ratio(
...
+        items = []
+        for node, input_port in nodes:
+            for tensor_collector in statistic_points.get_algo_statistics_for_node(
+                node.node_name,
+                self._backend_entity.get_filter_fn_for_statistics(input_port, self._algorithm_key),
+                self._algorithm_key,
+            ):
+                stats = tensor_collector.get_statistics()
+                shape = stats[SHAPE_BRANCH_KEY]
+                shape = tuple() if shape is None else tuple(shape.tolist())
```
daniil-lyakhov: When shape could be None?

src/nncf/experimental/common/tensor_statistics/collectors.py
```diff
@@ -43,13 +43,25 @@ class TensorReducerBase(ABC):
...
+        :param keep_axes: Axes to preserve during the reduction operation. These will be used in
+            `_reduce_out_of_place()` to calculate the reduction axes once the tensor shape is known.
         :param inplace: Whether should be calculated inplace or out of place.
         """
+        if reduction_axes is not None and keep_axes is not None:
+            msg = "Only one of `reduction_axes` or `keep_axes` should be specified, not both."
+            raise nncf.ValidationError(msg)
+
         self._reduction_axes = reduction_axes
+        self._keep_axes = keep_axes
```
ljaljushkin: Do these 2 variables mutually exclusive? 
If yes, I'd rename `_reduction_axes` to `_axes` and added `_axes_mode` - REDUCTION or KEEP

andrey-churkin: Done

src/nncf/experimental/common/tensor_statistics/collectors.py
andrey-churkin: Done

src/nncf/experimental/common/tensor_statistics/collectors.py
andrey-churkin: Done

src/nncf/experimental/common/tensor_statistics/collectors.py
andrey-churkin: Done

tests/cross_fw/test_templates/test_smooth_quant.py
andrey-churkin: Done

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: https://github.com/openvinotoolkit/nncf/blob/7eed8fc9f84dbd23389f86fde52a3aa46c038aad/tests/cross_fw/test_templates/test_ptq_params.py#L453

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: It helps simplify the code and avoid duplication.

andrey-churkin:
@ljaljushkin @nikita-savelyevv @daniil-lyakhov Please review

ljaljushkin (approved):

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
nikita-savelyevv: If `shape` can be None only during testing and not in any real life scenario then I would suggest to properly mock the returned shape in tests, rather that adopting algorithm logic to support None shapes.

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
```diff
@@ -247,42 +285,82 @@ def _get_statistics_for_node(
...
+            # However, when using `keep_axes`, the reduction axes are determined during statistics collection.
+            # Therefore, `keep_axes` and `inplace` cannot be used together with the OpenVINO backend.
+            # For the ONNX backend, we can't calculate reduction axes before inference because the tensor shape
+            # (actually, only the number of dimensions (ndim) is required) is unknown for some operations.
+            if model_backend == BackendType.ONNX:
+                axes_mode = AxesMode.KEEP
+                axes = (self._backend_entity.get_activation_channel_axis(node_to_smooth, input_act_port),)
+            else:
+                axes_mode = AxesMode.REDUCTION
+                axes = self._calculate_input_reduction_axes(graph, node_to_smooth, input_act_port)
```
nikita-savelyevv:
```suggestion
            axes, axes_mode = self._backend_entity._get_tensor_collector_axes(graph, node_to_smooth, input_act_port)
```

Following up on the [suggestion ](https://github.com/openvinotoolkit/nncf/pull/3687#discussion_r2436390769) by @daniil-lyakhov, I believe this will not result in any duplication. In this case `_calculate_input_reduction_axes` just needs to be moved to parent backend class instead of the algorithm. What do you think?

andrey-churkin: done

src/nncf/quantization/algorithms/smooth_quant/algorithm.py
andrey-churkin: done

PR #3694, opened 2025-10-21
Fix/smooth quant param grid

### LINKED ISSUE #3613: `_get_smooth_quant_param_grid` is broken
### 🐛 Describe the bug

https://github.com/openvinotoolkit/nncf/blob/b34d24e22a9520f5205c4b153a11c1b3d3a58a65/nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py#L90

I think that this line is currently subtly broken.
Right now, `_get_smooth_quant_param_grid()` returns
```python
{
    'advanced_parameters:smooth_quant_alphas': [
        AdvancedSmoothQuantParameters(convolution=-1, matmul=(0.15,)),
        AdvancedSmoothQuantParameters(convolution=-1, matmul=(0.25,)),
        AdvancedSmoothQuantParameters(convolution=-1, matmul=(0.5,)),
        AdvancedSmoothQuantParameters(convolution=-1, matmul=(0.75,)),
        AdvancedSmoothQuantParameters(convolution=-1, matmul=(0.95,))
    ]
}
```
Notice that the `matmul` values are wrapped in an extra `tuple` instead of just being plain `floats` like they are supposed to. This is happening because `itertools.product(*iterables)` will always return tuples of length `len(iterables)` even when there is only 1 argument.

These `tuple` values will then cause a crash in `create_ptq_pipeline`:
```python
File "/.../site-packages/nncf/quantization/algorithms/post_training/pipeline.py", line 97, in create_ptq_pipeline
    if model_type == ModelType.TRANSFORMER and (sq_params.convolution >= 0 or sq_params.matmul >= 0):
TypeError: '>=' not supported between instances of 'tuple' and 'int'
```

### Environment

NNCF 2.17.0
OpenVINO 2025.2.0
Torch 2.7.1
ONNX 1.18.0

OS Ubuntu 22.04.3 LTS
Python 3.10.12
Install PyPI

### Minimal Reproducible Example

I don't have a proper MRE right now, but I think that any quantization pipeline with enabled hyperparameter tuning should demonstrate the crash. So, something along the lines of: 
```python
nncf.quantize_with_accuracy_control(
    model=openvino_model,
    ...,
    advanced_accuracy_restorer_parameters=nncf.AdvancedAccuracyRestorerParameters(
        tune_hyperparams=True,
        ...,
    ),
)
```

Also, I think that the issue if fairly obvious from the `_get_smooth_quant_param_grid` code mentioned above. I was able to fix the issue by adding the missing tuple unpacking
```diff
-            AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v in itertools.product(alpha_values)
+            AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v, in itertools.product(alpha_values)
```

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Aryanpagaria:
### Changes

Fixed a subtle bug in _get_smooth_quant_param_grid where matmul values were returned as single-element tuples instead of plain floats.

Replaced:
AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v in itertools.product(alpha_values)
With:
AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v, in itertools.product(alpha_values)
This ensures matmul values are floats, preventing TypeError in the PTQ pipeline.

### Reason for changes
itertools.product always returns tuples, even for a single iterable.
Previously, this caused a crash when the code tried to compare sq_params.matmul >= 0 because tuples cannot be compared to integers.
The change fixes this by unpacking the single-element tuples to get plain float values.

### Related tickets
Closes #3613

### Tests

Verified manually that _get_smooth_quant_param_grid() now returns AdvancedSmoothQuantParameters with float matmul values, e.g., 0.15, 0.25, instead of (0.15,).
Results of test :
0.15 <class 'float'>
0.25 <class 'float'>
0.5  <class 'float'>
...

Ran existing pytest suite — all tests passed successfully.

No new tests were required as the logic is unchanged, only the tuple unpacking was fixed.

### DIFF
--- a/src/nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
+++ b/src/nncf/quantization/algorithms/hyperparameter_tuner/param_grid.py
@@ -87,7 +87,8 @@ def _get_smooth_quant_param_grid() -> ParamGrid:
     alpha_values = [0.15, 0.25, 0.5, 0.75, 0.95]
     return {
         "advanced_parameters:smooth_quant_alphas": [
-            AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v in itertools.product(alpha_values)
+            AdvancedSmoothQuantParameters(matmul=alpha_v) for alpha_v, in itertools.product(alpha_values)
+
         ]
     }
 


ljaljushkin:
@andrey-churkin, could you please take a look?

src/nncf/openvino/quantization/backend_parameters.py
```diff
@@ -1,17 +1,7 @@
-# Copyright (c) 2025 Intel Corporation
```
andrey-churkin, line 1: Could you briefly explain why it was removed?

src/nncf/experimental/torch/fx/quantization/backend_parameters.py
```diff
@@ -1,31 +0,0 @@
-# Copyright (c) 2025 Intel Corporation
```
andrey-churkin, line 1: Could you briefly explain why it was removed?

andrey-churkin:
@Aryanpagaria Please rebase your branch to resolve the conflicts

andrey-churkin:
Fixed https://github.com/openvinotoolkit/nncf/pull/3754

PR #3706, opened 2025-10-29
Support 3D Weights in SE Algorithm 

anzr299:
### Related tickets

175212

### Tests

PR Performance Job: post_training_weight_compression_performance - 57
Develop Branch Performance Job: post_training_weight_compression_performance - 58
WC Conformance Test: https://github.com/openvinotoolkit/nncf/actions/runs/19161281090: Pass

Model: Qwen/Qwen3-30B-A3B
NNCF Backend: OpenVINO
Higher is better.
Task: gsm8k
Limit: 100
Max New Tokens: 10000
OpenVINO version: 2026.0.0.dev20251111 (with WA for 176465)
n-shots: 5(default)

Precision Type | Filter | Value | Stderr
-- | -- | -- | --
INT4 SYM Per-Channel (with Scale estimation) Calibrated on wikitext with 128 samples | flexible-extract | 0.66 | 0.0476
  | strict-match | 0.38 | 0.0488
INT4 SYM Per-Channel | flexible-extract | 0.77 | 0.0423
  | strict-match | 0.28 | 0.0451
FP16 | flexible-extract | 0.91 | 0.0288
  | strict-match | 0.86 | 0.0349
  
WWB Results with Reasoning Disabled:
 INT4 Sym Per-Channel: 0.826173(vs FP16)
 INT4 Sym Per-Channel with SE: 0.938537(vs FF16)

Model: openai/gpt-oss-20b
NNCF Backend: Torch
Higher is better.

```
time \
accelerate launch -m lm_eval --model hf \
--model_args "{\"pretrained\":\"${MODEL_DIR}\",\"enable_thinking\":false}" \
--tasks gsm8k_cot_llama \
--fewshot_as_multiturn \
--apply_chat_template=True \
--device cuda \
--limit 100 \
--output_path $EXP_DIR \
--gen_kwargs max_new_tokens=1024,temperature=0.6,top_p=0.95,top_k=20 \
```

gpt-oss-20b | strict-match | flexible-extract | 
-- | -- | -- |
bf16 | 0.96 | 0.96 |  
int4_sym_gs32_experts_int8_the_rest | 0.96 | 0.94 |  
int4_sym_gs32_int8_the_rest | 0.64 | 0.96 |  
int4_sym_gs32_int8_the_rest_SE_32tokens_128samples_64se_samples | 0.79 | 0.96 |  
int4_sym_gs32_int8_the_rest_SE_128tokens_128samples_64se_samples | 0.81 | 0.96 |  
int4_sym_gs32_int8_the_rest_SE_half_128tokens_128samples_64se_samples | 0.83 | 0.97 |  
int4_sym_gs32_int8_the_rest_SE_half_256tokens_256samples   _64se_samples | 0.79 | 0.95 | 
int4_sym_gs32_int8_the_rest_SE_256tokens_256samples_64se_samples | 0.90 | 0.96 |  
int4_sym_gs32_int8_the_rest_SE_256tokens_256samples_256se_samples | 0.95 | 0.94 |


src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -340,9 +360,10 @@ def calculate_quantization_params(
                 )
             q_weights_ = fns.zeros_like(original_weight) + out
 
-            q_outs = fns.matmul(fns.transpose(q_weights_, (1, 0, 2)), X)
+            q_outs = fns.matmul(fns.transpose(q_weights_, weight_transpose_for_matmul_with_act), X)
             ideal_scale_diffs = fns.mean((fp_outs - q_outs) ** 2, axis=-1)
-            ideal_scale_diffs = fns.transpose(ideal_scale_diffs, (1, 0))
+            # ideal_scale_diffs = fns.transpose(ideal_scale_diffs, transpose_min_max_scale_diffs_dims)
+            ideal_scale_diffs = maybe_reshape_scale_diffs(ideal_scale_diffs, is_moe)
```
ljaljushkin: Suggesting extending test for Scale Estimation with a single layer model that has 3D weights and takes 3D activations with a few experts in batch dimension:

https://github.com/openvinotoolkit/nncf/blob/f6530762b6a5964346c39367e4b08b80a6d5cc35/tests/cross_fw/test_templates/template_test_weights_compression.py#L241-L262
       reference = self.get_scale_estimation_ref()

anzr299: Done

src/nncf/quantization/algorithms/weight_compression/activation_stats.py
```diff
@@ -27,17 +27,32 @@ def process_stats(stats: WCTensorStatistic, subset_size: int) -> tuple[Tensor, T
         s - maximum channel magnitude across samples [HiddenDim]
         X - average channel magnitude across tokens in the sequence [HiddenDim, min(SampleSize, ~subset_size)]
     """
-    X = fns.stack(stats.mean_values)  # [SampleSize, HiddenDim]
-    X_full = fns.transpose(X)  # [HiddenDim, SampleSize]
+    X = fns.stack(
+        stats.mean_values
+    )  # [SampleSize, HiddenDim] for 2-D or [SampleSize, No. of Experts, HiddenDim] for 3-D
+    X_dim = len(X.shape)
```
andreyanufr: Maybe the name n_dims would be more descriptive.

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -196,11 +196,14 @@ def calculate_quantization_params(
         X = X.astype(TensorDataType.float32)
         weight = weight.astype(TensorDataType.float32)
         eps = fns.finfo(weight).eps
+        is_moe = len(weight.shape) == 3
```
ljaljushkin: I wouldn't assume that 3D matrix multiplication is used for MoE only. Although we haven't had any other examples yet, I would keep it a little more general, for example, using the `is_batched_matmul` or `is_3d_weight` terms.

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -196,11 +196,14 @@ def calculate_quantization_params(
...
 
         was_transposed = False
-        if reduction_axis == 0:
-            weight = fns.transpose(weight)
-            reduction_axis = 1
+        if reduction_axis == 0 or (reduction_axes == 1 and is_moe):
+            # MoE: [num_experts, out_features, hidden_dimension] → [num_experts, hidden_dimension, out_features]
+            # Standard: [out_features, in_features] → [in_features, out_features]
+            weight = fns.transpose(weight, axes=(0, 2, 1)) if is_moe else fns.transpose(weight)
```
ljaljushkin: I believe it's possible to reduce number of `if` by some generalization or extracting common formula. 
e.g. here you can transpose with axes=(-1, -2)

ljaljushkin:
I've tried this PR for gpt-oss-20b compression and it significantly improved accuracy on gsm8k by using 128 samples from wikitext (seqlen=32). 

image

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Alright, I will use a more general name

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: hm, okay I will try to come up with some generalization for n-dimension otherwise just 3d & 2d
Although in this case, it expects the axes tuple to also have the same number of dimensions as tensor

ljaljushkin: > t expects the axes tuple to also have the sa

yeah, unlikely numpy can't do the same via transpose:
image

it can only swap:
np.empty((2,4,3)).swapaxes(-1,-2).shape

anzr299: Atleast in the case of pytorch it uses torch.permute for when axes is defined https://github.com/openvinotoolkit/nncf/blob/f6530762b6a5964346c39367e4b08b80a6d5cc35/src/nncf/tensor/functions/torch_numeric.py#L383

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Done. Changed is_moe to is_3d_weight

src/nncf/quantization/algorithms/weight_compression/activation_stats.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -1107,11 +1132,17 @@ def get_statistic_points(
                 statistic_point = self._backend_entity.target_point(
                     TargetType.POST_LAYER_OPERATION, node.node_name, port_id=output_port_id
                 )
-                # Reduce activations across all but the last dimension. The last dimension is assumed to be the hidden
-                # size dimension.
+                weight_dims = self._get_weight_dim_from_act_node(graph, node, output_port_id)
+                # by Default, Reduce activations across all but the last dimension. The last dimension is
```
ljaljushkin:
```suggestion
                # By default, reduce activations across all but the last dimension. The last dimension is
```

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -196,10 +196,22 @@ def calculate_quantization_params(
...
         weight = weight.astype(TensorDataType.float32)
         eps = fns.finfo(weight).eps
+        is_3d_weight = len(weight.shape) == 3
+
+        axes = {
+            'weight_transpose': (0, 2, 1) if is_3d_weight else None,
+            'matmul_transpose': (0, 1, 3, 2) if is_3d_weight else (1, 0, 2),
+            'squeeze': -1 if is_3d_weight else 0,
+            'act_hidden_dim': 1 if is_3d_weight else 0,
+            'scale_diffs_transpose': None if is_3d_weight else (1, 0),
```
ljaljushkin:
```suggestion
```

src/nncf/quantization/algorithms/weight_compression/activation_stats.py
```diff
@@ -27,17 +27,32 @@ def process_stats(stats: WCTensorStatistic, subset_size: int) -> tuple[Tensor, T
...
+    )  # [SampleSize, HiddenDim] for 2-D or [SampleSize, No. of Experts, HiddenDim] for 3-D
+    n_dim = len(X.shape)
+    if n_dim == 2:
+        X_full = fns.transpose(X)  # [HiddenDim, SampleSize]
+    if n_dim == 3:
+        X_full = fns.transpose(X, axes=(1, 2, 0))  # [No. of Experts, HiddenDim, SampleSize]
 
     # prevent high memory and time consumption
-    if X_full.shape[1] > subset_size:
+    subset_axis = 1 if n_dim == 2 else 2  # axis for subset_size dimension
```
ljaljushkin: I believe, it's possible achieve the same without if-condition by slightly changing transpose:

```python
    # Stack: [SampleSize, HiddenDim] for 2-D or [SampleSize, No. of Experts, HiddenDim] for 3-D
    X = fns.stack(stats.mean_values)

    # Move SampleSize to the last axis: [HiddenDim, SampleSize] or [No. of Experts, HiddenDim, SampleSize]
    # General approach: move axis 0 to the end
    axes = list(range(1, len(X.shape))) + [0]
    X_full = fns.transpose(X, axes=axes)

    # The sample dimension is always the last axis after transpose
    sample_axis = -1

    # Prevent high memory and time consumption by sampling
    if X_full.shape[sample_axis] > subset_size:
        # Activations were reduced across all but the last dimension
        lens = [reduce(mul, shape[:-1], 1) for shape in stats.shape_values]
        step = X_full.shape[sample_axis] // subset_size
        sorted_idxs = [i[0] for i in sorted(enumerate(lens), key=lambda x: -x[1])][::step]
        idxs = [idx for idx in sorted_idxs if idx < X_full.shape[sample_axis]][:subset_size]

        # Create index slices for all dimensions except the last one
        # This works for both 2D and 3D (and theoretically any dimensionality)
        index_slices = [slice(None)] * (len(X_full.shape) - 1) + [idxs]
        X = X_full[tuple(index_slices)]
    else:
        X = X_full

    # Compute max magnitude along the sample axis (last axis)
    # Result: [HiddenDim] or [No. of Experts, HiddenDim]
    s = fns.max(fns.abs(X_full), axis=sample_axis)
```

ljaljushkin (changes requested):

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -196,10 +196,22 @@ def calculate_quantization_params(
...
+        }
 
         was_transposed = False
-        if reduction_axis == 0:
-            weight = fns.transpose(weight)
+        if reduction_axis == 0 or (reduction_axis == 2 and is_3d_weight):
+            # Weights
+            # MoE: [num_experts, out_features, hidden_dimension] -> [num_experts, hidden_dimension, out_features]
+            # Default: [out_features, in_features] -> [in_features, out_features]
+            weight = fns.transpose(weight, axes=axes['weight_transpose'])
```
ljaljushkin:
```suggestion
            weight = fns.moveaxis(weight, -1, -2)
```

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -236,15 +248,23 @@ def calculate_quantization_params(
         denum = fns.sum(importance, axis=2, keepdims=True)
         importance = importance / (denum + eps)
 
-        X, _ = reshape_weight_for_grouped_quantization(X, 0, group_size)
+        X, _ = reshape_weight_for_grouped_quantization(X, axes['act_hidden_dim'], group_size)
```
ljaljushkin: why not just -2?

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -236,15 +248,23 @@ def calculate_quantization_params(
...
         best_diffs = None
         result_scale = None
-        fp_outs = fns.matmul(fns.transpose(original_weight, (1, 0, 2)), X)
-        q_outs = fns.matmul(fns.transpose(q_weights, (1, 0, 2)), X)
+
+        fp_outs = fns.matmul(fns.transpose(original_weight, axes['matmul_transpose']), X)
```
ljaljushkin: could you please add more comments about shapes for weights, activations in different cases (2d,3d,group-wise)? 
not trivial why 'matmul_transpose': (0, 1, 3, 2) if is_3d_weight else (1, 0, 2) is working correctly for all them

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -196,10 +196,22 @@ def calculate_quantization_params(
         X = X.astype(TensorDataType.float32)
         weight = weight.astype(TensorDataType.float32)
         eps = fns.finfo(weight).eps
+        is_3d_weight = len(weight.shape) == 3
+
+        axes = {
+            'weight_transpose': (0, 2, 1) if is_3d_weight else None,
+            'matmul_transpose': (0, 1, 3, 2) if is_3d_weight else (1, 0, 2),
```
ljaljushkin: please use variables, not dict. most of the variables might be even not needed, if we comments below are valid.

ljaljushkin (changes requested):

src/nncf/quantization/algorithms/weight_compression/activation_stats.py
anzr299: Alright, this is much clever. I will update it

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/activation_stats.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Changed, it.
Done

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: to avoid this difference and make generalization easier,
I made changes to the code such that the hidden dimension is always in the last dimension. It uses the transpose flow to achieve this. I have generalized it now.

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
anzr299: Done

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
ljaljushkin: Cool! Could you please run performance tests via dedicated job in jenkins to track possible improvement/degradation in compression time and memory foorprint?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -1086,6 +1086,51 @@ def _collect_statistics(
...
+        :param graph: Model graph.
+        :return: List of port ids of all the weights that feed the node with weight.
+        """
+        node_with_weight_edge = graph.get_output_edges_by_port_id(act_node, output_port_id)
+        for node_order, edge in enumerate(node_with_weight_edge):
+            is_node_with_weights = self._backend_entity.is_node_with_weights(edge.to_node, graph)
+            if not is_node_with_weights:
+                continue
+            node_with_weight = node_with_weight_edge[node_order].to_node
+            break
```
```suggestion
            if is_node_with_weights:
                node_with_weight = node_with_weight_edge[node_order].to_node
                break
```

anzr299: Done

ljaljushkin:
Since there's known issue with CPU Plugin transformation: ticket CVS-176465, 
suggest failing for openvino backend for model with 3D weights if version less or equal 2025.4: 
https://github.com/openvinotoolkit/nncf/blob/6b2e7f76a8530da3b95ea2f3566483850403da59/tests/cross_fw/shared/openvino_version.py#L28

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -851,6 +859,19 @@ def get_weight_compression_parameters(
                             f"node name: {node.node_name}. The node will be in {self._backup_mode} mode."
                         )
 
+                    model_backend = get_backend(model)
+                    ov_version = self._maybe_get_ov_version()
+                    if (
+                        model_backend == BackendType.OPENVINO
+                        and len(weight_shape) == 3
+                        and ov_version
+                        and ov_version <= "2025.4"
```
ljaljushkin: it won't work with "2025.4.0". Suggest using `<2026`. If it will be fixed in 2025.4.1 we will simply delete it. 
https://pypi.org/project/openvino/2025.3.0/

anzr299: Done

ljaljushkin (approved):
Great extension of data-aware methods! Thanks, Aamir!

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -851,6 +859,22 @@ def get_weight_compression_parameters(
                             f"node name: {node.node_name}. The node will be in {self._backup_mode} mode."
                         )
 
+                    model_backend = get_backend(model)
+                    ov_version = self._maybe_get_ov_version()
+                    if (
+                        model_backend == BackendType.OPENVINO
+                        and len(weight_shape) == 3
+                        and ov_version
+                        and ov_version <= "2026"
```
daniil-lyakhov: Str comparison could lead to errors in some point. Please look for a similar check in the nncf repo and adjust accordingly

daniil-lyakhov (commented):
Great PR! Minor

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: Sure, done

andrey-churkin (approved):

PR #3712, opened 2025-10-31
[pt] fix processing named tuple

AlexanderDokuchaev:
### Changes

Support using named tuples as outputs of model and functions.

### DIFF
--- a/src/nncf/torch/function_hook/hook_executor_mode.py
+++ b/src/nncf/torch/function_hook/hook_executor_mode.py
@@ -452,7 +452,10 @@ class FunctionHookMode(TorchFunctionMode):
                 for idx, value in enumerate(output):
                     output[idx] = self.process_post_function_hooks_for_value(value, op_meta, idx)
                 if cls_tuple is not None:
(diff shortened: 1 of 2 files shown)


src/nncf/torch/function_hook/hook_executor_mode.py
```diff
@@ -452,7 +452,10 @@ def execute_post_hooks(self, output: Any, op_meta: OpMeta) -> Any:
                 for idx, value in enumerate(output):
                     output[idx] = self.process_post_function_hooks_for_value(value, op_meta, idx)
                 if cls_tuple is not None:
-                    output = cls_tuple(output)
+                    if hasattr(cls_tuple, "_fields"):  # likely a namedtuple
```
ljaljushkin, line 455: please add a ticket to extend for other structures, like dict

AlexanderDokuchaev: Dont think that it's need to functions, i have no seen it.
For output of the model 176194

ljaljushkin (approved):

PR #3714, opened 2025-10-31
[OpenVINO] Introduce a way to control resetting of the state of stateful OV models

nikita-savelyevv:
### Changes

Added `nncf.definitions.NNCF_DATASET_RESET_STATE_KEY` constant to specify when to reset model state. This constant is used by OpenVINO backend to control resetting of internal model state between model inferences. This key can be added to a dataset sample input dictionary with either `True` or `False` value. With `True` value, the model state will be reset before inference on the corresponding sample, and with `False` the state will not be reset.

For an example of usage please see https://github.com/huggingface/optimum-intel/pull/1505.

### Reason for changes

Without this logic static quantization quality of stateful Whisper models is poor because a state of a stateful model must be cleared with the same schedule as it is done during calibration input data collection.

### Related tickets

172705

### Tests

Added `tests/openvino/native/test_engine.py::test_stateful_model_inference_with_controlled_resetting`.

### DIFF
--- a/src/nncf/data/dataset.py
+++ b/src/nncf/data/dataset.py
@@ -30,6 +30,11 @@ class Dataset:
     usually contains both examples and labels. So transformation function should extract
     the examples from the data item.
 
+    A special input key nncf.definitions.NNCF_DATASET_RESET_STATE_KEY can be used by OpenVINO backend to control
+    resetting of internal model state between model inferences. This key can be added to a dataset sample input
+    dictionary with either `True` or `False` value. With `True` value, the model state will be reset before inference
+    on the corresponding sample, and with `False` the state will not be reset.
+
     :param data_source: The iterable object serving as the source of data items.
     :param transform_func: The function that is used to extract the model's input
         from the data item. The data item here is the data item that is returned from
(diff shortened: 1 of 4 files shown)


nikita-savelyevv:
Example usage: https://github.com/huggingface/optimum-intel/pull/1505

src/nncf/data/dataset.py
```diff
@@ -38,6 +38,12 @@ class Dataset:
         will be passed into the model as-is.
     """
 
+    # RESET_STATE_KEY is a special input key used by OpenVINO backend to control resetting of internal model state
+    # between model inferences. This key can be added to a dataset sample input dictionary with either
+    # `True` or `False` value. With `True` value, the model state will be reset before inference on the corresponding
+    # sample, and with `False` the state will not be reset.
+    RESET_STATE_KEY = "nncf_reset_state"
```
AlexanderDokuchaev: Looks like incorrect place for definition of variable.
May be better define it in `src/nncf/definitions.py` ?
And for users add note in docstring of Dataset that it can be use stateful models

AlexanderDokuchaev (changes requested):

src/nncf/data/dataset.py
nikita-savelyevv: Moved to `src/nncf/definitions.py`

PR #3716, opened 2025-11-02
[PT] Strip pruned model

AlexanderDokuchaev:
### Changes

Add strip function for pruned model
Simplify strip to avoid build graph for `IN_PLACE` and `DQ` format.

### Related tickets

176026

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/19017695102

### DIFF
--- a/tests/torch2/function_hook/pruning/helpers.py
+++ b/tests/torch2/function_hook/pruning/helpers.py
@@ -22,6 +22,7 @@ class ConvModel(nn.Module):
     def __init__(self) -> None:
         super().__init__()
         self.conv = nn.Conv2d(3, 3, 3)
+        self.conv.weight.data = torch.arange(1, 82, dtype=torch.float32).view(3, 3, 3, 3)
 
     def forward(self, x: torch.Tensor):
         x = self.conv(x)


ljaljushkin:
Nice feature! 
Will be extremely useful to reduce graph building overhead for MoE models with a lot of linear layers.

tests/torch2/function_hook/pruning/rb/test_strip.py
```diff
@@ -0,0 +1,43 @@
...
+        # Set mask
+        pruning_module.mask[0] *= -1
+        pruned_weight = pruning_module(pruned_model.conv.weight)
+
+    striped_model = nncf.strip(pruned_model, strip_format=nncf.StripFormat.IN_PLACE, do_copy=False)
+    hook_storage = get_hook_storage(striped_model)
+
+    assert not list(hook_storage.named_hooks())
+    assert torch.equal(striped_model.conv.weight, pruned_weight)
+    assert torch.count_nonzero(striped_model.conv.weight) == 54
```
ljaljushkin, line 43: I'd also check count for binary mask or after applying it

src/nncf/torch/function_hook/pruning/strip.py
```diff
@@ -0,0 +1,59 @@
...
+        if hook_type != "post_hooks" or port_id != 0:
+            msg = f"Unexpected place of SparsityBinaryMask: {hook_type=}, {op_name=}, {port_id=}"
+            raise nncf.InternalError(msg)
+
+        module_name, weight_attr_name = split_const_name(op_name)
+        module = get_module_by_name(module_name, model)
+        weight_param = getattr(module, weight_attr_name)
+
+        weight_param.requires_grad = False
+        weight_param.data = hook_module(weight_param)
```
ljaljushkin, line 53: maybe `with torch.no_grad():` is also needed

AlexanderDokuchaev: https://github.com/openvinotoolkit/nncf/pull/3716/files#diff-596206cbb3e55fa75a2947ee6daf7d3dbddb5e545e0f990278ecbfbfa43bcee6R27
Used decorator `@torch.no_grad()`

tests/torch2/function_hook/pruning/rb/test_strip.py
AlexanderDokuchaev: No think
Mask set directly by `pruning_module.mask[0] *= -1` 
`RBPruningMask` and `binary_mask` are tested in test_modules.py

src/nncf/torch/function_hook/pruning/strip.py
ljaljushkin: Oh, I didn't notice for pruning it's global for the function, because for quantization it's local via context manager. different strip method look really the same, maybe some common function would avoid different implementations

tests/torch2/function_hook/pruning/rb/test_strip.py
ljaljushkin: yes, agree. but I'm more concerned about initial weights. what if initialization is changed. the test doesn't check initial state.

ljaljushkin (approved):

PR #3717, opened 2025-11-03
[PT] Pruning statistic

AlexanderDokuchaev:
### Changes

Add `nncf.pruning_statistic` to collect and display information about pruning parameters.

```python
    pruning_stat = nncf.pruning_statistic(pruned_model)
    print(pruning_stat)
```

```
┍━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━┯━━━━━━━━━━━━━━━━━┑
│ Parameter's name   │ Shape        │   Pruning ratio │
┝━━━━━━━━━━━━━━━━━━━━┿━━━━━━━━━━━━━━┿━━━━━━━━━━━━━━━━━┥
│ conv.weight        │ (3, 3, 3, 3) │           0.506 │
├────────────────────┼──────────────┼─────────────────┤
│                    │              │                 │
├────────────────────┼──────────────┼─────────────────┤
│ Masked parameters  │              │           0.506 │
├────────────────────┼──────────────┼─────────────────┤
│ All parameters     │              │           0.488 │
┕━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━┷━━━━━━━━━━━━━━━━━┙
```

### Related tickets

174484

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/19045571789/job/54392808345

### DIFF
--- a/src/nncf/pruning/prune_model.py
+++ b/src/nncf/pruning/prune_model.py
@@ -83,7 +83,7 @@ class ModelPruningStatistic:
     def __str__(self) -> str:
         total = [
             [None, None, None, None],
-            ["Masked parameters", None, self.pruning_ratio],
+            ["Prunable parameters", None, self.pruning_ratio],
             ["All parameters", None, self.global_pruning_ratio],
         ]
 


daniil-lyakhov (commented):
Great job! One thing - I think we need to specify that the total ratio is relative to the amount of prunable weight, not to all model weights. 

Someone would definitely expect model size to shrink by a factor of 2 not knowing that the big embedding layer, for instance, is not prunable and contributes significantly to the total amount of weight

ljaljushkin:
> Great job! One thing - I think we need to specify that the total ratio is relative to the amount of prunable weight, not to all model weights.
> 
> Someone would definitely expect model size to shrink by a factor of 2 not knowing that the big embedding layer, for instance, is not prunable and contributes significantly to the total amount of weight

what about reporting 2 types of total sparsity rate then? 
2 possible solutions:
1)
Effective sparsity — relative to all model weights
Targeted sparsity — relative to prunable weights
2) 
Prunable sparsity — ratio of zeroed weights among prunable parameters only
Overall sparsity — ratio of zeroed weights among all model parameters

AlexanderDokuchaev:
@daniil-lyakhov  @ljaljushkin  
Added global_pruning_ration to statistics.
And use ` Masked parameters` and `All parameters` in table

Issue #3481, opened 2025-05-06
[FastBiasCorrection] Support models without the batch dimension


daniil-lyakhov:
### 🚀 Feature request

Currently FastBiasCorrection algorithm does not support models without the batch dimension:
Reproducer:

From the branch:
https://github.com/daniil-lyakhov/nncf/tree/dl/fbc_no_batch_repro

```
python -m pytest /home/dlyakhov/Projects/nncf/tests/openvino/native/test_fast_bias_correction.py -k OneDimMM
```
Output:

```
    @pytest.mark.parametrize(
        "model_cls, ref_bias",
        (
            (ConvTestModel, [0.0288348, 1.0838453]),
            (ConvBNTestModel, [0.08396978, 1.1676897]),
            (OneDimMM, [0.95773065, 1.3218939, 0.81694865]),
        ),
    )
    def test_update_bias(self, model_cls, ref_bias, tmpdir):
        model = self.backend_specific_model(model_cls(), tmpdir)
        dataset = get_static_dataset(model_cls.INPUT_SIZE, self.get_transform_fn(), self.fn_to_type)
    
        quantization_algorithm = self.get_quantization_algorithm()
        graph = NNCFGraphFactory.create(model)
>       quantized_model = quantization_algorithm.apply(model, graph, dataset=dataset)

tests/cross_fw/test_templates/test_fast_bias_correction.py:121: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
nncf/quantization/algorithms/post_training/algorithm.py:113: in apply
    return self._pipeline.run_from_step(model, dataset, graph, 0, step_index_to_statistics)
nncf/quantization/algorithms/pipeline.py:164: in run_from_step
    step_model = self.run_step(step_index, step_statistics, step_model, step_graph)
nncf/quantization/algorithms/pipeline.py:119: in run_step
    current_model = pipeline_step[-1].apply(current_model, current_graph, step_statistics)
nncf/quantization/algorithms/fast_bias_correction/algorithm.py:182: in apply
    input_blob = self._backend_entity.create_input_data(
[...]

PR #3721, opened 2025-11-04
Support models without a batch dimension

Thakor-Yashpal:
### Changes

- Modified `nncf/quantization/algorithms/fast_bias_correction/openvino_backend.py`.
- Updated the `create_input_data` method to add a specific check for 1D input shapes (i.e., `len(shape) == 1`).
- When a 1D model is detected, the function now bypasses the per-channel loop and directly uses the single mean tensor from the `data` list.

### Reason for changes

The `create_input_data` method was looping based on the channel dimension size (3) but should have been based on the statistics list size (1)

### Related tickets

Closes the feature request: "Currently FastBiasCorrection algorithm does not support models without the batch dimension"
https://github.com/openvinotoolkit/nncf/issues/3481 

### Tests

Verified the fix by running the reproducer command provided in the issue, which now passes:

```bash
python -m pytest tests/openvino/native/test_fast_bias_correction.py -k OneDimMM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,6 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-my_project_env
 # PyInstaller
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
@@ -87,7 +86,7 @@ celerybeat-schedule
 .venv
 venv/
 ENV/
-
+my_project_env/
 # Spyder project settings
 .spyderproject
 .spyproject
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-
+my_project_env
 # PyInstaller
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by MaximProshin** [comment]
@daniil-lyakhov , please take a look

File: .gitignore [hunk 1]
```diff
@@ -24,7 +24,7 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-
+my_project_env
```
daniil-lyakhov:
```suggestion
```

src/nncf/version.py
```diff
@@ -9,7 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.19.0"
+__version__ = "2.19.0.dev0+1489448a3dirty"
```
daniil-lyakhov:
```suggestion
__version__ = "2.19.0"
```

src/nncf/quantization/algorithms/fast_bias_correction/openvino_backend.py
```diff
@@ -126,3 +126,156 @@ def extract_submodel(self, model_transformer, input_id, output_id):
             output_ids=[output_id],
             node_mapping=self._node_mapping,
         )
+# Copyright (c) 2025 Intel Corporation
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
```
daniil-lyakhov: Looks like some issues with the code merging, there are two different classes with the same name in the PR

daniil-lyakhov (changes requested):
@Thakor-Yashpal, thank you for the contribution! Please apply the comments and add the test reproducer to the PR

MaximProshin:
@Thakor-Yashpal , do you have any update with this PR? Do you plan to finalize it?

PR #3734, opened 2025-11-11
Add deprecation warning for TensorFlow backend

AlexanderDokuchaev:
### Reason for changes

Removing support TF backend in the next support

### DIFF
--- a/src/nncf/tensorflow/__init__.py
+++ b/src/nncf/tensorflow/__init__.py
@@ -57,5 +57,5 @@ from nncf.tensorflow.sparsity.rb import algorithm as rb_sparsity_algorithm
 from nncf.tensorflow.utils.state import ConfigState
 
 warning_deprecated(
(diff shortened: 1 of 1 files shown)


src/nncf/tensorflow/__init__.py
```diff
@@ -54,3 +55,7 @@
 from nncf.tensorflow.sparsity.magnitude import algorithm as magnitude_sparsity_algorithm
 from nncf.tensorflow.sparsity.rb import algorithm as rb_sparsity_algorithm
 from nncf.tensorflow.utils.state import ConfigState
+
+warning_deprecated(
+    "The TensorFlow backend is deprecated and will be removed in the next release."
```
MaximProshin: will be removed in the next release. => will be removed in a future release.

AlexanderDokuchaev: updated

MaximProshin (approved):

PR #3550, opened 2025-06-18
[OV] Optimized compression to MXFP4 data type

nikita-savelyevv:
### Changes

Added optimized compression to MXFP4 data type for OpenVINO backend.

| Model                | Memory Before (MiB) | Memory After (MiB)         | Time Before (sec) | Time After (sec)         |
|----------------------|---------------------|----------------------------|-------------------|--------------------------|
| llama-3.2-1b bf16    | 2778.55             | 2548.37 (-8.29%)           | 66.33             | 24.95 (-62.40%)          |
| llama-3.2-1b fp16    | 3434.61             | 2963.03 (-13.73%)          | 62.12             | 24.85 (-59.98%)          |
| llama-3.2-1b fp32    | 2041.79             | 1576.43 (-22.81%)          | 62.72             | 25.85 (-58.77%)          |
|                      |                     |                            |                   |                          |
| phi4-mini bf16       | 6384.81             | 5725.66 (-10.33%)          | 197.36            | 66.22 (-66.44%)          |
| phi4-mini fp16       | 8863.53             | 8375.75 (-5.51%)           | 195.85            | 66.93 (-65.82%)          |
| phi4-mini fp32       | 4406.82             | 3897.91 (-11.54%)          | 195.25            | 68.83 (-64.76%)          |
|                      |                     |                            |                   |                          |
| llama-3.1-8b bf16    | 7297.72             | 6096.06 (-16.46%)          | 413.86            | 135.25 (-67.32%)         |
| llama-3.1-8b fp16    | 7946.93             | 8311.89 (+4.58%)           | 413.64            | 136.05 (-67.11%)         |
| llama-3.1-8b fp32    | 7310.48             | 5043.89 (-30.03%)          | 411.45            | 140.66 (-65.81%)         |

### Reason for changes

Improving user experience.

### Related tickets

164717

### Tests

Extended optimized compression tests.

PR #3737, opened 2025-11-14
[OpenVINO] Optimized weight compression for FP4 mode

nikita-savelyevv:
### Changes

Added optimized weight compression through OpenVINO models for FP4 compression mode. Results should be similar to MXFP4 (#3550).

### Reason for changes

Improving UX.

### Tests

Extended `tests/openvino/optimized_functions/test_compression_functions.py`

### DIFF
--- a/src/nncf/openvino/optimized_functions/models.py
+++ b/src/nncf/openvino/optimized_functions/models.py
@@ -704,7 +704,7 @@ def _build_integer_quantize_dequantize_weight_model(
             compressed_weight = ov_results[0]
             scale = ov_parameters[1]
 
-    decompressed_weight = opset.multiply(convert_op(compressed_weight, ov.Type.f32), scale)
+    decompressed_weight = opset.multiply(scale, convert_op(compressed_weight, ov.Type.f32))
 
     ov_results = [decompressed_weight] + ov_results if return_compressed_weight else [decompressed_weight]
 
--- a/tests/openvino/optimized_functions/test_compression_functions.py
+++ b/tests/openvino/optimized_functions/test_compression_functions.py
@@ -26,6 +26,7 @@ from nncf import Dataset
 from nncf.common.factory import NNCFGraphFactory
 from nncf.common.utils.caching import ResultsCache
 from nncf.common.utils.caching import cache_results
+from nncf.openvino.cpu_info import is_arm_cpu
 from nncf.openvino.graph.node_utils import get_const_value_as_ov_tensor
 from nncf.quantization.algorithms.weight_compression.config import WeightCompressionConfig
 from nncf.quantization.algorithms.weight_compression.weight_lowering import MIN_INPUT_SIZE_FOR_OPTIMIZED_COMPRESSION
@@ -124,6 +125,10 @@ def openvino_available(available: bool):
         yield
 
 
+@pytest.mark.xfail(
+    is_arm_cpu(),
+    reason="Due to a bug in CPU plugin compression models can fail at compilation on ARM CPUs. Ticket: 164135.",
+)
 @pytest.mark.parametrize(
     "weight_shape,is_disabled",
     [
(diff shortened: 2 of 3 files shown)


tests/openvino/optimized_functions/test_compression_functions.py
```diff
@@ -123,10 +124,6 @@ def openvino_available(available: bool):
         yield
 
 
-@pytest.mark.xfail(
-    is_arm_cpu(),
-    reason="Due to a bug in CPU plugin compression models can fail at compilation on ARM CPUs. Ticket: 164135.",
-)
```
nikita-savelyevv, line 129: This issue is fixed since OpenVINO 2025.2.

src/nncf/openvino/optimized_functions/functions.py
```diff
@@ -108,16 +108,15 @@ def do_float_quantization(
...
-    NF4 format uses 16 levels in [-1, 1] range, while MXFP4 uses 16 levels in [-6, 6].
+    NF4 format uses 16 levels in [-1, 1] range, while FP4/MXFP4 uses 16 levels in [-6, 6].
 
     :param weight: Weight array to compress.
     :param config: Weight compression configuration.
     :param reduction_axes: Axes, along which to reduce (collect) different statistics.
     :param precomputed_scale: Optional precomputed scale.
-    :return: Returns quantized (for MXFP8_E4M3, FP4 and FP8_E4M3 normalized)
-        weight tensor and corresponding scale tensor.
+    :return: Returns quantized weight tensor and corresponding scale tensor.
```
nikita-savelyevv, line 117: MXFP8_E4M3 and FP8_E4M3 are actually not supported by optimized compression, no need to mention this.

src/nncf/quantization/algorithms/weight_compression/constants.py
```diff
@@ -33,7 +33,7 @@
     dtype=np.float32,
 )
 
-MXFP4_QUANTILES = np.array(
+F4E2M1_QUANTILES = np.array(
```
nikita-savelyevv, line 36: mxfp4 is a compression format with f4e2m1 weight, f8e8m0 scale and group size 32. While this grid is defined according to f4e2m1 data type, irrespective of other parameters. So renamed.

src/nncf/openvino/optimized_functions/models.py
```diff
@@ -626,8 +625,15 @@ def _build_float_quantization_model(
         eps = np.finfo(np.float32).eps
         scale = opset.select(opset.less(opset.abs(scale), eps), eps, scale)
 
+        # Equals 1.0 for NF4
```
daniil-lyakhov, line 628: Hmmm, is the 6.0 max val correct here?

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
```diff
@@ -183,7 +187,7 @@ def do_float_quantization(
     if scale is None:
         scale = calculate_float_quantization_params(weight, reduction_axes, config)
     norm_weight = _calculate_normalized_weight(weight, scale)
-    if config.mode in [CompressWeightsMode.NF4, CompressWeightsMode.MXFP4]:
+    if config.mode in [CompressWeightsMode.NF4, CompressWeightsMode.MXFP4, CompressWeightsMode.FP4]:
```
daniil-lyakhov, line 196: What do you think if we make a tuple like `MODES_TO_OPTIMIZE = (CompressWeightsMode.NF4, CompressWeightsMode.MXFP4, CompressWeightsMode.FP4)` to avoid duplicated definition of the same list

daniil-lyakhov (commented):
Great contribution! A couple minor comments

src/nncf/openvino/optimized_functions/models.py
nikita-savelyevv: Yes https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/weight_lowering.py#L102

src/nncf/quantization/algorithms/weight_compression/weight_lowering.py
nikita-savelyevv: Good idea! Added. However in this particular case, this logic is not connected with which data types are supported by optimized compression, but rather which data types `_calculate_float_quantized_weight` function supports.

tests/openvino/optimized_functions/test_compression_functions.py
nikita-savelyevv: Decided to resolve this in a separate PR: https://github.com/openvinotoolkit/nncf/pull/3743

src/nncf/openvino/optimized_functions/models.py
daniil-lyakhov: But what the comment stands for than?

daniil-lyakhov (approved):

src/nncf/openvino/optimized_functions/models.py
nikita-savelyevv: What I meant was we don't divide by max_val for nf4 data type because max_val equals to 1 for it

PR #3725, opened 2025-11-06
[Rebase] Support transposed input for data-aware Weights Compression

daniil-lyakhov:
### Changes

Rebase of the PR #3296

### Reason for changes

To make it possible to run AWQ/Scale estimation for the mamba models

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/3230

PR #3746, opened 2025-11-19
[AWQ] Expand scale dims to match activation dims

daniil-lyakhov:
### Changes

AWQ: multiply scale shape is expanded to match the activation shape length: Mamba models have an AWQ pattern which has 3 dims, and insertion of default 2d scale leads to an error during inference

### Reason for changes

To support AWQ algo for mamba models

### Related tickets

173277

### Tests

tests/cross_fw/test_templates/template_test_weights_compression.py::test_awq_scale_reference is updated to test the non mergable AWQ branch, testing the branch + new reshape implementation

### DIFF
--- a/src/nncf/quantization/algorithms/weight_compression/awq.py
+++ b/src/nncf/quantization/algorithms/weight_compression/awq.py
@@ -194,11 +194,20 @@ class AWQ(Algorithm):
                     merge_weight = self._backend_entity.get_weight(merge_node, port_id, model, graph)
                     merge_weight = (merge_weight * a_scale).astype(weight_dtype)
                     self._backend_entity.set_weight(merge_node, port_id, model, graph, merge_weight)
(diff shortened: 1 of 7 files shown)


src/nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -194,11 +194,20 @@ def apply(
                     merge_weight = self._backend_entity.get_weight(merge_node, port_id, model, graph)
                     merge_weight = (merge_weight * a_scale).astype(weight_dtype)
                     self._backend_entity.set_weight(merge_node, port_id, model, graph, merge_weight)
-                a_scale = fns.transpose(a_scale)
```
daniil-lyakhov, line 197: As this `a_scale` is always being squeezed before the usage, there is no need to do transpose here

daniil-lyakhov:
Will be covered in #3725

PR #3747, opened 2025-11-20
[Torch FX] Fix "None" Tensor Shape In NNCFGraph Edges

anzr299:
### Changes

Modify nncf graph builder for torch fx to obtain the tensor shape for edges for some metatypes like avgpool1d, median metatype etc. 
These are cases where there are multiple output nodes.

### Reason for changes

Fix failing tests in executorch openvino backend.

### Related tickets

176898

PTQ Conformance test #756: Pass


src/nncf/experimental/torch/fx/nncf_graph_builder.py
```diff
@@ -75,10 +75,15 @@ def _map_fx_unique_metatypes(node: torch.fx.Node, metatype: om.OperatorMetatype)
...
+        PT_METATYPE_TO_FX_METATYPE_MAPPING = {
+            om.PTEmbeddingMetatype: om.PTAtenEmbeddingMetatype,
+            om.PTEmbeddingBagMetatype: om.PTAtenEmbeddingBagMetatype,
+        }
+        if metatype in PT_METATYPE_TO_FX_METATYPE_MAPPING:
+            fx_metatype = PT_METATYPE_TO_FX_METATYPE_MAPPING[metatype]
+            weight_node = node.args[fx_metatype.weight_port_ids[0]]
             if weight_node.op == "get_attr":
-                return om.PTAtenEmbeddingMetatype
+                return fx_metatype
```
daniil-lyakhov, line 86: Please find a way to test it in NNCF

anzr299: Sure, I will extend the embedding test for embedding bag

anzr299: Done

src/nncf/torch/graph/operator_metatypes.py
```diff
@@ -839,6 +856,30 @@ class PTRELU6Metatype(PTOperatorMetatype):
...
+@PT_OPERATOR_METATYPES.register()
+class PTAMaxMetatype(PTOperatorMetatype):
+    name = "AMaxOp"
+    module_to_function_names = {NamespaceTarget.TORCH: ["amax"], NamespaceTarget.ATEN: ["amax"]}
+
+
+@PT_OPERATOR_METATYPES.register()
+class PTArgMaxMetatype(PTOperatorMetatype):
+    name = "ArgMaxOp"
+    module_to_function_names = {NamespaceTarget.TORCH: ["argmax"], NamespaceTarget.ATEN: ["argmax"]}
```
daniil-lyakhov: Usually we skip such kind of the metatypes as such nodes are not used in any pattern matching or propagation, so it does not matter this node has a valid metatype or the unknown metatype.

Do this metatypes participate in any kind of algorithms? If not, I would suggest to remove the unused metatypes

src/nncf/torch/graph/operator_metatypes.py
```diff
@@ -759,6 +775,7 @@ class PTAvgPool3dMetatype(PTOperatorMetatype):
     hw_config_names = [HWConfigOpName.AVGPOOL]
 
 
+@PT_OPERATOR_METATYPES.register()
```
daniil-lyakhov, line 778: Good catch!

src/nncf/experimental/torch/fx/nncf_graph_builder.py
```diff
@@ -190,7 +196,16 @@ def get_edge_params(
...
                 tensor = source_node.meta["val"][0]
-            elif source_nncf_node.metatype in [om.PTSplitMetatype, om.PTMaxMetatype, om.PTMinMetatype]:
+            elif source_nncf_node.metatype in [
+                om.PTSplitMetatype,
+                om.PTMaxMetatype,
+                om.PTMinMetatype,
+                om.PTMedianMetatype,
+                om.PTAdaptiveMaxPool1dMetatype,
+                om.PTAdaptiveMaxPool2dMetatype,
+                om.PTAdaptiveMaxPool3dMetatype,
```
daniil-lyakhov, line 206: Great!

src/nncf/torch/graph/operator_metatypes.py
anzr299: It doesn't participate in any algorithm; I added it here since I was checking all operators in executorch op test and adding them one by one to provide full coverage. I left it mid-way due to no real benefit and left the ones I did add. I will remove it.

anzr299: Done

AlexanderDokuchaev (approved):

PR #3748, opened 2025-11-20
Optimized compression for FP8 modes

nikita-savelyevv:
### Changes

Added optimized OpenVINO weights compression for fp8e4m3 data type.

`optimum-cli export openvino` time:

| Model                     | Memory Before (MiB) | Memory After (MiB)        | Time Before (sec) | Time After (sec)         |
|---------------------------|---------------------|----------------------------|-------------------|---------------------------|
| Llama-3.2-1B        | 2328.03             | 2394.14 (+2.84%)           | 63.52             | 14.03 (-77.92%)           |
| Phi-4-mini           | 5608.48             | 5197.70 (-7.33%)           | 187.34            | 28.05 (-85.03%)           |
| Llama-3.1-8B | 9918.52             | 8443.87 (-14.86%)          | 399.14            | 48.48 (-87.86%)           |

### Reason for changes

UX improvement. 

### Tests

Extended existing tests.

https://github.com/openvinotoolkit/nncf/actions/runs/19767009608


src/nncf/quantization/algorithms/weight_compression/fp8_conversion.py
```diff
@@ -0,0 +1,149 @@
...
+        # (f8_bits already contains the sign)
+        pass
+
+    return f8_bits & 0xFF
+
+
+_f16_to_f8e4m3_bits_vec = np.vectorize(_f16_to_f8e4m3_bits_scalar, otypes=[np.uint8])
+
+
+def fp32_to_fp8e4m3_values(x: np.ndarray) -> np.ndarray:
```
nikita-savelyevv: The reference conversion implementation looks a bit ugly unfortunately. It also can't be done within our `nncf.Tensor` framework yet, because `_f16_to_f8e4m3_bits_scalar` has to be vectorized with NumPy.

The reason behind all this, is that fp32 ->f8e4m3 conversion is done through fp16, i.e. fp32 -> fp16 -> f8e4m3. It is implemented this way on OpenVINO (and OneDNN) side because it is a more hardware friendly way. Please see the following links:
1. OpenVINO implementation: https://github.com/openvinotoolkit/openvino/blame/master/src/core/src/type/float8_e4m3.cpp
2. PR with addition of 2-step conversion: https://github.com/openvinotoolkit/openvino/pull/28501

nikita-savelyevv: Resolved

src/nncf/quantization/algorithms/weight_compression/config.py
```diff
@@ -46,7 +46,17 @@ def num_bits(self):
...
+            8
+            if self.mode
+            in [
+                CompressWeightsMode.INT8_SYM,
+                CompressWeightsMode.INT8_ASYM,
+                CompressWeightsMode.FP8_E4M3,
+                CompressWeightsMode.MXFP8_E4M3,
+            ]
+            else 4
+        )
```
AlexanderDokuchaev:
```suggestion
        if self.mode in [
            CompressWeightsMode.INT8_SYM,
            CompressWeightsMode.INT8_ASYM,
            CompressWeightsMode.FP8_E4M3,
            CompressWeightsMode.MXFP8_E4M3,
        ]:
            return 8
        return 4
```

src/nncf/openvino/optimized_functions/functions.py
```diff
@@ -116,7 +116,7 @@ def do_float_quantization(
     :param precomputed_scale: Optional precomputed scale.
     :return: Returns quantized weight tensor and corresponding scale tensor.
     """
-    assert config.mode in [CompressWeightsMode.NF4, CompressWeightsMode.MXFP4, CompressWeightsMode.FP4]
```
AlexanderDokuchaev, line 120: Using list of supported modes is preferable, as it will not need update on adding new modes.
And it can be defined as variable like `SUPPORTED_MODES`.

AlexanderDokuchaev (approved):

src/nncf/quantization/algorithms/weight_compression/config.py
nikita-savelyevv: Changed

src/nncf/openvino/optimized_functions/functions.py
nikita-savelyevv: Introduced variables `OPTIMIZED_COMPRESSION_COMPATIBLE_INT_MODES`, `OPTIMIZED_COMPRESSION_COMPATIBLE_FLOAT_MODES` and their union `OPTIMIZED_COMPRESSION_COMPATIBLE_MODES`. Reused the last one at `weight_lowering.py`.

PR #3751, opened 2025-11-21
Fix onnx Attention and torch SDPA quantization handling

### LINKED ISSUE #3750: `MULTIHEAD_ATTENTION_OUTPUT` ignored patterns don't match "proper" SDPA / Attention
### 🐛 Describe the bug

Currently, the `MULTIHEAD_ATTENTION_OUTPUT` ignore patterns for [`onnx`](https://github.com/openvinotoolkit/nncf/blob/09246b2bc7c244ae873171f08994be5ec8ac6623/src/nncf/onnx/quantization/ignored_patterns.py#L88) and [`torch`](https://github.com/openvinotoolkit/nncf/blob/09246b2bc7c244ae873171f08994be5ec8ac6623/src/nncf/torch/quantization/ignored_patterns.py#L84) only work for "decomposed" versions of attention by matching against `MATMUL` and `SOFTMAX` nodes in particular arrangements.

This means that `torch` models using the fused [`torch.nn.functional.scaled_dot_product_attention`](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) operator and `onnx` models using the [`Attention`](https://onnx.ai/onnx/operators/onnx__Attention.html) node (opsets 23+) don't get matched.

The edge between `MATMUL` and `SOFTMAX` doesn't need to be matched, since it is already "hidden inside" the SDPA / Attention nodes. However, the other `MATMUL` input should correspond to the `V` (third) input of SDPA / Attention.

I am willing to look into contributing a fix for this, but I am not 100% sure if I can fully figure this out on my own.

<details>
<summary>

### Environment

</summary>

```
nncf==2.18.0
torch==2.8.0
about-time==4.2.1
alive-progress==3.3.0
anyio==4.11.0
attrs==25.4.0
autograd==1.8.0
certifi==2025.11.12
click==8.3.0
cma==4.4.0
coloredlogs==15.0.1
contourpy==1.3.3
cycler==0.12.1
Deprecated==1.3.1
dill==0.4.0
filelock==3.20.0
flatbuffers==25.9.23
fonttools==4.60.1
fsspec==2025.10.0
graphemeu==0.7.2
h11==0.16.0
hf-xet==1.2.0
httpcore==1.0.9
httpx==0.28.1
huggingface_hub==1.1.4
humanfriendly==10.0
idna==3.11
Jinja2==3.1.6
joblib==1.5.2
jsonschema==4.25.1
jsonschema-specifications==2025.9.1
kiwisolver==1.4.9
markdown-it-py==4.0.0
MarkupSafe==3.0.3
matplotlib==3.10.7
mdurl==0.1.2
ml_dtypes==0.5.3
mpmath==1.3.0
natsort==8.4.0
networkx==3.4.2
ninja==1.13.0
nncf==2.18.0
numpy==2.2.6
nvidia-cublas-cu12==12.8.4.1
nvidia-cuda-cupti-cu12==12.8.90
nvidia-cuda-nvrtc-cu12==12.8.93
nvidia-cuda-runtime-cu12==12.8.90
nvidia-cudnn-cu12==9.10.2.21
nvidia-cufft-cu12==11.3.3.83
nvidia-cufile-cu12==1.13.1.3
nvidia-curand-cu12==10.3.9.90
nvidia-cusolver-cu12==11.7.3.90
nvidia-cusparse-cu12==12.5.8.93
nvidia-cusparselt-cu12==0.7.1
nvidia-nccl-cu12==2.27.3
nvidia-nvjitlink-cu12==12.8.93
nvidia-nvshmem-cu12==3.3.20
nvidia-nvtx-cu12==12.8.90
onnx==1.19.1
onnx-ir==0.1.12
onnxruntime==1.23.2
onnxscript @ git+https://github.com/ruro/onnxscript.git@ae22c2ff1f9816b3559f65b7019cd9f9ad4203ce
openvino-telemetry==2025.2.0
packaging==25.0
pandas==2.3.3
pillow==12.0.0
protobuf==6.33.1
psutil==7.1.3
pydot==3.0.4
Pygments==2.19.2
pymoo==0.6.1.5
pyparsing==3.2.5
python-dateutil==2.9.0.post0
pytz==2025.2
PyYAML==6.0.3
referencing==0.37.0
rich==14.2.0
rpds-py==0.29.0
safetensors==0.7.0
scikit-learn==1.7.2
scipy==1.16.3
setuptools==80.9.0
shellingham==1.5.4
six==1.17.0
sniffio==1.3.1
sympy==1.14.0
tabulate==0.9.0
threadpoolctl==3.6.0
timm==1.0.22
torch==2.8.0
torchvision==0.23.0
tqdm==4.67.1
triton==3.4.0
typer-slim==0.20.0
typing_extensions==4.15.0
tzdata==2025.2
wrapt==2.0.1
```

```
Additionally:
OS                  NixOS 25.11
Python              3.13.5
Install             PyPI
RAM                 32.00 GB
CPU                 12th Gen Intel(R) Core(TM) i9-12900HK
CUDA                12.8
```

</details>

### Minimal Reproducible Example

```python
import torch
import nncf
import timm

sdpa = timm.layers.attention.Attention(1, 1)
input_sample = {
    "x": torch.zeros(1, 1, 1),
}

sdpa = nncf.quantize(
    sdpa,
    calibration_dataset=nncf.Dataset([input_sample]),
    model_type=nncf.ModelType.TRANSFORMER,
    preset=nncf.QuantizationPreset.PERFORMANCE,
    target_device=nncf.TargetDevice.CPU,
)
```

### Are you going to submit a PR?

- [x] Yes I'd like to help by submitting a PR!

ruro:
### Changes

- added `ONNXAttentionMetatype` for the opset 23 `Attention` ONNX node
- fixed `scaled_dot_product_attention` quantization in `torch2` for the case when `Q`, `K` and `V` are parallel edges coming from the same input node

### Reason for changes

See #3750

### Related tickets

Fixes #3750

### Tests

- `tests/onnx/quantization/test_graphs.py::test_synthetic_models_graph[AttentionModel]` attention_model dot

- `tests/torch2/function_hook/quantization/test_quantized_graphs.py::test_quantized_graphs[unbind_scaled_dot_product_attention_model]` unbind_scaled_dot_product_attention_model dot

### DIFF
--- a/tests/onnx/models.py
+++ b/tests/onnx/models.py
@@ -11,11 +11,21 @@
 
 import numpy as np
 import onnx
+import pytest
+from packaging import version
 
 from nncf.common.utils.registry import Registry
 from tests.onnx.common import get_random_generator
 
-OPSET_VERSION = 13
+DEFAULT_OPSET_VERSION = 13
+OPSET_AVAILABILITY = [
+    # (opset_version, onnx_version) see https://github.com/onnx/onnx/releases
+    (24, "1.19.0"),
+    (23, "1.18.0"),
+    # ONNX versions below 1.17.0 are not tested,
+    # so opsets 22 and below are always available
+]
+
 ALL_SYNTHETIC_MODELS = Registry("ONNX_SYNTHETIC_MODELS")
 
 
@@ -29,11 +39,32 @@ def create_initializer_tensor(
 
 
 class ONNXReferenceModel:
+    required_opset_version = DEFAULT_OPSET_VERSION
+
     def __init__(self, onnx_model, input_shape: list[list[int]], graph_path):
         self.onnx_model = onnx_model
         self.onnx_model.ir_version = 9
         self.input_shape = input_shape
         self.path_ref_graph = graph_path
+
+    @classmethod
+    def get_pytest_marks(cls):
+        required_opset = cls.required_opset_version
+        onnx_version = version.parse(onnx.__version__)
+
+        for opset_version, available_since in OPSET_AVAILABILITY:
+            is_required = required_opset >= opset_version
+            is_available = onnx_version >= version.parse(available_since)
+
+            if is_required and not is_available:
+                model_name = cls.__name__
+                return pytest.mark.skip(
+                    reason=f"Model {model_name} requires opset {required_opset} "
+                    f"that is available since ONNX v{available_since}, but ONNX "
+                    f"v{onnx.__version__} is currently installed",
+                )
+
+        return ()
 
 
 @ALL_SYNTHETIC_MODELS.register()
@@ -161,7 +192,7 @@ class LinearModel(ONNXReferenceModel):
             ],
         )
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape], "linear_model.dot")
@@ -208,7 +239,7 @@ class MultiInputOutputModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape_1, input_shape_2, input_shape_3], "multi_input_output_model.dot")
@@ -251,7 +282,7 @@ class DoubleInputOutputModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape_1, input_shape_2], "double_input_output_model.dot")
@@ -286,7 +317,7 @@ class ModelWithIntEdges(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape], "int_edges_model.dot")
@@ -334,7 +365,7 @@ class OneConvolutionalModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         self.conv_bias = conv1_B
@@ -391,7 +422,7 @@ class OneConvolutionalIdentityBiasModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         self.conv_bias = conv1_B
@@ -468,7 +499,7 @@ class ReshapeWeightModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape], "reshape_weight_model.dot")
@@ -546,7 +577,7 @@ class WeightSharingModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape], "weight_sharing_model.dot")
@@ -598,7 +629,7 @@ class OneInputPortQuantizableModel(ONNXReferenceModel):
         )
 
         op = onnx.OperatorSetIdProto()
-        op.version = OPSET_VERSION
+        op.version = self.required_opset_version
         model = onnx.helper.make_model(graph_def, opset_imports=[op])
         onnx.checker.check_model(model)
         super().__init__(model, [input_shape], "one_input_port_quantizable_model.dot")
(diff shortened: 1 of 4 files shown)


ruro:
Hm. iirc `onnx` added support for opset 23 in version `1.18.0`. So the new test is currently failing in CI due to
```
onnx==1.17.0; python_version < '3.13'
onnx==1.18.0; python_version >= '3.13'
```

Do you have any preferences if I should mark this test as
```python
@pytest.mark.skipif(
    version.parse(onnx.__version__) < version.parse("1.18.0"),
    reason="Opset 23 was added in onnx 1.18.0",
)
```
or bump the version or something else?

andrey-churkin:
> Hm. iirc `onnx` added support for opset 23 in version `1.18.0`. So the new test is currently failing in CI due to
> 
> ```
> onnx==1.17.0; python_version < '3.13'
> onnx==1.18.0; python_version >= '3.13'
> ```
> 
> Do you have any preferences if I should mark this test as
> 
> ```python
> @pytest.mark.skipif(
>     version.parse(onnx.__version__) < version.parse("1.18.0"),
>     reason="Opset 23 was added in onnx 1.18.0",
> )
> ```
> 
> or bump the version or something else?

Hi @ruro, thanks for your contribution. We currently support multiple versions of ONNX, and the `Attention` operator was added in opset 23, which corresponds to ONNX 1.18.0. I believe we should run this test only for ONNX versions >= 1.18.0.

src/nncf/common/insertion_point_graph.py
```diff
@@ -236,9 +236,10 @@ def _get_default_pre_hook_ip_list(nncf_graph: NNCFGraph) -> list[PreHookInsertio
 
             for pred_node in pred_nodes:
                 input_edge = nncf_graph.get_edge(pred_node, nncf_node)
-                allowed_pre_hook_insertion_points.append(
-                    PreHookInsertionPoint(nncf_node.node_name, input_edge.input_port_id)
-                )
+                input_port_ids = [input_edge.input_port_id] + input_edge.parallel_input_port_ids
+                node_name = nncf_node.node_name
+                for input_port_id in input_port_ids:
+                    allowed_pre_hook_insertion_points.append(PreHookInsertionPoint(node_name, input_port_id))
```
andrey-churkin, line 242: @ruro Could you please briefly explain why these changes are necessary?

@daniil-lyakhov Please take a look

ruro: I've outlined my reasoning in the last two comments in #3750. The short version is that parallel edges aren't directly represented in the `PTNNCFGraph` (because it's not a Multi graph and doesn't allow repeated edges), but are instead stored in the `parallel_input_port_ids` property.

In this case, `unbind` has 3 outputs that are passed as `q`, `k` and `v` inputs of the `sdpa` node. Each of these 3 edges should be considered separately for the purposes of quantizer insertion/propagation, but the previous logic only added insertion points for "real" edges, ignoring any extra parallel edges.

Let me know if anything is unclear.

daniil-lyakhov: Great contribution! Could you please share a netron/ nncf graph visualization of the brand new supported subgraph? (nncf graph visualization api: https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/common/graph/graph.py#L611-L613)

This part of the code is the core logic of the NNCF, we need to figure out all possible side effects of this change

ruro: I am not sure, what you mean by "netron/nncf graph". The second image in the PR body is the expected graph for unbind+sdpa after applying quantization. Does that work?

ruro: Also, here are the before and after graphs, obtained by performing a `torch.onnx.export` (with `opset_version=23`) of a quantized `timm.layers.attention.Attention` module:

before

after

(The edge without the q/dq nodes is the `V` input of `Attention` as expected)

tests/onnx/models.py
```diff
@@ -1275,12 +1306,48 @@ def __init__(self):
...
         super().__init__(model, [x_input_shape, y_input_shape], "activation_matmul_model.dot")
 
 
+@ALL_SYNTHETIC_MODELS.register()
+class AttentionModel(ONNXReferenceModel):
+    #         Q   K   V
+    #          \  |  /
+    #         Attention
+    #             |
+    #            OUT
```
daniil-lyakhov, line 1321: Can we have the same model as in the torch test? So we can cover parallel edges in the ONNX case. Thank you!

ruro: In torch
```python
query, key, value = x.unbind(0)
return nn.functional.scaled_dot_product_attention(query, key, value)
```
results in a graph that looks like
```
           X
           |
        unbind
        /  |  \
       |   |   |
        \  |  /
scaled_dot_product_attention
           |
          OUT
```
But I believe that the "equivalent" ONNX graph converts `unbind` to a combination of `Split` and `Squeeze`, so the resulting graph looks something like
```
           X
           |
       ,-Split-,
      /    |    \
     /     |     \
    /      |      \
Squeeze Squeeze Squeeze
    \      |      /
     \     |     /
      \    |    /
       Attention
           |
          OUT
```
so there are no true "parallel" edges (in this case) because of the extra `Squeeze` nodes.

<s>

I suppose, I could make `X` a tensor of shape `3, 8, 16` and then apply the `Split` along the 0 (batch) axis. That graph **would** have parallel edges, but it won't be truly equivalent to the one in the torch test.

Which would you prefer for the ONNX test: a graph that is strictly equivalent to the torch one, but with no parallel edges, or a slightly different graph that does have parallel edges?

</s>

---

Actually, on some further investigation, parallel edges are just completely broken right now in ONNX. They aren't even created/recorded properly during the conversion from `onnx.ModelProto` to `NNCFGraph`. Compare how the torch backend creates edges

https://github.com/openvinotoolkit/nncf/blob/f438c627b9c13bbde091c41fded54fb441c752df/src/nncf/torch/function_hook/nncf_graph/nncf_graph_builder.py#L204-L212

how the openvino backend creates edges

https://github.com/openvinotoolkit/nncf/blob/f438c627b9c13bbde091c41fded54fb441c752df/src/nncf/openvino/graph/nncf_graph_builder.py#L116-L124

and how the ONNX backend does it

https://github.com/openvinotoolkit/nncf/blob/f438c627b9c13bbde091c41fded54fb441c752df/src/nncf/onnx/graph/nncf_graph_builder.py#L441-L448

The ONNX backend just doesn't handle `parallel_input_port_ids`.

When there are multiple edges with the same `in_node_id` and `output_node_id` (aka parallel edges), attempting to run `add_edge_between_nncf_nodes` on them will only end up adding a single edge (the last `add_edge_between_nncf_nodes` call will overwrite the previous ones).

I think that fixing this issue for the ONNX backend is a bit outside the scope of this PR.

daniil-lyakhov: Great catch! Could you please open a ticket with all the details above and add a comment to the ONNX test case? Perhaps we can convert it to another GFI

PR #3752, opened 2025-11-21
[Data-Aware WC] Fix Statistics Processing for Data Aware Methods

anzr299:
### Changes

Fix logic to sample activation values from the collected statistics which resolves a bug.

### Reason for changes

An error was introduced with the addition of SE support for 3D weights #3706. This was spotted by the failing test https://github.com/openvinotoolkit/nncf/actions/runs/19549771001

### Tests
1. Extended Scale estimation test `test_scale_estimation()` in `tests/cross_fw/test_templates/template_test_weights_compression.py` to include the case where calibration dataset size > scale estimation subset size
2. Added `test_process_stats` in weights compression template test for testing the process_stats function.

Example Test: https://github.com/openvinotoolkit/nncf/actions/runs/19580896526
WC Conformance Test: https://github.com/openvinotoolkit/nncf/actions/runs/19577215441

Precision Type | Filter | Value | Stderr
-- | -- | -- | --
INT4 SYM Per-Channel (with Scale estimation) After Fix | flexible-extract | 0.61 | 0.0490
  | strict-match | 0.40 | 0.0492
INT4 SYM Per-Channel (with Scale estimation) Before fix | flexible-extract | 0.66 | 0.0476
  | strict-match | 0.38 | 0.0488
INT4 SYM Per-Channel | flexible-extract | 0.77 | 0.0423
  | strict-match | 0.28 | 0.0451
FP16 | flexible-extract | 0.91 | 0.0288
  | strict-match | 0.86 | 0.0349

### DIFF
--- a/tests/cross_fw/test_templates/template_test_weights_compression.py
+++ b/tests/cross_fw/test_templates/template_test_weights_compression.py
@@ -663,19 +663,21 @@ class TemplateWeightCompression(ABC):
         return 1
 
     @pytest.mark.parametrize(
-        "mean_values_shape,num_samples,subset_size,expected_s_shape,expected_X_shape",
+        "mean_values_shape,num_samples,subset_size,expected_s_shape,expected_X_shape,expected_indices",
         [
             # 2D Activations
-            ((8,), 10, 5, (8,), (8, 5)),
-            ((8,), 5, 10, (8,), (8, 5)),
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev (changes requested):
Please add a unit test for process_stats function

AlexanderDokuchaev (approved):

PR #3755, opened 2025-11-24
Allow data-free compression with 3D weights

nikita-savelyevv:
### Changes

Update the condition introduced in https://github.com/openvinotoolkit/nncf/pull/3706 to allow data-free compression with 3D weights.

### Reason for changes

Only data-aware compression is not possible. Caught with optimum-intel tests https://github.com/huggingface/optimum-intel/actions/runs/19623687915/job/56188570802.

### DIFF
--- a/src/nncf/quantization/algorithms/weight_compression/algorithm.py
+++ b/src/nncf/quantization/algorithms/weight_compression/algorithm.py
@@ -877,7 +877,7 @@ class WeightCompression(Algorithm):
                     ):
                         # MoE operations are usually matmuls, so the check for matmul metatype is done
                         # This is to avoid raising the error for non-MoE cases with 3D weights.
-                        msg = f"""NNCF algorithms do not support 3D weights with current version of 
+                        msg = f"""NNCF compression algorithms do not support 3D weights with current version of 
                                 Openvino {ov_version} due to a known issue in statistics collection Ticket - 176465. 
                                 Node with weight: {node.node_name}."""
                         raise nncf.UnsupportedModelError(msg)


src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -869,12 +869,19 @@ def get_weight_compression_parameters(
                         and ov_version
                         and version.parse(ov_version) <= version.parse("2026")
                         and node.metatype in self._backend_entity.matmul_metatypes
+                        and (
+                            self._awq
+                            or self._scale_estimation
+                            or self._gptq
+                            or self._lora_correction
+                            or self._sensitivity_metric == SensitivityMetric.HESSIAN_INPUT_ACTIVATION
```
anzr299:
```suggestion
                            self._data_aware_mixed_precision
                            or self._data_aware_compression
```
A small change. perhaps we can use these attributes to check data aware

anzr299 (commented):
Great catch, Thank you for the correction!

src/nncf/quantization/algorithms/weight_compression/algorithm.py
nikita-savelyevv: Thanks! Unfortunately, `self._data_aware_mixed_precision` can't be used because only HAWQ is affected. Regarding `self._data_aware_compression`, that's a good point. I've also kept `self._awq` since both data-aware and data-free AWQ are affected.

ljaljushkin: Is data-free really affected? The issue was only in statistics collection, data-free awq shouldn't collect activ. stats.

anzr299: It is possibly the common steps where the reduction axes is not appropriate
https://github.com/openvinotoolkit/nncf/blob/d1d65fcb76024ccf288d6417941fe0f03cecde5c/src/nncf/quantization/algorithms/weight_compression/awq.py#L186
and finally in the merging/insertion-of-inverse-of-scale logic:
https://github.com/openvinotoolkit/nncf/blob/d1d65fcb76024ccf288d6417941fe0f03cecde5c/src/nncf/quantization/algorithms/weight_compression/awq.py#L192

PR #3759, opened 2025-11-26
[NNCF] Enable data-aware weight compression for MatMul with transpose_b=False

### LINKED ISSUE #3494: [Good First Issue][NNCF]: Support not transposed weight for data-aware weight compression methods
### Context

Matmul operation in OpenVINO assumes an implicit shape alignment for input arguments. It applies transpositions specified by optional `transpose_a ` and `transpose_b` attributes: [OV spec](https://docs.openvino.ai/2024/documentation/openvino-ir-format/operation-sets/operation-specs/matrix/matmul-1.html). 
Currently, [weight compression](https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/post_training_compression/weights_compression/Usage.md) in NNCF does not support `transpose_b`=False. 
Here's the [test](https://github.com/openvinotoolkit/nncf/pull/3296/files#diff-223ea638f7751f7c0c3e8f867ec9c8c132a3ccd62a9dcea2a5d158836c71c222R1484).
Potentially, it affects Mixed-Precision, AWQ, Scale Estimation, GPTQ and Lora Correction algorithms.

### What needs to be done?

The task is to enable data-aware weight compression methods (Mixed-Precision, AWQ, Scale Estimation, Lora Correction, GPTQ) for models with matrix multiplications having not transposed weight.

[test_compression_with_transpose](https://github.com/openvinotoolkit/nncf/pull/3296/files#diff-223ea638f7751f7c0c3e8f867ec9c8c132a3ccd62a9dcea2a5d158836c71c222R1484) shouldn't raise an error for `transpose_b=False`

### Example Pull Requests

https://github.com/openvinotoolkit/nncf/issues/3230
https://github.com/openvinotoolkit/nncf/pull/3296

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers
- [How to link your Pull Request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue-using-the-pull-request-sidebar)

### Contact points

@ljaljushkin 

### Ticket

_No response_

Varshith-Yadav:
### Changes
fixes #3494 
Added full support for data-aware weight compression when `MatMul` nodes use `transpose_b=False`.
Updated and validated `test_compression_with_transpose` to ensure it passes for `transpose_b=False`.

### Reason for changes
Previously, NNCF’s weight compression flow assumed that the weight input of `MatMul `operations was always transposed (`transpose_b=True`).

### Tests
`pytest tests/openvino/native/quantization/test_weights_compression.py -v`
(All tests pass; test_scale_estimation[True] remains the expected XFAIL for ticket 176465.)


ljaljushkin:
@Varshith-Yadav, thank you for the contribution! 
Is it possible to avoid many if conditions and do transpose once to keep original logic for transposed weights?

ljaljushkin:
please also add unit tests.
at least you can copy-paste from #3725: https://github.com/openvinotoolkit/nncf/pull/3725/files#diff-223ea638f7751f7c0c3e8f867ec9c8c132a3ccd62a9dcea2a5d158836c71c222R1955-R1979 and make sure exception is not raised for transpose_b=False

ljaljushkin:
> @Varshith-Yadav, thank you for the contribution! Is it possible to avoid many if conditions and do transpose once to keep original logic for transposed weights?

I have reconsidered and now believe that transposing each weight can extend the total compression duration. What about implementing and utilizing a "slice_weight" method with a transpose parameter?

Varshith-Yadav:
@ljaljushkin 
That makes sense. I agree that explicitly transposing the full weight tensor could introduce unnecessary overhead

I will update the implementation to use a `slice_weight `helper method . This way, we can fetch the necessary channels dynamically based on the `transpose_b `parameter without physically reshaping the underlying tensor.

I'll proceed with this approach and update the PR shortly.

Varshith-Yadav:
@ljaljushkin 
I've updated the implementation as requested. I added a slice_weight helper in utils.py to handle the data access without performing a full transpose, and refactored the GPTQ logic to use it.

I also added a new test file test_utils_slice_weight.py to verify the helper works correctly for both Numpy and PyTorch tensors with different transpose_b settings.

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -27,6 +27,13 @@
 from nncf.quantization.algorithms.weight_compression.config import WeightCompressionParameters
 from nncf.quantization.algorithms.weight_compression.parameters import CompressedWeight
 from nncf.quantization.algorithms.weight_compression.scale_estimation import ScaleEstimation
+from nncf.quantization.algorithms.weight_compression.utils import (
+    assign_weight_column,
+    assign_weight_slice,
+    extract_weight_column,
+    slice_weight,
+    zero_mask_columns,
```
ljaljushkin: I believe you need just 2 methods

```python
def get_weight_slice(weight: Tensor, slice_obj: Union[int, slice, Tensor], is_transposed: bool) -> Tensor:
     return weight[:, slice_obj] if is_transposed else weight[slice_obj, :]
```

```python
def set_weight_slice(weight: Tensor, slice_obj: Union[int, slice, Tensor], value: Tensor, is_transposed: bool) -> None:
    if is_transposed:
        weight[:, slice_obj] = value
    else:
        weight[slice_obj, :] = value
```

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -212,18 +219,20 @@ def _quantize_weights(
...
-            msg = "Transpose is not supported"
-            raise RuntimeError(msg)
 
         weight_tensor = self._backend_entity.get_weight(
             wc_params.node_with_weight, wc_params.weight_port_id, model, graph
         )
         weight_tensor = fns.astype(weight_tensor, TensorDataType.float32)
+        
+        # Get transpose_b value to handle weight shape correctly
+        transpose_b = wc_params.node_with_weight.layer_attributes.constant_attributes[wc_params.weight_port_id]["transpose"]
```
ljaljushkin: the same issue should be in other data-aware algorithms: awq, lora_correction, scale_estimation
Support copy-pasting a test for `transpose_b=False` + all these methods and check whether it fails: https://github.com/openvinotoolkit/nncf/pull/3725/files#diff-223ea638f7751f7c0c3e8f867ec9c8c132a3ccd62a9dcea2a5d158836c71c222R1960-R1961

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -27,6 +27,13 @@
 from nncf.quantization.algorithms.weight_compression.config import WeightCompressionParameters
 from nncf.quantization.algorithms.weight_compression.parameters import CompressedWeight
 from nncf.quantization.algorithms.weight_compression.scale_estimation import ScaleEstimation
+from nncf.quantization.algorithms.weight_compression.utils import (
```
ljaljushkin: `utils` name violates the code style: https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#474-file-naming
Possible name: tensor_slicing.py

ljaljushkin (changes requested):

ljaljushkin:
also recommend configuring automatic code formating: https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#2-automating-code-formatting

Varshith-Yadav:
@ljaljushkin Thanks for the detailed feedback! I have updated the PR with the following changes:

Refactored tensor_slicing: I implemented the simplified get_weight_slice and set_weight_slice helpers exactly as suggested (using generic slicing) in src/nncf/quantization/algorithms/weight_compression/tensor_slicing.py.

Algorithm Support: I updated GPTQ, AWQ, Scale Estimation, and LoRA Correction to use these new helpers. They now correctly identify the reduction axis based on transpose_b to handle non-transposed weights.

Testing:

Added test_compress_weights_algorithms_transpose_b_false  which successfully verifies that all 4 algorithms work on a model with transpose_b=False without crashing.

Added a new test file tests/openvino/native/test_weight_compression_utils.py to unit-test the helpers with both Numpy and PyTorch tensors.

Formatting: Ran pre-commit to apply the automatic Ruff formatting.

Ready for review!

ljaljushkin:
Thanks @Varshith-Yadav!
@daniil-lyakhov, could you please help with the review?

src/nncf/quantization/algorithms/weight_compression/tensor_slicing.py
daniil-lyakhov, line 1: I'm not sure we really need to make a separate function for that. An example on how to do slicing using the build in slice: https://github.com/openvinotoolkit/nncf/pull/3725/files#diff-cefaf6a4a2cb473c23106efa01889f05dc899e43c0dfc74ef8e8d60830e8a467R276-R281

src/nncf/quantization/algorithms/weight_compression/lora_correction.py
```diff
@@ -121,6 +121,12 @@ def calculate_adapters(
         layer_name = wc_params.node_with_weight.node_name
         layer_statistics = self._statistics[layer_name]
         is_debug = self._debug_interface is not None
+
+        # Get transpose_b value to handle weight shape correctly
+        transpose_b = wc_params.node_with_weight.layer_attributes.constant_attributes[wc_params.weight_port_id][
+            "transpose"
+        ]
```
daniil-lyakhov: Looks like an openvino/onnx specific code, please introduce a backend method to get this value from the model

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -212,18 +214,22 @@ def _quantize_weights(
...
 
         weight_tensor = self._backend_entity.get_weight(
             wc_params.node_with_weight, wc_params.weight_port_id, model, graph
         )
         weight_tensor = fns.astype(weight_tensor, TensorDataType.float32)
 
+        # Get transpose_b value to handle weight shape correctly
+        transpose_b = wc_params.node_with_weight.layer_attributes.constant_attributes[wc_params.weight_port_id][
+            "transpose"
+        ]
```
daniil-lyakhov: Openvino/ONNX specific code as well

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -141,6 +141,9 @@ def apply(
 
             weight = self._backend_entity.get_weight(wp.node_with_weight, weight_port_id, model, graph)
 
+            # Get transpose_b value to handle weight shape correctly
+            transpose_b = wp.node_with_weight.layer_attributes.constant_attributes[weight_port_id]["transpose"]
+
```
daniil-lyakhov: Openvino/ONNX specific code in a common algorithm

src/nncf/quantization/algorithms/weight_compression/awq.py
```diff
@@ -220,6 +222,9 @@ def _data_aware_step(self, wp, weight, statistics, prev_weight=None, prev_statis
         assert isinstance(wp.reduction_axes, tuple) and len(wp.reduction_axes) == 1
         reduction_axis = wp.reduction_axes[0]
 
+        # Get transpose_b value to handle weight shape correctly
+        transpose_b = wp.node_with_weight.layer_attributes.constant_attributes[weight_port_id]["transpose"]
+
```
daniil-lyakhov: Onnx/Openvino specific code in a common algorithm

daniil-lyakhov (commented):
@Varshith-Yadav, thank you for your contribution! 
My initial comments are below. Besides that, I believe we have to expand tests for each compression algorithm with the `transpose_b` option for each backend. I'm working on a similar issue right now (support of `transpose_a`), and when I'll finish my tests I'll share them with you so you can do the same in your PR.

Thank you!

Varshith-Yadav:
@daniil-lyakhov Thanks for the review!

I have removed the specific OpenVINO attribute checks from the common algorithms. I introduced `get_weight_transpose_b` in the Backend interface and implemented it in all backends (openvino_backend.py, onnx_backend.py, torch_backend.py, torch_fx_backend.py).

I removed the custom utils.py and switched to using standard Python slice() objects and inline if/else checks as suggested.

I've kept my current integration test (`test_compress_weights_algorithms_transpose_b_false` in `test_utils_slice_weight.py`) included for now to verify the fix works. I am happy to update/replace this with your standardized test pattern once you share it.

daniil-lyakhov:
@Varshith-Yadav, please rebase

Issue #3041, opened 2024-10-28
[Good First Issue][NNCF]: Replace common tensor_statistics by experimental tensor_statistics


kshpv:
### Context

Historically, NNCF has maintained two sets of tensor statistics: the old tensor statistics and the new tensor statistics. The new tensor statistics, located in the `experimental/common/tensor_statistics` directory, offer improved functionality and better performance. However, the codebase still contains references to the old tensor statistics, leading to redundancy and potential confusion.

### What needs to be done?

1. Search the codebase for all instances where the old tensor statistics are being used. This includes algorithms, tests, and any utility functions.
2. Modify the identified algorithms to utilize the new tensor statistics from `experimental/common/tensor_statistics`.
3. Update the corresponding tests to ensure they are compatible with the new tensor statistics.
4. Replace the contents of `common/tensor_statistics/collectors` with the corresponding implementations from `experimental/common/tensor_statistics`.
5. Once all references to the old tensor statistics have been updated, remove the `experimental/common/tensor_statistics`.
6. Remove `common/tensor_statistics/reduction` if it is not needed anymore.

In the end, there should be no `experimental/common/tensor_statistics`.

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers
- [How to link your Pull Request to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-to-an-issue-using-the-pull-request-sidebar)

### Contact points

@kshpv

olegkkruglov:
.take

AHB102:
@kshpv So basically functionally replace everything in old files(common) with the new files(Experimental) and then delete the new files(Experimental) ?

kshpv:
@AHB102, you also need to be sure that all tests passed and no regression is introduced

AHB102:
@kshpv Great! I've reviewed the example and will create a PR for a single-file change. We can go from there

kshpv:
@AHB102 I am really sorry, but @olegkkruglov already picked this GFI. I suggest you to pick another one. 
We have many things to do. I think we probably open new GFIs in the following days. 
If you are interested I can ping you. What do you say?

mlukasze:
[...]

PR #3761, opened 2025-11-27
[Collectors] Experimental collectors to common

daniil-lyakhov:
### Changes

* Redundant legacy tensor collectors are removed
* Experimental tensor collectors moved to common
* MyPy fixed for the collector classes
* Some reducres/aggregators tests are extended
* hash functions for some aggregators are fixed

### Reason for changes

Reduce amount of redundant code, move common tensor collectors to common directory

### Related tickets

150471
175176
https://github.com/openvinotoolkit/nncf/issues/3041

### DIFF
--- a/src/nncf/common/tensor_statistics/aggregator.py
+++ b/src/nncf/common/tensor_statistics/aggregator.py
@@ -13,7 +13,7 @@ from abc import ABC
 from abc import abstractmethod
 from itertools import islice
 from pathlib import Path
-from typing import Any, Optional, TypeVar, cast
+from typing import Any, Optional, TypeVar
 
 import nncf
 from nncf.common import factory
@@ -118,7 +118,12 @@ class StatisticsAggregator(ABC):
         """
         for _, statistic_point, tensor_collector in self.statistic_points.get_tensor_collectors():
             statistics = tensor_collector.get_statistics()
-            statistics = cast(TensorStatistic, statistics)
+            if not isinstance(statistics, TensorStatistic):
+                msg = (
+                    "Please use nncf.common.tensor_statistics.statistics.TensorStatistic"
+                    " initializing the TensorCollector to make it possible to dump and load statistics"
+                )
+                raise nncf.InternalError(msg)
             statistics_key = self._get_statistics_key(statistics, statistic_point.target_point)
             if statistics_key not in data:
                 msg = f"Not found statistics for {statistics_key}"
@@ -146,7 +151,12 @@ class StatisticsAggregator(ABC):
         data_to_dump = {}
         for _, statistic_point, tensor_collector in self.statistic_points.get_tensor_collectors():
             statistics = tensor_collector.get_statistics()
-            statistics = cast(TensorStatistic, statistics)
+            if not isinstance(statistics, TensorStatistic):
+                msg = (
+                    "Please use nncf.common.tensor_statistics.statistics.TensorStatistic"
+                    " initializing the TensorCollector to make it possible to dump and load statistics"
+                )
+                raise nncf.InternalError(msg)
             statistics_key = self._get_statistics_key(statistics, statistic_point.target_point)
             data = statistics.get_data(is_serialized=True)
             data_to_dump[statistics_key] = data
--- a/src/nncf/common/tensor_statistics/collectors.py
+++ b/src/nncf/common/tensor_statistics/collectors.py
@@ -183,7 +183,7 @@ class AggregatorBase:
         self._num_samples = num_samples
         self._collected_samples = 0
         self._window_size = window_size
-        self._container: Any = deque(maxlen=window_size)
+        self._container: deque[Any] = deque(maxlen=window_size)
 
     @property
     def num_samples(self) -> Optional[int]:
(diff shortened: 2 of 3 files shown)


src/nncf/common/tensor_statistics/aggregator.py
```diff
@@ -118,6 +118,7 @@ def _load_statistics(self, data: dict[str, Any]) -> None:
         """
         for _, statistic_point, tensor_collector in self.statistic_points.get_tensor_collectors():
             statistics = tensor_collector.get_statistics()
+            statistics = cast(TensorStatistic, statistics)
```
AlexanderDokuchaev: Can `get_statistics` return dict[str, any]? 
If it's correct, so there is need to use check `isinstance(statistics, TensorStatistic)` instead of cast

src/nncf/common/tensor_statistics/aggregator.py
```diff
@@ -145,6 +146,7 @@ def _prepare_statistics(self) -> dict[str, Any]:
         data_to_dump = {}
         for _, statistic_point, tensor_collector in self.statistic_points.get_tensor_collectors():
             statistics = tensor_collector.get_statistics()
+            statistics = cast(TensorStatistic, statistics)
```
AlexanderDokuchaev: same

src/nncf/common/tensor_statistics/collectors.py
```diff
@@ -9,370 +9,1287 @@
...
+            Aggregator never skips registration if num_samples is None.
+        :param window_size: Number of samples from the end of the list of collected samples to aggregate.
+        Aggregates all available collected statistics in case parameter is None.
+        """
+        self._aggregation_axes = (0,) if aggregation_axes is None else aggregation_axes
+        self._keepdims = True
         self._num_samples = num_samples
+        self._collected_samples = 0
+        self._window_size = window_size
+        self._container: Any = deque(maxlen=window_size)
```
AlexanderDokuchaev: Incorrect annotation so in result uses `cast` for `self.container` in subclasses

src/nncf/common/tensor_statistics/statistics.py
```diff
@@ -9,114 +9,300 @@
...
                 return False
-        return True
+            for pct in self.percentile_vs_values_dict:
+                if not fns.allclose(self.percentile_vs_values_dict[pct], other.percentile_vs_values_dict[pct]):
+                    return False
+            return True
+        return False
+
+    @classmethod
+    def from_config(cls, config: Union[dict[str, Any], dict[tuple[str, int], Any]]) -> TensorStatistic:
```
AlexanderDokuchaev: Config can be Any? 
Looks like there is some problems with annotation of function.
Dont use cast instead of isinstance check.

src/nncf/common/tensor_statistics/collectors.py
daniil-lyakhov: Container should be Any as it could be Any
https://github.com/daniil-lyakhov/nncf/blob/dl/exp_stats_to_prod/src/nncf/common/tensor_statistics/collectors.py#L927

src/nncf/common/tensor_statistics/aggregator.py
daniil-lyakhov: Fixed

src/nncf/common/tensor_statistics/aggregator.py
daniil-lyakhov: Fixed

src/nncf/common/tensor_statistics/statistics.py
daniil-lyakhov: Didn't do isinstance check for dict and all dict content, but fixed other comments. Please suggest how to check instance if really needed

src/nncf/common/tensor_statistics/aggregator.py
AlexanderDokuchaev: Can get_statistics return dict[str, any]?
As it dont falls with nncf.InternalError after last changes,  is it mean that `get_statistics` always return `TensorStatistic`?

src/nncf/common/tensor_statistics/collectors.py
AlexanderDokuchaev: It shows that HAWQAggregator violates the principles of inheritance.

src/nncf/common/tensor_statistics/statistics.py
AlexanderDokuchaev: `config: dict[Any, Any]` ?

src/nncf/common/tensor_statistics/aggregator.py
daniil-lyakhov: get_statistics still could return a dict, indeed
https://github.com/daniil-lyakhov/nncf/blob/dl/exp_stats_to_prod/src/nncf/common/tensor_statistics/collectors.py#L372-L373

src/nncf/common/tensor_statistics/collectors.py
daniil-lyakhov: As you wish, done

src/nncf/common/tensor_statistics/statistics.py
daniil-lyakhov: Done

daniil-lyakhov:
@AlexanderDokuchaev, please take a look

PR #3772, opened 2025-12-04
Avoid unnecessary memory allocations in Scale Estimation algo

nikita-savelyevv:
### Changes

Remove expressions like `q_weights_ = fns.zeros_like(original_weight) + out` by assigning directly to `q_weights`. The replaced expression basically does copying of `out`, but it takes additional time and is not actually needed.

| Model                     | SE Time Before (sec) | SE Time After (sec)                | Total Time Before (sec) | Total Time After (sec)                   |
|---------------------------|-------------------------------------|--------------------------------------------------|--------------------------|-------------------------------------------|
| TinyLlama-1.1B-Chat-v1.0 | 176                                 | 146 (-17.05%)                                   | 226.49                  | 194.44 (-14.13%)                          |
| Phi-4-mini               | 720                                 | 546 (-24.17%)                                   | 809.00                  | 638.88 (-21.02%)                          |
| Llama-3.1-8B             | 1664                                | 1394 (-16.24%)                                  | 1805.76                 | 1547.19 (-14.33%)                         |

Total time is the time `optimum-cli export openvino` command takes to finish.

Example command: 
```
optimum-cli export openvino -m TinyLlama/TinyLlama-1.1B-Chat-v1.0 --weight-format int4 --dataset wikitext2 --scale-estimation $save_dir
```

### Reason for changes

Improving UX.

### Related tickets

CVS-177225

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/19929440594


src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -211,26 +211,25 @@ def calculate_quantization_params(
         cur_config = deepcopy(config)
         cur_config.group_size = group_size
 
-        original_weight = fns.zeros_like(weight) + weight
```
nikita-savelyevv, line 214: Values of original weight are not changed in-place below, so should be safe to reuse the same tensor.

anzr299 (approved):

PR #3780, opened 2025-12-09
Fix Bandit B404: Consider possible security implications associated with the subprocess module

jszczepa:
### Changes

Ignore possible warning/errors detected by Bandit scans associated with usage of subprocess module.

### Reason for changes

Bandit scan detect issues that needs to be addressed.

### Related tickets

CVS-177669

### Tests

-

### DIFF
--- a/src/custom_version.py
+++ b/src/custom_version.py
@@ -48,7 +48,7 @@ from __future__ import annotations
 import contextlib
 import os
 import re
-import subprocess
+import subprocess  # nosec
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


AlexanderDokuchaev:
@jszczepa Why is it necessary to add `# nosec` to every line if B404 is already ignored through [pyproject.toml](https://github.com/openvinotoolkit/nncf/blob/develop/pyproject.toml#L201)?

MaximProshin:
@jszczepa , please respond to the question above.

PR #3782, opened 2025-12-10
[OpenVINO Quantizer] Fix Method To Get WC Config from OVQuantizer in OV Adapter

anzr299:
### Changes
in openvino adapter, call the get method of OVQUantizer to obtain the weights compression config

### Reason for changes
Openvino adapter called an internal dictionary to obtain WC algorithm's comrpession config directly. I believe this is a bad approach, instead I have created another method in OVQuantizer such that this can be called to obtain the internal weights compression algorithm's config


src/nncf/experimental/torch/fx/quantization/quantizer/openvino_adapter.py
```diff
@@ -46,4 +46,4 @@ def get_weight_compression_parameters(
         return self._quantizer.get_nncf_weight_compression_parameters(model, nncf_graph)
 
     def get_weight_compression_config(self) -> dict[str, Any]:
-        return self._quantizer.weight_compression_configuration
+        return self._quantizer.get_wc_algo_configuration()
```
anzr299: This method is defined in the quantizer here: https://github.com/anzr299/executorch/blob/e9fd7656e1390e2670d7f9ac178827571627117f/backends/openvino/quantizer/quantizer.py#L138

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -1309,3 +1309,43 @@ def _get_statistics_for_weights_compression(
...
+            "lora_correction": self._lora_correction,
+            "backup_mode": self._backup_mode,
+            "compression_format": self._compression_format,
+            "advanced_parameters": self._advanced_parameters,
+        }
+
+    @classmethod
+    def from_config(
+        cls,
+        wc_config: dict[str, Any],
```
```suggestion
        config: dict[str, Any],
```

daniil-lyakhov (changes requested):
Please add a test, example tests: https://github.com/openvinotoolkit/nncf/blob/develop/tests/torch/test_serialization.py

src/nncf/experimental/torch/fx/quantization/quantizer/openvino_adapter.py
```diff
@@ -46,4 +46,4 @@ def get_weight_compression_parameters(
         return self._quantizer.get_nncf_weight_compression_parameters(model, nncf_graph)
 
     def get_weight_compression_config(self) -> dict[str, Any]:
-        return self._quantizer.weight_compression_configuration
+        return self._quantizer.get_quantizer_specific_internal_algo_config()
```
daniil-lyakhov: `get_compression_config`?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -1309,3 +1309,43 @@ def _get_statistics_for_weights_compression(
...
+            "all_layers": self._all_layers,
+            "sensitivity_metric": self._sensitivity_metric,
+            "awq": self._awq,
+            "subset_size": self._subset_size,
+            "scale_estimation": self._scale_estimation,
+            "gptq": self._gptq,
+            "lora_correction": self._lora_correction,
+            "backup_mode": self._backup_mode,
+            "compression_format": self._compression_format,
+            "advanced_parameters": self._advanced_parameters,
```
daniil-lyakhov: Please make a copy of mutable objects there and check it doesn't affect the original params in the test

daniil-lyakhov: Config should be json serializable, please check how to serialize the advanced_parameters there

src/nncf/experimental/torch/fx/quantization/quantizer/openvino_adapter.py
```diff
@@ -46,4 +46,4 @@ def get_weight_compression_parameters(
         return self._quantizer.get_nncf_weight_compression_parameters(model, nncf_graph)
 
     def get_weight_compression_config(self) -> dict[str, Any]:
-        return self._quantizer.weight_compression_configuration
```
AlexanderDokuchaev, line 49: Why was this not detected during testing?

src/nncf/quantization/algorithms/weight_compression/algorithm.py
```diff
@@ -391,6 +391,22 @@ def __init__(
     def available_backends(self) -> list[BackendType]:
         return [BackendType.OPENVINO, BackendType.TORCH, BackendType.TORCH_FX, BackendType.ONNX]
 
+    @property
```
AlexanderDokuchaev, line 394: In what context will this be used?

src/nncf/experimental/torch/fx/quantization/quantizer/openvino_adapter.py
anzr299: Before this PR it was a class variable in the quantizer. That was a bad design so we made it a seperate method inside the quantizer.

src/nncf/quantization/algorithms/weight_compression/algorithm.py
anzr299: This will be used by the quantizer to access WC algorithm's properties such as the mode, group size and so on

src/nncf/experimental/torch/fx/quantization/quantizer/openvino_adapter.py
AlexanderDokuchaev: Could you share code of quantizer?

anzr299: Yes,
https://github.com/pytorch/executorch/blob/e74c4f29679ad4935f3167cd99efa752ecf5a666/backends/openvino/quantizer/quantizer.py#L131

PR #3800, opened 2025-12-18
[PT] Fast path for wrapped model without hooks

AlexanderDokuchaev:
### Changes

1. Add fast path of forward without added hooks
2. Optimized `FunctionHookMode`:
   - Reduce calling named module/children
   - Not use ref, where it is useless
   - Remove debug code 

### Related tickets

CVS-177676

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/20350759282
https://github.com/openvinotoolkit/nncf/actions/runs/20350773599

### DIFF
--- a/src/nncf/torch/function_hook/hook_storage.py
+++ b/src/nncf/torch/function_hook/hook_storage.py
@@ -234,7 +234,7 @@ class HookStorage(nn.Module):
 
     def is_empty(self) -> bool:
         """
-        Check if there are no hook stored.
+        Check if there are no hooks stored.
 
         :return: True if both pre-hooks and post-hooks are empty, False otherwise.
         """


src/nncf/torch/function_hook/hook_executor_mode.py
```diff
@@ -294,23 +273,16 @@ def get_current_executed_op_name(self, fn_name: str) -> str:
...
-
-    def register_op(self, fn_name: str) -> None:
-        """
-        Register an operation call for the current module and increment call counter.
-
-        :param fn_name: The function name of the operation.
-        """
-        module_name = self.get_current_relative_name()
-        op_name = generate_normalized_op_name(module_name, fn_name)
         self.op_calls[op_name] += 1
```
ljaljushkin, line 286: Can it be confusing that just getting name via `get_current_executed_op_name` increase counter of calls? 
I'd increase the counter explicitly outside of this function or keep it inside `register_op`.

src/nncf/torch/function_hook/hook_executor_mode.py
```diff
@@ -294,23 +273,16 @@ def get_current_executed_op_name(self, fn_name: str) -> str:
         """
         Get the name of the current operation being executed.
 
+        Note: should be called only ones because it updates op_calls.
```
ljaljushkin:
```suggestion
        Note: should be called only once because it updates op_calls.
```
IMO, this warning can be easily overlooked

src/nncf/torch/function_hook/hook_executor_mode.py
AlexanderDokuchaev: Renamed function to 
`get_next_op_call_name` and update docstring to be clear

PR #3808, opened 2025-12-22
[PT] FBC for linear

AlexanderDokuchaev:
### Changes

Add support linear operation for FBC 

### Related tickets

CVS-111111

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/20462273440
manual/job/post_training_quantization/766/

### DIFF
--- a/src/nncf/torch/function_hook/extractor.py
+++ b/src/nncf/torch/function_hook/extractor.py
@@ -194,7 +194,7 @@ def extract_linear(
     graph: PTNNCFGraph,
     input_node: NNCFNode,
     output_node: NNCFNode,
-) -> Optional[nn.Module]:
+) -> ExtractedFunc:
     """
     Extracts a linear layer from an NNCF graph and constructs an ExtractedFunc module.
 


tests/cross_fw/test_templates/helpers.py
```diff
@@ -163,13 +163,13 @@ def forward(self, x):
...
-    INPUT_SIZE = [1, 1, 4, 4]
+    INPUT_SIZE = [1, 1, 3, 3]
 
     def __init__(self):
         super().__init__()
-        self.fc = nn.Linear(4, 2)
-        self.fc.weight.data = torch.Tensor([[0.1, 0.2, 0.3, 0.2], [0.3, -0.1, 0.2, 0.4]])
-        self.fc.bias.data = torch.Tensor([1.0, 1.1])
+        self.fc = nn.Linear(3, 2)
+        self.fc.weight.data = torch.Tensor([[0.1, 0.2, 0.3], [0.3, -0.1, 0.2]])
```
AlexanderDokuchaev, line 171: Updated FCTestModel because the previous values fell into a range in which quantized values differ between Torch and OpenVINO backends

src/nncf/torch/function_hook/extractor.py
```diff
@@ -189,17 +189,66 @@ def extract_conv(
     return nn.Sequential(conv_module, bn_module)
 
 
+def extract_linear(
+    model: nn.Module,
+    graph: PTNNCFGraph,
+    input_node: NNCFNode,
+    output_node: NNCFNode,
+) -> Optional[nn.Module]:
```
daniil-lyakhov: Why optional?

PR #3811, opened 2025-12-24
[PT] prune + quantize

AlexanderDokuchaev:
### Changes

- Support quantization after pruning
- Renamed `apply_magnitude_sparsity_binary_mask` to `apply_magnitude_binary_mask`
- `wrap_model` don not raise error if model already wrapped
- `find_const_node_in_constant_subgraph` can find weight after pruning operations 

### Related tickets

CVS-171577

### DIFF
--- a/tests/torch2/function_hook/pruning/magnitude/test_strip.py
+++ b/tests/torch2/function_hook/pruning/magnitude/test_strip.py
@@ -35,7 +35,7 @@ def test_strip():
     with torch.no_grad():
         pruned_weight = pruning_module(pruned_model.conv.weight)
 
-    striped_model = nncf.strip(pruned_model, strip_format=nncf.StripFormat.PRUNE_IN_PLACE, do_copy=False)
+    striped_model = nncf.strip(pruned_model, strip_format=nncf.StripFormat.IN_PLACE, do_copy=False)
     hook_storage = get_hook_storage(striped_model)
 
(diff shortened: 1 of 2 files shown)


daniil-lyakhov (commented):
Please add an example with the new functionality, otherwise ok

AlexanderDokuchaev:
@daniil-lyakhov Example is not a part of this PR

src/nncf/torch/function_hook/strip.py
```diff
@@ -58,7 +58,9 @@ def strip_model(model: TModel, example_input: Any = None, strip_format: StripFor
     elif strip_format == StripFormat.DQ:
         model = replace_quantizer_to_compressed_weight_with_decompressor(model)
     elif strip_format == StripFormat.IN_PLACE:
-        model = apply_compression_in_place(model)
+        model = apply_compression_in_place(model, (SymmetricQuantizer, AsymmetricQuantizer, BaseWeightsDecompressor))
+    elif strip_format == StripFormat.PRUNE_IN_PLACE:
+        model = apply_compression_in_place(model, (RBPruningMask, UnstructuredPruningMask))
```
ljaljushkin: I'd keep `IN_PLACE` only and fail with error if there's any operation on activations.

AlexanderDokuchaev: done

PR #3813, opened 2025-12-26
[CI] notebooks

AlexanderDokuchaev:
### Changes

- Format notebooks by pre-commit
- Update profiller.py according [PyGuide.md](https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md)
- Set strict version in tools/requirements.txt


ljaljushkin (changes requested):
although profiler compares activation, I'd restrict its usage for activations only. we can extend it to visualize weights as well. what about calling it `compression_profiler` ?

AlexanderDokuchaev:
@ljaljushkin Suggest to think about it, if somebody will extend it
compression_profiler is not clear name, and it can profiler can be run model without compression

PR #3817, opened 2025-12-30
nncf.build_graph

AlexanderDokuchaev:
### Changes

Replace `NNCFGraphFactory` to `nncf.build_graph`

### Reason for changes

Suggestion of new api function `nncf.build_graph` to provide a simple way to get NNCFGraph for users.

### DIFF
--- a/src/nncf/experimental/torch/fx/quantization/quantize_model.py
+++ b/src/nncf/experimental/torch/fx/quantization/quantize_model.py
@@ -21,6 +21,7 @@ from torch.fx import GraphModule
 from torch.fx.passes.infra.pass_manager import PassManager
 
 import nncf
+from nncf.common.factory import build_graph
 from nncf.common.logging import nncf_logger
 from nncf.common.quantization.structs import QuantizationPreset
 from nncf.data import Dataset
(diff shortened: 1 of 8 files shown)


src/nncf/quantization/algorithms/pipeline.py
```diff
@@ -11,8 +11,8 @@
 
 from typing import Optional, TypeVar, Union
 
-from nncf.common.factory import NNCFGraphFactory
 from nncf.common.factory import StatisticsAggregatorFactory
+from nncf.common.factory import build_graph
```
ljaljushkin, line 15: what about testing short import since it's API now?

AlexanderDokuchaev: Changed, use `nncf.build_graph` instead of direct import

src/nncf/openvino/quantization/quantize_model.py
```diff
@@ -87,7 +87,7 @@ def _extract_all_subgraphs(model: ov.Model, current_id: str) -> None:
         :param model: Model.
         :param current_id: Current graph id.
         """
-        graphs[current_id] = NNCFGraphFactory.create(model)
+        graphs[current_id] = build_graph(model)
```
andrey-churkin, line 90: My suggestion is to directly import the `build_graph()` method in the remaining code, and avoid using `nncf.build_graph()` in internal modules. What do you think?

andrey-churkin: I've missed the comment from @ljaljushkin. Anyway, in my opinion, it's better to use a direct import and test the API call somewhere in the tests.

AlexanderDokuchaev: ok, updated

daniil-lyakhov (approved):

PR #3822, opened 2026-01-02
[PT] Build graph for unwrapped model

AlexanderDokuchaev:
### Changes

Add support of unwrapped model to build graph

### DIFF
--- a/src/nncf/torch/function_hook/graph/build_graph_mode.py
+++ b/src/nncf/torch/function_hook/graph/build_graph_mode.py
@@ -376,13 +376,9 @@ def build_graph(model: nn.Module, *args: Any, **kwargs: Any) -> nx.MultiDiGraph:
         hook_storage = HookStorage()
         forward_fn = model.forward
 
-    with (
-        training_mode_switcher(model, is_training=False),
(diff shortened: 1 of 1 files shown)


src/nncf/torch/function_hook/graph/build_graph_mode.py
ljaljushkin, line 367:
```suggestion
   This function builds a directed graph `nx.MultiDiGraph` that represents the operations 
   and data flow within the model. The graph is constructed by leveraging model hooks via `GraphBuilderMode`.

Warning:
    This operation can be time-consuming for large models. Use with care in performance-critical contexts.
```

AlexanderDokuchaev: This message doesn’t seem to add any value. Since the function isn’t used directly (except by one other function), it won’t be visible to end users or developers.

ljaljushkin: oh, yes - it's torch only. I meant the same message but for common API function

daniil-lyakhov (approved):

PR #3835, opened 2026-01-12
 Support 3D Weights in GPTQ Algorithm

anzr299:
### Changes

The approach here is quite straightfoward, `_quantize_weights` works as usual for 2D weights. The difference is in calculate hessian where the hessian is 3D in both 3D and 2D weights case. By default hessian has the shape (1, hidden_dim, hidden_dim). 
Before this was just (hidden_dim, hidden_dim). For 3D, it is (num_experts/batch, hidden_dim, hidden_dim).

Now, this 3D hessian or "batched" hessian is looped over and the 2D weight is extracted and passed to the old `_quantize_weights` function as usual and scale/zp are returned. These scales and zp are then stacked together in a collector variable. For 2D case, it is flattened. For 3D the stacked scale, zp are returned.

NOTE: Scale Estimation + GPTQ support is not added for 3D weights yet

### Reason for changes

Support 3D weights for models like MoE in GPTQ

### Related tickets

175789 & 175212

### Tests

Model: Qwen/Qwen3-30B-A3B
NNCF Backend: OpenVINO
Higher is better.
Task: gsm8k
Limit: 100
Max New Tokens: 10000
OpenVINO version: 2026.0.0.dev20251111 (with WA for 176465)
n-shots: 5(default)

Precision Type | Filter | Value
-- | -- | --
INT4 SYM GS128 (with GPTQ) Calibrated on GSM8k with 128 samples | flexible-extract | 0.79 
  | strict-match | 0.64
INT4 SYM GS128 (with GPTQ after bug fix commit https://github.com/openvinotoolkit/nncf/pull/3835/commits/dc355fe35f459b72f0bed625098d080add74dfed) Calibrated on GSM8k with 128 samples | flexible-extract | 0.78 
  | strict-match | 0.57
INT4 SYM GS128 | flexible-extract | 0.55 
  | strict-match | 0.29 
FP32 | flexible-extract | 0.92 
  | strict-match | 0.82 

Comparison of accuracy with meta-llama/Llama-3.2-1B-Instruct on Develop and this branch
Variant | bits_per_byte | byte_perplexity | word_perplexity
-- | -- | -- | --
This Branch (GPTQ) | 0.7965 | 1.7368 | 19.1466
develop (GPTQ) | 0.7965 | 1.7368 | 19.1466


src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -12,10 +12,13 @@
 import math
 from typing import Optional, TypeVar
 
+import numpy as np
```
andreyanufr: @AlexanderDokuchaev  Is it possible to use np here ?

anzr299: if not I can use the same approach as https://github.com/openvinotoolkit/nncf/blob/d35c32b5c1bef6f4f7c121c4eac559f10a3b1f0c/src/nncf/quantization/algorithms/weight_compression/activation_stats.py#L46

anzr299: Done

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -221,10 +264,11 @@ def _quantize_weights(
             msg = "Transpose is not supported"
             raise RuntimeError(msg)
 
-        weight_tensor = self._backend_entity.get_weight(
-            wc_params.node_with_weight, wc_params.weight_port_id, model, graph
-        )
-        weight_tensor = fns.astype(weight_tensor, TensorDataType.float32)
+        if len(hessian.shape) == 3 and hessian.shape[0] == 1:
```
andreyanufr: In which model does this happen?

anzr299: For now this is just a safety check. Since for 3D case also we pass 2D hessian to this function. I added it when the older test called this function manually. Would it be better to remove it?

src/nncf/quantization/algorithms/weight_compression/gptq.py
```diff
@@ -278,6 +322,7 @@ def _quantize_weights(
                     else:
                         if self._scale_estimation and block_compression_config.num_bits == 4:
                             activations = [inp[..., (i1 + i) : (i1 + i + group_size)] for inp in inputs]
+                            # TODO(anazir): Make it work for 3D weights
```
ljaljushkin: I'd through an error as earlier as possible that GPTQ+SE doesn't work in 3D case

ljaljushkin (changes requested):
Nice addition!

src/nncf/quantization/algorithms/weight_compression/gptq.py
anzr299: I forgot to remove that TODO before
It works now

Issue #3230, opened 2025-01-30
[Good First Issue][NNCF]: Support transposed input for data-aware weight compression methods


ljaljushkin:
### Context

Matmul operation in OpenVINO assumes an implicit shape alignment for input arguments. It applies transpositions specified by optional `transpose_a ` and `transpose_b` attributes: [OV spec](https://docs.openvino.ai/2024/documentation/openvino-ir-format/operation-sets/operation-specs/matrix/matmul-1.html). 
Currently, [weight compression](https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/post_training_compression/weights_compression/Usage.md) in NNCF does not support `transpose_a`=True. 
Here's the [check](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/algorithm.py#L705-L706) and [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L1552).
Potentially, it affects Mixed-Precision, AWQ, Scale Estimation and Lora Correction algorithms.

### What needs to be done?

The task is to enable data-aware weight compression methods (Mixed-Precision, AWQ, Scale Estimation, Lora Correction) for models with transposed input matrix multiplications.
1. At least one function [`process_stats`](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/activation_stats.py#L21) should be corrected, [check](https://github.com/openvinotoolkit/nncf/blob/develop/nncf/quantization/algorithms/weight_compression/algorithm.py#L705-L706) - removed.
2. [test](https://github.com/openvinotoolkit/nncf/blob/develop/tests/openvino/native/quantization/test_weights_compression.py#L1552) should pass and be a [templated one](https://github.com/openvinotoolkit/nncf/blob/dd85320c5cbfd9af1460a5cffbcc3989b8cc610b/tests/cross_fw/test_templates/template_test_weights_compression.py) in order to check OV and Torch backend at once. 
3. Tests that used `LMLinearModel` with `transpose_a=False` by default should pass with `transpose_a=True`.
 

### Example Pull Requests

https://github.com/openvinotoolkit/nncf/pull/3179
https://github.com/openvinotoolkit/nncf/issues/3129

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/nncf/blob/develop/CONTRIBUTING.md)
[...]

PR #3839, opened 2026-01-13
[WC] Scale Estimation transpose_a support

daniil-lyakhov:
### Changes

* Support of conv `transpose_a` attribute for ONNX/OpenVINO backends in the Scale Estimation algorithm
* The static method `activations_to_wc_statistics` moved from the Scale Estimation algorithm to the GPTQ

### Reason for changes

* To enable Scale Estimation for the models with SMM (pattern has transpose_a in ONNX/OpenVINO)

### Related tickets

173277
https://github.com/openvinotoolkit/nncf/issues/3230

### Tests

tests/cross_fw/test_templates/template_test_weights_compression.py::test_scale_estimation updated with the `transpose_a` cases


src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -139,17 +139,21 @@ def apply(
...
-                raise nncf.UnsupportedModelError(msg)
-
             weight = self._backend_entity.get_weight(wp.node_with_weight, weight_port_id, model, graph)
 
+            activation_port_id = self._backend_entity.get_activation_port_id(wp.node_with_weight, graph)
+            act_shape = graph.get_input_edge_by_port_id(wp.node_with_weight, activation_port_id).tensor_shape
+            act_ch_axis = self._backend_entity.get_activation_channel_axis(
+                wp.node_with_weight, activation_port_id, act_shape
+            )
+            act_ch_axis %= len(act_shape)
```
AlexanderDokuchaev: For which case it's need?
All tests still pass if remove the line.

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -139,17 +139,21 @@ def apply(
                 continue
             _, weight_port_id = weight_data[0]
 
-            if self._backend_entity.matmul_has_transposed_activations(wp.node_with_weight, graph):
-                msg = "Transposed activations are not supported yet for the Scale Estimation algorithm"
-                raise nncf.UnsupportedModelError(msg)
-
             weight = self._backend_entity.get_weight(wp.node_with_weight, weight_port_id, model, graph)
 
+            activation_port_id = self._backend_entity.get_activation_port_id(wp.node_with_weight, graph)
```
AlexanderDokuchaev: This looks like copypast from awq.py.
Please think about to refactor it into a shared function.

tests/openvino/native/models.py
```diff
@@ -78,14 +78,18 @@ def _create_ov_model(self, input_shape=None, reshape_shape=None, matmul_w_shape=
 
 
 class SimpleMoEModel(OVReferenceModel):
-    def _create_ov_model(self, num_experts=2, hidden_dim=8, out_dim=16, seq_len=4):
+    def _create_ov_model(self, num_experts=2, hidden_dim=8, out_dim=16, seq_len=4, tranpsose_a: bool = False):
```
AlexanderDokuchaev:
```suggestion
    def _create_ov_model(self, num_experts=2, hidden_dim=8, out_dim=16, seq_len=4, transpose_a: bool = False):
```

tests/openvino/native/models.py
```diff
@@ -78,14 +78,18 @@ def _create_ov_model(self, input_shape=None, reshape_shape=None, matmul_w_shape=
...
         input_shape = [num_experts, seq_len, hidden_dim]
         input_1 = opset.parameter(input_shape, name="Input")
 
         weight_data = np.arange(0, num_experts * hidden_dim * out_dim, dtype=np.float32)
         weight_data = weight_data.reshape(num_experts, hidden_dim, out_dim)
 
-        matmul = opset.matmul(input_1, weight_data, transpose_a=False, transpose_b=False, name="MoE_MatMul")
+        if tranpsose_a:
+            transpose = opset.transpose(input_1, (0, 2, 1))
```
AlexanderDokuchaev, line 86: Please check, looks like it never runs

AlexanderDokuchaev (changes requested):

tests/openvino/native/models.py
daniil-lyakhov: Good catch,
It is, because of this skip https://github.com/daniil-lyakhov/nncf/blob/dl/sa_transpose_a/tests/openvino/native/quantization/test_weights_compression.py#L2369

I fixed the test and asked @anzr299 to remove the skip. Ticket 179366

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
daniil-lyakhov: Outdated

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
daniil-lyakhov: `WeightCompressionAlgoBackend.get_activation_channel_axis_and_shape` is instroduced, please check

daniil-lyakhov:
@AlexanderDokuchaev, @andreyanufr, please take a look

src/nncf/quantization/algorithms/weight_compression/backend.py
```diff
@@ -296,6 +296,23 @@ def get_activation_channel_axis(node: NNCFNode, port_id: int, input_shape: tuple
...
+        Returns the activation channel axis and activation tensor shape for a node with weights.
+
+        :param graph: NNCF graph containing the model structure and tensor metadata.
+        :param node: NNCF node with weights instance.
+        :return: A tuple consisting of activation channel axis and activation tensor shape.
+        """
+        activation_port_id = self.get_activation_port_id(node, graph)
+        act_shape = graph.get_input_edge_by_port_id(node, activation_port_id).tensor_shape
+        act_ch_axis = self.get_activation_channel_axis(node, activation_port_id, act_shape)
+        return act_ch_axis % len(act_shape), act_shape
```
andreyanufr, line 322: Please add comment why  % len(act_shape)

andreyanufr (approved):

src/nncf/quantization/algorithms/weight_compression/backend.py
daniil-lyakhov: Actually, onnx backend is requiring it, updated with a comment

src/nncf/quantization/algorithms/weight_compression/backend.py
```diff
@@ -296,6 +296,25 @@ def get_activation_channel_axis(node: NNCFNode, port_id: int, input_shape: tuple
...
+        Returns the activation channel axis and activation tensor shape for a node with weights.
+
+        :param graph: NNCF graph containing the model structure and tensor metadata.
+        :param node: NNCF node with weights instance.
+        :return: A tuple consisting of activation channel axis and activation tensor shape.
+        """
+        activation_port_id = self.get_activation_port_id(node, graph)
+        act_shape = graph.get_input_edge_by_port_id(node, activation_port_id).tensor_shape
+        act_ch_axis = self.get_activation_channel_axis(node, activation_port_id, act_shape)
+        # Mod the activation axis by the length of the activation shape
```
AlexanderDokuchaev: https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#424-block-and-inline-comments 

Comment that just describe build-in python operation is useless here 
Please explain why this is needed." (in which case get_activation_channel_axis returns positive or negative axis and why required here to to use only positive) 

And may be it it's problem of backend specific `get_activation_channel_axis` implementation, and will be better to fix it.

PS. For PT it's actual too, not only for ONNX.

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -266,17 +266,24 @@ def get_scale_estimation_ref(check_sampling_activation_stats_flow: bool) -> TTen
         Returns the reference output of calculate_quantization_params of ScaleEstimation.
         """
 
+    @pytest.mark.parametrize("transpose_a", [False, True])
```
AlexanderDokuchaev:
```suggestion
    @pytest.mark.parametrize("transpose_a", [False, True], ids=["no_tr_a", "tr_a"])
    @pytest.mark.parametrize("is_moe", [False, True], ids=["reg", "moe"])
    @pytest.mark.parametrize("check_sampling_activation_stats_flow", [False, True], ids=["full", "sampled"])
```
Its good practice to add ids for parameters to got readable name like, 
`test_scale_estimation[sampled-moe-tr_a]` instead of `test_scale_estimation[True-True-True]`

https://github.com/openvinotoolkit/nncf/blob/develop/docs/styleguide/PyGuide.md#62-parametrization

tests/onnx/quantization/test_weights_compression.py
```diff
@@ -493,23 +493,33 @@ def wrap_model(model: onnx.ModelProto, data: Any) -> onnx.ModelProto:
...
+
         mb = ModelBuilder()
         x = mb.add_input("input", (1, 4, 8))
         output = mb.add_output("output", (1, 4, 16))
         weights = np.arange(0, 16 * 8, dtype=np.float32).reshape(16, 8).T
-        mb.add_matmul(x, shape=(8, 16), output=output, data=weights)
+        if transpose_a:
+            squeeze = mb.add_squeeze(x)
+            transpose = mb.add_transpose(squeeze, (1, 0))
+            mb.add_gemm(transpose, shape=(8, 16), output=output, weight_data=weights, trans_a=1)
```
AlexanderDokuchaev, line 520: Why for OV used `transpose+matmul` but for onnx `squeeze + transpose + gemm`? 
Could it be defined by same operations?

tests/torch/function_hook/quantization/test_weights_compression.py
```diff
@@ -583,11 +583,11 @@ def get_sequential_matmul_model(transpose_a: bool) -> torch.nn.Module:
         return SequentialMatmulModel()
 
     @staticmethod
-    def get_model_for_test_scale_estimation():
+    def get_model_for_test_scale_estimation(transpose_a: bool):
```
AlexanderDokuchaev, line 592: Please add assert transpose_a=False to avoid any confusion from signature of method and result

And looks like using `pytest.skip` inside model creation function instead of using extra `transpose_a_supported` fixture, is shorter and easier too understand and support

```python
class TestPTTemplateWeightCompression(TemplateWeightCompression):
    @staticmethod
    def get_model_for_test_scale_estimation(transpose_a: bool):
        if transpose_a:
            pytest.skip("transpose_a=True is not supported for PT backend")
        return LinearModel(torch.arange(0, 8 * 16, dtype=torch.float32).reshape(16, 8))
```

instead of 

```python 
class TemplateWeightCompression(ABC):

    @abstractmethod
    @pytest.fixture
    def transpose_a_supported(self) -> bool:

    @pytest.mark.parametrize("transpose_a", [False, True])
    @pytest.mark.parametrize("is_moe", [False, True])
    @pytest.mark.parametrize("check_sampling_activation_stats_flow", [False, True])
    def test_scale_estimation(
        self, mocker, transpose_a, is_moe, check_sampling_activation_stats_flow, transpose_a_supported
    ):
        """Checks that scales match the reference."""
        if transpose_a and not transpose_a_supported:
            msg = "Transpose a is not supported for the current backend"
            pytest.skip(msg)


class TestPTTemplateWeightCompression(TemplateWeightCompression):
    @pytest.fixture
    def transpose_a_supported(self) -> bool:
        return False

    @staticmethod
    def get_model_for_test_scale_estimation(transpose_a: bool):
        return LinearModel(torch.arange(0, 8 * 16, dtype=torch.float32).reshape(16, 8))
```

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -195,7 +195,7 @@ def calculate_quantization_params(
         """
         reduction_axis = reduction_axes[0]
 
-        s, X = process_stats(statistics, subset_size)
+        s, X = process_stats(statistics, subset_size, act_ch_axis=act_ch_axis)
```
AlexanderDokuchaev, line 198: As far as I understand, this is the main change in this PR.  
But when I revert it, all tests still pass.

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
```diff
@@ -185,6 +184,7 @@ def calculate_quantization_params(
         :param weight: The weight tensor that is being quantized.
         :param reduction_axes: Tuple specifying the axes along which the reduction is performed for quantization.
         :param config: Configuration parameters for the weight compression, including quantization settings.
+        :param act_ch_axis: The activation channel axis.
```
AlexanderDokuchaev:
```suggestion
        :param act_ch_axis: The activation channel axis. Defaults to -1.
```

AlexanderDokuchaev (changes requested):

src/nncf/quantization/algorithms/weight_compression/backend.py
daniil-lyakhov: Why Normalization is Required:
Downstream code in AWQ and activation stats processing compares act_ch_axis with positive dimension indices. For example [in awq.py line 227:](https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/quantization/algorithms/weight_compression/awq.py#L224-L226)

Why Not Fix Backends Instead:
While it would be cleaner to standardize backend implementations, the current approach is actually correct because:

* The abstract method get_activation_channel_axis allows backends flexibility to use their natural conventions
* The normalization happens at a single point (get_activation_channel_axis_and_shape) that explicitly documents it returns positive axes
* This avoids changing multiple backend implementations and maintains backward compatibility

tests/onnx/quantization/test_weights_compression.py
daniil-lyakhov: Gemm operation does not support batch dimentions (in contrast to OpenVINO backend), to align them one have to rewrite all the tests to eliminate the batch dimentions for all backends.

I can do so but I don't see any value in that and it a lot of effort work

src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
daniil-lyakhov: Done

tests/torch/function_hook/quantization/test_weights_compression.py
daniil-lyakhov: Done

tests/cross_fw/test_templates/template_test_weights_compression.py
```diff
@@ -322,10 +321,46 @@ def get_orig_weight(model: TModel) -> Tensor:
     def get_decompressed_weight(compressed_model: TModel, input: TTensor) -> Tensor:
         """Returns decompressed weight"""
 
+    @pytest.mark.parametrize("transpose_a", [False, True], ids=["no_tr_a", "tr_a"])
+    def test_scale_estimation_act_ch_axis_param(self, mocker, transpose_a):
```
AlexanderDokuchaev, line 339: It seems this test focuses on the implementation instead of verifying the behavior after the changes

AlexanderDokuchaev (changes requested):

tests/cross_fw/test_templates/template_test_weights_compression.py
daniil-lyakhov: The process_stats function itself is completely tested here https://github.com/openvinotoolkit/nncf/blob/5bcc5bdc2f470a91d2916911f6c7ca57c1cd7c24/tests/cross_fw/test_templates/template_test_weights_compression.py#L798-L817

The stressed test is checking that the process_stats recieves correct input parameters
I believe it is enough to cover the logic.
But if you don't think so please suggest a test case to cover it properly

daniil-lyakhov:
@AlexanderDokuchaev, the PR was rebased, the brand new test `test_scale_estimation_calculate_quantization_params` is updated to cover `act_ch_axis`

AlexanderDokuchaev (approved):

PR #3848, opened 2026-01-20
[POC] GPTQModel convertor

AlexanderDokuchaev:
### Changes

Add convertor linear moduole to gptqmodel format. 
Currently support only triton kernrel. 

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/21706864333

### DIFF
--- a/examples/llm_compression/torch/gptq_model_convertor/requirements_extra.txt
+++ b/examples/llm_compression/torch/gptq_model_convertor/requirements_extra.txt
@@ -1,1 +1,3 @@
 gptqmodel==5.6.12
+# TODO(AlexanderDokuchaev): remove transformers next release gptqmodel https://github.com/ModelCloud/GPTQModel/issues/2328
+transformers==4.57.6
--- a/tests/integration/gptq_model/requirements_extra.txt
+++ b/tests/integration/gptq_model/requirements_extra.txt
@@ -1,1 +1,3 @@
 gptqmodel==5.6.12
+# TODO(AlexanderDokuchaev): remove transformers next release gptqmodel https://github.com/ModelCloud/GPTQModel/issues/2328
+transformers==4.57.6


src/nncf/experimental/torch/gptqmodel/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```python
+from nncf.experimental.torch.gptqmodel.convertor import convert_model
+
+nncf_model = nncf.compress_weights(
+    model,
+    dataset=nncf.Dataset([example_input]),
+    mode=nncf.CompressWeightsMode.INT8_ASYM,
+)
+converted_model = convert_model(nncf_model)
```
ljaljushkin, line 23: I’d prefer using CompressionFormat here. Do we need a separate function for this?
```
compression_format: CompressionFormat = CompressionFormat.GPTQ_MODEL,
```

AlexanderDokuchaev: It's just poc for only for gptqmodel. 
Which protentional format do you expect in `src/nncf/experimental/torch/gptqmodel` module?

AlexanderDokuchaev:
Speedup in example to more 2x 
```
┍━━━━━━━━━━━━━━━━━━━━━━━━┯━━━━━━━━━━━━┑
│                        │   Time (s) │
┝━━━━━━━━━━━━━━━━━━━━━━━━┿━━━━━━━━━━━━┥
│ Original               │      0.962 │
├────────────────────────┼────────────┤
│ Compressed by NNCF     │      5.674 │
├────────────────────────┼────────────┤
│ Converted to GPTQModel │      2.513 │
┕━━━━━━━━━━━━━━━━━━━━━━━━┷━━━━━━━━━━━━┙
```

src/nncf/experimental/torch/gptqmodel/convertor.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Replaces the linear layer weights with compressed INT8 weights to GPTQ format.
+
+    :param linear: The linear layer with compressed weight.
+    :param decompressor_module: The decompressor module.
+    :return: The converted GPTQModel linear layer with compressed weights.
+    """
+    num_bits = 8
+    group_size = -1
+    if isinstance(decompressor, INT8AsymmetricWeightsDecompressor):
```
ljaljushkin, line 45: In my view, it would be more convenient if the compress_weight API could produce GPTQ‑formatted weights directly, without requiring an additional call to the convert method.  
The current workflow—first inserting a Decompressor module and then replacing it with TritonV2QuantLinear—feels redundant. Since the scale and zero‑point are already computed during compression, it seems feasible to instantiate TritonV2QuantLinear immediately. Is there a technical reason this approach wouldn’t work?

From my understanding, the convert API is mainly useful when loading a previously saved NNCF‑compressed PyTorch checkpoint. In practice, though, we rarely save intermediate compressed checkpoints and typically run compression from scratch.

I’d appreciate your thoughts on whether the extra conversion step is necessary and what constraints might justify the current design.

ljaljushkin (approved):
looking forward to seeing support for int4 and the ability to save and load checkpoint in the next PR's

PR #3854, opened 2026-01-22
[TorchAO] Migrate torch.ao -> torchao

anzr299:
### Changes

Migration of torch.ao to torchao.

Notable differences in torch.ao and torchao implementation which affect test references:

1. Constant folding in torch.ao produced the new constant with the name _frozen_param{index}. In the new torchao implementation it keeps the same name as the node being replaced 
2. The old deprecated XNNPACK Quantizer was being used before which is now changed to point to the new XNNPACKQuantizer implementation in https://github.com/pytorch/executorch/tree/main/backends/xnnpack/quantizer. This implementation now explicitly quantizes the `conv_transpose2D` operator and its weights. Earlier only the activations were quantized for this op

Reference graphs changed due to above:

    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/mobilenet_v3_small.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/resnet18.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/swin_v2_t.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/synthetic_transformer.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/unet.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/vit_b_16.dot
    tests/executorch/data/ao_export_quantization_OpenVINOQuantizer/yolo11n_sdpa_block.dot
    tests/executorch/data/XNNPACKQuantizer/unet.dot (Due to difference Number 2)
    tests/executorch/data/XNNPACKQuantizer/unet_ref_qconfig.json ((Due to difference Number 2))

### Tests

PTQ-791 - Pass
Weights Compression Conformance - https://github.com/openvinotoolkit/nncf/actions/runs/21260554857 - Pass
WC Examples - https://github.com/openvinotoolkit/nncf/actions/runs/21277782722 - Pass
Install Tests - https://github.com/openvinotoolkit/nncf/actions/runs/21260085794 - Pass


JacobSzwejbka:
Oh nice! Thank you! I was just about to put up this change myself. Its blocking https://github.com/pytorch/executorch/pull/16752

anzr299:
> Oh nice! Thank you! I was just about to put up this change myself. Its blocking [pytorch/executorch#16752](https://github.com/pytorch/executorch/pull/16752)

Great! Are you aware of the release schedule for executorch? 
generally we upgrade the pytorch version and at the same time torchvision.
Can we expect Executorch to also be the same? Since I cannot see any tags for a new version

JacobSzwejbka:
> > Oh nice! Thank you! I was just about to put up this change myself. Its blocking [pytorch/executorch#16752](https://github.com/pytorch/executorch/pull/16752)
> 
> Great! Are you aware of the release schedule for executorch? generally we upgrade the pytorch version and at the same time torchvision. Can we expect Executorch to also be the same? Since I cannot see any tags for a new version

Yeah we release around the same time as PyTorch/PyTorch iirc. We cut a branch recently for 1.1 https://github.com/pytorch/executorch/issues/16365

constraints.txt
```diff
@@ -4,6 +4,7 @@ openvino==2025.4.1
 # Pytorch
 torch==2.9.0
 torchvision==0.24.0
+torchao==0.14.0
```
mergennachin: does it work with torchao 0.15? 

we just pinned the main line and release branch to 0.15

anzr299: We usually upgrade the torch and related dependency versions on nncf develop branch in a follow up PR after release branch is prepared which is soon.

Although, I suppose for Executorch release which uses torch 2.10 `torch.ao` used in nncf latest commits should be fine?

The nncf commit used here https://github.com/pytorch/executorch/pull/16752 doesn't depend on `torchao` at all and instead uses `torch.ao`. This could be a problem with 2.11 I agree. This PR should unblock that.

daniil-lyakhov: Aamir, as we discussed it with Alexander Suslov, the main idea was to get rid of torch.ao and do not use any external dependencies (like torchao)
The task is to remove `torch.ao` imports and replace them with code in the nncf codebase, no external dependencies

anzr299: Ah I see, since I couldn't see a ticket for the migration, it was a side-effect requirement for CVS-176783 which requires testing OVQuantizer from Executorch.
I will instead port torch.ao helper functions to NNCF.

tests/torch/fx/test_model_transformer.py
```diff
@@ -31,6 +31,7 @@
 from nncf.experimental.torch.fx.node_utils import get_graph_node_by_name
 from nncf.experimental.torch.fx.node_utils import get_node_args
 from nncf.experimental.torch.fx.node_utils import get_tensor_constant_from_node
+from nncf.experimental.torch.fx.transformations import _get_model_device
```
AlexanderDokuchaev: If you define function that that starts with "_" (like "private") then it should not be imported anywhere.

src/nncf/experimental/torch/fx/transformations.py
```diff
@@ -430,7 +444,11 @@ def insert_one_qdq(model: torch.fx.GraphModule, target_point: PTTargetPoint, qua
                 # With extra check of scale and zero_point being scalar, it makes
                 # sure that the default overload can be used.
                 # TODO(dlyakhov): maybe need more complex attr name here
-                qparam_node = create_getattr_from_value(model, graph, target_node.name + key, value_or_node)
+                tensor_device = value_or_node.device
+                # Passing device is neccesary to avoid large models to be cached by torchao.
```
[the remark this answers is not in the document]
AlexanderDokuchaev: it is actual

tests/torch/fx/test_model_transformer.py
anzr299: Ah, yes. I found a function to get model device in pytorch backend now. I am using that.
`from nncf.torch.utils import get_model_device`

examples/llm_compression/torch/distillation_qat_with_lora/requirements.txt
```diff
@@ -1,5 +1,6 @@
 tensorboard==2.13.0
 torch==2.9.0
+torchao==0.14.0
```
AlexanderDokuchaev: Actual version of torchao is 0.16
In the PR used two 0.14 and 0.15 
Please update to 0.16 everywhere

anzr299: Done.
I have updated it to 0.15.0 instead of 0.16.0.
I will update this version with torch version upgrade. Should be more suitable with torch 2.10.0 I believe

PR #3856, opened 2026-01-23
[doc] Update doc

AlexanderDokuchaev:
### Changes

Removed out date part about `nncf.register_module` that was removed.
Move quantization.md to `docs/usage/` as common description for ptq and qat.


ljaljushkin:
I tried to walk through nncf readme and failed to find references of algorithms to our nncf examples. We have a table with openvino tutorials only. Is it intentional? cc: @MaximProshin @AlexanderDokuchaev 

At least, make sense to refer to pruning example here:
https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/training_time_compression/pruning/Usage.md

ljaljushkin (approved):

Issue #3871, opened 2026-02-01
Support for latest numpy releases


leoll2:
### 🚀 Feature request

In `pyproject.toml`, the numpy version is currently constrained to <2.3.0, which makes it difficult to integrate `nncf` with projects that require newer numpy versions.

### Feature Use Case

A common use case is projects based on Python 3.14: numpy only supports this Python from 2.3 onwards.

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

rutiktambe:
I’ve opened a PR to relax the NumPy upper bound for Python 3.14 compatibility , Please review when convenient :
https://github.com/rutiktambe/nncf/pull/1

AlexanderDokuchaev:
@leoll2 The upcoming 3.0.0 release will support `numpy>=1.24.0,<2.5.0`

PR #3875, opened 2026-02-02
Bump numpy

AlexanderDokuchaev:
### Changes

Set `numpy>=1.24.0, <2.5.0 `

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/3871

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/21585144658/job/62191338216

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -2,7 +2,6 @@ name: weekly
 permissions: read-all
 
 on:
-  pull_request:
   workflow_dispatch:
   schedule:
     - cron: '0 0 * * 0'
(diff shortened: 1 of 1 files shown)


tests/common/accuracy_control/test_evaluator.py
```diff
@@ -80,12 +80,6 @@ class ModeTestStruct:
             values_for_each_item=[np.array(0.3)],
             expected_is_metric_mode=True,
         ),
-        # Return: (ConvertibleToFloat, List[ConvertibleToFloat])
-        ModeTestStruct(
-            metric_value=np.array([0.1]),
```
AlexanderDokuchaev, line 85: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    metric_value = metric_value if metric_value is None else float(metric_value)

PR #3913, opened 2026-02-10
[SDL] Enable Bandit B404 test

jszczepa:
### Changes

This modification helps to apply SDL policy related to Bandit scans (see: https://github.com/intel-innersource/applications.security.bandit-config/blob/main/ipas_default.config)

### Reason for changes

Each use of subprocess module need to be reviewed against cmd injection by potential attacker.

### Related tickets

N/A

### Tests

N/A


AlexanderDokuchaev (changes requested):
These changes crash the examples when run follow steps from README.

AlexanderDokuchaev:
I have rejected these changes, as the proposed implementation does not align with the original goals and introduces more technical overhead than actual value.

1. A separate shared file, symblic links (that was restricted by SDL rules) and additional imports in examples are being introduced solely to add `# nosec` comments. This feels unnecessarily complex for solving a localized issue.

2. Bandit artifacts configured via `pyproject.toml` are only used during development and are not part of the SDL pipeline, so these changes provide no practical benefit for production checks.

3. Rule `B404` triggers on the mere import of the `subprocess` module, even though `subprocess` is the standard and recommended way to execute shell commands in Python. As a result, the rule generates a significant number of false positives.

4. This PR itself clearly demonstrates the issue: enabling this rule leads exclusively to false positive findings, and the only resulting code changes are mass additions of `# nosec` comments to suppress them.

5. There are already more targeted rules, such as `B602–B607`, which analyze the actual command execution paths and provide more accurate and actionable results.

PR #3958, opened 2026-02-27
Fix weight layout detection for MatMul with transpose in OpenVINO backend

### LINKED ISSUE #3230: [Good First Issue][NNCF]: Support transposed input for data-aware weight compression methods

naraen-ram:
Fixes #3230 

## Description

Fix incorrect weight layout detection for MatMul layers when transpose is applied via OpenVINO graph rather than constant attributes.

Previously, constant_layer_attrs["transpose"] did not reflect graph-level transpose nodes, which could lead to incorrect channel axis detection during weight compression.

This change checks input_attributes metadata to correctly determine transpose state before computing layout.

## Testing

Reproduced issue using custom Transpose → MatMul OpenVINO model.

Verified using:
pytest tests/openvino/native/quantization/test_weights_compression.py -k matmul

Result:
12 passed, 0 failed

## Impact

Fixes weight compression correctness for:
- AWQ
- Mixed precision
- Scale estimation
- LoRA correction


naraen-ram:
Hi @ljaljushkin ,

Just a follow-up on this PR. I’d appreciate a review whenever convenient.

This contribution is part of my preparation for **GSoC 2026**, so getting feedback would help me move forward. Happy to make any requested changes.

Thanks!

daniil-lyakhov:
Hi @naraen-ram, thank you for your contribution.
1) Please describe a use case when this fails a model compression/quantization
2) Layer attributes should not reflect neighboring nodes in any way, the transpose layer attribute is openvino MatMul specific parameter, it does not correlate with anything

naraen-ram:
Hi @daniil-lyakhov,

During my investigation of the issue, I reproduced a scenario where the transpose appears as a separate `Transpose → MatMul` pattern in the OpenVINO graph rather than through the `transpose_a / transpose_b` attributes of the MatMul operator. In this case, the graph contains a Transpose node feeding into MatMul, while the MatMul attributes still report `transpose_a=False`. My initial implementation attempted to account for this situation during layout detection.

After revisiting the issue description again, I realized that the original task specifically refers to supporting `MatMul(transpose_a=True)` cases rather than transpose as a separate node. 

I'll reproduce a model that produces `transpose_a=True` in the MatMul attributes and adjust the implementation accordingly.

Thanks!

naraen-ram:
Hi @daniil-lyakhov, I've updated the PR.

The root cause was that when `transpose_a=True`, the hidden/channel dimension shifts to axis -2 instead of -1. The existing code just raised `UnsupportedModelError` in this case rather than handling it.

I removed the guards that were raising errors for transposed activations in `algorithm.py` (LoRA), `scale_estimation.py`, and `gptq.py`. For scale estimation, just removing the guard wasn't enough — the algorithm also needed to figure out which axis the channel dimension was actually on. So instead of erroring out, it now looks that up at runtime and passes it along to `process_stats`.

I renamed the test from `test_compression_skipped_with_transposed_activations` to `test_compression_works_with_transposed_activations` and rewrote to verify compression succeeds. The input shape was also fixed from `(2, 4)` to `(2, 24, 16)` since the old shape was only chosen to trigger a raise quickly.

**Skipped Test Cases:**
`lora_correction`: Fixing this requires reworking how LoRA adapter matrices are shaped, which goes beyond the core fix here.
`gptq`: The Hessian calculation assumes the hidden dim is always the last axis, which breaks with transpose_a=True. Left as a potential follow-up.

Let me know what you think...
Thanks!

naraen-ram:
Hi @ljaljushkin @daniil-lyakhov ,

Just checking in on this PR. I’d appreciate a review whenever convenient.

This contribution is part of my preparation for GSoC 2026, so getting feedback would help me move forward. Happy to make any requested changes.

Thanks!

Issue #3574, opened 2025-07-07
[NNCF] Use `CommonStatefulClassesRegistry` to load the `SingleConfigQuantizationPoint`


daniil-lyakhov:
### 🚀 Feature request

As discussed https://github.com/openvinotoolkit/nncf/pull/3487#discussion_r2189101804
one have to use a registry to load the `SingleConfigQuantizationPoint`

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

rutiktambe:
Yes, I’d be happy to work on this.

Based on the discussion in #3487, my understanding is that `SingleConfigQuantizationPoint` should be loaded via `CommonStatefulClassesRegistry` instead of being constructed directly, to align with the existing registry-based pattern.

I can take this on and start with:
- wiring `SingleConfigQuantizationPoint` loading through `CommonStatefulClassesRegistry`,
- keeping the change minimal and backward-compatible,
- and adding or updating tests if needed.

If that matches the intended direction, I’ll proceed with a PR.

abhayuvi:
.take

PR #3978, opened 2026-03-12
Register SingleConfigQuantizationPoint in CommonStatefulClassesRegistry

abhayuvi:
### Changes

`SingleConfigQuantizationPoint` is now decorated with
`@CommonStatefulClassesRegistry.register()`. `SingleConfigQuantizerSetup.from_state()`
is updated to load `SingleConfigQuantizationPoint` via the registry instead of calling
it directly. A new test file `tests/common/quantization/test_quantizer_setup.py` is
added to cover the registry lookup and get_state/from_state round-trips.

### Reason for changes

`SingleConfigQuantizationPoint` had a `from_state()` method but was missing its
`@CommonStatefulClassesRegistry.register()` decoration, unlike all other stateful
classes in the same file (`WeightQuantizationInsertionPoint`,
`ActivationQuantizationInsertionPoint`, `TargetPoint`). This made
`SingleConfigQuantizerSetup.from_state()` bypass the registry and call the class
directly, inconsistent with the established pattern.

### Related tickets

#3574

### Tests

- Added `tests/common/quantization/test_quantizer_setup.py`
- Covers: registry lookup for `SingleConfigQuantizationPoint`, get_state/from_state
  round-trip for both weight and activation insertion points, and unified scale group
  preservation across a `SingleConfigQuantizerSetup` round-trip
- All 222 existing tests in `tests/common/quantization/` continue to pass

### DIFF
--- a/src/nncf/common/quantization/quantizer_setup.py
+++ b/src/nncf/common/quantization/quantizer_setup.py
@@ -152,6 +152,7 @@ class SCQPointStateNames:
     NAMES_OF_QUANTIZED_OPS = "directly_quantized_operator_node_names"
 
 
+@CommonStatefulClassesRegistry.register()
 class SingleConfigQuantizationPoint(QuantizationPointBase):
     _state_names = SCQPointStateNames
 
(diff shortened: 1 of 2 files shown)


abhayuvi:
Hi @daniil-lyakhov , just wanted to check in on this PR. Could you let me know if the implementation looks good or if there are any changes needed? Happy to address any feedback right away.

daniil-lyakhov:
Hi @abhayuvi, thank you for your contribution! I believe that this issue should fix this part of the code https://github.com/openvinotoolkit/nncf/blob/develop/src/nncf/common/quantization/quantizer_setup.py#L198-L205
Could you please look at that?

PR #3997, opened 2026-03-24
Initialize result_scale with default scale in ScaleEstimation

RinZ27:
### Changes

Initialized `result_scale` with the default `scale` value at the start of the `calculate_quantization_params` function. I noticed that if both `initial_steps` and `scale_steps` are set to zero, the variable remains uninitialized, which would cause an error during the subsequent `fns.squeeze` call.

### Reason for changes

This ensures the algorithm is robust even when optimization steps are disabled by the user. While the default configuration uses several steps, I want to make sure the code doesn't crash if someone decides to skip the estimation process entirely.

### Related tickets

None

### Tests

Verified the fix by manually checking the logic for the edge case where steps are zero. Everything is green in the local environment and the change is a simple safety initialization that doesn't affect the default behavior.

### DIFF
--- a/src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
+++ b/src/nncf/quantization/algorithms/weight_compression/scale_estimation.py
@@ -248,7 +248,7 @@ class ScaleEstimation:
 
         X, _ = reshape_weight_for_grouped_quantization(X, -2, group_size)
         best_diffs = None
-        result_scale = None
+        result_scale = scale
         fp_outs = fns.matmul(fns.moveaxis(weight, -2, -3), X)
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev:
Hi @RinZ27 
Thanks a lot for your PR! 
However, this change requires a bit more work in the codebase and adding some new tests.
So I went ahead and created [a new PR](https://github.com/openvinotoolkit/nncf/pull/4039)  based on this one.

RinZ27:
@AlexanderDokuchaev Appreciate you taking this forward in #4039 to ensure broader application and proper test coverage. Handling the fix in a more comprehensive PR seems like the right move. I'll be following the progress there. Thanks for the update!

PR #4039, opened 2026-04-14
Initialize result_scale with default scale in ScaleEstimation

AlexanderDokuchaev:
### Changes

Support initial_steps and scale_steps to be zero for `calculate_quantization_params`
Add tests case for differ steps

### Reason for changes

https://github.com/openvinotoolkit/nncf/pull/3997

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/24425425305
https://github.com/openvinotoolkit/nncf/actions/runs/24425387669


andreyanufr (approved):
Good changes.

PR #3793, opened 2025-12-16
Rework HWConfig

AlexanderDokuchaev:
### Changes

Replace JSON-based configuration loading with a Python-style HW Config.

### Reason for changes

Simplifies modification and maintenance by keeping configuration logic in Python code instead of external JSON files.

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/20596514331
manual/job/post_training_quantization/767/
https://github.com/openvinotoolkit/nncf/actions/runs/20596794339

PR #4047, opened 2026-04-20
Use HWConfigOpName instead of str in HW setups 

AlexanderDokuchaev:
### Changes

- Use `HWConfigOpName` enum instead of string literals
- Add missed HWConfigOpName: NORMALIZEL2, VARIADICSPLIT, SHUFFLECHANNELS, BROADCAST, CONVERTLIKE 
- Sort HWConfigOpName values 
- Use auto()
- Rename class and add _ to values

### Reason for changes

This ensures consistent operation naming and reduces errors due to typos. 

### Related tickets

https://github.com/openvinotoolkit/nncf/pull/3793

### Tests

https://github.com/openvinotoolkit/nncf/actions/runs/24678941855


daniil-lyakhov (approved):

src/nncf/common/hardware/opset.py
```diff
@@ -13,52 +13,65 @@
...
     REDUCESUM = "ReduceSum"
-    REDUCEL2 = "ReduceL2"
-    INTERPOLATE = "Interpolate"
-    MVN = "MVN"
     RESHAPE = "Reshape"
-    CONCAT = "Concat"
-    FLATTEN = "Flatten"
-    SQUEEZE = "Squeeze"
-    UNSQUEEZE = "Unsqueeze"
+    SCALED_DOT_PRODUCT_ATTENTION = "ScaledDotProductAttention"
```
andrey-churkin: What do you think about adding `_` in other cases as well? One possible rule could be: if we have several capital letters, we split the words. For example:

- `AVG_POOL`
- `CONVERT_LIKE`
- `DEPTH_WISE_CONVOLUTION`

I’m not insisting - just thinking it might improve readability.

AlexanderDokuchaev: why not
Done, and add using auto()

AlexanderDokuchaev: And renamed to HWOpName

andrey-churkin (approved):

Issue #4033, opened 2026-04-12
IgnoredScope silently fails to protect weights when targeting Constant node names


t8:
## Describe the bug

`nncf.compress_weights()` with `IgnoredScope(names=[...])` targeting a Constant node name (the weight tensor itself) does not prevent the weight from being compressed. NNCF reports "1 ignored nodes were found" -- it finds the node -- but the weight is still compressed through the consuming operation. No warning or error is emitted.

**What happens:**
1. User specifies the weight's node name (e.g., `self.sensitive_embed.weight`)
2. NNCF confirms it found the node ("1 ignored nodes were found")
3. The weight is compressed anyway

**Workaround:** specify the consuming operation name (e.g., `__module.sensitive_embed/aten::linear/MatMul`) instead of the Constant name. A regex pattern like `.*sensitive_embed.*` also works because it matches both the Constant and the consuming MatMul.

**Root cause:** `get_nodes_to_compress()` only iterates over operation metatypes (MatMul, Embedding, etc.). The IgnoredScope name matching runs against the NNCFGraph which includes Constants, so the Constant IS found and counted. But the compression loop only checks if each operation node is in the ignored set. The Constant is in the set but is never checked because Constants are not iterated.

**Results:**

| IgnoredScope target | Type | "Ignored nodes found" | Weight dtype after compression |
|---|---|---|---|
| `self.sensitive_embed.weight` | Constant node | 1 found | **INT4 (not protected)** |
| `__module.sensitive_embed/aten::linear/MatMul` | MatMul op | 1 found | float32 (protected) |
| `.*sensitive_embed.*` (regex) | Pattern | 2 found | float32 (protected) |

**Suggested improvements:**

1. If a Constant node name appears in IgnoredScope, also exclude the consuming operation from compression. This makes `IgnoredScope(names=["weight_name"])` work as users expect.
2. Alternatively, emit a warning when an IgnoredScope name matches only Constant nodes (not operations), since it will have no effect on compression.

Happy to submit a PR for either approach if the team agrees on a direction.

## Environment

- NNCF: 3.0.0
- OpenVINO: 2026.0.0
- PyTorch: 2.11.0+cpu
- OS: Windows 11
- Python: 3.11.9
- CPU: Intel Core Ultra X7 358H
- RAM: 32 GB

## Minimal Reproducible Example

```python
import torch, torch.nn as nn, openvino as ov, nncf

class Model(nn.Module):
    def __init__(self):
        super().__init__()
[...]

PR #4049, opened 2026-04-21
[WC] Propagate IgnoredScope Constant matches to consuming weighted ops

t8:
### Changes

- Add `propagate_ignored_constants_to_weighted_consumers` helper in `nncf.common.scopes`. For each ignored node whose metatype is in the caller-supplied `constant_metatypes`, the helper performs a forward BFS through `passthrough_metatypes` / `passthrough_node_types` and adds any reached weighted op to the ignored set. Traversal halts at weighted ops and at non-passthrough non-weighted nodes, so ignoring an arbitrary intermediate op (e.g. a ReLU or an indices-side Gather) does not drag its downstream weighted consumers into the ignored set.
- Add `constant_metatypes`, `passthrough_metatypes`, and `passthrough_node_types` properties to `WeightCompressionAlgoBackend` (empty by default) and implement them for the OV, Torch, FX, and ONNX backends. The lists mirror the reverse-walk conventions already used by `get_operation_const_op` and `find_const_node_in_constant_subgraph`.
- Call the helper from `WeightCompression.get_ignored_node_names`, `WeightCompression._get_ignored_scope_weight_statistics`, and the Torch `prune` entry point so the resolved ignored set includes consuming weighted ops.
- Emit an info log listing the added node names whenever the helper expands the ignored set; the log only fires when expansion was nontrivial.

### Reason for changes

`IgnoredScope(names=[...])` silently failed to protect weights when the name targeted a Constant node. NNCF reported "N ignored nodes were found" but the weight was still compressed or pruned, because `nncf.compress_weights` and `nncf.prune` iterate only over weighted operation nodes and the Constant in the ignored set was never compared against the compression loop. Users had to look up the consuming op's name and pass that instead, which is non-obvious and easy to get wrong.

Verified the bug still reproduces on the latest pypi release, NNCF 3.1.0, using the reproducer in #4033. See that issue for the original report and maintainer discussion.

### Related tickets

#4033

### Tests

New unit tests for the helper in `tests/common/test_scopes.py` cover:

- Direct `Constant -> MatMul` propagation
- Single-hop passthrough (`Constant -> Reshape -> MatMul`)
- Chained passthrough (`Constant -> Identity -> Identity -> MatMul`)
- Shared-weight Constants feeding multiple MatMuls
- Intermediate compute ops (ReLU) that must NOT propagate
- Parameter nodes (source-only but not Constants)
- `passthrough_node_types` string-based traversal
- Halt at non-passthrough non-weighted nodes
- Empty `constant_metatypes` as kill-switch

Integration tests:

- `tests/openvino/native/quantization/test_weights_compression.py`: new `IgnoredScope(names=...)` parametrizations on `IntegerModel` for `matmul_1_data`, `matmul_2_data`, `gather_2_data` Constant names, plus a new synthetic `Constant -> Reshape -> MatMul` model exercising the passthrough walk end-to-end.
- `tests/onnx/quantization/test_weights_compression.py`: synthetic model with an explicit `Constant` op feeding MatMul (the pattern where ONNX creates a graph node, as opposed to initializer tensors).
- `tests/torch/function_hook/pruning/magnitude/test_algo.py`: new `test_prune_with_ignored_scope` covering Conv weight Constants on `TwoConvModel`.

Pre-submission checks run locally:

- `ruff check` (v0.14.14 per `.pre-commit-config.yaml`): clean
- `ruff format --check`: clean
- `mypy --strict` on the two strict-checked files I modified (`src/nncf/common/scopes.py`, `src/nncf/torch/function_hook/pruning/prune_model.py`): clean
- `tests/common`: 900 passed
- `tests/openvino/native/quantization/test_weights_compression.py`: 666 passed, 1 xfailed, 3 pre-existing failures in `test_lora_adapters_reduce_noise` verified against clean `develop`
- `tests/onnx/quantization/test_weights_compression.py`: 123 passed, 3 skipped
- `tests/torch/function_hook/pruning/magnitude/test_algo.py`: 13 passed (2 CUDA-only skipped on macOS)

Known limitation: ONNX weight initializers are not represented as NNCFGraph nodes (only explicit `Constant` ops are), so the filed bug pattern does not surface in typical ONNX workflows today. The ONNX backend properties are wired up for API consistency and to cover the `Constant` op path.


AlexanderDokuchaev:
Thanks for the PR! 
This solution works well. We've decided to go with a slightly more minimal approach and have opened two follow-up PRs based on your findings. 
https://github.com/openvinotoolkit/nncf/pull/4063
https://github.com/openvinotoolkit/nncf/pull/4064
Closing this in favor of those, thanks again for the solid groundwork!

PR #4054, opened 2026-04-28
[PT] Check allowed modules on `load_from_config`

AlexanderDokuchaev:
### Changes

- Added `allowed_modules` argument to `load_from_config`
- Removed duplcation defention in model_creation.py
- Added `api` marker
- Updated `src/nncf/torch/__init__py`, use as for imported objects

### Reason for changes

Possible loading any modules after https://github.com/openvinotoolkit/nncf/pull/4050

### Related tickets

CVS-185593

### Tests

[Test examples](https://github.com/openvinotoolkit/nncf/actions/runs/25043257889) - success

### DIFF
--- a/src/nncf/torch/function_hook/serialization.py
+++ b/src/nncf/torch/function_hook/serialization.py
@@ -9,8 +9,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from fnmatch import fnmatchcase
 from importlib import import_module
-from typing import Any, TypedDict, TypeVar, cast
+from typing import Any, Sequence, TypedDict, TypeVar, cast
 
 from torch import nn
 
@@ -23,6 +24,7 @@ from nncf.torch.utils import get_model_device
 from nncf.torch.utils import is_multidevice
 
 COMPRESSION_STATE_ATTR = "compression_state"
+DEFAULT_ALLOWED_MODULES = ("nncf.*",)
 TModel = TypeVar("TModel", bound=nn.Module)
 
 
@@ -88,7 +90,9 @@ MODULE_NAME_MAP = {
 }
 
 
-def load_from_config(model: TModel, config: dict[str, Any]) -> TModel:
+def load_from_config(
+    model: TModel, config: dict[str, Any], allowed_modules: Sequence[str] = DEFAULT_ALLOWED_MODULES
+) -> TModel:
     """
     Initialize model with compressed modules from config file.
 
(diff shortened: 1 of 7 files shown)


andrey-churkin (approved):

src/nncf/torch/function_hook/serialization.py
```diff
@@ -88,7 +101,19 @@ def get_config(model: nn.Module) -> dict[str, Any]:
...
+    NNCF_PT_CATEGORY,
+    [
+        FunctionCallTelemetryExtractor("nncf.torch.load_from_config"),
+    ],
+)
+def load_from_config(
+    model: TModel,
+    config: dict[str, Any],
+    *,
+    allowed_modules: tuple[str] = DEFAULT_ALLOWED_MODULES,
```
daniil-lyakhov, line 115: In what scenario someone would change this parameter? Should we just use DEFAULT_ALLOWED_MODULES all the time?

src/nncf/torch/function_hook/serialization.py
```diff
@@ -136,23 +163,48 @@ def load_from_config(model: TModel, config: dict[str, Any]) -> TModel:
...
-def restore_module(module_path: str, cls_name: str, module_config: dict[str, Any]) -> nn.Module:
+def _is_allowed_module(module_path: str, allowed_modules: tuple[str, ...]) -> bool:
+    """
+    Check if the module path matches any of the allowed patterns.
+
+    :param module_path: The import path of the module.
+    :param allowed_modules: Allowed patterns for module import paths.
+    :return: True if the module path is allowed, False otherwise.
+    """
+    return any(fnmatchcase(module_path, pattern) for pattern in allowed_modules)
```
daniil-lyakhov: As this function is only called once, could we just inline it?

src/nncf/torch/function_hook/serialization.py
AlexanderDokuchaev: Restricting to nncf.* blocks user-defined modules. 
Yes it’s rare cases, but easy to support.

src/nncf/torch/function_hook/serialization.py
AlexanderDokuchaev: yes, done

src/nncf/torch/function_hook/serialization.py
daniil-lyakhov: It won't hurt, agree

daniil-lyakhov (approved):

PR #4063, opened 2026-05-05
[pruning] Ignored scope by name of constant node

AlexanderDokuchaev:
### Changes

Add a check for ignored names to each pruned weight, rather than applying the check only at the operation node level.
Moved some logic to `get_prunable_parameters` function to easy validate ignored scope.

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/4033

### Tests

[Test examples](https://github.com/openvinotoolkit/nncf/actions/runs/25376728927) - success

### DIFF
--- a/src/nncf/torch/function_hook/pruning/magnitude/algo.py
+++ b/src/nncf/torch/function_hook/pruning/magnitude/algo.py
@@ -27,7 +27,7 @@ TModel = TypeVar("TModel", bound=nn.Module)
 
 def apply_magnitude_pruning(
     model: TModel,
-    parameters: list[str],
+    parameters: set[str],
     mode: PruneMode,
     ratio: float,
 ) -> TModel:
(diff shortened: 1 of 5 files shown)


daniil-lyakhov (approved):

anzr299:
As I understand, the ignored nodes returned from get_ignored_node_names_from_ignored_scope(...) returns weight node names. 
But this weight names cannot be directly applied to ignore in pruning.
With your changes the pruning produces set of "nodes to prune" which are not operation names but weight names as returned by ignored scope.
Is this a correct understanding?

AlexanderDokuchaev:
@anzr299  `get_ignored_node_names_from_ignored_scope` returns any node that matcched by ignored scope, not only weight nodes.
only this  `if const_node is not None and const_node.node_name not in ignored_names:` the change behevior.

anzr299 (approved):

PR #4064, opened 2026-05-06
[wc] Ignored scope by name of constant node

AlexanderDokuchaev:
### Changes

Add a check for ignored names to each compressed weight, rather than applying the check only at the operation node level.

### Related tickets

https://github.com/openvinotoolkit/nncf/issues/4033

### Tests

[Weight compression](https://github.com/openvinotoolkit/nncf/actions/runs/25424539107) - success

[Test examples](https://github.com/openvinotoolkit/nncf/actions/runs/25424551853) - success

### DIFF
--- a/tests/onnx/requirements.txt
+++ b/tests/onnx/requirements.txt
@@ -1,6 +1,7 @@
 -c ../../constraints.txt
 onnx
 onnxruntime
+onnxscript
 openvino
 pandas
 pytest-cov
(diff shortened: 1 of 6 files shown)


daniil-lyakhov:
Pls rebase

tests/onnx/requirements.txt
```diff
@@ -1,6 +1,7 @@
 -c ../../constraints.txt
 onnx
 onnxruntime
+onnxscript
```
daniil-lyakhov, line 4: Why do we need onnxscript exactly?

AlexanderDokuchaev: To export torch model to onnx

daniil-lyakhov (approved):

andreyanufr (approved):

PR #4072, opened 2026-05-15
Add FP8 support for the ONNX backend

andrey-churkin:
### Changes

- Add support for `nncf.CompressWeightsMode.FP8_E4M3` mode in the `nncf.compress_weights()` method for the ONNX backend.
- Add support for quantization using `nncf.QuantizationMode.FP8_E4M3` and `nncf.QuantizationMode.FP8_E5M2` modes in the `nncf.quantize()` method for the ONNX backend. 
- Add an example `‎examples/llm_compression/onnx/smollm2_360m_fp8`
### Reason for changes

Add support for FP8 quantization and weight compression in the ONNX backend.

### Related tickets

- CVS-180789

### Tests

[Weight compression](https://github.com/openvinotoolkit/nncf/actions/runs/25908094808) - success

[Test examples](https://github.com/openvinotoolkit/nncf/actions/runs/28224288278) - success

### DIFF
--- a/src/nncf/quantization/algorithms/min_max/onnx_backend.py
+++ b/src/nncf/quantization/algorithms/min_max/onnx_backend.py
@@ -12,7 +12,6 @@
 
 import numpy as np
 
-import nncf
 from nncf.common.graph.graph import NNCFGraph
 from nncf.common.graph.graph import NNCFNode
 from nncf.common.graph.operator_metatypes import OperatorMetatype
@@ -33,6 +32,7 @@ from nncf.onnx.graph.transformations.commands import ONNXQuantizerInsertionComma
 from nncf.onnx.graph.transformations.commands import ONNXTargetPoint
 from nncf.onnx.hardware.config import ONNXHWConfig
 from nncf.onnx.quantization.default_quantization import DEFAULT_ONNX_QUANT_TRAIT_TO_OP_DICT
+from nncf.onnx.quantization.quantizer_parameters import convert_fc_params_to_onnx_params
 from nncf.onnx.quantization.quantizer_parameters import convert_fq_params_to_onnx_params
 from nncf.parameters import ModelType
 from nncf.parameters import TargetDevice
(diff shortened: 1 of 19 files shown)


src/nncf/quantization/algorithms/min_max/onnx_backend.py
```diff
@@ -158,11 +158,20 @@ def create_unified_scales_quantizers_insertion_commands(
...
-        msg = "FakeConvert insertion not implemented in ONNX backend!"
-        raise nncf.InternalError(msg)
+        axis = None
+        if quantizer_config.per_channel:
+            node = nncf_graph.get_node_by_name(target_point.target_node_name)
+            axis = (
+                get_weight_quantization_axis(node, target_point.port_id) if target_point.is_weight_target_point() else 1
+            )
+        onnx_parameters = convert_fc_params_to_onnx_params(parameters, axis)
+        nncf_input_node_next_nodes = ONNXMinMaxAlgoBackend._get_input_edges_mapping(nncf_graph)
```
daniil-lyakhov, line 175: Potential point for an optimization in future. Maybe we could add a comment to highligt it

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
```diff
@@ -363,8 +364,14 @@ def _add_dequantize_linear_layer(
                 zero_point = pack_4_bits(zero_point)
 
         # Create initializers for the quantized weights, scale, and zero point
+        if weight_dtype == onnx.TensorProto.FLOAT8E4M3FN:
+            np_dtype = helper.tensor_dtype_to_np_dtype(weight_dtype)
+            vals = onnx.numpy_helper.saturate_cast(np.asarray(quantized_weights), np_dtype).flatten()
+        else:
+            vals = quantized_weights
```
daniil-lyakhov: Two similar code blocks, maybe worth a private method?

daniil-lyakhov (commented):
No major comments, please add some tests

src/nncf/quantization/algorithms/min_max/onnx_backend.py
andrey-churkin: I added a comment. However, as you can see, the same approach is used for almost all commands in the ONNX backend, so I guess there is a reason why it was implemented this way. In any case, it's something we should look into, but I think it should be addressed in a follow-up PR.

src/nncf/quantization/algorithms/weight_compression/onnx_backend.py
andrey-churkin: I've rewritten it slightly. Given that it's only two lines, I don't think introducing a separate method provides much value.

daniil-lyakhov (approved):
LGTM

andreyanufr (approved):

andreyanufr (approved):

PR #4075, opened 2026-05-15
[mypy] Add generic types for registry

AlexanderDokuchaev:
### Changes

Add generic type for Registry class
Removed `is_subtype` argument from `OperatorMetatypeRegistry.registry` method, cause it was need only for legacy operation types.
Removed `add_name_as_attr` as not used.
Changes names of registry objects to be more informable.

### Reason for changes

Code quality

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ files = [
     "src/nncf/api",
     "src/nncf/data",
     "src/nncf/common",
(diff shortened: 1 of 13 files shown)


daniil-lyakhov (commented):
Looks good, maybe we could allign the `get_subtypes` method across backends? Torch does list copy, and other backends don't. Otherwise ok

anzr299 (approved):

PR #4143, opened 2026-07-18
Stop getitem quantizer propagation on numel change

tej-1916:
## Summary

Fixes TorchFX quantizer propagation through `operator.getitem` when the selected tensor has a different number of elements than the source tensor.

## Root cause

TorchFX graph conversion always classified `operator.getitem` as `PTGatherMetatype`, and gather is quantization-agnostic. As a result, activation quantizers could propagate backward through split/getitem boundaries and land on the producer tensor even when downstream weighted ops consumed smaller selected tensors.

## Changes

- Add a TorchFX metadata check that proves whether `operator.getitem` preserves tensor `numel`.
- Mark unsafe or unproven getitems with an internal NNCF node propagation-barrier attribute while preserving their public `PTGatherMetatype`.
- Route marked getitems through existing autogenerated ignored-node handling and exclude them as activation post-hook insertion points.
- Add focused TorchFX regression tests for different-numel and same-numel getitem behavior.

## Validation

- `python -m pytest tests/torch/fx/test_quantizer_config.py -k "getitem_different_numel or getitem_same_numel" -vv`
- `python -m pytest tests/torch/fx/test_quantizer_config.py -q`
- `python -m pytest tests/torch/fx/test_models.py -k "yolo26 or yolo11n or split or getitem" -q`
- `python -m ruff format --check src/nncf/common/graph/graph.py src/nncf/experimental/torch/fx/nncf_graph_builder.py src/nncf/quantization/algorithms/min_max/algorithm.py tests/torch/fx/test_quantizer_config.py`
- `python -m ruff check src/nncf/common/graph/graph.py src/nncf/experimental/torch/fx/nncf_graph_builder.py src/nncf/quantization/algorithms/min_max/algorithm.py tests/torch/fx/test_quantizer_config.py`
- `git diff --check`


tej-1916:
The implementation and focused regression tests are ready for review. CI workflows currently show action required; could a maintainer please approve the workflow runs when convenient? Thank you.

anzr299:
Thank you for the PR!
Fixed in #4155

PR #4155, opened 2026-07-28
[TorchFX] Remove __getitem__ after split node from Inference Graph

anzr299:
### Changes

Explicitly remove getitem node which comes after a split metatype like chunk from the inference graph so that the solver does not see it.

### Reason for changes

Some TorchFX operations are returning a tuple instead of a torch.Tesnor. In case an algorithm requests a statistic from such node nncf will raise an error like tuple has not is_empty method. This affects the yolon11 model, current workaround is the ignored scope with the type of operation which returns a tuple

### Related tickets

CVS-191629


tests/torch/test_models/synthetic.py
```diff
@@ -671,6 +671,28 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return torch.cat(list(chunks), dim=1)
 
 
+class SplitBlockModel(nn.Module):
+    """
+    Like the YOLO C2f split block, with a weighted op on a single branch only.
+    conv -> chunk -> conv -> cat
+        \                     /
+         \                   /
+           ->   chunk     ->
```
AlexanderDokuchaev:
```suggestion
    conv -> chunk -> conv -> cat
               \_____________/
```

AlexanderDokuchaev (approved):

Issue #3576, opened 2025-07-07
[quantize_pt2e] Support batched statistics


daniil-lyakhov:
### 🚀 Feature request

External quantizers could mark a multiply node as quantizable (see https://github.com/openvinotoolkit/nncf/pull/3558). NNCF is hardcoded the weight nodes via the `weight_port_id` parameter in metatypies, so  the logic of the quantization is leaked to the metatypes.

The task is to figure out a way to correctly identify weight of a node without hardcoded parameter. For example, NNCF should identify which kind of multiply is present: activations x activations, weights x activations or activations x weights
The batched statistics should be enabled and tested for a custom quantizers (like x86)

### Feature Use Case

Quantization with quantize_pt2e + custom quantizers with batch size > 1

### Are you going to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Aryanpagaria:
Can I work on this @daniil-lyakhov

daniil-lyakhov:
Hello, @Aryanpagaria! Thank you for your interest. I believe this task is too complicated to be a GFI, it requires an engineering input. But because of a low priority, we don't have a general plan yet

CodersAcademy006:
Hey @daniil-lyakhov i would also love to contribute to this isse and work on it. Thanks

PR #4173, opened 2026-08-15
Fix PT2E Mul weight detection for parameters

ricardoasouz:
## Summary

Fixes PT2E/TorchAO weight detection for `aten.mul.Tensor` when a learned parameter can appear on either operand.

Previously, `TorchAOQuantizerAdapter` relied on `metatype.weight_port_ids` to determine whether a `get_attr` node represented a weight. `PTMulMetatype` has no fixed weight port because both

- `activation * weight`
- `weight * activation`

are valid.

As a result, a `torch.nn.Parameter` feeding `Mul` could be classified as an activation.

This change adds a Mul-specific fallback: when the source `get_attr` corresponds to a named `torch.nn.Parameter`, it is treated as a weight. Other operations continue to use the existing `weight_port_ids` behavior.

## Tests

Added coverage for:

- `activation * Parameter` -> `WeightQuantizationInsertionPoint`
- `Parameter * activation` -> `WeightQuantizationInsertionPoint`
- `activation * buffer` -> remains `ActivationQuantizationInsertionPoint`
- custom TorchAO quantizer with batch size 2
- automatic `batchwise_statistics=True`
- explicit verification that quantization is not silently skipped
- explicit Q/DQ presence in the resulting FX graph

Validation:

- focused issue coverage: 4 passed
- ViT regression checks: 2 passed
- `tests/executorch/test_ptq.py`: 57 passed
- `make pre-commit`: passed
- `git diff --check`: passed

## Notes

The behavior change is intentionally limited to `PTMulMetatype` so parameters used by other operations continue to follow their existing metatype-defined weight-port semantics.

Related to #3576.

## AI-assisted development

AI tools were used during codebase exploration, hypothesis generation, and drafting. I personally reviewed, tested, and validated the final implementation and can explain and maintain the submitted changes.


ricardoasouz:
Hi @anzr299, just rebased this onto the latest develop and ran tests/executorch/test_ptq.py again. All 57 tests pass. Let me know if there’s anything else you’d like me to check. Thanks!

ricardoasouz:
Rebased onto the latest develop and revalidated using the repository-pinned ExecuTorch environment with Torch 2.13.0.
tests/executorch/test_ptq.py: 57 passed.
This includes the new Mul parameter detection cases for both operand orders, the buffer case, and the batchwise statistics regression test.

ricardoasouz:
Updated this PR after re-checking it end-to-end. In the previous revision, the Mul parameter was classified as a weight, but `WeightQuantizationInsertionPoint` carried no port, and `PTMulMetatype` has no static weight ports, so MinMax created no target point, and the parameter ended up unquantized (on `develop`, it is still quantized as an activation). The weight insertion point now carries the input port found by the adapter, and MinMax uses it. End-to-end tests cover both operand orders and the buffer case; `tests/executorch/test_ptq.py`: 60 passed. The broader variant for other ops is described in the notes; happy to follow up if you prefer that direction.

PR #4176, opened 2026-08-21
[OpenVINO] Add Support for GroupedMM in AWQ and Scale Estimation for OpenVINO Backend

anzr299:
### Changes

GroupedMatMul applies one weight matrix per group to a single shared 2D activation. Data-aware compression assumed a 3D weight always comes with a matching 3D activation, so AWQ, scale estimation and GPTQ failed.

#### AWQ

For weights, The experts axis is folded into the output channels, the search then minimizes the error over all experts at once, and the scale is broadcast back over the experts axis.

For activation, the same scale is applied over the hidden dim of the activation in non mergible case, For mergible scale, it is reshaped but still only applied over the hidden dim of the previous matmul weights.

#### Scale estimation

Broadcasts the pooled statistics over the group axis, so the scale stays per group.

### Reason for changes

to support grouped_mm based MoE models in OpenVINO backend.

### Related tickets

191895

### Tests

test_scale_estimation and test_awq_scale_reference extended for this case.
test_node_utils extended for the weight channel axes and activation channel axis for GroupedMatMul.

WWB Eval

Model: Qwen3.6 35B A3B
mode: INT4_ASYM; GS: 64; 

AWQ=True
MoE op | WWB Accuracy
-- | --
BatchedMM | 0.956345
GroupedMM | 0.952167

Scale Estimation=True
MoE op | WWB Accuracy
-- | --
BatchedMM | 0.9534
GroupedMM | 0.9587

[Test examples](https://github.com/openvinotoolkit/nncf/actions/runs/32956643990) - success

[Weight compression](https://github.com/openvinotoolkit/nncf/actions/runs/32975768854) - success


tests/openvino/native/quantization/test_weights_compression.py
```diff
@@ -2519,7 +2524,84 @@ def get_scale_estimation_ref(check_sampling_activation_stats_flow):
         )[check_sampling_activation_stats_flow]
 
     @staticmethod
-    def get_moe_scale_estimation_ref(check_sampling_activation_stats_flow):
+    def get_moe_scale_estimation_ref(check_sampling_activation_stats_flow, grouped_mm=False):
```
andreyanufr: Is it possible to move reference to json file ? We have examples in tests.

anzr299: That is a good idea. I will move it.

anzr299: I have combined all the refs for SE in 1 function (earlier MoE refs had another getter). Then moved it to a reference JSON file.

andreyanufr (approved):

andrey-churkin (approved):

PR #4184, opened 2026-09-05
fix(quantization): compare quantizer setup groups in both directions

Anai-Guo:
### What

`QuantizerSetupBase.equivalent_to` calls two of its three helpers twice in the **same** direction:

```python
return (
    _compare_qps(self, other)
    and _compare_qps(other, self)
    and _compare_shared_input_groups(self, other)
    and _compare_shared_input_groups(self, other)   # <- (other, self)
    and _compare_unified_scale_groups(self, other)
    and _compare_unified_scale_groups(self, other)  # <- (other, self)
)
```

`_compare_qps` is already checked both ways; the two group comparisons are not. Each of them only asserts that *every group of `first` has a match in `second`*, so running it twice in the same direction leaves the subset case unchecked: a setup whose grouping is a strict subset of the other's compares equal.

### Why the id map has to be split too

Simply flipping the arguments is not enough. All three helpers read the shared `this_qp_id_to_other_qp_id_dict`, which `_compare_qps(other, self)` overwrites with the *other → this* mapping. A `(other, self)` group comparison needs exactly that map, while the `(self, other)` one needs the *this → other* map. So `_compare_qps` now writes into a caller-supplied dict and each direction is given its own; the translation logic itself is unchanged.

### Reproduction

Two setups with identical quantization points, where `other` has one extra shared-input group:

```python
qps = {0: qp_a, 1: qp_b, 2: qp_c}
a = setup(qps, shared_input_operation_set_groups={0: {0, 1}})
b = setup(qps, shared_input_operation_set_groups={0: {0, 1}, 1: {2}})
a.equivalent_to(b)   # True   <- wrong, b groups qp_c and a does not
b.equivalent_to(a)   # False  <- the direction that is already checked
```

Running `equivalent_to` extracted from `develop` against duck-typed setups, before and after the patch:

```
extra shared-input group in `other`     current=True    patched=False   expected=False  OK
extra unified-scale group in `other`    current=True    patched=False   expected=False  OK
[control] equivalent setups             current=True    patched=True    expected=True   OK
[control] extra group in `self`         current=False   patched=False   expected=False  OK
[control] differing QPs                 current=False   patched=False   expected=False  OK
[control] no groups                     current=True    patched=True    expected=True   OK
```

Every control keeps its previous result; only the previously-unchecked direction changes.

`ruff check` and `ruff format --check` pass with the pinned `v0.15.18` and the repo's `pyproject.toml`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/nncf/common/quantization/quantizer_setup.py
+++ b/src/nncf/common/quantization/quantizer_setup.py
@@ -356,8 +356,13 @@ class QuantizerSetupBase:
 
     def equivalent_to(self, other: "QuantizerSetupBase") -> bool:
         this_qp_id_to_other_qp_id_dict: dict[QuantizationPointId, QuantizationPointId] = {}
+        other_qp_id_to_this_qp_id_dict: dict[QuantizationPointId, QuantizationPointId] = {}
 
-        def _compare_qps(first: "QuantizerSetupBase", second: "QuantizerSetupBase") -> bool:
+        def _compare_qps(
+            first: "QuantizerSetupBase",
+            second: "QuantizerSetupBase",
+            id_map: dict[QuantizationPointId, QuantizationPointId],
+        ) -> bool:
             for this_qp_id, this_qp in first.quantization_points.items():
                 matches: list[QuantizationPointId] = []
                 for other_qp_id, other_qp in second.quantization_points.items():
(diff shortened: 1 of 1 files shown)


AlexanderDokuchaev:
@Anai-Guo 
Thanks for catching this!
 Since this method is only used in tests, the fix actually belongs in the test itself. I'll make that change in a follow-up PR.

PR #4223, opened 2026-09-17
[PT] Add structured 2:4 magnitude pruning

### LINKED ISSUE #4222: [Feature request] Structured M:N (2:4) magnitude-based pruning
### Is your feature request related to a problem? Please describe.

NNCF currently supports pruning via `nncf.prune` with the modes `UNSTRUCTURED_MAGNITUDE_LOCAL`, `UNSTRUCTURED_MAGNITUDE_GLOBAL` and `UNSTRUCTURED_REGULARIZATION_BASED` (see [Pruning usage docs](https://github.com/openvinotoolkit/nncf/blob/develop/docs/usage/training_time_compression/pruning/Usage.md)). All of these produce unstructured sparsity: zeros are scattered arbitrarily through the weight tensors.

There is currently no support for structured M:N sparsity (for example 2:4, two zeros in every consecutive group of four weights). As discussed in [#3810](https://github.com/openvinotoolkit/nncf/discussions/3810), where this direction was raised as an opportunity for contribution, M:N patterns are widely supported by hardware and runtimes as a sparsity format, and they can be produced with a magnitude criterion using the same kind of workflow NNCF already has.

### Describe the solution you'd like

Add structured M:N magnitude-based pruning to the existing `nncf.prune` flow, starting with 2:4:

- For every applicable group of N consecutive weights along a chosen structural dimension, zero out the M weights with the smallest magnitudes and keep the remaining N-M weights. The invariant is enforced independently for every group, rather than only globally.
- For 2:4 this means each group of four weights contains exactly two zeros and two retained values, selected by magnitude.

Design points I would like feedback on before/while implementing:

1. **API representation.** Following the existing `PruneMode` enum convention, something like a new mode `STRUCTURED_MAGNITUDE_2_4` (similar to how `CompressWeightsMode` enumerates format variants such as `MXFP4`, `NVFP4`, ...), with an optional extensible `sparsity_pattern=(m, n)` parameter kept internal for now. Alternatively a fully generic `(m, n)` public parameter from the start.
2. **Supported layer types.** Initially `nn.Linear` (incl. `MatMul`/`Addmm` weights) and `nn.Conv2d`/`Conv1d`/`Conv3d` weights. Depthwise convolutions and embeddings would be excluded, and unsupported metatypes would be skipped with a warning so the rest of the model is still pruned.
3. **Grouping dimension.** For linear/matmul weights, group along the output-channel (row) dimension, with in-features contiguous. For convolutions, reuse the existing channel-axes helpers so groups follow the output-channel × spatial position layout. The grouping should be built from the same channel/reduction axis information the weight-compression code already uses (`get_weight_compression_reduction_axes`).
4. **Non-divisible dimensions.** The cleanest policy seems to be raising `nncf.ValidationError` with a clear message when the structural dimension is not divisible by N, consistent with the fail-fast validation style used elsewhere (e.g. weight compression). Happy to follow a different precedent if one exists.
5. **Interaction with existing features.** `ignored_scope` filtering should keep working unchanged; the new mode should produce statistics via `nncf.pruning_statistic`; `nncf.strip(..., strip_format=StripFormat.IN_PLACE)` should bake the masks into the weights; checkpoint save/load via `get_config`/`load_from_config` should work through the existing `StatefulModuleInterface` mechanism.
6. **Testing.** Group-level invariant tests (not just overall zero ratio), magnitude selection with positive/negative and tied magnitudes, non-divisible dimensions, ignored scopes, save/load, strip, statistics, and integration tests on Linear/Conv models.

### Describe alternatives you've considered

- A standalone external script/tool outside of NNCF: does not help NNCF users and cannot leverage the existing graph analysis, ignored scopes and statistics.
- Supporting only a global sparsity ratio and calling it 2:4: this would not guarantee the pattern per group and would not be consumable by hardware expecting the format.

### Additional context

This feature follows up on the structured sparsity direction raised in [#3810](https://github.com/openvinotoolkit/nncf/discussions/3810). I searched the current issues and PRs and did not find an equivalent feature request or work in progress for M:N / 2:4 sparsity; if one exists, please point me to it and I will close this.

The goal is to reuse the existing pruning infrastructure (function-hook masks, `get_prunable_parameters`, pruning statistics, strip) rather than introducing a parallel pruning framework. No hardware acceleration or new kernels are proposed — only producing the weight pattern itself, which is a prerequisite for acceleration on runtimes/hardware that support the format. I'm happy to open a PR implementing this once the general direction is confirmed.

uvv-01:
### What does this PR do?

This PR adds structured magnitude-based pruning with a fixed 2:4 sparsity pattern to the PyTorch backend, exposed as a new `nncf.PruneMode.STRUCTURED_MAGNITUDE_2_4` mode of the existing `nncf.prune` API. In every consecutive group of four weights along the structural dimension of the weight tensor, the two weights with the smallest absolute values are zeroed and the other two are kept, so each group contains exactly two zeros and two non-zero values.

### Why is this needed?

NNCF currently offers unstructured magnitude pruning (local/global) and regularization-based pruning, but no support for structured M:N sparsity patterns. Structured patterns such as 2:4 (semi-structured sparsity) can be consumed by hardware and runtime implementations that support them, which makes them a natural extension of the pruning stack. This follows up on the structured sparsity direction raised in discussion #3810 and is tracked by issue #4222.

### What is implemented?

- `PruneMode.STRUCTURED_MAGNITUDE_2_4` in the public `PruneMode` enum.
- A `StructuredPruningMask` module that applies a fixed binary mask on the weight, mirroring the existing unstructured mask module.
- Grouping of weight tensors into `(out_channels, num_groups, pattern_n)` views based on the same channel/reduction axes the weight compression algorithms use: Linear and MatMul weights are grouped along the output-channel dimension, convolution weights along the flattened reduction dimension of every output channel.
- Magnitude selection per group via a stable sort, which makes the retained weights deterministic for tied magnitudes.
- Validation: weight tensors whose structural dimension is not divisible by the pattern group size fail with a `ValidationError`; operations that do not support structured pruning are skipped with a warning; ratio-based magnitude schedulers explicitly reject the new mode since the pattern fixes the sparsity level (the `ratio` argument is ignored for this mode).
- Integration with pruning statistics, `nncf.strip` and model save/load (`get_config` / `load_config`) for the new mask module.
- Documentation in the pruning usage guide.

### Testing

- New unit tests covering the per-group 2:4 invariant (not just the overall zero ratio), magnitude selection with positive/negative weights, deterministic behavior for tied magnitudes, non-divisible dimensions, invalid pattern validation, unsupported operations, ignored scopes, statistics, save/load round-trip, strip and forward-pass checks, plus an end-to-end `nncf.prune` integration test on a model with Linear and Conv2d layers: `pytest tests/torch/function_hook/pruning/` (95 passed).
- The broader `tests/torch/function_hook` suite (749 passed on the affected areas) and the pruning documentation example tests were run locally. The only failures observed (`test_export_onnx`, two weight-compression error-message tests) were confirmed to fail identically on the unmodified upstream `develop` branch in this environment.
- `ruff check` / `ruff format --check` clean; `mypy` reports no new errors compared to the unmodified `develop` branch in this environment.

### Limitations

- Only the PyTorch backend is covered; other backends are unchanged.
- 2:4 is the only pattern exposed for now; the internal grouping is parameterized by `(m, n)` so additional M:N patterns can be added later.
- A sparsity pattern alone does not guarantee acceleration on every backend or hardware target; the fine-tuning recommendation from the unstructured pruning docs applies here as well.

### Related issue

Closes #4222


uvv-01:
@jpablomch @clinty could you please check the workflow and tell me if this need any changes.

AlexanderDokuchaev:
Hi @uvv-01 
Could you please explain which inference type this should be used for? 
And provide some metrics with accuracy and performance?

uvv-01:
Hi @AlexanderDokuchaev , thanks for reviewing this.

The current implementation is intended as a PyTorch-side pruning method that produces weights with a fixed 2:4 structured sparsity pattern. The goal of this PR is to add the pruning capability to NNCF, without claiming automatic speedups on every backend.

The intended workflow is to apply "nncf.prune()" with "STRUCTURED_MAGNITUDE_2_4", optionally fine-tune, then strip the masks and export to a backend/hardware that supports 2:4 sparsity.

I agree that practical evidence would strengthen the PR. I’ll add a small benchmark covering baseline vs. pruned accuracy and latency, export results where supported, and document the hardware, runtime, model, input shape, and measurement procedure. I’ll also clarify that speedups depend on backend and hardware support.

For the initial experiment, I plan to use a small representative PyTorch model. If you have a preferred model, backend, or device, I’d be happy to evaluate on that instead.

